PR #13, opened 2018-08-13
*: extract scale down as scaler interface

weekface:
This PR only extract pd's scale down logic, when accepted, i will do other, include defer deletion of PVC.


weekface:
Closed in favor of #20

PR #16, opened 2018-08-13
add upgrade e2e test

xiaojingchen:
this pr add e2e test case to check upgrade feature
PTAL

### DIFF
--- a/tests/e2e/upgrade.go
+++ b/tests/e2e/upgrade.go
@@ -20,7 +20,7 @@ func testUpgrade() {
 	Expect(err).NotTo(HaveOccurred())
 
 	By("Then members should upgrade by order: pd ==> tikv ==> tidb")
-	err = wait.Poll(5*time.Second, 5*time.Minute, memberUpgraded)
+	err = wait.Poll(5*time.Second, 10*time.Minute, memberUpgraded)
 	Expect(err).NotTo(HaveOccurred())
 
 	By("Then all members should running")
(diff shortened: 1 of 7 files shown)


tests/e2e/upgrade.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func upgrade() (bool, error) {
+	tc, err := cli.PingcapV1alpha1().TidbClusters(ns).Get(clusterName, metav1.GetOptions{})
+	if err != nil {
+		logf("failed to upgrade tidbcluster, error: %v", err)
+		return false, err
+	}
+
+	tc.Spec.PD.Image = strings.Replace(tc.Spec.PD.Image, getImageTag(tc.Spec.PD.Image), "v2.0.0", -1)
```
tennix: Do we have TiDB `v2.0.0` docker images in our CI? It's better to define version as a constant.

xiaojingchen: I have pushed v2.0.0 to registry manually.

tests/e2e/upgrade.go
```diff
@@ -0,0 +1,169 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap.com/v1alpha1"
+	"github.com/pingcap/tidb-operator/pkg/controller"
+	apps "k8s.io/api/apps/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	upgradeVersion = "v2.0.0"
```
weekface: use v2.0.5 instead, downgrade may be unsafe

xiaojingchen: the v2.0.5 version is not exist

PR #19, opened 2018-08-14
update local dind documentation

gregwebs:
* Add a warning about problems with flannel which I experienced on Ubuntu
* Use Kubernetes version 11
* Add restart instructions
* update dind-run-operators, most importantly to always pull

### DIFF
--- a/docs/local-dind-tutorial.md
+++ b/docs/local-dind-tutorial.md
@@ -26,9 +26,9 @@ Before deploying a TiDB cluster to Kubernetes, make sure the following requireme
 1. Use DinD to install and deploy a multiple-node Kubernetes cluster:
 
     ```sh
-    $ wget https://cdn.rawgit.com/kubernetes-sigs/kubeadm-dind-cluster/64a2befa65ce23475158b65793e56d4bc1ae0a79/fixed/dind-cluster-v1.11.sh
-    $ chmod +x dind-cluster-v1.11.sh
-    $ CNI_PLUGIN="flannel" NUM_NODES=4 ./dind-cluster-v1.11.sh up
+    $ wget https://cdn.rawgit.com/kubernetes-sigs/kubeadm-dind-cluster/64a2befa65ce23475158b65793e56d4bc1ae0a79/fixed/dind-cluster-v1.10.sh
+    $ chmod +x dind-cluster-v1.10.sh
+    $ CNI_PLUGIN="flannel" NUM_NODES=4 ./dind-cluster-v1.10.sh up
     ```
 
     > **Note:** `CNI_PLUGIN=flannel` prevents issues with restarting docker. However, if you have network issues preventing tidb startup, you can re-create the cluster without setting `CNI_PLUGIN`.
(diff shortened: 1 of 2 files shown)


docs/local-dind-tutorial.md
```diff
@@ -26,17 +26,18 @@ Before deploying a TiDB cluster to Kubernetes, make sure the following requireme
...
     ```
 
+    > **Note:** `CNI_PLUGIN=flannel` prevents issues with restarting docker. However, if you have network issues preventing tidb startup, you can re-create the cluster without setting `CNI_PLUGIN`.
     > **Note:** If you fail to pull the Docker images due to the firewall, you can try the following method (the Docker images used are pulled from [UCloud Docker Registry](https://docs.ucloud.cn/compute/uhub/index)):
 
     ```sh
     $ git clone https://github.com/pingcap/kubeadm-dind-cluster
     $ cd kubeadm-dind-cluster
-    $ NUM_NODES=4 tools/multi_k8s_dind_cluster_manager.sh rebuild e2e-v1.10
+    $ NUM_NODES=4 tools/multi_k8s_dind_cluster_manager.sh rebuild e2e-v1.11
```
xiaojingchen: the pingcap/kubeadm-dind-cluster  doesn't support v1.11 now

```
xiaojingdeMacBook-Pro:kubeadm-dind-cluster xiaojingchen$ ./tools/multi_k8s_dind_cluster_manager.sh rebuild
usage:
  ./tools/multi_k8s_dind_cluster_manager.sh rebuild namespace
  ./tools/multi_k8s_dind_cluster_manager.sh clean namespace
  ./tools/multi_k8s_dind_cluster_manager.sh up namespace
  ./tools/multi_k8s_dind_cluster_manager.sh down namespace
  ./tools/multi_k8s_dind_cluster_manager.sh push namespace
The possible namespace values are: e2e-v1.7|stability-v1.7|e2e-v1.8|stability-v1.8|e2e-v1.9|stability-v1.9|e2e-v1.10|stability-v1.10
```

tennix:
Why upgrade kubernetes to v1.11? All our documents and e2e-tests are tested in v1.10. If we upgrade this we have to make a thorough test for v1.11. For now I think we should keep v1.10 and upgrade in a later time when all related works are tested.
@lilin90 tested flannel on Ubuntu, and there's no problem with it. What's your Ubuntu and Docker version?
image

hack/dind-run-operators.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+helm init --upgrade
 ./manifests/local-dind/setup.sh
 kubectl apply -f ./manifests/crd.yaml
-helm install charts/tidb-operator --name tidb-operator --namespace=tidb-operator
+helm install charts/tidb-operator --name tidb-operator --namespace=tidb-admin --set "imagePullPolicy=Always"
```
tennix: If we use this script in the tutorial, we should keep the namespace of tidb-operator consistent.

tennix: BTW, we plan to modify dind-cluster.sh to adjust for our test, because we can hardly deploy a docker registry for development with the upstream dind-cluster.sh. `kubectl port-forward` works for Linux but not Mac.

gregwebs: Do you like the name "tidb-admin"? With "operator", it only makes sense to run the operator (which is what we are doing at the moment). With "tidb-admin" it would make sense to run additional utilities under that name. I think it captures everything a user needs to know: "tidb": its for tidb. "admin", its for administrative duties for tidb.

I have seen port-forward working fine on Mac. But it doesn't keep up with changes properly?

tennix: I'm OK with tidb-admin. kubectl port-forward works for normal service but not for docker registry, since Docker for Mac runs Docker daemon inside a VM and port-forward can only bind localhost. We can't push Docker images through Docker daemon to VM's localhost.

PR #20, opened 2018-08-15
*: extract scale in as scaler interface

weekface:
This PR only extract pd's scale down logic, when accepted, i will do other, include defer deletion of PVC.

### DIFF
--- a/pkg/manager/member/pd_scaler.go
+++ b/pkg/manager/member/pd_scaler.go
@@ -30,13 +30,13 @@ func NewPDScaler(pdControl controller.PDControlInterface) *pdScaler {
 	return &pdScaler{pdControl}
 }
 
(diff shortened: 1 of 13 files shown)


pkg/manager/member/pd_scaler.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/pingcap/tidb-operator/pkg/controller"
+	apps "k8s.io/api/apps/v1beta1"
+)
+
+type pdScaler struct {
+	pdControl controller.PDControlInterface
+}
+
+// NewPDScaler returns a *pdScaler
+func NewPDScaler(pdControl controller.PDControlInterface) *pdScaler {
```
tennix: It's better to return a `Scaler` for exported function. So we don't have to add type assertion `var _ Scaler = &pdScaler{}`

PR #28, opened 2018-08-17
Scale e2e test

xiaojingchen:
- add e2e test for scale feature
- fix upgrade e2e test errors

PTAL


weekface:
To confirm pod's UID not changed when scales in/out. You can do it this way:

Before scales in/out, make a map:
```
podUIDMap = map[string]string{}
pods, err := kubeCli.CoreV1().Pods(ns).List(...)
...
for _, pod := range pods {
        podUIDMap[pod.GetName()] = pod.UID
}
```

After scales in/out:
```
pods, err := kubeCli.CoreV1().Pods(ns).List(...)
...
for _, pod := range pods {
        if uid, ok := podUIDMap[pod.GetName()]; ok && uid != pod.UID {
                Fail ....
        }
}
```

tests/e2e/scale.go
```diff
@@ -0,0 +1,215 @@
...
+		return false, nil
+	}
+
+	if len(podUIDsBeforeScale) == len(podUIDs) {
+		return false, fmt.Errorf("the length of pods before scale equals the length of pods after scale")
+	}
+
+	for podName, uidAfter := range podUIDs {
+		if uidBefore, ok := podUIDsBeforeScale[podName]; ok && uidBefore != uidAfter {
+			return false, fmt.Errorf("pod: [%s] have be recreated", podName)
```
weekface:
```
return true, fmt.Errorf("pod: [%s] have be recreated", podName)
```

tests/e2e/scale.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+
+	return true, nil
+}
+
+// scaleInSecurity confirms member scale in securely
+func scaleInSecurely() (bool, error) {
```
tennix: `Safely` not `Securely`

PR #31, opened 2018-08-17
Readability Improvements

morgo:
Changed helm instructions to include install.
Improved readability by just following commands (don't double clone repo by accident).
Added colons to text prior to commands.
Clarified further cases of tidb cluster vs kubernetes cluster.


morgo:
Made obsolete by #32

PR #33, opened 2018-08-19
chore(development env): unified the DinD development environment

onlymellb:
The purpose of this PR is to unify the development environment of developers at home and abroad by setting environment variables `KUBE_REPO_PREFIX`. This script will do the following at startup.

- Launch a k8s v1.10 cluster with flannel network plugin
- Install a docker registry in DinD env, the defaults port of the registry is `5000`, which can be changed by setting the environment variable `REGISTRY_PORT`.
- Deploy tiller in k8s 
- Deploy local volume provisioner in k8s
- Create PV mount points on all Kube nodes except the master node, you can control the number of created PVS by setting the environment variable PV_NUMS. All PV directories will be mounted in the host's `YOU-TIDB-OPERATOR-DIR/data` directory.
- Create TidbCluster CRD object and e2e namespace `tidb-operator-e2e`
- Start and stop an existing cluster with the commands `./dind-cluster-v1.10.sh stop` and `./dind-cluster-v1.10.sh start`
- Rebuild the k8s cluster with the commands `./dind-cluster-v1.10.sh clean` and `./dind-cluster-v1.10.sh up`

PTAL @tennix @weekface @xiaojingchen @gregwebs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ images/tidb-operator-e2e/tidb-cluster/
 images/tidb-operator-e2e/tidb-operator/
 *.tar
 tmp/
+data/
 .idea
 
 # ginkgo test coverage
--- a/checkout-pr-branch.sh
+++ b/checkout-pr-branch.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# This script is used to checkout a tidb-operator PR branch in a forked repo.
+if [[ -z ${1:-} ]]; then
+echo "\
+This script is used to checkout a tidb-operator PR branch in a forked repo
+
+Usage:
+    checkout-pr-branch.sh [github-username]:[pr-branch]
+
+The argument can be copied directly from github PR page.
+The local branch name would be [github-username]/[pr-branch].\
+" >&2
+exit
+fi
+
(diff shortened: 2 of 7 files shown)


tennix:
I think it's better to mount a host directory as registry directory too.

tennix:
Kubernetes dashboard can't be accessed by http://localhost:8080/ui

gregwebs:
This looks great! Seems like we need some updates to the docs, including the DinD tutorial.
Right now I get an error at the end of running this.

``` sh
./manifests/local-dind/dind-cluster-v1.10.sh: line 1840: replace_repo_prefix: unbound variable
error: no objects passed to apply
➜  tidb-operator git:(onlymellb/unified-development-env) 
```

gregwebs:
Flannel previously failed for me, but seems to be working now when using this :)

gregwebs:
One problem I have always had with these setup scripts is that I cannot change NUM_NODES without cleaning. It would be great to address that in another PR.

onlymellb:
@gregwebs I have fixed the problem without variable `replace_repo_prefix` definition, In this PR if you have already started the k8s cluster, you can't dynamically scale out the node by setting the NUM_NODES environment variable.

PR #38, opened 2018-08-21
Fix the issue that failed to enter PD/TiKV/TiDB's debug mode

onlymellb:
This PR cancels the setting of the `errexit` option in a shell script in order to avoid  `source /etc/podinfo/annotations` failure, and also adds the logic of pd's domain name resolution.

PTAL @tennix @weekface @xiaojingchen @gregwebs


charts/tidb-cluster/templates/pd-configmap.yaml
```diff
@@ -35,10 +35,26 @@ data:
...
+    threshold=30
+    while true;do
+        nslookup ${HOSTNAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc 2>/dev/null
+        [[ $? -eq 0 ]] && break
+        echo "nslookup domain ${HOSTNAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc failed"
+        sleep ${period}
+        elapseTime=$(( elapseTime+period ))
+        if [[ ${elapseTime} -ge ${threshold} ]];then
+            echo "nslookup domain ${HOSTNAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc timeout"
+            break
```
weekface: change `break` to `exit 1`?

gregwebs: echo should also be to stderr

gregwebs:
But I already see `source ... || true` in some of the scripts so why remove `errexit`? I usually use an echo messages like `source ... || echo "could not source annotations`.

I am very much against turning off `errexit`. `errexit` is the only way I know to write reliable shell scripts.

tennix:
`source file` execute every line of the file in a shell, if the first line fails, the `true` will be executed and we'll never go to debug mode.

tennix:
No need to add `|| true` for `source`

PR #41, opened 2018-08-22
Fix service's nodePort will changing every 30s

xiaojingchen:
This pr rewrites the service comparison logic to avoid frequent update services causing the nodePort to always be changing


weekface:
You should add an unit test spec to confirm the logic is correct.

Issue #44, opened 2018-08-22
Add tolerations onto pods of tidb cluster


mathspanda:
Pods of tidb clusters are necessary to be deployed on dedicated nodes under certain scenarios. Some taints may be added to these dedicated nodes to ensure that other pods will not be scheduled onto them. So maybe we should apply corresponding tolerations to tidb pods?

weekface:
Yes, operator should allow users to add tolerations to TiDB pods.

weekface:
@mathspanda Could you send us a PR to add this feature? You can follow [CONTRIBUTING](https://github.com/pingcap/tidb-operator/blob/master/docs/CONTRIBUTING.md) guide for how to contribute.

mathspanda:
sure, i will send a PR.

weekface:
Close in favor of #49

PR #49, opened 2018-08-24
add tolerations configuration to tidb pods

mathspanda:
This PR allows users to add tolerations to TiDB pods.

Fix issue #44.


mathspanda:
@weekface PTAL.

Issue #52, opened 2018-08-24
Remove noisy error log message


weekface:
When we don't set the [`resources`](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/values.yaml#L48) for TidbCluster, tidb-operator will print these noisy error messages:
```
E0824 07:53:15.726431       1 util.go:161] failed to parse CPU resource  to quantity: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
E0824 07:53:15.726524       1 util.go:166] failed to parse memory resource  to quantity: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
E0824 07:53:15.772255       1 util.go:161] failed to parse CPU resource  to quantity: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
E0824 07:53:15.773356       1 util.go:166] failed to parse memory resource  to quantity: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

We should remove these noisy error messages when the resource is empty.

PR #54, opened 2018-08-24
[ISSUE #52] Remove noise error log message

huanwei:
Signed-off-by: huanwei <huan@harmonycloud.cn>

Fix [issue #52  Remove noise error log message](https://github.com/pingcap/tidb-operator/issues/52)

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -157,30 +157,38 @@ func ResourceRequirement(spec v1alpha1.ContainerSpec, defaultRequests ...corev1.
 		if rr.Requests == nil {
 			rr.Requests = make(map[corev1.ResourceName]resource.Quantity)
 		}
-		if q, err := resource.ParseQuantity(spec.Requests.CPU); err != nil {
-			glog.Errorf("failed to parse CPU resource %s to quantity: %v", spec.Requests.CPU, err)
-		} else {
-			rr.Requests[corev1.ResourceCPU] = q
+		if spec.Requests.CPU != nil {
+			if q, err := resource.ParseQuantity(spec.Requests.CPU); err != nil {
+				glog.Errorf("failed to parse CPU resource %s to quantity: %v", spec.Requests.CPU, err)
+			} else {
+				rr.Requests[corev1.ResourceCPU] = q
+			}
 		}
-		if q, err := resource.ParseQuantity(spec.Requests.Memory); err != nil {
-			glog.Errorf("failed to parse memory resource %s to quantity: %v", spec.Requests.Memory, err)
-		} else {
-			rr.Requests[corev1.ResourceMemory] = q
+		if spec.Requests.Memory != nil {
+			if q, err := resource.ParseQuantity(spec.Requests.Memory); err != nil {
+				glog.Errorf("failed to parse memory resource %s to quantity: %v", spec.Requests.Memory, err)
+			} else {
+				rr.Requests[corev1.ResourceMemory] = q
+			}
 		}
 	}
 	if spec.Limits != nil {
 		if rr.Limits == nil {
 			rr.Limits = make(map[corev1.ResourceName]resource.Quantity)
 		}
-		if q, err := resource.ParseQuantity(spec.Limits.CPU); err != nil {
-			glog.Errorf("failed to parse CPU resource %s to quantity: %v", spec.Limits.CPU, err)
-		} else {
-			rr.Limits[corev1.ResourceCPU] = q
+		if spec.Limits.CPU != nil {
+			if q, err := resource.ParseQuantity(spec.Limits.CPU); err != nil {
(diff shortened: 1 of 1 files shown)


huanwei:
PTAL. @weekface

pkg/util/util.go
```diff
@@ -157,30 +157,38 @@ func ResourceRequirement(spec v1alpha1.ContainerSpec, defaultRequests ...corev1.
 		if rr.Requests == nil {
 			rr.Requests = make(map[corev1.ResourceName]resource.Quantity)
 		}
-		if q, err := resource.ParseQuantity(spec.Requests.CPU); err != nil {
-			glog.Errorf("failed to parse CPU resource %s to quantity: %v", spec.Requests.CPU, err)
-		} else {
-			rr.Requests[corev1.ResourceCPU] = q
+		if spec.Requests.CPU != nil {
```
weekface: `spec.Requests.CPU`'s type is [`string`](https://github.com/weekface/tidb-operator/blob/fea4bd22345e55bf2967c63530ee831c8c32b5b7/pkg/apis/pingcap.com/v1alpha1/types.go#L156), can't compared with `nil`, replace `nil` with `""` instead

pkg/util/util.go
```diff
@@ -157,30 +157,38 @@ func ResourceRequirement(spec v1alpha1.ContainerSpec, defaultRequests ...corev1.
...
+				glog.Errorf("failed to parse CPU resource %s to quantity: %v", spec.Requests.CPU, err)
+			} else {
+				rr.Requests[corev1.ResourceCPU] = q
+			}
 		}
-		if q, err := resource.ParseQuantity(spec.Requests.Memory); err != nil {
-			glog.Errorf("failed to parse memory resource %s to quantity: %v", spec.Requests.Memory, err)
-		} else {
-			rr.Requests[corev1.ResourceMemory] = q
+		if spec.Requests.Memory != nil {
```
weekface: ditto

pkg/util/util.go
```diff
@@ -157,30 +157,38 @@ func ResourceRequirement(spec v1alpha1.ContainerSpec, defaultRequests ...corev1.
...
 	}
 	if spec.Limits != nil {
 		if rr.Limits == nil {
 			rr.Limits = make(map[corev1.ResourceName]resource.Quantity)
 		}
-		if q, err := resource.ParseQuantity(spec.Limits.CPU); err != nil {
-			glog.Errorf("failed to parse CPU resource %s to quantity: %v", spec.Limits.CPU, err)
-		} else {
-			rr.Limits[corev1.ResourceCPU] = q
+		if spec.Limits.CPU != nil {
```
weekface: ditto

pkg/util/util.go
```diff
@@ -157,30 +157,38 @@ func ResourceRequirement(spec v1alpha1.ContainerSpec, defaultRequests ...corev1.
...
+				glog.Errorf("failed to parse CPU resource %s to quantity: %v", spec.Limits.CPU, err)
+			} else {
+				rr.Limits[corev1.ResourceCPU] = q
+			}
 		}
-		if q, err := resource.ParseQuantity(spec.Limits.Memory); err != nil {
-			glog.Errorf("failed to parse memory resource %s to quantity: %v", spec.Limits.Memory, err)
-		} else {
-			rr.Limits[corev1.ResourceMemory] = q
+		if spec.Limits.Memory != nil {
```
weekface: ditto

gregwebs:
The zero value comparison in this case should be the empty string rather than nil.

huanwei:
@weekface @gregwebs Thanks for your review.

tennix:
@huanwei Thanks for your contribution!
The noisy error log message comes from `resource.ParseQuantity`, by grepping `ParseQuantity` you'll find 3 other places calling `ParseQuantity`:
https://github.com/pingcap/tidb-operator/blob/5adb50444336af11806a21d17ff0f45155bb464d/pkg/controller/controller_utils.go#L79
https://github.com/pingcap/tidb-operator/blob/5adb50444336af11806a21d17ff0f45155bb464d/pkg/manager/member/pd_member_manager.go#L366
https://github.com/pingcap/tidb-operator/blob/5adb50444336af11806a21d17ff0f45155bb464d/pkg/manager/member/tikv_member_manager.go#L277
The latter two are parsing storage requests for PVC, it must be non-empty because most platform needs this value to allocate a PV, so the error message is OK here. But the first one, tikv storage capacity is not required to be non-empty, if it's empty we set it to `zero`, but we got noisy error messages. So you need to fix this too.

weekface:
@tennix 
This PR only fix the `CPU` and `Memory` noisy error messages problem. It is enough for this PR. The `storage` problem should be fixed in another PR i think.

tennix:
OK, @huanwei can you send another PR to fix the issue I listed above?

huanwei:
@tennix Sure. I will submit a new PR to handle the storage problem.

Issue #73, opened 2018-09-02
status of tidbcluster CR always updated


mathspanda:
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/tidbcluster/tidb_cluster_control.go#L80

Variable `oldStatus` is a pointer type, while `tc.Status` is not. According to the comment of function `reflect.DeepEqual`, values of distinct types are never deeply equal. So maybe the result of `DeepEqual` in this line is always false, and then status of tidbcluster CR  is always updated?

weekface:
Good Catch! Thank you.

Further moreover:
1. Replace `reflect.DeepEqual` with `apiequality.Semantic.DeepEqual` which is supplied by K8s.
https://github.com/pingcap/tidb-operator/blob/fe1792b2e8fbbc6728c204900529d07f2eb34b54/pkg/manager/member/utils.go#L174
2. Add unit test to cover this situation

Would you like to resolve this problem?

mathspanda:
Sure.

weekface:
Thank you for your feedback.

PR #75, opened 2018-09-03
Fix deepequal result in UpdateTidbCluster always false

mathspanda:
1、Fix deepequal result in UpdateTidbCluster always false
2、Replace reflect.DeepEqual with apiequality.Semantic.DeepEqual
3、Add unit test to cover this situation

Fix issue #73

### DIFF
--- a/pkg/controller/tidbcluster/tidb_cluster_control.go
+++ b/pkg/controller/tidbcluster/tidb_cluster_control.go
@@ -14,12 +14,11 @@
 package tidbcluster
 
 import (
-	"reflect"
-
 	"github.com/golang/glog"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap.com/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/manager"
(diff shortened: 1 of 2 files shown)


mathspanda:
@weekface PTAL

weekface:
I am sorry i can't understand your test codes fully. In order to test `UpdateTidbCluster` method fully, we must mock `updateTidbCluster `(line 75) and `updateTidbClusterStatus`(line 81) , these two methods are hard to mock now. So for simplicity, i think just test `apiequality.Semantic.DeepEqual` is enough.

What about just test it like this now:

```
a := v1alpha1.TidbClusterStatus{}
b := a.DeepCopy()

g.Expect(apiequality.Semantic.DeepEqual(&a, b)).To(Equal(true))
// change b
b.PD.Phase = "xxx"
g.Expect(apiequality.Semantic.DeepEqual(&a, b)).To(Equal(false))

```

mathspanda:
THX for comments. Maybe i make things harder than necessary. Fixed.

PR #80, opened 2018-09-05
Add a new RequeueError type

### LINKED ISSUE #50: Need a new error type to requeue the items
Now the tidbcluster [Controller](https://github.com/weekface/tidb-operator/blob/b7ef27008d4f50d9be261d68f731c27ece7cc430/pkg/controller/tidbcluster/tidb_cluster_controller.go#L47) has a attribute `queue` which is used for tracking the items that should be processed.

If the [`sync`](https://github.com/weekface/tidb-operator/blob/b7ef27008d4f50d9be261d68f731c27ece7cc430/pkg/controller/tidbcluster/tidb_cluster_controller.go#L203) method returns an error, we [log](https://github.com/weekface/tidb-operator/blob/b7ef27008d4f50d9be261d68f731c27ece7cc430/pkg/controller/tidbcluster/tidb_cluster_controller.go#L204) the error message and [requeue](https://github.com/weekface/tidb-operator/blob/b7ef27008d4f50d9be261d68f731c27ece7cc430/pkg/controller/tidbcluster/tidb_cluster_controller.go#L205) the item, else [forget](https://github.com/weekface/tidb-operator/blob/b7ef27008d4f50d9be261d68f731c27ece7cc430/pkg/controller/tidbcluster/tidb_cluster_controller.go#L207) it.

So if we have a task that needs several `sync`s, it needs to requeue this item after the first `sync`,   but don't log the error (because it is not an error, we just want to process the item again),  we should defines a new error type (maybe named `RequeueError`), if ther error returns from the `sync` method is a `RequeueError` type, don't log the error message.

weekface:
This PR resolves #50 .


gregwebs:
It looks like this removes aggregate error creation because they will not pass the test `IsRequeueError`. However, an error will also not pass that test if it is wrapped in any way. I think we should traverse errors as per https://gist.github.com/gregwebs/5a14a83970d607411310a3e0281d55be.

gregwebs:
I have the function `FInd` now which will work with the k8s aggregate error: https://github.com/lysu/errors/pull/2

weekface:
This sounds good. We should be waiting for that [feature](https://github.com/lysu/errors/pull/2).

weekface:
Switched to `pingcap/errors`, @gregwebs PTAL

gregwebs:
We are maintaining backwards compatibility for pingcap/errors with pkg/errors, so it is better to import pingcap/errors as pkg/errors as [done in TiDB](https://github.com/pingcap/tidb/pull/7151/files#diff-836546cc53507f6b2d581088903b1785R103)

PR #83, opened 2018-09-05
*: Rename label annotation keys

### LINKED ISSUE #66: label's key names are not uniform between monitor and pd/tidb/tikv
the monitor's deployments and pods create by helm, their label keys is difference with pd/tikv/tidb's label keys.  

monitor's labels:
```app=tidb-cluster,chart=tidb-cluster-0.1.0,component=monitor,heritage=Tiller,release=tidb-cluster-e2e```
pd's labels:
```cluster.pingcap.com/app=pd,cluster.pingcap.com/owner=tidbCluster,cluster.pingcap.com/tidbCluster=demo-cluster```

we need to unify their label keys naming conventions for management and query.
1. should decide whether to use long or short names
long name: ```cluster.pingcap.com/app```
short name: ```app```
2. should define the unified label key names
```owner=tidb-operator```
```cluster=tidb cluster name```
```app=pd/tidb/tikv/monitor```
3.  limited only use lowercase characters and ‘-’

The above content needs to be discussed，please makes your suggestions

### PR DESCRIPTION
**Pull request by tennix**
This PR fixes #66. The labels and annotations are renamed according to [Kubernetes recommendation](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) and [Helm recommendation](https://github.com/helm/helm/blob/master/docs/chart_best_practices/labels.md).

With this PR, we can not only query all resources easily with unified labels, but also distingush resources managed by Tiller with TiDB Operator itself easily. The common labels are:
* `app.kubernetes.io/instance`: the unique cluster name.
* `app.kubernetes.io/name`: the entire app name, for tidb cluster it's always `tidb-cluster`.
* `app.kubernetes.io/component`: the component name, `pd`/`tikv`/`tidb`/`monitor` etc.
* `app.kubernetes.io/managed-by`: the resource manager name, `Tiller` or `tidb-operator`

### CODE REVIEW

File: pkg/manager/meta/meta_manager_test.go [hunk 1]
```diff
@@ -208,8 +209,8 @@ func TestMetaManagerSync(t *testing.T) {
 			pvcUpdateErr:  false,
 			pvUpdateErr:   false,
 			podChanged:    true,
-			pvcChanged:    true,
-			pvChanged:     true,
+			pvcChanged:    false,
```
weekface: Why change this to false?

tennix: PVC's labels are [inherited from StatefulSet label selector](https://github.com/kubernetes/kubernetes/blob/v1.10.0/pkg/controller/statefulset/stateful_set_utils.go#L152) other than from the pod, but the statefulset controller won't work in this unit test. So PVC and PV won't be changed.

onlymellb: PVC and PV would be changed because some of the labels such as memberId、clusterId、storeId are synchronized from the pod.

charts/tidb-cluster/templates/monitor-job.yaml
```diff
@@ -4,18 +4,18 @@ kind: Job
...
   labels:
-    app: {{ template "tidb-cluster.name" . }}
-    component: monitor-configurator
-    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+    app.kubernetes.io/name: {{ template "tidb-cluster.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Values.clusterName }}
+    app.kubernetes.io/component: monitor-configurator
```
xiaojingchen: this is just a job, not need to define a new component for it, ```monitor``` will be better

tennix: If the labels are the same with monitor deployment, then monitor service would select it which doesn't make sense.

PR #84, opened 2018-09-06
Manage tidb service by helm instead of tidb-operator

### LINKED ISSUE #36: Add annotations and labels for service exposed by TiDB cluster
Load Balancer service on public cloud usually needs to add some annotations or labels to configure as documented [here](https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types). 
TiDB Operator needs to add annotations and labels for service exposed by TiDB cluster.

### LINKED ISSUE #65: Manage tidb service by charts
Services of tidb cluster include two types, the headless services and external access services. now they all maintained by tidb operator.

The headless services are used for network communication between internal pods and are invisible to users, so they should be maintained by the operator itself, but the tidb service is exposed to the user. It should be modified by the user if needed, so we should move the tidb service to charts and manage it by Helm.

tennix:
This PR fixes #65 and fixes #36. It move tidb service management from controller-manager to Charts and allows users to specify extra arguments for using public cloud LoadBalancer.

Since this already migrated to new labels, it should wait until #83 get merged to master.


charts/tidb-cluster/templates/tidb-service.yaml
```diff
@@ -0,0 +1,40 @@
...
+  ports:
+  - name: mysql-client
+    port: 4000
+    targetPort: 4000
+    protocol: TCP
+    {{- if .Values.tidb.service.mysqlNodePort }}
+    nodePort: {{ .Values.tidb.service.mysqlNodePort }}
+    {{- end }}
+  {{- if .Values.tidb.service.exposeStatus }}
+  - name: status
```
xiaojingchen: this port should be export as a headless service, no need be added to this service

tennix: Sometimes users need this service to query meta info no matter which tidb-server is just like mysql port.

Issue #82, opened 2018-09-05
TiKV pod sidecar container pushgateway image in values.yaml is never used


tennix:
TiKV pod sidecar container pushgateway in `charts/tidb-cluster/values.yaml` is not rendered in `charts/tidb-cluster/templates/tidb-cluster.yaml`. 
https://github.com/pingcap/tidb-operator/blob/5107beff8306cfe943279eeaf1bdfdf2ada8e953/charts/tidb-cluster/values.yaml#L102-L109

This section needs to be added in `charts/tidb-cluster/templates/tidb-cluster.yaml`.

Besides the pushgateway image is never used in TiKV pod spec because `controller.GetPushgatewayImage` function is wrong.
https://github.com/pingcap/tidb-operator/blob/5107beff8306cfe943279eeaf1bdfdf2ada8e953/pkg/manager/member/tikv_member_manager.go#L333

weekface:
Close in favor of #85

PR #85, opened 2018-09-06
fix tikvpromgateway not used in tidbcluster templates

mathspanda:
1. Add tikvpromgateway values in tidbcluster template
2. Fix function `controller.GetPushgatewayImage`

Fix issue: #82


mathspanda:
@weekface @tennix PTAL.

PR #96, opened 2018-09-12
*: Setting TZ env instead of mounting /etc/localtime

### LINKED ISSUE #88: Setting timezone correctly for TiDB cluster
We're using alpine Linux as the base image for pd/tikv/tidb Docker image. Currently, in kubernetes we mount host machine `/etc/localtime` to pod, so the log timestamp is the same with the host machine.

But tidb/tikv also need to know timezone name which cannot retrieve only from `/etc/localtime`. Normally on a real machine, /etc/localtime is a soft link to `/usr/share/zoneinfo/<timezone-name>`. With the soft link, TiDB and TiKV can retrieve timezone name. Also, some Linux distros set timezone name in `/etc/timezone` or `/etc/TZ` and Alpine Linux uses `TZ` environment variable.

So on Kubernetes, we need to use a soft link for `/etc/localtime` instead of just mount host's `/etc/localtime`. Besides we'd better set `TZ` environment and `/etc/TZ` too to make programs funtion well in containers.

*Note*: Previous released Docker images on DockerHub doesn't contain `tzdata` package which will never work for this. Only newer version >= v2.0.7 contains `tzdata` package. So if users want to use older version of TiDB, then they have to build PD/TiKV/TiDB Docker images themselves to include `tzdata` package.

tennix:
This PR fixes #88. It also removes the mounting /etc/localtime solution for log timestamp.

With this PR the logs for pods will be in the same timezone with TiDB instead of with host machine. This may introduce confusion for users when using `kubectl logs` to view different pod logs in the same Kubernetes cluster, unfortunately this is unavoidable. But assuming there's a log collector, users should reference the collecting timestamp instead of timestamp in the log message.

### DIFF
--- a/charts/tidb-cluster/templates/monitor-configmap.yaml
+++ b/charts/tidb-cluster/templates/monitor-configmap.yaml
@@ -14,12 +14,12 @@ data:
     global:
       scrape_interval: 15s
       evaluation_interval: 15s
-    {{- if .Values.alertmanagerURL }}
+    {{- if .Values.monitor.prometheus.alertmanagerURL }}
     alerting:
       alertmanagers:
       - static_configs:
         - targets:
-          - {{ .Values.alertmanagerURL }}
+          - {{ .Values.monitor.prometheus.alertmanagerURL }}
     {{- end }}
     scrape_configs:
       - job_name: 'tidb-cluster'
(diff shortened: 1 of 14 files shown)


tennix:
We should, but this PR doesn't require it because it only sets `TZ` environment variable. We should do it in another PR.

PR #100, opened 2018-09-17
add full backup support

### LINKED ISSUE #98: TiDB cluster full backup feature
TiDB uses [mydumper](https://github.com/pingcap/mydumper) to achieve the full backup.

With TiDB Operator, we need to use helm to create a `PVC` and a `CronJob`, the `CronJob` will create `Pods` which will all use the same PVC and PV.

In every `Pods` created by `CronJob`, we should use `mydumper` to backup the whole TiDB cluster's data to the different unique data directory in `PV`.

weekface:
This PR resolves #98 , add the full backup feature.

### DIFF
--- a/charts/tidb-cluster/templates/fullbackup-cronjob.yaml
+++ b/charts/tidb-cluster/templates/fullbackup-cronjob.yaml
@@ -1,0 +1,72 @@
+{{- if .Values.fullbackup.create }}
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ .Values.clusterName }}-fullbackup
+  labels:
+    app.kubernetes.io/name: {{ template "tidb-cluster.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Values.clusterName }}
+    app.kubernetes.io/component: fullbackup
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  schedule: "{{ .Values.fullbackup.schedule }}"
+  concurrencyPolicy: Forbid
+  suspend: {{ .Values.fullbackup.suspend }}
+  startingDeadlineSeconds: {{ .Values.fullbackup.startingDeadlineSeconds }}
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          initContainers:
+          - name: get-ts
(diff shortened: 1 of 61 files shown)


charts/tidb-cluster/values.yaml
```diff
@@ -171,6 +171,35 @@ monitor:
   #   value: tidb
   #   effect: "NoSchedule"
 
+fullbackup:
+  create: false
+  binlogImage: pingcap/tidb-binlog:latest
+  mydumperImage: pingcap/tidb-enterprise-tools:latest
```
gregwebs: Can we use a separate mydumper image? Particularly since mydumper is its own GPL licensed software and we are iterating to push our changes upstream. It would also remove the need to configure `mydumperBin`

gregwebs:
Looking good! There should be some associated docs changes with this PR.

weekface:
https://github.com/pingcap/SRE/pull/102

charts/tidb-cluster/values.yaml
weekface: Removed `mydumperBin`.

weekface: We can change the image to a separate mydumper image when it is available.

gregwebs: :+1: I will make a mydumper image available

weekface:
The full backup associated docs may come in another PR. @gregwebs

weekface:
We must split it to a new chart, because this is an issue of helm: https://github.com/helm/helm/issues/1193.

iamxy:
The binlogctl tool has been added into related docker image

charts/tidb-cluster/templates/fullbackup-cronjob.yaml
```diff
@@ -0,0 +1,70 @@
...
+metadata:
+  name: {{ .Values.clusterName }}-fullbackup
+  labels:
+    app.kubernetes.io/name: {{ template "tidb-cluster.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Values.clusterName }}
+    app.kubernetes.io/component: fullbackup
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  schedule: "{{ .Values.fullbackup.schedule }}"
```
tennix: `{{ .Values.fullbackup.schedule | quote }}`

PR #99, opened 2018-09-14
PD graceful upgrade

xiaojingchen:
this PR fixes #94  and #70

@weekface @tennix @gregwebs @onlymellb PTAL

PR #104, opened 2018-09-25
TiKV graceful upgrade

### LINKED ISSUE #93: TiKV graceful upgrade
TiKV graceful upgrade steps:
1. call the pd's api `pd/api/v1/schedulers` with the value `{"evict-leader-scheduler":storeID}`, then pd will start a scheduler to evict leaders in the tikv.
2. wait util all leaders in the tikv have be evicted
3. call call the pd's api `pd/api/v1/schedulers/{storeID}` to delete the scheduler
4. check the tikv's status until they are all available then  continue to upgrade the next tikv pod

xiaojingchen:
this PR add tikv graceful upgrade feature fixes #93 
it should be merged after the PR: #99 

@weekface @tennix @gregwebs @onlymellb PTAL

### DIFF
--- a/pkg/manager/member/tikv_upgrader.go
+++ b/pkg/manager/member/tikv_upgrader.go
@@ -19,6 +19,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/golang/glog"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap.com/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/controller"
@@ -40,7 +41,9 @@ type tikvUpgrader struct {
 }
 
 // NewTiKVUpgrader returns a tikv Upgrader
-func NewTiKVUpgrader(pdControl controller.PDControlInterface, podControl controller.PodControlInterface, podLister corelisters.PodLister) Upgrader {
+func NewTiKVUpgrader(pdControl controller.PDControlInterface,
+	podControl controller.PodControlInterface,
+	podLister corelisters.PodLister) Upgrader {
 	return &tikvUpgrader{
 		pdControl:  pdControl,
 		podControl: podControl,
@@ -61,31 +64,20 @@ func (tku *tikvUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Stateful
 	}
 
 	if !tc.Status.TiKV.SyncSuccess {
-		force, err := tku.needForce(tc)
-		if err != nil {
-			return err
-		}
-		if force {
-			setUpgradePartition(newSet, 0)
-			return nil
-		}
 		return fmt.Errorf("tidbcluster: [%s/%s]'s tikv status sync failed,can not to be upgraded", ns, tcName)
 	}
 
 	tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
+	setUpgradePartition(newSet, int(*oldSet.Spec.UpdateStrategy.RollingUpdate.Partition))
+
 	upgradePod, err := tku.findUpgradePod(tc)
 	if upgradePod == nil || err != nil {
-		setUpgradePartition(newSet, int(*oldSet.Spec.UpdateStrategy.RollingUpdate.Partition))
 		return err
-	}
-
-	if upgradePod == nil {
-		return nil
 	}
 
 	storeIDStr, exist := upgradePod.Labels[label.StoreIDLabelKey]
 	if !exist {
-		return fmt.Errorf("Tidbcluster: [%s/%s]'s tikv Pod:[%s] does not contain storeID", ns, tcName, upgradePod.GetName())
+		return fmt.Errorf("tidbcluster: [%s/%s]'s tikv Pod:[%s] does not contain storeID", ns, tcName, upgradePod.GetName())
 	}
 	upgradeOrdinal, err := getOrdinal(upgradePod)
 	if err != nil {
@@ -98,30 +90,13 @@ func (tku *tikvUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Stateful
 	}
 
 	if store, exist := tc.Status.TiKV.Stores[storeIDStr]; exist {
-		evictLeaderBeginTimeStr, evicting := upgradePod.Annotations[EvictLeaderBeginTime]
 		storeID, err := strconv.ParseUint(storeIDStr, 10, 64)
 		if err != nil {
 			return err
 		}
-		if !evicting && store.LeaderCount != 0 {
-			err = tku.pdControl.GetPDClient(tc).BeginEvictLeader(storeID)
-			if err != nil {
-				return err
-			}
-			// updete upgradePod
-			upgradePod.Annotations[EvictLeaderBeginTime] = time.Now().Format("2006-01-02 15:04:05")
-			_, err = tku.podControl.UpdatePod(tc, upgradePod)
-			if err != nil {
-				return err
-			}
-			return nil
-		}
+		_, evicting := upgradePod.Annotations[EvictLeaderBeginTime]
 
-		evictLeaderBeginTime, err := time.Parse("2006-01-02 15:04:05", evictLeaderBeginTimeStr)
-		if err != nil {
-			return err
-		}
-		if store.LeaderCount == 0 || time.Now().After(evictLeaderBeginTime.Add(EvictLeaderTimeOut)) {
+		if tku.readyToUpgrade(upgradePod, store) {
 			err = tku.pdControl.GetPDClient(tc).EndEvictLeader(storeID)
 			if err != nil {
 				return err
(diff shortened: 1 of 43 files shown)


pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,162 @@
...
+	"github.com/pingcap/tidb-operator/pkg/controller"
 	apps "k8s.io/api/apps/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
 )
 
-type tikvUpgrader struct{}
+const (
+	EvictLeaderBeginTime = "evictLeaderBeginTime"
+	EvictLeaderTimeOut   = 3 * time.Minute
```
weekface: Does 3 minutes enough?

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,162 @@
...
+	}
+
+	tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
+	setUpgradePartition(newSet, *oldSet.Spec.UpdateStrategy.RollingUpdate.Partition)
+
+	if tc.Status.TiKV.StatefulSet.CurrentReplicas == 0 {
+		return controller.RequeueErrorf("tidbcluster: [%s/%s]'s tikv doesn't have old version pod to upgrade", ns, tcName)
+	}
+
+	if tc.Status.TiKV.StatefulSet.UpdatedReplicas+tc.Status.TiKV.StatefulSet.CurrentReplicas != tc.Status.TiKV.StatefulSet.Replicas {
```
weekface: Use `tc.PDAllPodsStarted`

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,162 @@
...
+		return controller.RequeueErrorf("tidbcluster: [%s/%s]'s tikv pods are not all created", ns, tcName)
+	}
+	for i := tc.Status.TiKV.StatefulSet.Replicas; i > tc.Status.TiKV.StatefulSet.CurrentReplicas; i-- {
+		store := tku.getStoreByOrdinal(tc, i-1)
+		if store == nil || store.State != metapb.StoreState_name[int32(metapb.StoreState_Up)] {
```
weekface: Use `v1alpha1.TiKVStateUp`

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,162 @@
...
+	}
+
+	upgradeOrdinal := tc.Status.TiKV.StatefulSet.CurrentReplicas - 1
+	upgradePodName := tikvPodName(tc, upgradeOrdinal)
+	upgradePod, err := tku.podLister.Pods(ns).Get(upgradePodName)
+	if err != nil {
+		return err
+	}
+
+	for _, store := range tc.Status.TiKV.TombstoneStores {
```
weekface: What is this block means?

The `TombstoneStores` may contain all the ordinals.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,162 @@
...
+				}
+				return nil
+			}
+		}
+	}
+
+	return nil
+}
+
+func (tku *tikvUpgrader) readyToUpgrade(upgradePod *corev1.Pod, store v1alpha1.TiKVStore) bool {
```
weekface: I think this method should have two return values: (bool, error)

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,162 @@
...
+	tcName := tc.GetName()
 	if tc.Status.PD.Phase == v1alpha1.UpgradePhase {
 		_, podSpec, err := GetLastAppliedConfig(oldSet)
 		if err != nil {
 			return err
 		}
 		newSet.Spec.Template.Spec = *podSpec
-	} else {
-		tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
+		return nil
```
weekface: Why return directly？

xiaojingchen: tikv can't be upgraded when pd is upgrading

pkg/manager/member/tikv_upgrader.go
xiaojingchen: just mean directly to upgrade when the store's state is Tombstone.

weekface: This is not correct.

`TombstoneStores ` will contain the ordinal if we had offlined this ordinal. Then there will be no graceful upgrade at all.

pkg/manager/member/tikv_upgrader.go
xiaojingchen: This method does not guarantee that all region leaders can be transferred. Setting timeout is just to avoid long-term blocking, and it is not necessary to set it too long. I believe that 3 minutes is enough for most of the region leaders to transfer.

pkg/manager/member/tikv_upgrader.go
xiaojingchen: That's Right,the `TombstoneStores` saves all history records.

xiaojingchen: I will fix it now

pkg/manager/member/tikv_upgrader.go
xiaojingchen: the error of the method is not needed to throw out.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,154 @@
...
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap.com/v1alpha1"
+	"github.com/pingcap/tidb-operator/pkg/controller"
 	apps "k8s.io/api/apps/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
 )
 
-type tikvUpgrader struct{}
+const (
+	EvictLeaderBeginTime = "evictLeaderBeginTime"
```
tennix: Add a comment for exported variable or constant.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,154 @@
...
 			return err
 		}
 		newSet.Spec.Template.Spec = *podSpec
-	} else {
-		tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
+		return nil
+	}
+
+	if !tc.Status.TiKV.Synced {
+		return fmt.Errorf("Tidbcluster: [%s/%s]'s tikv status sync failed,can not to be upgraded", ns, tcName)
```
tennix: Add space after the comma.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,154 @@
...
+				setUpgradePartition(newSet, upgradeOrdinal)
+				return nil
+			}
+
+			if !evicting {
+				err = tku.pdControl.GetPDClient(tc).BeginEvictLeader(storeID)
+				if err != nil {
+					return err
+				}
+				// updete upgradePod
```
tennix: Mark pod for leader eviction. It's better to abstract the following code as a function to make it more understandable.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,154 @@
...
+	for _, store := range tc.Status.TiKV.Stores {
+		if store.PodName == upgradePodName {
+			storeID, err := strconv.ParseUint(store.ID, 10, 64)
+			if err != nil {
+				return err
+			}
+			_, evicting := upgradePod.Annotations[EvictLeaderBeginTime]
+
+			if tku.readyToUpgrade(upgradePod, store) {
+				err = tku.pdControl.GetPDClient(tc).EndEvictLeader(storeID)
```
tennix: The upgrade will not take effect immediately, the `setUpgradePartition` will update StatefulSet only after the end of the current loop. Before the pod is upgraded, the leaders may be migrated to the pod again.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -14,26 +14,196 @@
...
+		tku.endEvictLeader(tc, 0)
+		return nil
+	}
+
+	if !tc.TiKVAllPodsStarted() {
+		return controller.RequeueErrorf("tidbcluster: [%s/%s]'s tikv pods are not all created", ns, tcName)
+	}
+	for i := tc.Status.TiKV.StatefulSet.Replicas; i > tc.Status.TiKV.StatefulSet.CurrentReplicas; i-- {
+		store := tku.getStoreByOrdinal(tc, i-1)
+		if tc.Status.TiKV.StatefulSet.UpdatedReplicas != tc.Status.TiKV.StatefulSet.Replicas-tc.Status.TiKV.StatefulSet.CurrentReplicas {
```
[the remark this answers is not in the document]
xiaojingchen: it is to ensure the upgraded pod has created

pkg/manager/member/tikv_upgrader_test.go
```diff
@@ -131,5 +538,65 @@ func newTidbClusterForTiKVUpgrader() *v1alpha1.TidbCluster {
...
+				Labels:    l,
+			},
+			Status: corev1.PodStatus{Phase: corev1.PodRunning},
+		})
+	}
+	return pods
+}
+
+func getTiKVPodName(i int) string {
+	return fmt.Sprintf("%s-%d", controller.TiKVMemberName("upgrader"), i)
```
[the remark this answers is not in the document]
xiaojingchen: the method of https://github.com/pingcap/tidb-operator/pull/104/files#diff-969d53a49525b7fa111454dd3e70b96fR202 needs a TidbCluster object as param, but in test file haven't a TidbCluster object some time, so use a simple method to replace it

PR #108, opened 2018-09-30
backup up to gcp

weekface:
This PR add backup to gcp support to operator.

### DIFF
--- a/charts/tidb-cluster/templates/fullbackup-cronjob.yaml
+++ b/charts/tidb-cluster/templates/fullbackup-cronjob.yaml
@@ -31,7 +31,7 @@ spec:
               mountPath: "/savepoint-dir"
           containers:
           - name: fullbackup
-            image: pingcap/tidb-enterprise-tools:latest
+            image: {{ .Values.fullbackup.mydumperImage }}
             command:
             - /bin/sh
             - -c
@@ -43,23 +43,39 @@ spec:
               mkdir -p /data/${dirname}/
               cp /savepoint-dir/savepoint /data/${dirname}/
 
-              /mydumper \
+              /usr/bin/mydumper \
                 --outputdir=/data/${dirname} \
                 --host=`eval echo '${'$host'}'` \
                 --port=4000 \
                 --user={{ .Values.fullbackup.user }} \
                 --password={{ .Values.fullbackup.password }} \
                 {{ .Values.fullbackup.options }}
+
+            {{- if .Values.fullbackup.gcp }}
+              /usr/local/bin/uploader \
+                --cloud=gcp \
+                --bucket={{ .Values.fullbackup.gcp.bucket }} \
+                --backup-dir=/data/${dirname}
+            {{- end }}
             volumeMounts:
             - name: savepoint-dir
               mountPath: "/savepoint-dir"
             - name: data
               mountPath: "/data"
+          {{- if .Values.fullbackup.gcp }}
+            - name: gcp-credentials
+              mountPath: "/gcp"
+              readOnly: true
+          {{- end }}
             ENV:
             - name: MY_POD_NAME
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+          {{- if .Values.fullbackup.gcp }}
+            - name: GOOGLE_APPLICATION_CREDENTIALS
+              value: /gcp/credentials.json
+          {{- end }}
           restartPolicy: OnFailure
           volumes:
           - name: savepoint-dir
@@ -67,4 +83,10 @@ spec:
           - name: data
             persistentVolumeClaim:
               claimName: {{ .Values.clusterName }}-fullbackup
+        {{- if .Values.fullbackup.gcp }}
+          - name: gcp-credentials
+            secret:
+              secretName: {{ .Values.clusterName }}-fullbackup-secret
+        {{- end }}
+
 {{- end }}
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -174,7 +174,8 @@ monitor:
 fullbackup:
   create: false
   binlogImage: pingcap/tidb-binlog:latest
-  mydumperImage: pingcap/tidb-enterprise-tools:latest
+  # https://github.com/tennix/tidb-cloud-backup
+  mydumperImage: tennix/tidb-cloud-backup:latest
   # storageClassName is a StorageClass provides a way for administrators to describe the "classes" of storage they offer.
   # different classes might map to quality-of-service levels, or to backup policies,
   # or to arbitrary policies determined by the cluster administrators.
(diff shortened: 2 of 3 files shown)


weekface:
`snapshot` argument?

You can add custom arguments at here: https://github.com/pingcap/tidb-operator/pull/108/files#diff-b8571ebe161ca414e35a16dc83482915R195

charts/tidb-cluster/values.yaml
```diff
@@ -174,7 +174,8 @@ monitor:
 fullbackup:
   create: false
   binlogImage: pingcap/tidb-binlog:latest
-  mydumperImage: pingcap/tidb-enterprise-tools:latest
+  # https://github.com/tennix/tidb-cloud-backup
+  mydumperImage: tennix/tidb-cloud-backup:latest
```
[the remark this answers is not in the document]
weekface: @tennix

gregwebs:
mydumper doesn't give a consistent snapshot of tidb unless [tidb-mvc-snapshot](https://github.com/pingcap/mydumper/blob/master/mydumper.c#L167) is given. I think we are also supposed to turn on some of the lock reducing options since we are using the snapshot already.

gregwebs:
Assuming this is the existing mydumper, you should discuss with the tools team to make sure this is being ran correctly. Alternatively you can ask them to build our [new version of mydumper](https://github.com/morgo/mydumper/tree/tidb-support) that does the right thing without any arguments. Maybe that is what this image is using?

tennix:
I'll update tidb-cloud-backup image to use https://github.com/pingcap/mydumper instead of the official mydumper.

gregwebs:
Tha't fine. It's think its even better to use our [new mydumper](https://github.com/morgo/mydumper/tree/tidb-support). You can ask @morgo about it. We are just waiting to get it officially upstreamed so that we no longer have a fork.

charts/tidb-cluster/templates/fullbackup-cronjob.yaml
```diff
@@ -43,28 +43,50 @@ spec:
               mkdir -p /data/${dirname}/
               cp /savepoint-dir/savepoint /data/${dirname}/
 
-              /mydumper \
+              /usr/bin/mydumper \
```
tennix: Mydumper and uploader are already in `$PATH`, so it's unnecessary to specify the full path.

charts/tidb-cluster/values.yaml
```diff
@@ -191,10 +192,17 @@ fullbackup:
...
-  options: "--rows=10000"
+  options: "--chunk-filesize=100"
   # user must set tidb cluster user and password
   user: root
   password:
+  # backup to gcp
+  gcp: {}
+    # base64 format of the service account json file data
+    # https://cloud.google.com/docs/authentication/production#obtaining_and_providing_service_account_credentials_manually
+    # and then: cat credentials.json | base64
```
tennix: On Linux `base64` without any arguments will result in a multiline encoding string which is not allowed in `Secret`.
As stated by Kubernetes documentation https://kubernetes.io/docs/concepts/configuration/secret/
> Encoding Note: The serialized JSON and YAML values of secret data are encoded as base64 strings. Newlines are not valid within these strings and must be omitted. When using the base64 utility on Darwin/macOS users should avoid using the -b option to split long lines. Conversely Linux users should add the option -w 0 to base64 commands or the pipeline base64 | tr -d '\n' if -w option is not available.

charts/tidb-cluster/templates/fullbackup-cronjob.yaml
```diff
@@ -43,28 +43,50 @@ spec:
...
+              mydumper \
                 --outputdir=/data/${dirname} \
                 --host=`eval echo '${'$host'}'` \
                 --port=4000 \
                 --user={{ .Values.fullbackup.user }} \
                 --password={{ .Values.fullbackup.password }} \
                 {{ .Values.fullbackup.options }}
+
+            {{- if .Values.fullbackup.gcp }}
+              /usr/local/bin/uploader \
```
tennix: This also doesn't need to specify the full path.

images/tidb-operator-e2e/tidb-cluster-values.yaml
```diff
@@ -5,9 +5,12 @@
...
   create: true
 
-clusterName: demo-cluster
+# clusterName is the TiDB cluster name, it is required and should be unique
+# if multiple clusters are deployed in the same namespace.
+clusterName: demo
 
-# nodeSelectorRequired: true
+# timezone is the default system timzone for TiDB
+timezone: Asia/Shanghai
```
gregwebs: Is this change supposed to be in this PR?

weekface: No, it just keep this file sync with `charts/tidb-cluster/values.yaml`

gregwebs: Its odd to pick a default. Can we use the timezone of the host machine or the helm user as the default?

weekface: @tennix

tennix: We can't get the host machine's timezone when installing cluster.

gregwebs: Can we leave it empty and [require that it be set](https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md#know-your-template-functions)?

tennix: The default value is just an example and users can change it when necessary just like the storage class field. I'm OK with other values or leave it empty as a required value.

gregwebs: I think we need to assume that the values won't be changed unless explicitly called out either in the tutorial, or as a field in the MarketPlace, etc. I agree that if we do give it a value, UTC is best.

charts/tidb-cluster/values.yaml
```diff
@@ -194,10 +195,17 @@ fullbackup:
   # https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#starting-deadline
   startingDeadlineSeconds: 3600
   # https://github.com/maxbube/mydumper/blob/master/docs/mydumper_usage.rst#options
-  options: "--rows=10000"
+  options: "--chunk-filesize=100"
   # user must set tidb cluster user and password
   user: root
```
gregwebs: Can you change this from `root` to `backup`? Right now it seems like using backup exposes the credentials? Can we move the password into the secret system?

gregwebs:
We are going to start distributing the new mydumper merged by the tools team (today?). So lets just make sure to use that one (which will ensure a consistent backup), but that could be in a follow-up PR.

tennix:
Yes, the mydumper image update can be done separately.

PR #111, opened 2018-10-09
configure the operator ServiceAccount

gregwebs:
This allows the serviceAccount to be configured from values.yaml. I am not sure if this should be directly connected to rbac.create or not.


tennix:
When RBAC is enabled, we should create a ServiceAccount for controller-manager. I don't think it's necessary to allow users to specify the ServiceAccount for controller-manager, since we've already created the ServiceAccount named `tidb-controller-manager` when installing tidb-operator.

gregwebs:
I added a commit that also makes the ServiceAccount creation optional (even if rbac is true).

xiaojingchen:
@gregwebs  we need to change images/tidb-operator-values.yaml also when the values.yaml changed, otherwise it can not pass e2e test , cause e2e test use the file to replace the charts/tidb-operator/values.yaml.

charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -107,15 +107,17 @@ roleRef:
   name: tidb-controller-manager
   apiGroup: rbac.authorization.k8s.io
 {{- end }}
+{{- if .Values.controllerManager.ServiceAccount }}
 ---
 kind: ServiceAccount
 apiVersion: v1
 metadata:
-  name: tidb-controller-manager
+  name: {{ .Values.controllerManager.ServiceAccount }}
```
tennix: It's `serviceAccount` not `ServiceAccount`.

PR #121, opened 2018-10-15
bump tidb to v2.0.7 & allow user specified config files

tennix:
This PR bumps TiDB to v2.0.7 as the default TiDB version. And also allow users to specify config files in values.yaml. Normal users don't have to customize config files in values.yaml, so it's hidden there.

### DIFF
--- a/charts/tidb-cluster/templates/pd-configmap.yaml
+++ b/charts/tidb-cluster/templates/pd-configmap.yaml
@@ -112,6 +112,9 @@ data:
     echo "/pd-server ${ARGS}"
     exec /pd-server ${ARGS}
   config-file: |-
+    {{- if .Values.pd.config }}
+{{ .Values.pd.config | indent 4 }}
+    {{- else }}
     # PD Configuration.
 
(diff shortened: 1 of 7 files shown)


tennix:
It's easier to handle configuration files in values.yaml in a program with helm client. The program can load a configuration template in memory and render some customized values and then uses helm client to manage the release. So the program doesn't have to render the configuration into a file. See here https://github.com/helm/helm/blob/v2.11.0/cmd/helm/install.go#L289-L292 for an example.

tennix:
@onlymellb PTAL and update our CI environment to provide TiDB v2.0.7 and v2.1.0-rc.3 Docker images.

gregwebs:
Right. What I am suggesting is changing the else part of the below from a file inlined in the yaml to a .Files.Get. So it can still be over-ridden in values.yaml.

```
  config-file: |-
    {{- if .Values.tidb.config }}
{{ .Values.tidb.config | indent 4 }}
    {{- else }}
```

PR #122, opened 2018-10-15
use the parameterized ServiceAccount in monitor-job.yaml

gregwebs:
Looks like I missed this in my last PR.


tennix:
The Job doesn't require any special permissions. So the `default` service account in a namespace should work.

gregwebs:
great then

PR #123, opened 2018-10-16
Marketplace

gregwebs:
This is the guide that will be available to users for a manual MarketPlace install. There are two commits here. The second is all the charts, so you probably want to look at the first.

We could copy the charts that are already in tidb-operator, but we need to make sure we get a particular version.
So the workflow could be:

1) update repo
2) copy charts from a git tag that we made for a chart release 

Let me know what you think is best.


marketplace/gcp/tidb-operator-enterprise/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM gcr.io/cloud-marketplace-tools/k8s/deployer_helm/onbuild
```
tennix: add new line

marketplace/gcp/tidb-operator-enterprise/chart/tidb-mp/values.yaml
```diff
@@ -0,0 +1,51 @@
...
+  clusterName: db
+  rbac:
+    create: false
+  
+  fullbackup:
+    # Need to get the backup images pushed to GCR first.
+    create: false
+    password:
+
+  monitor:
```
tennix: How does this works if not running as a sidecar?

marketplace/gcp/tidb-operator-enterprise/chart/tidb-mp/requirements.yaml
```diff
@@ -0,0 +1,10 @@
+dependencies:
+- name: tidb-crd
+  version: 0.1.0
+  repository: ../../deps/tidb-crd
```
tennix: Is this a relative path? There seems no such directory

marketplace/gcp/tidb-operator-enterprise/chart/tidb-mp/templates/application.yml
```diff
@@ -0,0 +1,248 @@
...
+      kind: PersistentVolumeClaim
+
+metadata:
+  name: "{{ .Release.Name }}"
+  namespace: "{{ .Release.Namespace }}"
+  labels:
+    app.kubernetes.io/name: "{{ .Release.Name }}"
+  annotations:
+    marketplace.cloud.google.com/deploy-info: '{partner_id: "pingcap-public", product_id: "tidb-operator-enterprise", partner_name: "PingCAP"}'
+    kubernetes-engine.cloud.google.com/icon: |-
```
tennix: Could you update the logo icon? This icon looks weird on marketplace page.

tennix:
If we're only using a specific tag of tidb-operator, perhaps we can provide a document in this repo or somewhere else. And only contain tools and scripts in this repo to generate a deployer. When we release a new version, we can run the build step to build a deployer and send the deployer to GKE marketplace.
In this way, we can avoid storing the charts in multiple places in this repo.

tennix:
Seems the replicas can't be customized, are you going to add a document to tell users how to scale themselves?

gregwebs:
> If we're only using a specific tag of tidb-operator, perhaps we can provide a document in this repo or somewhere else. And only contain tools and scripts in this repo to generate a deployer

Yes, that's exactly what this is.

> When we release a new version, we can run the build step to build a deployer and send the deployer to GKE marketplace.

This suggested approach would be better for us if we could still expose the values.yml to the user. Although we are only requires to have them set the PARAMETERS for schema.yml, I think we have too many other configurations someone would want to change (as per your comment about replicas). We can probably figure out how to expose the values.yaml, but I don't know how to right now or if it is really a supported workflow. There is a downside in that it hides away the helm chart from the power user, so it is harder for them to understand what the configuration is doing.

marketplace/gcp/tidb-operator-enterprise/chart/tidb-mp/values.yaml
gregwebs: It is running as a sidecar in just the monitor pod. It just counts its own time right now, its not quite what we want yet.

gregwebs:
> What is the difference between charts and marketplace/gcp/tidb-operator-enterprise/chart

This puts together a wrapping application that is a one-click deploy in MarketPlace. The only difference between charts and marketplace/gcp/tidb-operator-enterprise/chart/tidb-mp/charts is one additional template file for the billing agent.

marketplace/gcp/tidb-operator-enterprise/chart/tidb-mp/values.yaml
tennix: Maybe we should consider a general approach to run sidecar containers in tidb-operator @weekface

tennix:
The chart files are duplicated. For now, I know you added billing agent in charts directory. In the future, if the charts can be copied unmodified to the marketplace, we should git ignore it.
I think advanced users may just use helm to install charts directly if we give a link to this repo. So we don't have to expose all the values to users.

marketplace/gcp/tidb-operator-enterprise/chart/tidb-mp/values.yaml
gregwebs: The problem with the `toYaml extraContainers` approach that I originally tried is that the extraContainers cannot be helm templated. However, that may work fine for other sidecar use cases.

gregwebs:
We can actually ignore the charts instead of commit them. We just need to make sure that we get a properly released version of the charts (git tag) and not just take master.

tennix:
The `TidbCluster` crd problem still exists, you need to use the latest code to regenerate this pr.

gregwebs:
okay, done now. It now only contains the installer chart. There are some chart tweaks to get pushed upstream into tidb-operator now.

PR #128, opened 2018-10-17
[DNM] fix pdclient connection unavailable bug

### LINKED ISSUE #127: pkg/controller PDControl's PDClient http connection cache bug
Current `GetPDClient` default return the PDClient that was cached
```
func (pdc *defaultPDControl) GetPDClient(tc *v1alpha1.TidbCluster) PDClient {
	pdc.mutex.Lock()
	defer pdc.mutex.Unlock()
	namespace := tc.GetNamespace()
	tcName := tc.GetName()
	key := pdClientKey(namespace, tcName)
	if _, ok := pdc.pdClients[key]; !ok {
		pdc.pdClients[key] = NewPDClient(pdClientURL(namespace, tcName), timeout)
	}
	return pdc.pdClients[key]
}
```
But in k8s the http connection may be unavailabe when upgrade and failover happen. than the cached pdclient call api failed will cause a long time of blocking.

xiaojingchen:
this PR fixes #127 

@weekface  @tennix @onlymellb @gregwebs PTAL


weekface:
Reopen it, because of #154

PR #131, opened 2018-10-18
remove noisy events

weekface:
- Remove noisy events: `LastHeartbeatTime` updated very frequently, if no other updates, don't emit an `Event`.
- Remove `-enable-tolerant` args:  https://github.com/pingcap/tidb-binlog/pull/366


pkg/controller/tidbcluster_control.go
```diff
@@ -97,6 +101,22 @@ func (rtc *realTidbClusterControl) recordTidbClusterEvent(verb string, tc *v1alp
 	}
 }
 
+func deepEqualExceptHeartbeatTime(new *v1alpha1.TidbClusterStatus, old *v1alpha1.TidbClusterStatus) bool {
```
tennix: ditto

pkg/controller/tidbcluster_control.go
```diff
@@ -51,7 +53,7 @@ func NewRealTidbClusterControl(cli versioned.Interface,
 	}
 }
 
-func (rtc *realTidbClusterControl) UpdateTidbCluster(tc *v1alpha1.TidbCluster) (*v1alpha1.TidbCluster, error) {
+func (rtc *realTidbClusterControl) UpdateTidbCluster(tc *v1alpha1.TidbCluster, new *v1alpha1.TidbClusterStatus, old *v1alpha1.TidbClusterStatus) (*v1alpha1.TidbCluster, error) {
```
tennix: Don't use `new` as variable name.

pkg/controller/tidbcluster_control.go
```diff
@@ -78,7 +80,9 @@ func (rtc *realTidbClusterControl) UpdateTidbCluster(tc *v1alpha1.TidbCluster) (
 
 		return updateErr
 	})
-	rtc.recordTidbClusterEvent("update", tc, err)
+	if !deepEqualExceptHeartbeatTime(new, old) {
```
tennix: `new` and `old` are pointers, this method actually modifies the status. Even it's not used anymore, you should avoid modification.

PR #135, opened 2018-10-19
Marketplace update

gregwebs:
There is one change here to make the template functions the same as "chart.name". Another change to add a field `ubbagent` to values.yaml

The rest are changes to help with configuring service account and permissions. The operator rbac section was changed because nodes, pv, and pvc require cluster-wide permission.


charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -1,68 +1,81 @@
...
   resources: ["tidbclusters"]
   verbs: ["*"]
+{{- end }}
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["persistentvolumes"]
+  verbs: ["get", "list", "watch", "patch","update"]
+- apiGroups: [""]
```
tennix: The `persistentvolumeclaims` should be in the `{{ if .Values.clusterScoped }}` block.

I thought this was wrong at first glance. `nodes` and `persistentvolumes` are cluster wide resources and users cannot access these resources if not giving cluster scope permission. But I believe tidb-operator cannot work correctly without these permissions now. So users still needs cluster admin permission in order to create clusterrole/clusterrolebinding. This is a requirement whether user specify `clusterScoped` or not.

charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -1,68 +1,81 @@
...
-  name: {{ .Release.Namespace }}:tidb-controller-manager
+  name: {{ .Release.Name }}:tidb-controller-manager
   apiGroup: rbac.authorization.k8s.io
-{{- else }}
+{{- if (not .Values.clusterScoped) }}
+---
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1beta1
 metadata:
   name: tidb-controller-manager
```
tennix: We should also add release name in role and rolebinding name.

PR #137, opened 2018-10-19
binglog remove enable tolerant config

xiaojingchen:
Binlog have remove the `enable tolerant` config in the PR: https://github.com/pingcap/tidb-binlog/pull/366 .
So the charts should  remove `enable tolerant`  too


weekface:
This pr had removed this argument: https://github.com/pingcap/tidb-operator/pull/131

PR #149, opened 2018-10-26
separation startup scripts and application config from yaml files

onlymellb:


charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -14,3 +14,11 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
+
+{{- define "tidb-cluster.utils.template" -}}
```
gregwebs: I would like to give the template helpers generic names if they are not specific to our chart. This makes it easier to re-use them in different charts.

tennix: Maybe `chart.configmap-template`?

onlymellb: Maybe use `helm-toolkit.utils.template` instead @gregwebs @tennix ?

onlymellb:
The pump/drainer startup script has been separated.

charts/tidb-cluster/templates/_helpers.tpl
tennix: ok, lgtm.

tennix:
For the configuration files, I think it's better to add the filename suffix for toml, yaml format.

tennix:
I've run `helm template charts/tidb-cluster --name=demo --namespace=tidb` and got the following errors:
```
Error: render error in "tidb-cluster/templates/tikv-configmap.yaml": template: tidb-cluster/templates/tikv-configmap.yaml:13:42: executing "tidb-cluster/templates/tikv-configmap.yaml" at <include "helm-toolki...>: error calling include: template: no template "helm-toolkit.utils.template" associated with template "gotpl"
```
Have you tested the templates?

PR #164, opened 2018-11-05
e2e: meta sync

weekface:
This pr adds the e2e specs to ensure meta informations are synced properly. 


tests/e2e/create.go
```diff
@@ -347,3 +352,111 @@ func reclaimPolicySynced(tc *v1alpha1.TidbCluster) (bool, error) {
...
+		switch component {
+		case label.PDLabelVal:
+			var memberID string
+			members, err := pdCli.GetMembers()
+			if err != nil {
+				logf(err.Error())
+				return false, nil
+			}
+			for _, member := range members.Members {
+				if member.Name == podName {
```
tennix: What if there is no member name matching pod name?

tests/e2e/create.go
```diff
@@ -347,3 +352,111 @@ func reclaimPolicySynced(tc *v1alpha1.TidbCluster) (bool, error) {
...
+		case label.TiKVLabelVal:
+			var storeID string
+			stores, err := pdCli.GetStores()
+			if err != nil {
+				logf(err.Error())
+				return false, nil
+			}
+			for _, store := range stores.Stores {
+				addr := store.Store.GetAddress()
+				if strings.Split(addr, ".")[0] == podName {
```
tennix: Also what if no matching?

tests/e2e/create.go
weekface: Then this case will fail: https://github.com/pingcap/tidb-operator/pull/164/files/b036f31f8548f392e2e94579a8cb896f05b0a9f4#diff-c1c833494c542d5b15d71a65d557b44bR402

tests/e2e/create.go
weekface: ditto

tests/e2e/create.go
```diff
@@ -347,3 +352,111 @@ func reclaimPolicySynced(tc *v1alpha1.TidbCluster) (bool, error) {
...
+			Expect(pod.Labels[label.StoreIDLabelKey]).To(Equal(storeID))
+		case label.TiDBLabelVal:
+			continue outerLoop
+		}
+
+		var pvcName string
+		for _, vol := range pod.Spec.Volumes {
+			if vol.PersistentVolumeClaim != nil {
+				pvcName = vol.PersistentVolumeClaim.ClaimName
+				break
```
xiaojingchen: if the pod have multiple pvcs?

weekface: currently, there is not multiple pvcs

tests/e2e/create.go
weekface: done

tests/e2e/create.go
weekface: done

PR #168, opened 2018-11-06
use {{ .Release.Name }} as cluster name

tennix:
GKE marketplace needs to set `{{ .Release.Name }}` to its managed resources labels. But we don't have that in tidb-cluster resources. Helm [recommends] use `{{ .Release.Name }}` as instance label. We've already use instance labels for pod int tidb-operator code
https://github.com/pingcap/tidb-operator/blob/4d0f6f955dd65d808ffdc4d1e05c0682934fef5e/pkg/label/label.go#L32-L34
To keep consistent, we should use release name for cluster name too.

### DIFF
--- a/charts/tidb-cluster/templates/NOTES.txt
+++ b/charts/tidb-cluster/templates/NOTES.txt
@@ -1,12 +1,12 @@
 1. Watch tidb-cluster up and running
-   watch kubectl get pods --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }} -o wide
+   watch kubectl get pods --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Release.Name }} -o wide
 2. List services in the tidb-cluster
-   kubectl get services --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }}
+   kubectl get services --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Release.Name }}
 3. Access tidb-cluster using MySQL client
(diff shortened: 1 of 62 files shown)


charts/tidb-cluster/values.yaml
```diff
@@ -7,10 +7,6 @@
 rbac:
   create: true
 
-# clusterName is the TiDB cluster name, it is required and should be unique
-# if multiple clusters are deployed in the same namespace.
-clusterName: demo
```
weekface: change `images/tidb-operator-e2e/values.yaml` also.

charts/tidb-cluster/templates/NOTES.txt
```diff
@@ -1,12 +1,12 @@
 1. Watch tidb-cluster up and running
-   watch kubectl get pods --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }} -o wide
+   watch kubectl get pods --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Release.Name }} -o wide
 2. List services in the tidb-cluster
-   kubectl get services --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }}
+   kubectl get services --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Release.Name }}
 3. Access tidb-cluster using MySQL client
-   kubectl port-forward -n {{ .Release.Namespace }} svc/{{ .Values.clusterName }}-tidb 4000:4000 &
```
gregwebs: I think this should stay as clustername as a ux enhancement. Marketplace suggests a release name of pingcap-tidb-operator-1

PR #169, opened 2018-11-07
add more unit tests and give priority to syncing reclaim policy 

weekface:
- add more unit tests to https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/tidbcluster/tidb_cluster_control.go, coverage from 29.8% to 94.7%
- give highest priority to syncing reclaim policy
- fix a Makefile bug
- fix potential panic
- fix a scheduler bug: when a deleted pod is recreated again, it should be rescheduled to the original node


pkg/scheduler/predicates/ha_test.go
```diff
@@ -252,6 +252,18 @@ func TestHAFilter(t *testing.T) {
 				g.Expect(getSortedNodeNames(nodes)).To(Equal([]string{"kube-node-1", "kube-node-2", "kube-node-3"}))
 			},
 		},
+		{
+			name:      "5 pods scheduled to 3 nodes(and then delete the oridinal 1 pod), the created pod ordinal 1 should rescheduled to node 3",
```
onlymellb: s/oridinal/ordina/g

onlymellb:
```suggestion
			name:      "5 pods scheduled to 3 nodes(and then delete the ordinal 1 pod), the created pod ordinal 1 should rescheduled to node 3",
```

onlymellb:
LGTM

PR #171, opened 2018-11-07
Set password for TiDB when first deployed

### LINKED ISSUE #167: Set TiDB password when creating TiDB cluster
By default, TiDB bootstraps with an empty root password which is unsafe since we don't support network policy yet (see #53). We can use a batchjob to set a password for TiDB. For reference, [here](https://github.com/helm/charts/tree/master/stable/mysql) is the MySQL chart which uses `Secret` for password.

tennix:
This PR closes #167. It uses a `BatchJob` to set TiDB password when first deployed. When the job is completed, users can use the password to access TiDB.

### DIFF
--- a/charts/tidb-cluster/templates/NOTES.txt
+++ b/charts/tidb-cluster/templates/NOTES.txt
@@ -3,8 +3,9 @@
 2. List services in the tidb-cluster
    kubectl get services --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }}
 3. Access tidb-cluster using MySQL client
+   PASSWORD=$(kubectl get secret -n {{ .Release.Namespace }} {{ .Values.clusterName }}-tidb -ojsonpath="{.data.password[*]}"| awk '{print $6}')
    kubectl port-forward -n {{ .Release.Namespace }} svc/{{ .Values.clusterName }}-tidb 4000:4000 &
-   mysql -h 127.0.0.1 -P 4000 -u root -D test
+   mysql -h 127.0.0.1 -P 4000 -u root -D test -p${PASSWORD}
 {{- if .Values.monitor.create }}
 4. View monitor dashboard for TiDB cluster
    kubectl port-forward -n {{ .Release.Namespace }} svc/{{ .Values.clusterName }}-grafana 3000:3000
--- a/charts/tidb-cluster/templates/NOTES.txt
+++ b/charts/tidb-cluster/templates/NOTES.txt
@@ -2,12 +2,16 @@
    watch kubectl get pods --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }} -o wide
 2. List services in the tidb-cluster
    kubectl get services --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }}
-3. Access tidb-cluster using MySQL client
-   PASSWORD=$(kubectl get secret -n {{ .Release.Namespace }} {{ .Values.clusterName }}-tidb -ojsonpath="{.data.password[*]}"| awk '{print $6}')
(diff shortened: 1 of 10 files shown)


charts/tidb-cluster/templates/NOTES.txt
```diff
@@ -3,8 +3,9 @@
 2. List services in the tidb-cluster
    kubectl get services --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Values.clusterName }}
 3. Access tidb-cluster using MySQL client
+   PASSWORD=$(kubectl get secret -n {{ .Release.Namespace }} {{ .Values.clusterName }}-tidb -ojsonpath="{.data.password[*]}"| awk '{print $6}')
```
gregwebs: add a space after `-o`. Would be nice to get rid of the awk also.

charts/tidb-cluster/templates/tidb-secret.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Values.clusterName }}
+    app.kubernetes.io/component: tidb
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+type: Opaque
+data:
+  {{- if .Values.tidb.password }}
+  password: {{ printf "SET PASSWORD FOR 'root'@'%%' = '%s' ; FLUSH PRIVILEGES;" .Values.tidb.password | b64enc }}
```
gregwebs: I would much prefer to have *just* the password as the secret.

gregwebs:
If we take the existing password as input then this can also function as a general password reset (particularly if the user that the password is set for is also a variable)

tennix:
If users remember the current password, then sure it works. But sometimes they just forget the current password, so this method won't work.

gregwebs:
> If users remember the current password, then sure it works. But sometimes they just forget the current password, so this method won't work.

Okay, this does seem like a different workflow. There are other considerations also: ideally we would have the network (service) turned off until the password is set. In that line of thought I am actually wondering if this should be ran as an init container (that will wait indefinitely for TiDB to be ready) to ensure that the password change job is scheduled.

gregwebs:
sorry, I guess initContainer doesn't make sense. But probably the operator should ensure that this runs.

tennix:
This is a one-time job, the chart will serve us well for this task. We don't have to embed this into tidb-operator Go code.

charts/tidb-cluster/templates/tidb-secret.yaml
tennix: You can see the reasons why I store the sql command here instead of only password.
https://github.com/pingcap/tidb-operator/pull/171/files#diff-7b1a419604c0fcc0219f5b547c56eb40R27

gregwebs:
One more thought on doing this the secure way: add the concept of a "ready" label that the service uses as a label selector. That label is added when this job is complete so that TiDB would not be available as an endpoint until after

tennix:
The password setter itself uses TiDB service, so we can't handle it this way unless we use TiDB pod IP, but this is unstable.

charts/tidb-cluster/templates/tidb-secret.yaml
gregwebs: The non-secret part could be a ConfigMap then. Encrypting the SQL means that editing this SQL requires accessing the password.

tennix: For other SQL initialization setups, I think we can store them in another ConfigMap instead of storing them in this Secret.

onlymellb:
LGTM

PR #178, opened 2018-11-09
Migrate to Go module

### LINKED ISSUE #156: Consider using Go 1.11 module to manage dependencies
Go 1.11 introduces [module](https://golang.org/cmd/go/#hdr-Download_and_install_packages_and_dependencies) which enables builtin support to manage dependencies. The [dep](https://github.com/golang/dep) is difficult to update dependencies when involves Kubernetes, particularly we've encountered this error several times `err (*pkgtree.NonCanonicalImportRoot); required by (root)`.
We've experimentally migrated other projects to Go modules and it works pretty well even involved kubernetes dependency. So it's time to migrate to Go modules.

tennix:
This PR closes #156. With Go module, we can remove the vendor directory. But it's still reserved for faster and stable compilation in CI and our laptop development environment.


gregwebs:
Lgtm

PR #185, opened 2018-11-19
add default value for tidb / tikv / pd log level

cwen0:
if `log-level` is empty in tikv config file, starting tikv will fail.  
In this pr, if the `log-level` is empty,  it will be set to `info`. 
@tennix @weekface @xiaojingchen PTAL

### DIFF
--- a/charts/tidb-cluster/templates/config/_pd-config.tpl
+++ b/charts/tidb-cluster/templates/config/_pd-config.tpl
@@ -26,7 +26,7 @@ cert-path = ""
 key-path = ""
 
 [log]
-level = "{{ .Values.pd.logLevel }}"
+level = "{{ .Values.pd.logLevel | default `info` }}"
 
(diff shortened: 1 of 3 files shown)


charts/tidb-cluster/templates/config/_pd-config.tpl
```diff
@@ -26,7 +26,7 @@ cert-path = ""
 key-path = ""
 
 [log]
-level = "{{ .Values.pd.logLevel }}"
+level = "{{ .Values.pd.logLevel | default `info` }}"
```
weekface: does `default 'info'` works?

tennix:
```suggestion
level = {{ .Values.pd.logLevel | default "info" | quote }}
```

cwen0:
@tennix @weekface PTAL again, thanks!

Issue #115, opened 2018-10-12
Scaling does not work with tidb:latest


morgo:
I deployed a GKE cluster using the tutorial, with one exception:

```
helm install ./charts/tidb-cluster -n tidb --namespace=tidb --set pd.storageClassName=pd-ssd,tikv.storageClassName=pd-ssd,pd.image=pingcap/pd:latest,tidb.image=pingcap/tidb:latest,tikv.image=pingcap/tikv:latest
```

This step no longer works:

```
helm upgrade tidb charts/tidb-cluster --set pd.storageClassName=pd-ssd,tikv.storageClassName=pd-ssd,tikv.replicas=5
```

The nodes are in `CrashLoopBackOff` status.

morgo:
It looks like pd is failing.  I tried scaling just tidb:

```
helm upgrade tidb charts/tidb-cluster --set tidb.replicas=4

tocker@cloudshell:~/tidb-operator (morgan-test-213703)$ kubectl get po -n tidb
NAME                              READY     STATUS             RESTARTS   AGE
demo-monitor-865db5758b-t99g2     2/2       Running            0          44m
demo-monitor-configurator-vnxkk   0/1       Completed          0          44m
demo-pd-0                         1/1       Running            0          44m
demo-pd-1                         1/1       Running            0          43m
demo-pd-2                         0/1       CrashLoopBackOff   4          2m
demo-tidb-0                       1/1       Running            0          42m
demo-tidb-1                       1/1       Running            0          42m
demo-tidb-2                       1/1       Running            0          2m
demo-tidb-3                       1/1       Running            0          2m
demo-tikv-0                       2/2       Running            0          43m
demo-tikv-1                       2/2       Running            0          43m
demo-tikv-2                       2/2       Running            0          43m

```

morgo:
Stack driver is showing:
```
[...]

PR #187, opened 2018-11-20
add user guide

### LINKED ISSUE #161: User guide
For GCP MarketPlace we should have a guide following these guidelines:
https://cloud.google.com/marketplace/docs/partners/kubernetes-solutions/create-app-package#user-guide
Currently our MarketPlace guide just explains how to do a CLI install.

I think most of the what we are suppose to have is general documentation that we should have which applies to any tidb-operator deployment.

tennix:
This PR adds a general user guide for tidb-operator.  It closes #161


docs/user-guide.md
```diff
@@ -0,0 +1,225 @@
...
+
+```shell
+$ sudo mkdir -p /mnt/disks/disk0
+$ sudo mkfs.ext4 /dev/nvme0n1
+$ sudo mount -t ext4 -o nodelalloc /dev/nvme0n1 /mnt/disks/disk0
+```
+
+To auto-mount disks when your operating system is booted, you should edit `/etc/fstab` to include these mounting info.
+
+After mounting all data disks on Kubernetes nodes, you can deploy [local-volume-provisioner](https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume) to automatically provision the mounted disks as Local PersistentVolumes.
```
weekface: The user must use local-volume-provisioner  v2.2.0 or greater for `RetainPolicy` support.

docs/user-guide.md
```diff
@@ -0,0 +1,225 @@
...
+
+```shell
+$ releaseName="tidb-cluster"
+$ namespace="tidb"
+$ clusterName="demo" # Make sure this is the same as variable defined in charts/tidb-cluster/values.yaml
+```
+
+### Deploy TiDB cluster
+
+After TiDB Operator and Helm is deployed correctly, TiDB cluster can be deployed using following command:
```
weekface:
```suggestion
After TiDB Operator and Helm are deployed correctly, TiDB cluster can be deployed using the following command:
```

docs/user-guide.md
```diff
@@ -0,0 +1,225 @@
...
+To destroy TiDB cluster, run the following command:
+
+```shell
+$ helm delete ${releaseName} --purge
+```
+
+The above command only delete the running pods, the data is persistent. If you do not need the data anymore, you can run the following command to clean the data:
+
+```shell
+$ kubectl delete pvc -n ${namespace} -l app.kubernetes.io/instance=${releaseName}
```
weekface: use these labels instead:

```
app.kubernetes.io/instance
app.kubernetes.io/managed-by
```

gregwebs:
This looks great! I will review it some more, but just wanted to add a note on structure.

I would like to have an operations guide that assumes everything is installed and working. Its odd for someone using GKE to see a section on local volume provisioner. Can we split this user guide to installation and operation?
Most installation details (in particular the local volume provisioner here) are particular to the K8s setup. Many are being covered in the tutorials and not present in this guide.

tennix:
Users on GKE also need to deploy local volume provisioner if they want to use local volume, but the setup is a bit different. I'll split this document into two files.

docs/user-guide.md
tennix: The following command uses the manifests we provided to deploy local volume provisioner. So it's already >= v2.2.0

docs/operation-guide.md
```diff
@@ -0,0 +1,129 @@
...
+
+    * Using LoadBalancer
+
+        When you set `tidb.service.type` to `LoadBalancer` and the underlining Kubernetes support LoadBalancer, then a LoadBalancer will be created for TiDB service. You can access it via the external IP with port `4000`. Some cloud platforms support internal load balancer via service annotations, for example you can add annotation `cloud.google.com/load-balancer-type: Internal` to `tidb.service.annotations` to create an internal load balancer for TiDB on GKE.
+
+    * Using NodePort
+
+        You can access TiDB via any node's IP with tidb service node port. The node port is the port after `4000`, usually greater than `30000`.
+
+## Scale TiDB cluster
```
gregwebs: We need to add a note that all changes should be made to values.yaml rather than using `-set` as per #115.

docs/operation-guide.md
```diff
@@ -0,0 +1,129 @@
...
+
+If TiDB cluster is running on GKE, the backup data can be uploaded to GCS bucket. A bucket name and base64 encoded service account credential that has bucket read/write access must be provided. The comments in charts/tidb-cluster/values.yaml is self-explanatory for GCP backup.
+
+### Incremental backup
+
+To enable incremental backup, set `binlog.pump.create` and `binlog.drainer.create` to `true`. By default the incremental backup data is stored in protobuffer format in a PV. You can change `binlog.drainer.destDBType` from `pb` to `mysql` or `kafka` and configure the corresponding downstream.
+
+## Restore
+
+Currently, tidb-operator only supports restoring from full backup in GCS bucket. The `restore` section in charts/tidb-cluster/values.yaml should have enough comments as document.
```
gregwebs: I don't think the restore workflow makes sense. The helm chart is for persistent objects, but restore is a one-off operation. There could be multiple restores of different databases or different restores at different times, and I don't know how `helm upgrade` will handle it. Can we separate out the restore operation from the helm chart?

gregwebs: The restore section itself documents what fields can be changed, but not any of the workflow concerns.

docs/operation-guide.md
```diff
@@ -0,0 +1,129 @@
...
+        You can access TiDB via any node's IP with tidb service node port. The node port is the port after `4000`, usually greater than `30000`.
+
+## Scale TiDB cluster
+
+To scale TiDB cluster, just modify the `replicas` of PD, TiKV and TiDB. And then run the following command:
+
+```shell
+$ helm upgrade ${releaseName} charts/tidb-cluster
+```
+
```
gregwebs: How about changing cpu/memory?

gregwebs:
LGTM. I have some comments, but I would also be happy to merge and address them afterwards.

docs/operation-guide.md
tennix: For local volume, changing cpu/memory may cause pod pending if the node doesn't have enough resouce.

docs/operation-guide.md
```diff
@@ -0,0 +1,135 @@
...
+        When you set `tidb.service.type` to `LoadBalancer` and the underlining Kubernetes support LoadBalancer, then a LoadBalancer will be created for TiDB service. You can access it via the external IP with port `4000`. Some cloud platforms support internal load balancer via service annotations, for example you can add annotation `cloud.google.com/load-balancer-type: Internal` to `tidb.service.annotations` to create an internal load balancer for TiDB on GKE.
+
+    * Using NodePort
+
+        You can access TiDB via any node's IP with tidb service node port. The node port is the port after `4000`, usually greater than `30000`.
+
+## Scale TiDB cluster
+
+TiDB Operator has full support of horizontal scaling. But for vertical scaling, if you're using local volumes for PD and TiKV, then scaling up may cause pod pending if the node doesn't have enough resources. So it's not recommended to do vertical scaling.
```
gregwebs: What about vertical scale-down?

gregwebs: What about TiDB SQL? I should be able to scale those vertically?

I created a github issue focused on TiKV/PD here: https://github.com/pingcap/tidb-operator/issues/191

tennix: I've added the recommendations for TiDB cluster deployment.

tennix:
I think we should merge this PR and fix other issues in following PRs.

gregwebs:
Yes, good to merge

docs/operation-guide.md
gregwebs: A lot more thorough now, thanks! I still don't know from the guide if I can safely vertical scale TIDB SQL.

PR #188, opened 2018-11-20
keep scheduler name consistent as defined in values.yml

onlymellb:
The scheduler can't use scheduler policy configmap correctly when you change the schedulerName in values.yaml. This PR fix it.

### DIFF
--- a/charts/tidb-operator/templates/scheduler-deployment.yaml
+++ b/charts/tidb-operator/templates/scheduler-deployment.yaml
@@ -1,7 +1,7 @@
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
-  name: tidb-scheduler
+  name: {{ .Values.scheduler.schedulerName }}
   labels:
     app.kubernetes.io/name: {{ template "chart.name" . }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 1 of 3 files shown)


charts/tidb-operator/templates/scheduler-deployment.yaml
```diff
@@ -26,7 +26,7 @@ spec:
       serviceAccount: {{ .Values.scheduler.serviceAccount }}
     {{- end }}
       containers:
-      - name: tidb-scheduler
+      - name: {{ .Values.scheduler.schedulerName }}
```
tennix: Container name is used internally, using `tidb-scheduler` is handy for `kubectl exec` and `kubectl logs`. So it's unnecessary to change this.

onlymellb: I think it's better to keep the name consistent, and it's not necessary to fix the container name to executing `kubectl exec` and `kubectl logs`

tennix: If using `tidb-scheduler` as container name, the two container names are always `tidb-scheduler` and `kube-scheduler`. When debugging we can always use the same name without thinking what it is. Besides container name should reflect the actual image it uses other than using the installer specified name.

onlymellb: If you want to use `tidb-scheduler` as container name, you can keep the schedulerName as default. But if you want to change the default value then keeping the container name consistent with it makes sense.

PR #190, opened 2018-11-21
refine defer deleting pvc codes

weekface:
* refine defer deleting pvc codes, remove `wait.Poll`
* add more unit tests, `scaler.go` coverage from 82.8% to 95.2%

@tennix @xiaojingchen PTAL

### DIFF
--- a/pkg/manager/member/scaler_test.go
+++ b/pkg/manager/member/scaler_test.go
@@ -1,0 +1,139 @@
+// Copyright 2018 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


pkg/manager/member/scaler_test.go
```diff
@@ -0,0 +1,139 @@
...
+			pvcIndexer.Add(test.pvc)
+		}
+		if test.deletePVCErr {
+			pvcControl.SetDeletePVCError(fmt.Errorf("delete pvc failed"), 0)
+		}
+
+		err := gs.deleteAllDeferDeletingPVC(tc, setName, test.memberType, test.from, test.to)
+		test.expectFn(g, err)
+	}
+	tests := []testcase{
```
xiaojingchen: these cases lack of  normal case.

tennix:
`deleteAllDeferDeletingPVC` can't distinguish the different `continue` cases. I think we should add a helper return value for unit testing.

PR #197, opened 2018-11-26
fix pd_control: not rely on error text

weekface:
- just return an error if delete store failed
- use `GetEvictLeaderSchedulers` instead

@tennix @xiaojingchen PTAL

### DIFF
--- a/pkg/controller/pd_control.go
+++ b/pkg/controller/pd_control.go
@@ -397,6 +397,13 @@ func (pc *pdClient) BeginEvictLeader(storeID uint64) error {
 		return nil
 	}
 
+	// pd will return an error with the body contains "scheduler existed" if the scheduler already exists
+	// this is not the standard response.
+	// so these lines are just a workaround for now:
+	//   - make a new request to get all schedulers
+	//   - return nil if the scheduler already exists
+	//
+	// when PD returns standard json response, we should get rid of this verbose code.
 	evictLeaderSchedulers, err := pc.GetEvictLeaderSchedulers()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/controller/pd_control.go
```diff
@@ -407,15 +396,24 @@ func (pc *pdClient) BeginEvictLeader(storeID uint64) error {
 	if res.StatusCode == http.StatusOK {
 		return nil
 	}
-	err2 := readErrorBody(res.Body)
-	if strings.Contains(err2.Error(), schedulerExisted) {
-		return nil
+
+	evictLeaderSchedulers, err := pc.GetEvictLeaderSchedulers()
```
xiaojingchen: this should be exec before create evict leader, if the scheduler leader exists, then not need to post sheduler leader

pkg/controller/pd_control.go
```diff
@@ -428,11 +426,19 @@ func (pc *pdClient) EndEvictLeader(storeID uint64) error {
 	if res.StatusCode == http.StatusOK || res.StatusCode == http.StatusNotFound {
 		return nil
 	}
+
 	err2 := readErrorBody(res.Body)
-	if strings.Contains(err2.Error(), schedulerNotFound) {
-		return nil
+	evictLeaderSchedulers, err := pc.GetEvictLeaderSchedulers()
```
xiaojingchen: ditto

pkg/controller/pd_control.go
```diff
@@ -328,13 +324,6 @@ func (pc *pdClient) DeleteStore(storeID uint64) error {
 		return err
 	}
 
-	// FIXME: We should not rely on error text
-	// Remove a tombstone store should returns "store has been removed"
-	bodyStr := string(body)
-	if strings.Contains(bodyStr, "store has been removed") {
-		return nil
-	}
-
```
gregwebs: PD now actually returns a structured json response with an error code and an error code as a header.

weekface: Yes, but the older version don't have the error code and header.

Return the error is ok for this situation. The `DeleteStore` function is reentrant.

PR #200, opened 2018-11-28
 modify tidb-operator charts to adapt k8s v1.12

onlymellb:
This PR makes itidb-operator running in k8s v1.12 correctly.


onlymellb:
Reference PR https://github.com/kubernetes/kubernetes/pull/60398

charts/tidb-operator/values.yaml
```diff
@@ -4,6 +4,9 @@
 # Also see rbac.create and controllerManager.serviceAccount
 clusterScoped: true
 
+# kubeVersion is the k8s cluster version
+kubeVersion: v1.10.2
```
gregwebs: It seems almost guaranteed that this will not be set properly: is there any way to automatically get this setting?

tennix: Yes, this version should be retrieved from current cluster. The variable `Capabilities.KubeVersion` is just for this purpose.
https://github.com/helm/helm/blob/master/docs/chart_template_guide/builtin_objects.md
> Capabilities.KubeVersion provides a way to look up the Kubernetes version. It has the following values: Major, Minor, GitVersion, GitCommit, GitTreeState, BuildDate, GoVersion, Compiler, and Platform.

tennix:
The 1.10 image sha256sum should be updated @onlymellb

onlymellb:
The dind k8s v1.10 image has been updated.

PR #201, opened 2018-11-28
Orphan pods cleaner

weekface:
- add orphan pods cleaner: delete the pod if its pvc is not exist
- add unit tests to orphan pods cleaner
- only delete one defer deleting pvc when scale out
- e2e: set upgrade timeout duration to 10 minutes(upgrading all 5 tikvs may need more than 5 minutes)

@tennix @xiaojingchen PTAL

### DIFF
--- a/pkg/manager/member/orphan_pods_cleaner.go
+++ b/pkg/manager/member/orphan_pods_cleaner.go
@@ -1,0 +1,109 @@
+// Copyright 2018 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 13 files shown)


pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -0,0 +1,109 @@
...
+			continue
+		}
+
+		_, err := opc.pvcLister.PersistentVolumeClaims(ns).Get(pvcName)
+		if err == nil {
+			skipReason[podName] = skipReasonOrphanPodsCleanerPVCIsFound
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			return skipReason, err
```
xiaojingchen: why not use `continue` here?

weekface: If there is an error, return directly.
Is this reasonable?

PR #206, opened 2018-11-30
[WIP] fix grafana config

tennix:
* mount monitor configmap to grafana container
* upgrade grafana to 4.6.5
* fix grafana configuration behind reverse proxy
* make grafana deployment optional
* add e2e test for monitor


tennix:
The Grafana docker startup script run.sh was misconfigured to `chown` configuration directory which in Kubernetes is mounted as read only volume. See here https://github.com/grafana/grafana-docker/issues/51
This was fixed about a year ago, but latest 4.6.5 docker image which was released 17 days ago still has this problem. So the configmap is not mounted and environment variables are used instead.

PR #227, opened 2018-12-11
separate backup/restore to another chart

### LINKED ISSUE #214: Separate backup/restore to another chart
The backup and restore lifecycle is different from tidb-cluster, we should separate to an individual chart.

weekface:
fixes #214 

- keep scheduled fullbackup in the tidb-cluster chart, named it scheduled-backup
- separate restore to a new chart
- add ad-hoc backup to the new chat

@tennix @xiaojingchen PTAL


docs/operation-guide.md
```diff
@@ -128,11 +134,16 @@ To create a full backup job, modify `fullbackup` section in `values.yaml` file.
...
-If TiDB cluster is running on GKE, the backup data can be uploaded to GCS bucket. A bucket name and base64 encoded service account credential that has bucket read/write access must be provided. The comments in `values.yaml` is self-explanatory for GCP backup.
+```shell
+$ releaseName="tidb-backup"
+$ namespace="tidb"
+$ helm install charts/tidb-backup --name=${releaseName} --namespace=${namespace}
+$ watch kubectl get po -n ${namespace} -l app.kubernetes.io/component=fullbackup -o wide
+```
 
-### Incremental backup
+> **Note:** You must set the fullbackup PV's [reclaim policy](https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy) to `Retain` to safe your backup data.
```
gregwebs: safe -> save

gregwebs:
## Shared configuration

clusterName is now duplicated across charts. I think the only way to solve this is to read in clusterName from a shared source?

The k8s namespace used by the original install is implicitly needed for this one. I think that we should be recording the namespace once and reading it back rather than having the user type it in every time (this is helpful outside this change as per recent github issues).


## Persistent vs. ad-hoc

A scheduled backup is a persistent object and it could make sense to stay in the existing helm chart if you assume a user wants just 1 scheduled backup for their cluster, which seems like a very common use case.

The binlog pump is designed to have just one with multiple downstream connectors? Or one pump for each connector?

We should still support ad-hoc (no schedule) backup, just as restores are ad-hoc. This is separate from a scheduled backup.


## Ad-Hoc workflow

I am still not entirely clear on the workflow of using helm for these ad-hoc operations. It seems like it would at least run into minor issues of for example the restore job name already existing the second time someone wants to do a restore. I think it would help to map out our use cases/scenarios.

docs/operation-guide.md
```diff
@@ -128,11 +134,16 @@ To create a full backup job, modify `fullbackup` section in `values.yaml` file.
...
 
-To enable incremental backup, set `binlog.pump.create` and `binlog.drainer.create` to `true`. By default the incremental backup data is stored in protobuffer format in a PV. You can change `binlog.drainer.destDBType` from `pb` to `mysql` or `kafka` and configure the corresponding downstream.
+If TiDB cluster is running on GKE, the backup data can be uploaded to GCS bucket. A bucket name and base64 encoded service account credential that has bucket read/write access must be provided. The comments in `values.yaml` is self-explanatory for GCP backup.
```
xiaojingchen: no need to restrict that `TiDB cluster is running on GKE`

xiaojingchen: no need to restrict that `TiDB cluster is running on GKE`

docs/operation-guide.md
weekface: here, I mean to keep the data **safe**, the pv is retained when the pvc is deleted.

docs/operation-guide.md
weekface: @tennix PTAL

weekface: @tennix PTAL

docs/operation-guide.md
gregwebs: oh, then it would be "to keep your backup data safe".

docs/operation-guide.md
tennix: Currently, the backup data can be uploaded to GCS bucket if the cluster is running on GKE. More public cloud storage support is planned in the future.

weekface:
> clusterName is now duplicated across charts. I think the only way to solve this is to read in clusterName from a shared source?

I think the user must know the shared source name even use a shared source. @gregwebs 

> The binlog pump is designed to have just one with multiple downstream connectors? Or one pump for each connector?

Here is the [new tidb-binlog document](https://pingcap.com/docs/tools/tidb-binlog-cluster/), one drainer one downstream connector, is this right? @WangXiangUSTC 

> We should still support ad-hoc (no schedule) backup, just as restores are ad-hoc. This is separate from a scheduled backup.

Do we need add an one time fullbackup job? @tennix

gregwebs:
Pump is designed as one cluster. So both pump and a scheduled backup seem like a single persistent object that could match the users workflow well by being in our existing helm chart. Its definitely still possible that someone will want to create multiple scheduled backup jobs though.

For the clusterName the idea is something like: have a file clusterName.txt that charts reference by symlink. But there is probably a more standard way of doing this with helm.

gregwebs:
"being in our existing helm chart". I mean stay in the tidb-cluster chart rather than moved here.

weekface:
> Pump is designed as one cluster. So both pump and a scheduled backup seem like a single persistent object that could match the users workflow well by being in our existing helm chart. Its definitely still possible that someone will want to create multiple scheduled backup jobs though.
> 
> For the clusterName the idea is something like: have a file clusterName.txt that charts reference by symlink. But there is probably a more standard way of doing this with helm.
> 

@tennix PTAL 
@gregwebs may be confused on why we should separate backup/restore to another chart

docs/operation-guide.md
xiaojingchen: the cluster can uploaded to GCS bucket only need it has bucket read/write access, no matter it running on GKE or any other places

gregwebs:
@weekface I am actually the one that originally requested the separation. I only requested restore (and ad-hoc backup) though.

Fundamentally our existing helm chart lifecycle is for singlular persistent objects and the request is just to separate out those that are not of that category regardless of their functionality.

weekface:
what is `ad-hoc backup` means? One-time full backup?

- keep scheduled fullbackup in the `tidb-cluster` chart
- separate restore to a new chart
- add one-time fullbackup to the new chat

Is this right?

weekface:
@gregwebs Is [this solution](https://github.com/pingcap/tidb-operator/pull/227#issuecomment-446499238) ok?

charts/tidb-backup/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+  gcp:
+    # credentialsData is base64 format of the service account json file data
+    # https://cloud.google.com/docs/authentication/production#obtaining_and_providing_service_account_credentials_manually
+    # and then: cat credentials.json | base64 | tr -d '\n'
+    # credentialsData: ""
+    # bucket: ""
+    # # srcDir is the directory of backup data in bucket
+    # srcDir: ""
+
+metaInstance: "{{ $labels.instance }}"
```
tennix: No Prometheus configured for this chart, so these variables are unnecessary.

gregwebs:
yes, I use ad-hoc to mean not scheduled.

> keep scheduled fullbackup in the tidb-cluster chart
separate restore to a new chart
add one-time fullbackup to the new chat

Yes!

> @gregwebs Is this solution ok?

It seems like this link doesn't work now

weekface:
@tennix @xiaojingchen @gregwebs PTAL

weekface:
@tennix @xiaojingchen @gregwebs PTAL. It needs at least 2 approving reviews.

Issue #228, opened 2018-12-11
Using multiple statefulsets for each component


tennix:
Currently, each component has one statefulset, and the statefulset controls the component's behavior. This works well for normal cases. But have some limitations:
* When using local volume, only horizontal scaling is guaranteed, vertical scaling may fail if there are not enough resources on the node.
* Users can't specify an exact number of pods in a given AZ. For example deploy 3 TiKV in AZ-1, and 2 TiKV in AZ-2.
* For on-premise deployment, k8s nodes may not have the same size. To make most of the resources, it's better to allow different profiles for a single cluster.
* No isolation between different tidb-servers, a big AP query may affect TP query. The worst case is that AP query cause a tidb-server OOM, and this tidb-server is also serving TP request.

When switched to multiple statefulsets, all the above problems can be resolved.

gregwebs:
The components in question are just PD and TiKV?

> For on-premise deployment, k8s nodes may not have the same size. To make most of the resources, it's better to allow different profiles for a single cluster.

Is this resolvable by using the various resource requests parameters to allow specifying a range?

> Users can't specify an exact number of pods in a given AZ. For example deploy 3 TiKV in AZ-1, and 2 TiKV in AZ-2.

Is this resolvable by telling the operator this layout and having the operator interact with the scheduler? Essentially by default make TiKV not deployable and then change labels to get them deployed where we want?


In my existing thinking about multiple stateful sets, it was a temporary condition to allow a scaling transition. However, it sounds like this is leaning closer to having a separate stateful set per pod? I think what we need is just a StatefulPod and then have group behavior manage by our operator. So it seems like a StatefulSet could become our StatefulPod? Its also worth considering researching how others have tackled this challenge and seeing if we can come up with a better abstraction.

tennix:
No, the components include PD/TiKV/TiDB. See the last issue I listed.

The resource limits and requests may both different in a single cluster for on-premise deployment.

It is doable by hacking the scheduler extender to deploy an exact number of pods in a given AZ. But considering the other issues I listed, multiple statefulsets would be a more elegant approach.

By using multiple statefulsets, we're not making it separate statefulset per pod. Though it's possible if users deploy like that. We're still relying on StatefulSet controller to manage pods. Pods are still grouped as a statefulset, but are split as we needed (Geo-location, Heterogeneous nodes, Isolation workloads).

gregwebs:
I like where this is going! I think we need a more detailed design doc around all the different scenarios.

[...]

PR #240, opened 2018-12-18
Proposal: add multiple statefulsets support to TiDB Operator

weekface:
fixes: #228 


docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+Figure 2:
+
+``` yaml
+tikv:
+  replicas: 3
+  image: pingcap/tikv:v2.1.0
+  …
+  extra:
```
gregwebs: This should be renamed since `extra` is a potentially common key name. `sets` would be better.

weekface: There are a `default` set and zero or more extra sets, so what about `extraSets` or `otherSets`

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,201 @@
...
+}
+
+type TiKVSpec struct {
+  Name            string           `json:"name"`
+  Replicas        int32            `json:"replicas"`
+  …
+}
+
+type TiKVStatus struct {
+StatefulSet     []apps.StatefulSetStatus     `json:"statefulSet,omitempty"`
```
xiaojingchen: StatefulSet -> StatefulSets

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,201 @@
...
+StatefulSet     *apps.StatefulSetStatus     `json:"statefulSet,omitempty"`
+  …
+}
+```
+
+Figure 4:
+
+``` go
+type TidbClusterSpec struct {
+  TiKV            []TiKVSpec            `json:"tikv,omitempty"`
```
xiaojingchen: `TiKV` is a unreasonable name when its type is an array.

docs/design-proposals/multiple-statefulsets.md
gregwebs: Yes, that is better. I wan't sure if we would support having just sets.

gregwebs:
I am still very concerned that we are creating a highly-manual UX that doesn't actually match the user's workflow. To scale vertically, I just want to tell the operator to scale vertically. Actually change the yaml to create multiple stateful sets isn't what I want to do.

The advantage seems to be that
1) It makes it apparent to me that I need to increase the cluster capacity
2) I can write programs that can use this to automatically do what I want

docs/design-proposals/multiple-statefulsets.md
tennix: Yeah, that is the problem we just discussed. Removing the default statefulset would impact all the other sets that inherit the property from the default set.

weekface: Just sets +1

```yaml
pd:
  - name: name-1
    replicas: 3
    ...
  - name: name-2
    replicas: 2
    ...
```

docs/design-proposals/multiple-statefulsets.md
weekface: Any suggestion?

weekface: @tennix @onlymellb @gregwebs Is this ok?

tennix: I prefer `TiKVSets`

weekface: `TiKVSets` is ok for me. Do we need rename `TiKVSpec` to `TiKVSetSpec`?

gregwebs:
What would be very helpful to understand how this is designed to be used is to alter the user guide according to the proposal. Or otherwise list out all the different scenarios this is designed for and how it will be used in them.

docs/design-proposals/multiple-statefulsets.md
weekface: @tennix @onlymellb  @xiaojingchen @gregwebs PTAL

weekface:
I made some changes based on @tennix @onlymellb @xiaojingchen 's suggestions: https://github.com/pingcap/tidb-operator/pull/240/commits/e608b0765095c4df82447903f065a6b25fbaa5c7

weekface:
According to K8s [api compatibility](https://github.com/kubernetes/community/blob/master/contributors/devel/api_changes.md#on-compatibility):

> An API change is considered compatible if it:
> 
> - adds new functionality that is not required for correct behavior (e.g., does not add a new required field)
> - does not change existing semantics, including:
> - the semantic meaning of default values and behavior
> - interpretation of existing API types, fields, and values
> - which fields are required and which are not
> - mutable fields do not become immutable
> - valid values do not become invalid
> - explicitly invalid values do not become valid
> 
> Put another way:
> 
> - Any API call (e.g. a structure POSTed to a REST endpoint) that succeeded before your change must succeed after your change.
> - Any API call that does not use your change must behave the same as it did before your change.
> - Any API call that uses your change must not cause problems (e.g. crash or degrade behavior) when issued against an API servers that do not include your change.
> - It must be possible to round-trip your change (convert to different API versions and back) with no loss of information.
> - Existing clients need not be aware of your change in order for them to continue to function as they did previously, even when your change is in use.
> - It must be possible to rollback to a previous version of API server that does not include your change and have no impact on API objects which do not use your change. API objects that use your change will be impacted in case of a rollback.
> - If your change does not meet these criteria, it is not considered compatible, and may break older clients, or result in newer clients causing undefined behavior. Such changes are generally disallowed, though exceptions have been made in extreme cases (e.g. security or obvious bugs).
> 

So add an extra field to existing may be a better solution. For example: change`values.yaml` from figure 1 to figure 2, and change the TidbCluster object from figure 3 to figure 4:

Figure 1:
``` yaml
tikv:
  replicas: 3
  image: pingcap/tikv:v2.1.0
  …
```

Figure 2:
``` yaml
tikv:
  replicas: 3
  image: pingcap/tikv:v2.1.0
  …

tikvs:
  - name: name-1
    replicas: 3
    image: pingcap/tikv:v2.1.0
    ...
  - name: name-2
    replicas: 2
    image: pingcap/tikv:v2.1.0
    ...
```

Figure 3:
``` go
type TidbClusterSpec struct {
  TiKV            TiKVSpec            `json:"tikv,omitempty"`
  …
}
```

Figure 4:
``` go
type TidbClusterSpec struct {
  TiKV            TiKVSpec            `json:"tikv,omitempty"`
  TiKVs           []TiKVSpec            `json:"tikvs,omitempty"`
  …
}
```

tennix:
How do we handle the status of `TidbCluster`, should we merge all the statefulsets status? And how do we handle the order of all the statefulsets, the first PD acts as the seed and it's different from the rest.

weekface:
> How do we handle the status of TidbCluster, should we merge all the statefulsets status?

No. Add a new attribute `SetStatuses`.

> And how do we handle the order of all the statefulsets, the first PD acts as the seed and it's different from the rest.

Yes, the `pd` is the first order. Others are ordered by their index.

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,417 @@
...
+
+type TiKVSpec struct {
+  Name            string           `json:"name"`
+  Replicas        int32            `json:"replicas"`
+  …
+}
+
+type TiKVStatus struct {
+  StatefulSet     *apps.StatefulSetStatus     `json:"statefulSet,omitempty"`
+  StatefulSets    []apps.StatefulSetStatus    `json:"statefulSets,omitempty"`
```
xiaojingchen: `StatefulSets` type change to map[string]apps.StatefulSetStatus

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,417 @@
...
+* For TiKV: no change
+* For TiDB: N+1 Services should be created, N client services and 1 main client service
+
+### Upgrade:
+
+With multiple statefulsets supported, the user can modify the spec in the `values.yaml` then the Operator will range over all the statefulsets and select only one statefulset to upgrade at a time. So the basic progress is the same with older design.
+
+The user can modify all the statefulsets, but we only support upgrade one Pod at a time.
+
+For example, the user can change the images from v2.1.0(Figure 5) to v2.2.0(Figure 6) to upgrade the cluster from v2.1.0 to v2.2.0:
```
onlymellb: How do you deal with situations where a user has modified the name of an existing statefulset?

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,417 @@
...
+    image: pingcap/tikv:v2.2.0
+    …
+  - name: name-2
+    replicas: 2
+    image: pingcap/tikv:v2.2.0
+    …
+```
+
+The operator does not guarantee the order of multiple statefulsets’ upgrade. If the user care about the order indeed. Just: modify one statefulset, wait it done, and then modify the other one.
+
```
onlymellb: I think we can have a default upgrade order, such as we can upgrade the main statefulset first, and then upgrade the statefulset in the order of statefulset slice?

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,417 @@
...
+    image: pingcap/tikv:v2.1.0
+    …
+  - name: name-2
+    replicas: 2
+    image: pingcap/tikv:v2.1.0
+    …
+```
+
+The operator does not guarantee the order of multiple statefulsets’ scale in/out. If the user care the order indeed. Just: modify one statefulset, wait it done, and then modify the other one.
+
```
onlymellb: Ditto

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,417 @@
...
+### Failover:
+
+When a PD/TiKV/TiDB peer is failure, the failover module will add a new peer to the failed statefulset. Others are the same with older design.
+
+### Vertical scaling
+
+To do vertical scaling, the user must:
+
+- Add a new statefulset in the `tikvs` section, using more `requests` or `limits` and wait it done. Add a new `name-3`, change from figure 9 to figure 10;
+- Scale down the old statefulset(`name-1`) in the `tikvs` section to zero and wait it done. See figure 11;
```
onlymellb: Do we need to wait for the new statefulset name-3 to start completely before scale down

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,417 @@
...
+      requests:
+        cpu: 8000m
+        memory: 8Gi
+        storage: 8Gi
+    …
+```
+
+### Multiple zones
+
+We can use defferent `nodeSelector`s in different statefulsets to deploy tidb to multiple zones, figure 13:
```
onlymellb:
```suggestion
We can use different `nodeSelector`s in different statefulsets to deploy tidb to multiple zones, figure 13:
```

docs/design-proposals/multiple-statefulsets.md
weekface: Yes.

docs/design-proposals/multiple-statefulsets.md
weekface: Add a new statefulset and keep the old if the user changes the name of an existing one.

docs/design-proposals/multiple-statefulsets.md
weekface: **The user** must ensure that.

docs/design-proposals/multiple-statefulsets.md
onlymellb: That would cause the actual running statefulset to be one more than the user configured, so that's fine?

gregwebs: I think we should reject that change.

weekface: what do we need to do to reject it?

docs/design-proposals/multiple-statefulsets.md
weekface: Done.

docs/design-proposals/multiple-statefulsets.md
weekface: Done.

docs/design-proposals/multiple-statefulsets.md
```diff
@@ -0,0 +1,417 @@
...
+The same as TidbCluster object, from Figure 3 to Figure 4:
+
+Figure 3:
+
+``` go
+type TidbClusterSpec struct {
+  TiKV            TiKVSpec            `json:"tikv,omitempty"`
+  …
+}
+
```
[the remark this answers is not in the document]
weekface: Done.

docs/design-proposals/multiple-statefulsets.md
gregwebs: Its worth noting the difference between what we want to do and what is convenient to do. I think it is not possible to run custom validation logic for a CRD on update. One approach would be to use a phase field to indicate the CRD is an error state.

weekface: The use change the `values.yaml` file, not the CRD object. The CRD object is updated by us with the help of `helm`.

What we should consider first is how do we modify the CRD if the use change the name in `values.yaml`.

weekface: We can use `admission webhooks ` to validate the CRD object.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#admission-webhooks

weekface: this is added to the document.

gregwebs: Isn't that just for CRD creation? We would also need update.

weekface: [Update](https://github.com/kubernetes/kubernetes/blob/37555e6d24c2f951c40660ea59a80fa251982005/pkg/apis/admissionregistration/types.go#L259) is supported.

weekface:
> * What if the `name` filed in `tikvs` are the same? How do we handle this case?

Validation will fail.  Addressed.


> * For API compatibility, do you mean we should use different `apiVersion` to access the two different form of resources

I mean the API changes in this proposal are compatible.


> * For the pd/tikv service, you mentioned there is no change. Do you mean that different statefulsets using the same headless and non-headless service as the main statefulset, or service per statefulset?

No.
* For PD: N+1 Services should be created, N headless services and 1 client service
* For TiKV: N headless services should be created

Addressed.

Issue #239, opened 2018-12-18
operator was blocked when delete a pd member that is not exist


xiaojingchen:
some times the operator call pd api to delete pd member, but the pd member is not found, then the  operator was blocked.
the reasonable process is to return success directly and continue to perform other operations.

the tidb-operator's  log:
```
E1218 09:43:44.323228       1 tidb_cluster_controller.go:239] TidbCluster: example/example, sync failed failed 500 to delete member 4458752632436092868: "etcdserver: member not found"
, requeuing
E1218 09:43:45.640748       1 tidb_cluster_controller.go:239] TidbCluster: example/example, sync failed failed 500 to delete member 4458752632436092868: "etcdserver: member not found"
, requeuing
E1218 09:43:48.241726       1 tidb_cluster_controller.go:239] TidbCluster: example/example, sync failed failed 500 to delete member 4458752632436092868: "etcdserver: member not found"
, requeuing
E1218 09:43:53.396093       1 tidb_cluster_controller.go:239] TidbCluster: example/example, sync failed failed 500 to delete member 4458752632436092868: "etcdserver: member not found"
, requeuing
E1218 09:44:03.668888       1 tidb_cluster_controller.go:239] TidbCluster: example/example, sync failed failed 500 to delete member 4458752632436092868: "etcdserver: member not found"
, requeuing
E1218 09:44:12.470007       1 tidb_cluster_controller.go:239] TidbCluster: example/example, sync failed failed 500 to delete member 4458752632436092868: "etcdserver: member not found"
```

xiaojingchen:
The reason is that the operator does not take into account the non-existent situation when call PD delete API.
Recommended that call a get member API before deleting it; if the member is not found return success directly.

the related code:
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/pd_control.go#L330
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/pd_control.go#L348
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/pd_control.go#L306

gregwebs:
Should PD give a different error code?

weekface:
This may be a solution. https://github.com/pingcap/tidb-operator/issues/238#issuecomment-448185522

PD don't return a 404 not found response if the member is not found when we are deleting it.
So we must get the member again to ensure it is not found if deleting failed. Here is an example:

gregwebs:
[...]

Issue #238, opened 2018-12-18
Parse pd member id failed


xiaojingchen:
when operator call the api of delete pd member, happen the error of `out of range` error ,
the error logs:
```
E1218 09:52:42.464090       1 tidb_cluster_controller.go:239] TidbCluster: example2/example2, sync failed strconv.Atoi: parsing "12891273174085095651": value out of range, requeuing
```

weekface:
https://github.com/pingcap/tidb-operator/blob/4c183422345d8fe9e5a450f3729339ee01684767/pkg/manager/member/pd_failover.go#L160

"12891273174085095651" is too big for `Int`, so we must use `strconv.ParseUint` instead of `strconv.Atoi`

https://github.com/pingcap/tidb-operator/blob/1bc6486d27ac323d73a9de00c7a1cd187d658442/pkg/controller/pd_control.go#L345

Get the member again to ensure it is not found if deleting failed. Here is an example: https://github.com/pingcap/tidb-operator/blob/1bc6486d27ac323d73a9de00c7a1cd187d658442/pkg/controller/pd_control.go#L400-L415

PR #245, opened 2018-12-21
fix pd control bugs

xiaojingchen:
this PR fixs #238 #239 
But this is a temporary solution, the fundamental solution is making PD API returns error code 404 when no found member. 
@gregwebs @weekface @tennix PTAL

### DIFF
--- a/pkg/controller/pd_control.go
+++ b/pkg/controller/pd_control.go
@@ -304,6 +304,20 @@ func (pc *pdClient) GetStore(storeID uint64) (*StoreInfo, error) {
 }
 
 func (pc *pdClient) DeleteStore(storeID uint64) error {
+	var exist bool
+	stores, err := pc.GetStores()
+	if err != nil {
+		return err
+	}
+	for _, store := range stores.Stores {
+		if store.Store.GetId() == storeID {
+			exist = true
+			break
+		}
+	}
+	if !exist {
+		return nil
+	}
 	apiURL := fmt.Sprintf("%s/%s/%d", pc.url, storePrefix, storeID)
 	req, err := http.NewRequest("DELETE", apiURL, nil)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/controller/pd_control.go
```diff
@@ -316,7 +330,7 @@ func (pc *pdClient) DeleteStore(storeID uint64) error {
 	defer DeferClose(res.Body, &err)
 
 	// Remove an offline store should returns http.StatusOK
-	if res.StatusCode == http.StatusOK {
+	if res.StatusCode == http.StatusOK || res.StatusCode == http.StatusNotFound {
```
weekface: Does `DeleteStore` request's response will return `http.StatusNotFound `?

pkg/controller/pd_control.go
```diff
@@ -338,14 +366,28 @@ func (pc *pdClient) DeleteMemberByID(memberID uint64) error {
 		return err
 	}
 	defer DeferClose(res.Body, &err)
-	if res.StatusCode == http.StatusOK {
+	if res.StatusCode == http.StatusOK || res.StatusCode == http.StatusNotFound {
```
weekface: ditto

pkg/manager/member/pd_failover.go
```diff
@@ -157,12 +157,12 @@ func (pf *pdFailover) tryToDeleteAFailureMember(tc *v1alpha1.TidbCluster) error
 		return nil
 	}
 
-	memberID, err := strconv.Atoi(failureMember.MemberID)
+	memberID, err := strconv.ParseUint(failureMember.MemberID, 10, 64)
```
weekface: Add a unit test case:  don't fail when memberID is "12891273174085095651"

xiaojingchen:
@gregwebs 
I am not clear what the race condition means,  
Is it means two or more threads call the deleteMember method to delete one member at the same time, may cause `no found` error?

gregwebs:
@xiaojingchen yes, its possible between get and delete that it is deleted (by a different program than tidb-operator).

xiaojingchen:
This will only conflict once, and the caller will try again without any problem.

PR #246, opened 2018-12-21
Allow specifying tidb service load balancer IP

### LINKED ISSUE #242: support all internal load balancer description fields
It is easy enough to do

```
tidb.service.type=LoadBalancer
tidb.service.anotations=cloud.google.com/load-balancer-type=Internal
```

However, there are other available fields such as `loadBalancerIP` that are not as easy to add.

https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing

tennix:
Fixes #242 @gregwebs PTAL

### DIFF
--- a/charts/tidb-cluster/templates/tidb-service.yaml
+++ b/charts/tidb-cluster/templates/tidb-service.yaml
@@ -17,6 +17,9 @@ spec:
   {{- if .Values.tidb.service.externalTrafficPolicy }}
   externalTrafficPolicy: {{ .Values.tidb.service.externalTrafficPolicy }}
   {{- end }}
+  {{- if .Values.tidb.service.loadBalancerIP }}
+  loadBalancerIP: {{ .Values.tidb.service.loadBalancerIP }}
+  {{- end }}
   ports:
   - name: mysql-client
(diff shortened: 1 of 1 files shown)


gregwebs:
LGTM. It is possible to want to configure additional fields in the future though, so I am hoping that in the long-term we will come up with techniques to allow more extensibility. I think in this case part of the problem is that the YAML layout for the load balancer is flattened into the spec instead of nested under a key.

tennix:
I think we should only expose the most commonly used fields, the other fields are seldom used, only advanced users need to set those and they should already know how Helm works and hack the templates to satisfy their needs.

gregwebs:
Maintaining a fork is always a little problematic. But perhaps we can try out [tooling that helps with that]( https://www.replicated.com/ship/)

gregwebs:
CI seems to be out of disk

```
failed to watch file "/var/log/pods/d55fd326-0550-11e9-a901-02422767d21a/tidb-operator-e2e/0.log": no space left on device+ tail -5 result.log
```

weekface:
I have rebuild the dind env. You can try it again later.

PR #252, opened 2018-12-25
add block-cache-size to values.yaml

weekface:
This pr adds the two `block-cache-size` options to values.yaml


tennix:
There are some parameters related to CPU numbers need to be exposed in values.yaml
```toml
[readpool.storage]
high-concurrency = 4
normal-concurrency = 4
low-concurrency = 4
[readpool.coprocessor]
high-concurrency = 8
normal-concurrency = 8
low-concurrency = 8
[storage]
scheduler-worker-pool-size = 4
```
The three `concurrency` can be set to the same value.

PR #257, opened 2018-12-27
mark the build as dirty only if the source file has been modified

onlymellb:
When building the source code, we only need to pay attention to the source file. Therefore, in the released version information, it is necessary to mark it as dirty only when the source file is modified during building. The modification of other non-source files does not affect the build of the source code.


gregwebs:
What is the reason for this change? I usually have a requirement that after a build or a CI run nothing should show up as modified in git. This means someone building and testing the project will not end up in an unexpected situation. Usually the way I do this is just by using .gitignore, including dedicating some ignored directories for testing.

tennix:
I think we need to figure out why `make check` makes the working directory dirty other than only care about `*.go` files.

PR #260, opened 2019-01-04
new algorithm for scheduler HA predicate

### LINKED ISSUE #235: A new algorithm for tidb scheduler HA predicate
Now the algorithm of [HA predicate](https://github.com/pingcap/tidb-operator/blob/master/pkg/scheduler/predicates/ha.go#L54) is complicated and hard to understand:

1. First, we sort all the nodes we get from kube-scheduler by how many same kind of pod it contains, find the nodes that have least pods.
2. When scheduling the first replicas pods, we must ensure no previous pods on the nodes.
3. For later pods, we choose the nodes that have least pods.

So we need an easier solution. If a user cannot access the system, it is – from the users' point of view – unavailable. In TiDB Operator, we say HA, it means the whole system is still available when there is **one** Kube node down. 

Discussed with @tennix @onlymellb and @xiaojingchen. The new algorithm is:

- ~~multiple pods are not allowed to be scheduled concurrently in TiDB Operator: lower ordinal pods must be scheduled before the current pod~~
- return the node to kube-scheduler if there is only one node and the pod's pvc is bound 
- return these nodes that have least pods and its pods count is less than replicas/2 to kube-scheduler
- let kube-scheduler to make the final decision

Then all the pods count of Kube nodes is equal or less then replicas / 2, it is available(the raft group is still in quorum.) when this node is down.

weekface:
fixes #235 

### DIFF
--- a/pkg/scheduler/predicates/ha.go
+++ b/pkg/scheduler/predicates/ha.go
@@ -60,6 +60,10 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
 	component := pod.Labels[label.ComponentLabelKey]
 	tcName := getTCNameFromPod(pod, component)
 
+	if len(nodes) == 0 {
+		return nil, fmt.Errorf("kube nodes is empty")
+	}
+
 	if len(nodes) == 1 {
 		pvcName := fmt.Sprintf("%s-%s", component, podName)
 		pvc, err := h.pvcGetFn(ns, pvcName)
--- a/pkg/scheduler/predicates/ha.go
+++ b/pkg/scheduler/predicates/ha.go
@@ -15,31 +15,36 @@ package predicates
 
 import (
 	"fmt"
+	"sort"
+	"strings"
 
 	"github.com/golang/glog"
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap.com/v1alpha1"
+	"github.com/pingcap/tidb-operator/pkg/client/clientset/versioned"
 	"github.com/pingcap/tidb-operator/pkg/label"
-	"github.com/pingcap/tidb-operator/pkg/util"
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 )
 
-const (
-	replicas int32 = 3
-)
-
 type ha struct {
 	kubeCli   kubernetes.Interface
+	cli       versioned.Interface
 	podListFn func(ns, instanceName, component string) (*apiv1.PodList, error)
+	pvcGetFn  func(ns, pvcName string) (*apiv1.PersistentVolumeClaim, error)
+	tcGetFn   func(ns, tcName string) (*v1alpha1.TidbCluster, error)
 }
 
 // NewHA returns a Predicate
-func NewHA(kubeCli kubernetes.Interface) Predicate {
+func NewHA(kubeCli kubernetes.Interface, cli versioned.Interface) Predicate {
 	h := &ha{
 		kubeCli: kubeCli,
+		cli:     cli,
 	}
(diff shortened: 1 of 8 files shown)


pkg/scheduler/predicates/ha_test.go
```diff
@@ -39,229 +40,327 @@ func TestMapNil(t *testing.T) {
...
 			expectFn: func(nodes []apiv1.Node, err error) {
-				g.Expect(strings.Contains(err.Error(), "strconv.Atoi: parsing")).To(Equal(true))
+				g.Expect(err).To(HaveOccurred())
+				g.Expect(strings.Contains(err.Error(), "get tidbcluster failed")).To(BeTrue())
 			},
 		},
 		{
-			name:      "one pod, podName is wrong",
-			ordinal:   0,
+			name:      "one node, return zero node",
```
xiaojingchen:
```suggestion
			name:      "zero node, return zero node",
```

pkg/scheduler/predicates/ha.go
```diff
@@ -15,117 +15,111 @@ package predicates
...
 func (h *ha) Name() string {
 	return "HighAvailability"
 }
 
-// 1. First, we sort all the nodes we get from kube-scheduler by how many same kind of pod it contains,
-//    find the nodes that have least pods.
-// 2. When scheduling the first replicas pods, we must ensure no previous pods on the nodes.
-// 3. For later pods, we choose the nodes that have least pods.
+// 1. return the node to kube-scheduler if there is only one node and the pod's pvc is bound
+// 2. return these nodes that have least pods and its pods count is less than replicas/2 to kube-scheduler
```
```suggestion
// 2. return these nodes that have least pods and its pods count is less than (replicas+1)/2 to kube-scheduler
```

pkg/scheduler/predicates/ha.go
```diff
@@ -15,117 +15,111 @@ package predicates
...
 	glog.V(4).Infof("nodeMap: %+v", nodeMap)
 
-	var min int
-	var minInitialized bool
-	for _, podNameArr := range nodeMap {
-		count := len(podNameArr)
-		if !minInitialized {
-			minInitialized = true
-			min = count
+	min := -1
```
onlymellb: Is it better to initialize the min variable to a large enough value?

weekface: Large enough, how big is it?

onlymellb: The default number of pods that can be run on a kubelet is 110, so I think it's okay to be larger than this number, what do you think?

pkg/scheduler/predicates/ha.go
```diff
@@ -15,117 +15,111 @@ package predicates
...
+			continue
+		}
+		if podsCount < min {
+			min = podsCount
+			minNodeNames = make([]string, 0)
 		}
+		minNodeNames = append(minNodeNames, nodeName)
 	}
 
+	if len(minNodeNames) == 0 {
```
onlymellb: If the passed parameter `nodes []apiv1.Node` is empty slice, should we return quickly at the beginning of the function?

pkg/scheduler/predicates/ha.go
weekface: This number may be modified by the user. -1 is good enough i think.

pkg/scheduler/predicates/ha_test.go
```diff
@@ -39,229 +40,327 @@ func TestMapNil(t *testing.T) {
 
 	g.Expect(m["c"] == nil).To(Equal(true))
 	g.Expect(m["a"] == nil).To(Equal(false))
+
+	var i *int
+	g.Expect(i).To(BeNil())
```
tennix: What is this?

pkg/scheduler/predicates/ha_test.go
```diff
@@ -39,229 +40,327 @@ func TestMapNil(t *testing.T) {
...
+}
+
+func TestSortEmptyArr(t *testing.T) {
```
tennix: What's the purpose of this test?

PR #262, opened 2019-01-09
add tidb discovery service

### LINKED ISSUE #263: Introduce a discovery service for PD to properly get bootstrapped
Now the TiDB Operator has lots of hack for PD to bootstrap, including injecting an annotation `bootstrapping` to PD startup script to distinguish if the first pod is bootstrapping or recreated, and the startup script is too complicated. When introduce multiple statefulsets #240, the startup script can be even more complicated and could hardly cover all cases.

After discussion, we've decided to introduce a discovery service just like [etcd discovery](https://github.com/etcd-io/etcd/blob/master/Documentation/dev-internal/discovery_protocol.md#public-discovery-service) to aid PD to start with the correct arguments.

weekface:
This PR closes #263. Like [etcd discovery](https://github.com/etcd-io/etcd/blob/master/Documentation/dev-internal/discovery_protocol.md#public-discovery-service), this PR adds a TiDB Discovery service to help the new PD member to discover all other members in cluster bootstrap phase. The Operator will create a new TiDB Discovery Pod for every TiDB cluster.

PD start script:

-  When the PD starts up, first determine whether there is already data (the /var/lib/pd/member/wal directory exists), if there is data, start it directly, otherwise request the `Discovery` service together with the parameters that can mark their identity (advertise-peer-urls), waiting for the `Discovery` service to return the startup parameters: `-initial-cluster=` or `--join=` to start the service. If the request is failed, continue with the next request.

Discovery service:

After receiving the request, parse the `advertise-peer-urls` parameters. By querying the TidbCluser object to know the `Replicas` of the PD, the following processing logic:

- Record each peer and save its information to a map structure;
- If the count of map is equal to `Replicas` at this time, generate the `--initial-cluster=` args for the currently requested peer and return it to the client.
- If the count of map is not equal to `Replicas` at this time, request the PD members API and generate a `--join` args based on these members to return to the client.
- As long as one peer is started by the `--initial-cluster=` startup args, all subsequent peers can get the current cluster state through the PD members API, and then get the `--join` startup args.


charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -43,56 +42,36 @@ while true; do
...
-
-    ARGS="${ARGS}--join="
-    for i in $(seq 0 ${TOP});
-    do
-        [[ ${i} -eq ${ORDINAL} ]] && continue
-        ARGS="${ARGS}http://${SET_NAME}-${i}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc:2380"
-        if [[ ${i} -lt ${TOP} ]]
+    while true; do
+        sleep 5
+        wget -O /tmp/result http://${discovery_url}/new/${domain}:2380
```
tennix: Is it better to use base64 string as query parameter? Specify a timeout for the HTTP request.

charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -43,56 +42,36 @@ while true; do
...
-      TOP=$((ORDINAL-1))
-    fi
-
-    ARGS="${ARGS}--join="
-    for i in $(seq 0 ${TOP});
-    do
-        [[ ${i} -eq ${ORDINAL} ]] && continue
-        ARGS="${ARGS}http://${SET_NAME}-${i}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc:2380"
-        if [[ ${i} -lt ${TOP} ]]
+    while true; do
```
tennix: You can use `until` instead of `while`:
```shell
until result=$(wget -qO- -T 3 http://${discovery_url}/new/${domain}:2380 2>/dev/null); do
    echo "waiting for discovery service returns start args ..."
    sleep 2
done
ARGS="${ARGS}${result}"
```

pkg/discovery/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+		return "", fmt.Errorf("advertisePeerUrl is empty")
+	}
+	strArr := strings.Split(advertisePeerUrl, ".")
+	if len(strArr) != 4 {
+		return "", fmt.Errorf("advertisePeerUrl format is wrong: %s", advertisePeerUrl)
+	}
+
+	podName, peerServiceName, ns := strArr[0], strArr[1], strArr[2]
+	tcName := strings.TrimSuffix(peerServiceName, "-pd-peer")
+	tc, err := td.tcGetFn(ns, tcName)
```
tennix: Maybe better use the discovery pod itself namespace or verify if the two namespaces are equal.

pkg/discovery/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return "", err
+	}
+	keyName := fmt.Sprintf("%s/%s", ns, tcName)
+	replicas := tc.Spec.PD.Replicas
```
tennix: TODO: the replicas should be the total replicas of pd sets.

pkg/discovery/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if td.clusters == nil {
+		td.clusters = map[string]*clusterInfo{}
```
tennix: Initialize the `clusters` in `NewTiDBDiscovery` function instead of here.

gregwebs:
This makes sense. I feel like PD should already be able to do this though.

charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -20,16 +18,18 @@ then
...
-SERVICE_DOMAIN="${SERVICE_NAME}.${NAMESPACE}.svc"
-
 runmode=${runmode:-normal}
 if [[ X${runmode} == Xdebug ]]
 then
     echo "entering debug mode."
     tail -f /dev/null
 fi
 
+cluster_name=`echo ${PEER_SERVICE_NAME} | sed 's/-pd-peer//'`
```
onlymellb: Add a comment to describe the general form of variable  `PEER_SERVICE_NAME`?

pkg/discovery/server/mux.go
```diff
@@ -0,0 +1,69 @@
...
+	"net/http"
+
+	restful "github.com/emicklei/go-restful"
+	"github.com/golang/glog"
+	"github.com/pingcap/tidb-operator/pkg/client/clientset/versioned"
+	"github.com/pingcap/tidb-operator/pkg/discovery"
+)
+
+type server struct {
+	discovery *discovery.TiDBDiscovery
```
onlymellb: Why not define an interface with method `Discovery`,  if you have a program that uses this code in the future, it is convenient for mock testing.

pkg/discovery/discovery.go
```diff
@@ -0,0 +1,110 @@
...
+	strArr := strings.Split(advertisePeerUrl, ".")
+	if len(strArr) != 4 {
+		return "", fmt.Errorf("advertisePeerUrl format is wrong: %s", advertisePeerUrl)
+	}
+
+	podName, peerServiceName, ns := strArr[0], strArr[1], strArr[2]
+	tcName := strings.TrimSuffix(peerServiceName, "-pd-peer")
+	podNamespace := os.Getenv("MY_POD_NAMESPACE")
+	if ns != podNamespace {
+		return "", fmt.Errorf("the peer's namespace: %s is not equal to discovery namespace: %s", ns, podNamespace)
```
onlymellb: It is better to print the `advertisePeerUrl` information in the error log, especially when there may be multiple clusters under a namespace.

Issue #264, opened 2019-01-11
scheduler: scheduled the all tikv pods to same node 


xiaojingchen:
after deploy tidbcluster, found all tikv pods were scheduled to one node.
```
[tidb@172.16.30.59 ~]$ kubectl get po -n example3 -o wide
NAME                                  READY     STATUS      RESTARTS   AGE       IP              NODE
example3-monitor-configurator-cnhkd   0/1       Completed   0          7m        10.128.35.166   172.16.30.118
example3-monitor-d85c86ccc-vmk2s      2/2       Running     0          7m        10.128.32.186   172.16.30.59
example3-pd-0                         1/1       Running     0          7m        10.128.34.78    172.16.30.98
example3-pd-1                         1/1       Running     0          7m        10.128.35.187   172.16.30.118
example3-pd-2                         1/1       Running     0          7m        10.128.47.59    172.16.30.58
example3-tidb-0                       1/1       Running     0          6m        10.128.42.214   172.16.30.216
example3-tidb-1                       1/1       Running     0          6m        10.128.35.185   172.16.30.118
example3-tidb-initializer-mfjxt       0/1       Completed   0          7m        10.128.36.4     172.16.30.117
example3-tikv-0                       2/2       Running     1          7m        10.128.34.66    172.16.30.98
example3-tikv-1                       2/2       Running     0          7m        10.128.34.121   172.16.30.98
example3-tikv-2                       2/2       Running     0          7m        10.128.34.90    172.16.30.98
```
the tidb-scheduler log:
```
I0111 10:19:22.367319       1 scheduler.go:75] scheduling pod: example2/example2-pd-0
I0111 10:19:22.367365       1 scheduler.go:78] entering predicate: HighAvailability, nodes: [172.16.30.216 172.16.30.58 172.16.30.98 172.16.30.117 172.16.30.59 172.16.30.118]
I0111 10:19:22.379762       1 ha.go:103] []
I0111 10:19:22.379799       1 ha.go:104] nodeMap: map[172.16.30.58:[] 172.16.30.98:[] 172.16.30.117:[] 172.16.30.59:[] 172.16.30.118:[] 172.16.30.216:[]]
I0111 10:19:22.379832       1 scheduler.go:83] leaving predicate: HighAvailability, nodes: [172.16.30.216 172.16.30.58 172.16.30.98 172.16.30.117 172.16.30.59 172.16.30.118]
[...]

PR #266, opened 2019-01-15
Serial scheduling

weekface:
This pr fixes: #264, make sure only one pod can be scheduled at the same time.

### DIFF
--- a/pkg/scheduler/predicates/ha.go
+++ b/pkg/scheduler/predicates/ha.go
@@ -157,7 +157,7 @@ func (h *ha) realAcquireLock(pod *apiv1.Pod) (*apiv1.PersistentVolumeClaim, *api
 	var schedulingPVC *apiv1.PersistentVolumeClaim
 	for i := 0; i < len(pvcList.Items); i++ {
 		pvc := pvcList.Items[i]
-		if pvc.GetName() == currentPVCName && currentPVC == nil {
+		if pvc.GetName() == currentPVCName {
 			currentPVC = &pvcList.Items[i]
 		}
 		if pvc.Annotations[label.AnnPVCPodScheduling] != "" && schedulingPVC == nil {
--- a/pkg/scheduler/predicates/ha.go
+++ b/pkg/scheduler/predicates/ha.go
@@ -156,7 +156,7 @@ func (h *ha) realAcquireLock(pod *apiv1.Pod) (*apiv1.PersistentVolumeClaim, *api
 	var currentPVC *apiv1.PersistentVolumeClaim
 	var schedulingPVC *apiv1.PersistentVolumeClaim
 	items := pvcList.Items
-	for i, _ := range items {
+	for i := range items {
 		if items[i].GetName() == currentPVCName {
 			currentPVC = &items[i]
 		}
--- a/pkg/scheduler/predicates/ha.go
+++ b/pkg/scheduler/predicates/ha.go
@@ -16,6 +16,8 @@ package predicates
 import (
 	"fmt"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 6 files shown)


pkg/scheduler/predicates/ha.go
```diff
@@ -127,13 +142,69 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	currentPVCName := pvcName(component, podName)
+	var currentPVC *apiv1.PersistentVolumeClaim
+	var schedulingPVC *apiv1.PersistentVolumeClaim
+	for i := 0; i < len(pvcList.Items); i++ {
+		pvc := pvcList.Items[i]
+		if pvc.GetName() == currentPVCName && currentPVC == nil {
```
onlymellb: Why do you want to add the condition `currentPVC == nil`, will there be multiple PVCs with the same name?

weekface: Removed.

pkg/scheduler/predicates/ha.go
```diff
@@ -127,13 +142,69 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
...
+	podName := pod.GetName()
+	pvcList, err := h.pvcListFn(ns, instanceName, component)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	currentPVCName := pvcName(component, podName)
+	var currentPVC *apiv1.PersistentVolumeClaim
+	var schedulingPVC *apiv1.PersistentVolumeClaim
+	for i := 0; i < len(pvcList.Items); i++ {
```
```suggestion
	for _, pvc := range pvcList.Items {
```

weekface: Use this style to the golang [common mistakes](https://github.com/golang/go/wiki/CommonMistakes)

onlymellb: There is not use closures running as goroutines，If there is such a scene, then the way to use range can also avoid this problem.

weekface: `currentPVC ` and `schedulingPVC ` are pointers,  the `pvc` in the for range block is a variable,    if we use the for range style, then these two variables will get the same address.

weekface: For example:
```
package main

import "fmt"

func main() {
	arr := []string{"a", "b", "c"}

	var pa *string
	var pb *string
	for _, item := range arr {
		if item == "a" {
			pa = &item
		}
		if item == "b" {
			pb = &item
		}
	}
	fmt.Println(pa)
	fmt.Println(pb)
	fmt.Println(pa == pb)

	var ppa *string
	var ppb *string
	for i := 0; i < len(arr); i++ {
		if arr[i] == "a" {
			ppa = &arr[i]
		}
		if arr[i] == "b" {
			ppb = &arr[i]
		}
	}
	fmt.Println(ppa)
	fmt.Println(ppb)
	fmt.Println(ppa == ppb)
}
```

Outputs:
```
0xc00005e1c0
0xc00005e1c0
true
0xc000060180
0xc000060190
false
```

onlymellb: You can create a new variable to fix this problem. such as follow:
```
for i := 0; i < len(pvcList.Items); i++ { 
    pvc := pvcList.Items[i]
    ...
}
------
for _, pvc := range pvcList.Items {
    pvc := pvc
 ...
}
```

weekface: Style `for i:=0; i<n; i++` is equivalent to the `for range` style, but using `for i:=0; i<n; i++` is more clear at this situation.

PR #273, opened 2019-01-28
in values.yml, omit just the empty leaves

gregwebs:
This matches the golang source code representation of values.yml


gregwebs:
The e2e tests keep failing. I am having trouble understanding the logs.

tennix:
Maybe the CI DinD environment is broken, I've just rebuilt the DinD environment.

gregwebs:
thanks, it is passing now

PR #279, opened 2019-02-19
Add `ClusterIDLabelKey` label to TidbCluster

xiaojingchen:
this PR add `ClusterIDLabelKey` label to tidbCluster and add `ClusterID` field in `TidbCluster.Status`
@weekface @tennix PTAL

### DIFF
--- a/pkg/controller/tidbcluster_control.go
+++ b/pkg/controller/tidbcluster_control.go
@@ -15,7 +15,6 @@ package controller
 
 import (
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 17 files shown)


pkg/controller/tidbcluster_control.go
```diff
@@ -34,25 +36,46 @@ import (
...
+	pdControl PDControlInterface) TidbClusterControlInterface {
 	return &realTidbClusterControl{
 		cli,
 		tcLister,
 		recorder,
+		pdControl,
 	}
 }
 
+func (rtc *realTidbClusterControl) UpdateMetaInfo(tc *v1alpha1.TidbCluster) (*v1alpha1.TidbCluster, error) {
```
weekface: I don't think it is necessary to sync the `ClusterID` to `TidbCluster` labels.

PR #280, opened 2019-02-20
charts: backup to ceph object storage

LinuxGit:


charts/tidb-backup/templates/scripts/_start_restore.sh.tpl
```diff
@@ -21,3 +21,19 @@ downloader \
...
+{{- if .Values.restore.ceph }}
+downloader \
+  --cloud=ceph \
+  --bucket={{ .Values.restore.ceph.bucket }} \
+  --endpoint={{ .Values.restore.ceph.endpoint }} \
+  --srcDir={{ .Values.restore.ceph.srcDir }} \
+  --destDir=${dataDir}
+
+/loader \
+  -d ${dataDir}/{{ .Values.restore.ceph.srcDir }} \
```
gregwebs: This section looks the same as above except for srcDir. I think we can move `srcDir` up a level so that it will be `.Values.restore.srcDir`

charts/tidb-backup/templates/scripts/_start_backup.sh.tpl
```diff
@@ -19,3 +19,11 @@ uploader \
   --bucket={{ .Values.backup.gcp.bucket }} \
   --backup-dir=/data/${dirname}
 {{- end }}
+
+{{- if .Values.backup.ceph }}
+uploader \
```
gregwebs: Do we need an updated uploader docker image?

charts/tidb-backup/templates/scripts/_start_backup.sh.tpl
```diff
@@ -19,3 +19,11 @@ uploader \
...
+  --cloud=ceph \
+  --bucket={{ .Values.backup.ceph.bucket }} \
```
gregwebs: Is this the S3 API? If so can we just configure this as AWS rather than ceph?

LinuxGit: Yes. We use rook ceph operator with the S3 API. The config is a bit different, like endpoint
https://github.com/tennix/tidb-cloud-backup/blob/master/pkg/setup.go#L111

charts/tidb-backup/templates/scripts/_start_backup.sh.tpl
LinuxGit: Yes. The pr just merged on https://github.com/tennix/tidb-cloud-backup.

charts/tidb-backup/templates/scripts/_start_backup.sh.tpl
gregwebs: endpoint is actually supported by AWS though, its just that most users won't be specifying it but instead specifying a region. It looks like the main difference is that `--cloud=ceph` is translated to some different default configuration values:

```
WithS3ForcePathStyle(true).
		WithDisableSSL(true).
WithMaxRetries(20)
```

We should consider an upload that is of the format

```
aws
  bucket: backup
  ceph: true
```

Where `ceph: true` creates those above option defaults.

I am also okay with keeping it as is if you still think that is the best.

charts/tidb-backup/templates/scripts/_start_restore.sh.tpl
LinuxGit: Good idea. I've updated it.

charts/tidb-cluster/values.yaml
```diff
@@ -333,6 +333,10 @@ scheduledBackup:
   # and then: cat credentials.json | base64 | tr -d '\n'
   #  credentialsData: ""
   #  bucket: ""
+  # backup to ceph object storage
```
tennix: `images/tidb-operator-e2e/tidb-cluster-values.yaml` also has to be updated.

charts/tidb-backup/templates/scripts/_start_backup.sh.tpl
LinuxGit: OK. I'd like to keep it now. After finishing the backup feature of TEP, I may change it in the future.
Another difference is that I check and create the bucket on rook ceph object store, see https://github.com/tennix/tidb-cloud-backup/blob/master/pkg/setup.go#L119.
Ceph also support swift protocal, I'm not going to support it now.

tennix:
@xiaojingchen PTAL

charts/tidb-backup/templates/scripts/_start_restore.sh.tpl
```diff
@@ -9,15 +9,23 @@ host=`echo {{ .Values.clusterName }}_TIDB_SERVICE_HOST | tr '[a-z]' '[A-Z]' | tr
 downloader \
   --cloud=gcp \
   --bucket={{ .Values.restore.gcp.bucket }} \
-  --srcDir={{ .Values.restore.gcp.srcDir }} \
+  --srcDir={{ .Values.restore.srcDir }} \
   --destDir=${dataDir}
+{{- end }}
+
+{{- if .Values.restore.ceph }}
```
gregwebs: The current template allows for the downloader to be called twice. I think and `else` would be better.

iamxy:
LGTM

PR #281, opened 2019-02-21
Update setup.md

weekface:
Add bind mount guide to `Local Persistent Volume` section.


docs/setup.md
```diff
@@ -52,7 +60,9 @@ You can follow Helm official [documentation](https://helm.sh) to install Helm in
 
 ## Local Persistent Volume
 
-Local disks are recommended to be formatted as ext4 filesystem.
+Local disks are recommended to be formatted as ext4 filesystem. The local persistent volume directory must be a [mount point](https://unix.stackexchange.com/questions/198590/what-is-a-bind-mount): a whole disk mount or a bind mount:
```
tennix: The url should be bind mount not mount point. Also I think we should state the disadvantages of bind mount for TiDB. All the volumes has the size of the whole disk and there is no quota and isolation of bind mount volumes.

Issue #274, opened 2019-01-28
Flexible user password initialization


gregwebs:
# Problem

Our current initializer job that sets the user password suffers from a problem:
* It puts SQL into a secret with a password. When we retrieve the password in an automated way, we must run an `awk` command: it is a very fragile situation

Both the initializer job and other passwords (backup) have passwords passed through helm.
* This is insecure unless the user stores their helm config as a secret (not the default)
* This can present its own escaping issues

# Solving Escaping

The initializer job is structured the way it is to avoid escaping issues. However, we have determined two alternatives:
* Use bash and `mysqladmin`
* Use a docker image that has a mysql library installed.
  * We can use a dynamic language like Python to avoid a compilation step.
  * It should also be possible to use a pre-compiled go program, but then we would need to distribute our own image
* Secrets are given to the pod as either a file or an environment variable that can be read in by the program

# Avoiding helm

All passwords should be created as secrets by the user. The manifests just reference the secret names.

gregwebs:
Also, `gcp.credentialsData` in the backup chart should be created as a secret.

tennix:
With `py-mysqldb` installed on alpine linux image and the initial root password stored in `ROOT_PASSWORD` environment, the root password can be set with the following command:

`python -c "import os, MySQLdb; MySQLdb.connect(host='127.0.0.1', port=4000, user='root').cursor().execute(\"set password for 'root'@'%%' = %s; flush privileges;\", (os.environ.get('ROOT_PASSWORD'),))"`

There is already a Python mysqlclient image on DockerHub: https://github.com/tnir/mysqlclient.

As for your suggestion, I prefer to make the secrets optional: If users provide secret names for the passwords, then helm would not create these secrets. But users can also set passwords and credentials in the values.yaml and let helm to create these secrets.

gregwebs:
> I prefer to make the secrets optional: If users provide secret names for the passwords, then helm would not create these secrets. But users can also set passwords and credentials in the values.yaml and let helm to create these secrets.

There is a down-side to maintaining two ways of doing things, particularly where one is insecure by default. Supporting insecure by default techniques means we must spend time documenting how to do things in a secure way.

The insecure by default comes in two forms
* Helm will store this its chart data in an ordinary ConfigMap by default. There is a configuration option to store the entire configuration as a secret, but that still means that giving access to helm is equivalent to giving access to all the secrets.
[...]

PR #282, opened 2019-02-22
Don't initialize when there is no tidb.password

gregwebs:
## Changes

Don't create
* a secret
* an initializer job

## Motivation

From my reading, random password initialization is broken: https://github.com/helm/helm/issues/3053

The user can manage users from the MySQL connection
As a benefit this allows users manage their password such that it is never stored in K8s. Or they can create their own init job, etc.

## Testing

- [ ] Run with -set tidb.password=password
- [X] Run without setting the tidb password

 For both options:
* Perform output suggested in NOTES
* Verify whether secret/initializer are created

### DIFF
--- a/charts/tidb-cluster/templates/NOTES.txt
+++ b/charts/tidb-cluster/templates/NOTES.txt
@@ -1,17 +1,27 @@
+Cluster Startup
 1. Watch tidb-cluster up and running
-   watch kubectl get pods --namespace {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Release.Name }} -o wide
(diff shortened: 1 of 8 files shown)


gregwebs:
Relates to some discussion on #274

tennix:
The related documents should be updated too.

tennix:
What are the checkbox meaning? Will you support both of them in this PR?

gregwebs:
Yes. It is listing the manual testing cases. Perhaps you want to do the unchecked one?

tennix:
From #274, we've discussed that it's better to reference secret name created outside of helm. Are you going to do this in another PR or just leave it like this?

gregwebs:
secret name references would be a different PR.

PR #286, opened 2019-02-26
Flexible tidb initializer job with secret set outside of helm

### LINKED ISSUE #274: Flexible user password initialization

tennix:
This PR closes #274. It uses secret created outside of helm to avoid storing password in ConfigMap. And it also enables flexible initialization SQL statement.


charts/tidb-cluster/values.yaml
```diff
@@ -177,8 +180,9 @@ tidb:
       # cloud.google.com/load-balancer-type: Internal
 
 # mysqlClient is used to set password for TiDB
+# it must has Python MySQL client installed
```
gregwebs: Having this image be configurable may not be useful anymore. But we could rename it to pythonMysqlClient.

gregwebs:
This looks great! I will try it out. In #274 I also mentioned secrets used in backup.

charts/tidb-cluster/templates/tidb-initializer-job.yaml
```diff
@@ -23,22 +24,50 @@ spec:
...
+          import os, MySQLdb
+          host = {{ printf "%s-tidb" .Values.clusterName | quote }}
+          port = 4000
+          root_password = os.environ.get('ROOT_PASSWORD')
+          system_password = os.environ.get('SYSTEM_PASSWORD')
+          conn = MySQLdb.connect(host=host, port=port, user='root', connect_timeout=5)
+          if root_password:
+              conn.cursor().execute("set password for 'root'@'%%' = %s;", (root_password,))
+          if system_password:
+              conn.cursor().execute("create user 'system'@'%%' identified by %s;", (system_password,))
```
gregwebs: The system user is a special user in MySQL that cannot be created or altered. So for compatibility reasons I think we should avoid having a user named 'system'

charts/tidb-cluster/templates/tidb-initializer-job.yaml
```diff
@@ -23,22 +24,50 @@ spec:
...
+                  conn.cursor().execute(line)
+                  conn.commit()
+          {{- end }}
+        env:
+        - name: ROOT_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: {{ .Values.tidb.passwordSecret }}
+              key: root
+        - name: SYSTEM_PASSWORD
```
gregwebs: The setup here of the secret being a `user: password` mapping is nice. Users would not need to modify this yaml file if we could loop over the `user: password` pairs in the Python code to set them. Instead they just set the secret user-password pairs. I think this should be possible by mounting the secret as a file.

charts/tidb-cluster/values.yaml
tennix: It's still need to be configured especially when dockerhub is unavailable.

charts/tidb-cluster/templates/tidb-initializer-job.yaml
tennix: When there are multiple data in a secret, the secret volume is mounted as multiple files. But this is possible, we can mount secret as `/etc/tidb/secrets`, and having secret data named as `<user>_password`, then each user's password is storing in `/etc/tidb/secrets/<user>_password`. And then looping the files in Python script.

gregwebs: Yes! I would just have it as `<user>` rather than `<user>_password. The secret is just for TiDB passwords.

charts/tidb-cluster/values.yaml
```diff
@@ -145,10 +145,13 @@ tikvPromGateway:
 
 tidb:
   replicas: 2
-  # The password to access TiDB
-  # If set, the password will be stored both in helm and in a Secret
+  # The secret name of root password, you can create secret with following command:
+  # kubectl create secret generic tidb-secret --from-literal=root=<root-password>
   # If unset, the root password will be empty and you can set it after connecting
-  # password: "admin"
+  # passwordSecret: tidb-secret
```
gregwebs: It may be better to call this `passwordSecretName`.

PR #290, opened 2019-02-27
Support slow log tailing sidcar for tidb instance

### LINKED ISSUE #231: How to view TiDB slow log?
TiDB can output slow log to a separate log file. This makes it easy to view the slow logs. Otherwise, users have to grep slow log from all logs. However, users cannot view the slow logs by `kubectl` if TiDB pod emits logs to a separate log file.

Kubernetes [documentation](https://kubernetes.io/docs/concepts/cluster-administration/logging/#using-a-sidecar-container-with-the-logging-agent) suggests using sidecar to support multiple log files. The sidecar container can only tails the log file to stdout, so the resources used is small:
> order of couple of millicores for cpu and order of several megabytes for memory

So I think we should add sidecar container for TiDB pod for slow log.

aylei:
Arbitrary mount point given by user may be harmful, so we don't allow customizing the slow-query-log file location.

close #231 

Signed-off-by: Yeh-lei Wu <rayingecho@gmail.com>


pkg/apis/pingcap.com/v1alpha1/types.go
```diff
@@ -42,8 +42,10 @@ const (
 	TiDBMemberType MemberType = "tidb"
 	// TiKVMemberType is tikv container type
 	TiKVMemberType MemberType = "tikv"
-	//PushGatewayMemberType is pushgateway container type
+	// PushGatewayMemberType is pushgateway container type
 	PushGatewayMemberType MemberType = "pushgateway"
+	// TiDBLogTailerMemberType is tidb log tailer container type
```
weekface:
```suggestion
	// SlowLogTailerMemberType is tidb log tailer container type
```

charts/tidb-cluster/values.yaml
```diff
@@ -175,6 +175,16 @@ tidb:
     exposeStatus: true
     # annotations:
       # cloud.google.com/load-balancer-type: Internal
+  # separateSlowLog: true
```
weekface: Do we need the `separateSlowLog` key?

pkg/manager/member/tidb_member_manager.go
```diff
@@ -30,6 +31,12 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 )
 
+const (
+	slowQueryLogVolumeName = "slowlog"
+	slowQueryLogDir        = "/tmp/log/tidb"
```
tennix: I think it's better to store logs in `/var/log/tidb`.

charts/tidb-cluster/values.yaml
aylei: It's optional and default to `false`

pkg/manager/member/tidb_member_manager.go
```diff
@@ -240,6 +247,76 @@ func (tmm *tidbMemberManager) getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbClust
...
+			Image:           controller.GetSlowLogTailerImage(tc),
+			ImagePullPolicy: tc.Spec.TiDB.SlowLogTailer.ImagePullPolicy,
+			Resources:       util.ResourceRequirement(tc.Spec.TiDB.SlowLogTailer.ContainerSpec),
+			VolumeMounts: []corev1.VolumeMount{
+				{Name: slowQueryLogVolumeName, MountPath: slowQueryLogDir},
+			},
+			Command: []string{
+				"sh",
+				"-c",
+				fmt.Sprintf("touch %s; tail -n0 -f %s;", slowQueryLogFile, slowQueryLogFile),
```
onlymellb: Use `-F` option to follow the log file is better.

charts/tidb-cluster/templates/config/_tidb-config.tpl
```diff
@@ -59,7 +59,11 @@ format = "text"
 disable-timestamp = false
 
 # Stores slow query log into separated files.
+{{- if .Values.tidb.separateSlowLog }}
+slow-query-file = "/tmp/log/tidb/slowlog"
```
tennix: This also needs to be changed to `/var/log/tidb/slowlog`.

aylei: Good catch.
BTW, this is kind of a code smell but I can't figure out a better solution. I'll document this caveat or any better ideas?

aylei: I mean user may override the configuration by setting `tidb.config`

charts/tidb-cluster/values.yaml
```diff
@@ -175,6 +175,16 @@ tidb:
...
+  slowLogTailer:
+    image: busybox:1.26.2
+    resources:
+      limits: {}
+      #   cpu: 50m
+      #   memory: 10Mi
+      requests: {}
+      #   cpu: 50m
```
gregwebs: We should specify a request. Very few resources should be needed for this.

charts/tidb-cluster/templates/config/_tidb-config.tpl
tennix: I think we can specify the command line argument `-log-slow-query` of tidb-server, Its priority is higher than the configuration file.

charts/tidb-cluster/values.yaml
aylei: Is cpu 20m and memory 5MB a good default for both limit and request? (I've tested locally and seems that 10m cpu, 1MB memory is fairly enough)

tennix: OK, LGTM.

gregwebs: I would not use a limit or make the limit a bit higher at least until we have observed it in production..

tennix: Agreed, maybe 50Mi memory limit is enough. This is pretty small comparing to TiDB itself.

PR #297, opened 2019-03-05
Support override helm values in e2e test

aylei:
close #296 
Signed-off-by: Aylei <rayingecho@gmail.com>

Issue #296, opened 2019-03-05
Support overriding helm values in e2e test


aylei:
This can help us writing more complex e2e test case

aylei:
close via #298

PR #298, opened 2019-03-05
Ensure SLOW_LOG_FILE env variable is always set

### LINKED ISSUE #295: TiDB server failed to start due to un-set parameter
The log of failing pod:
```sh
/usr/local/bin/tidb_start_script.sh: line 40: SLOW_LOG_FILE: parameter not set
```

aylei:
close #295 

### DIFF
--- a/tests/e2e/test_helper.go
+++ b/tests/e2e/test_helper.go
@@ -89,6 +89,9 @@ var fixtures = []clusterFixture{
 		clusterSpec: clusterSpec{
 			ns:          "ns-2",
 			clusterName: "cluster-name-1",
+			values: map[string]string{
+				"tidb.separateSlowLog": "false",
+			},
 		},
 		cases: []testCase{
 			testCreate,
--- a/tests/e2e/test_helper.go
+++ b/tests/e2e/test_helper.go
@@ -14,6 +14,7 @@
 package e2e
 
 import (
+	"bytes"
 	"fmt"
 	"os/exec"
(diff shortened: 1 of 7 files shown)


weekface:
I think you should move #297 codes to this PR, or we can't confirm this PR can work as expected without the e2e tests.

aylei:
Branch issue-296 merged, close #297 #296 
&
/run-e2e-tests

aylei:
@weekface @tennix PTAL

tests/e2e/test_helper.go
```diff
@@ -42,36 +43,53 @@ var (
...
 		},
 	},
 	{
-		ns:          "ns-1",
-		clusterName: "cluster-name-2",
+		clusterSpec: clusterSpec{
+			ns:          "ns-1",
+			clusterName: "cluster-name-2",
+			values: map[string]string{
+				"tidb.separateSlowLog": "true",
```
weekface: Need a case with `"tidb.separateSlowLog": "false"`?

aylei: This option is default to false, but in case of we change the defaults later, I will explicitly declare that case.

weekface:
I rebuilt the dind env.

You should rerun the e2e after this job is completed: https://internal.pingcap.net/jenkins/job/rebuild_k8s_dind_test_env/82/

aylei:
ping @tennix @xiaojingchen

PR #301, opened 2019-03-06
refactor backup

tennix:
This PR refactors ad-hoc backup charts and uses external secrets to do the backup and restore.


charts/tidb-cluster/values.yaml
```diff
@@ -339,20 +339,30 @@ scheduledBackup:
...
+  # And then create the secret by: kubectl create secret generic gcp-backup-secret --from-file=./credentials.json
+  # secretName: gcp-backup-secret
+
   # backup to ceph object storage
   ceph: {}
-    # endpoint: ""
-    # bucket: ""
+  # endpoint: ""
+  # bucket: ""
+  # secretName
```
weekface: Remove this line?

charts/tidb-backup/values.yaml
```diff
@@ -5,61 +5,49 @@
...
+  # Read the following document to create the service account and download the credentials file as credentials.json:
+  # https://cloud.google.com/docs/authentication/production#obtaining_and_providing_service_account_credentials_manually
+  # And then create the secret by: kubectl create secret generic gcp-backup-secret --from-file=./credentials.json
+  # secretName: gcp-backup-secret
+
+# backup to or restore from ceph bucket,  the backup path is in the form of <clusterName>-<name>
+ceph: {}
+  # endpoint: ""
+  # bucket: ""
+  # secretName
```
weekface: ditto

weekface:
@shuijing198799 PTAL

gregwebs:
The mode parameter is creative :) But maybe it means that backup & restore should be separate charts?

I think the operation guide needs mention of using helm namespacing of ad-hoc backup/restore.

charts/tidb-backup/templates/backup-job.yaml
```diff
@@ -1,8 +1,8 @@
-{{- if .Values.backup.create }}
+{{- if eq .Values.mode "backup" }}
```
gregwebs: Can you validate the mode? Maybe have `{{- if neq .Values.mode "restore" && neq .Values.mode "backup" }} ERROR`

tennix:
The backup and restore have a lot of parameters in common, and the chart is really simple. Making it two separate charts is unnecessary.

charts/tidb-backup/templates/backup-job.yaml
tennix: Yes, that should be added. But I can't find any document about `fail` or `error` in Helm template.

charts/tidb-backup/templates/backup-job.yaml
```diff
@@ -45,35 +45,50 @@ spec:
...
         emptyDir: {}
       - name: data
         persistentVolumeClaim:
-          claimName: {{ .Values.clusterName }}-backup
-    {{- if .Values.backup.gcp }}
+          claimName: {{ .Values.clusterName }}-{{ .Values.name }}
+    {{- if .Values.gcp }}
       - name: gcp-credentials
         secret:
-          secretName: {{ .Values.clusterName }}-backup
```
shuijing198799: in the test cases we should initialize two cluster backup and restore to use the same pvc. but the claimName has a prefix clustername, this will not suit our first need.

charts/tidb-backup/templates/backup-job.yaml
gregwebs: I think there is a `fail` function. Helm says they have Sprig functions. http://masterminds.github.io/sprig/flow_control.html

docs/operation-guide.md
```diff
@@ -179,24 +190,36 @@ To create a scheduled full backup job, modify `scheduledBackup` section in `valu
...
 
 ## Restore
 
-Currently, tidb-operator only supports restoring from full backup in GCS bucket. The `restore` section in `values.yaml` should have enough comments as document.
+Restore is similar to backup. See the `values.yaml` file for details.
+
+Modified the variables in `values.yaml` and then create restore job using the following command:
+
+```shell
+$ helm install charts/tidb-backup --name=<backup-name> --namespace=${namespace}
```
gregwebs: I think one would likely be running this with `-set mode=restore` rather than modifying that in values.yaml.

docs/operation-guide.md
```diff
@@ -179,24 +190,36 @@ To create a scheduled full backup job, modify `scheduledBackup` section in `valu
...
 
-If TiDB cluster is running on GKE, the backup data can be uploaded to GCS bucket. A bucket name and base64 encoded service account credential that has bucket read/write access must be provided. The comments in `values.yaml` is self-explanatory for GCP backup.
+```shell
+$ kubectl create secret generic backup-secret -n ${namespace} --from-literal=user=<user> --from-literal=password=<password>
+```
+
+Then run the following command to create an ad-hoc backup job:
+
+```shell
+$ helm install charts/tidb-backup --name=<backup-name> --namespace=${namespace}
```
gregwebs: I would give an example where the backup name is unique:

```
helm install charts/tidb-backup --namespace=${namespace} --name=tidb-backup-$(date -u '+%F-%T') 
```

gregwebs: Similar for the restore below

gregwebs:
I agree that backup & restore are very similar. I would only make it two separate charts if we can share almost all the files. This could be done with symlinks. But helm charts also support sub-charts and passing values down (we do that for GCP marketplace), which would be another approach.

shuijing198799:
GLTM

charts/tidb-backup/templates/backup-job.yaml
```diff
@@ -45,35 +45,50 @@ spec:
...
+              key: password
       restartPolicy: OnFailure
       volumes:
       - name: savepoint-dir
         emptyDir: {}
       - name: data
         persistentVolumeClaim:
-          claimName: {{ .Values.clusterName }}-backup
-    {{- if .Values.backup.gcp }}
+          claimName: {{ .Values.name }}
```
shuijing198799: if the claim name change to {{.valuses.name}} the name in the bakcup-pvc.yaml name should change accord the persistentvolumeclaim.

PR #304, opened 2019-03-08
fix version typo

gregwebs:


charts/tidb-operator/templates/config/_scheduler-policy-json.tpl
```diff
@@ -14,7 +14,7 @@
 {{- if semverCompare "~1.11.0" .Capabilities.KubeVersion.GitVersion }}
     {"name": "CheckNodePIDPressure"},
 {{- end }}
-{{- if semverCompare "<1.12-0" .Capabilities.KubeVersion.GitVersion }}
+{{- if semverCompare "<1.12.0" .Capabilities.KubeVersion.GitVersion }}
```
onlymellb: This is normal usage. refer [here](https://github.com/helm/charts/blob/bc178b63c7df469f1bbccd86629f8cb573a32292/stable/nginx-ingress/templates/controller-service.yaml#L31)

gregwebs: So the purpose of this is to [find pre-release version](https://github.com/Masterminds/semver#working-with-pre-release-versions). I can close this issue, but I wonder if we know if the pre-release versions work?

PR #307, opened 2019-03-11
set the SLOW_LOG_FILE in the startup script

gregwebs:
Relying on go code to set an env variable is brittle, at least brittle enough that I ran into this. We can make sure it is set in the script itself.

Bash is painful, this is my standard technique for making sure variables are set to a default.

@aylei

I tested this as fixing the issue.


gregwebs:
#295 set this in go code.

aylei:
Thank you for your guidance @gregwebs !
&
LGTM

charts/tidb-cluster/templates/scripts/_start_tidb.sh.tpl
```diff
@@ -34,6 +34,7 @@ then
     ARGS="${ARGS} --enable-binlog=true"
 fi
 
+SLOW_LOG_FILE=${SLOW_LOG_FILE:-""}
 if [[ ! -z "${SLOW_LOG_FILE}" ]]
 then
     ARGS="${ARGS} --log-slow-query=${SLOW_LOG_FILE:-}"
```
tennix: Is it necessary to have `${SLOW_LOG_FILE:-}` after setting it on line 37 and checking it on line 38?

gregwebs: On line 38 it will be undefined. The purpose of line 37 is just to define it and set a default value if it is not defined. It is a little verbose and obscure as bash is, but I think this idiom is easy to get used to and understand whereas the other ones are trickier. And overall it is much safer to use these techniques and have undefined variables be an error.

gregwebs: Oh, so I think you are more directly asking about `:-`. That allows for the variable to be unset and not throw an undefined variable error.

```
${parameter:-word}
    If parameter is unset or null, the expansion of word is substituted. 
    Otherwise, the value of parameter is substituted.
```

Issue #278, opened 2019-02-16
Relax hyperkube version match requirement?


kkimdev:
Currently users are supposed to match hyperkube's version to Kubernetes' version by `--set scheduler.kubeSchedulerImage=gcr.io/google-containers/hyperkube:${KUBE_VERSION}`

However, some managed Kubernetes service provides automatic Kubernetes version upgrade which can introduce version mismatch overnight.  It would be great if this exact version match requirement can be relaxed.

I don't know the technical feasibility about this though.

gregwebs:
I think it should be possible for us to default to matching what K8s is running (and even track if it changes)

weekface:
@kkimdev The pods created by TiDB Operator are scheduled by TiDB Scheduler(kube-scheduler + tidb-scheduler). We recommend keeping consistent, but these two versions can be different actually.

gregwebs:
Major version changes of k8s will definitely break this setup.
We can use either helm Capabilities.KubeVersion or a MutatingAdmissionWebhook to set the image tag to match K8s so that the user does not configure this parameter.
The helm approach would require the user to do a helm upgrade when ugprading the K8s version.

Otherwise we would need some code to monitor the K8s version and change the image tag when the K8s version changes.

gregwebs:
We now default the helm chart to try to match the k8s version, and it works on GKE. @kkimdev can you try it out on your K8s install?

kkimdev:
It worked well on both gke and  kubeadm-dind-cluster.  Thanks!

PR #313, opened 2019-03-12
automatically set the scheduler K8s version

gregwebs:
This is tested to work on GKE. If this approach is approved, we will need to update some docs.

There is some relation to #278: after this change one should perform a helm upgrade when the K8s version changes (normally this would only matter for a major version change).

It is also possible to try to construct the string from Capabilities.KubeVersion.Major + Capabilities.KubeVersion.Minor.
However, that give me `1.12+` rather than `1.12.5`. So the GitVersion approach seems to match our user guide.


tennix:
This looks great! Previously we use hyperkube image built by coreos hosted on quay.io. And the image is `quay.io/coreos/hyperkube:v1.10.2_coreos.0`, but the `GitVersion` returned by apiserver is `v1.10.2+coreos.0`. So this approach doesn't works. But now after CoreOS is acquired by RedHat, [quay.io](https://quay.io/repository/coreos/hyperkube?tab=tags) doesn't have newer versions (>= v1.11) of hyperkube image.

I'm not sure if this can still work for other registry images and I guess it's quite fragile. But if it works for `gcr.io/google-containers/hyperkube` I think we can add this to values.yaml to improve GKE users' experience and notify user in the documentation that this is only tested on GKE.

gregwebs:
Users on other platforms besides GKE will still pull this pulbic image from the gcr.io without issue (unless they are behind the GFW). But yes, their GitVersion may not match. We should be able to make a more complicated function that can accomodate different environments. For example that can also remove the `+`.

PR #314, opened 2019-03-12
Change binlog/enable from bool to string

kolbe:
Change 0dada1ec37df88d0bcdc6b6cf687407ba9393d6f to tidb made binlog/enable a string instead of bool.


gregwebs:
We cannot merge this: we need a way to know whether we are using TiDB 2.0 or 3.0 and change the configuration file format appropriately.

I made a request to TiDB to maintain backwards compatibility: https://github.com/pingcap/tidb/issues/9686

tennix:
Yes, if https://github.com/pingcap/tidb/pull/9688 get merged, then the configuration file doesn't need to be changed.

PR #318, opened 2019-03-13
add slack

xiaojingchen:
@weekface @cwen0 @zyguan @shuijing198799 PTAL

### DIFF
--- a/tests/slack/slack.go
+++ b/tests/slack/slack.go
@@ -1,0 +1,137 @@
+package slack
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+var (
+	Channel    string
+	WebhookUrl string
+)
+
+type Field struct {
+	Title string `json:"title"`
+	Value string `json:"value"`
+	Short bool   `json:"short"`
+}
+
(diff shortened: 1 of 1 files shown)


gregwebs:
It doesn't matter much for this since it is just tests, but generally it is nice to publish messages to something like a Kafka queue and have a consumer publish them to Slack. That makes it easy for any other tooling to look at the events. That way only a single publisher need be written the tests and the Slack consumer can potentially be generic to publish from other tools as well.

xiaojingchen:
@gregwebs  that's right, but at present, we use slack to notify, and support other tools is too expensive.
in the other hand, stability testing takes a long time, so it is necessary to add notification mechanism.

tests/slack/slack.go
```diff
@@ -0,0 +1,137 @@
...
+	err := Send(WebhookUrl, "", payload)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func SendGoodMsg(msg string) error {
+	attachment := Attachment{
+		Title: "operator stability test success",
```
weekface:
```suggestion
		Title: "operator stability test succeeded",
```
?

PR #330, opened 2019-03-19
stability: use fault-trigger at e2e tests and add some log

cwen0:
@weekface @xiaojingchen @zyguan @shuijing198799 PTAL


tests/cmd/e2e/config.yaml
```diff
@@ -0,0 +1,23 @@
+nodes:
```
weekface: Remove this file?

PR #333, opened 2019-03-20
delete blockWrite.go use blockwrite.go instead

shuijing198799:
1.delete blockWrite.go


shuijing198799:
@cwen0 @zyguan @xiaojingchen @weekface  PTAL

PR #352, opened 2019-03-27
Update DinD k8s version to v1.12

weekface:
@tennix @onlymellb PTAL


gregwebs:
How about 1.13?

weekface:
We currently only do a lot of stability testing on 1.12, other versions have not yet

weekface:
@tennix PTAL

PR #354, opened 2019-03-27
use release name as default clusterName

tennix:
Making users specify two names for a cluster is verbose and error-prone, if users install two clusters in the same name without specifying the `clusterName` explicitly, then the second cluster would fail to create because of duplicated `clusterName`.

This PR uses release name as default `clusterName`, but also allow users to specify `clusterName` explicitly in order to make the pod name short if the release name is auto-generated long name. For example, if the release name uses UUID, then specifying a shorter `clusterName` explicitly would make the resource name more readable using `kubectl`.


charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -22,3 +22,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- $wtf := $context.Template.Name | replace $last $name -}}
 {{ include $wtf $context }}
 {{- end -}}
+
+{{- define "cluster.name" -}}
+{{- default .Release.Name .Values.clusterName }}
+{{- end -}}
```
weekface: No newline at the end of file

weekface:
Add an e2e test spec to cover this?

PR #376, opened 2019-04-09
update roadmap

tennix:


gregwebs:
This is missing mention of refactoring to CRDs. I suggest we release before the CRD refactor as 1.0 (without beta tags) and then after the CRD refactor as 2.0

PR #378, opened 2019-04-09
Add util class to support to add annotations to Grafana

qiffang:
**Function**
1. Support to add annotations to Grafana
2. Add a errorCount metrics to record how many error occurred.
Prometheus can scrape this metric by xxx:8083/metrics.

**Usage**
1. `cli, _ := NewGrafanaClient("http://$GrafanaIP:3000", "$GrafanaUserName", "$GrafanaPassword", "$PrometheusScrapePort")`
2. `annotation := Annotation{
	dashboardId: $dashboardId,
	panelId: $panelId,
	tags: []string{"tag"},
	timestampInMilliSec: time.Now().Unix() * 1000,
	text: "test",
	}`
3. `cli.AddAnnotation(annotation)`
4.  `cli.IncrErrorCount()`

**Test**
1. Add Unit Test Case - PASS
2. Check the Grafana Annotation - PASS
image
3. Check the metrics by prometheus exporter - PASS
image


pkg/util/annotationUtil.go
```diff
@@ -0,0 +1,151 @@
...
+
+	req, err := http.NewRequest("POST", cli.getAnnotationPath(), bytes.NewBuffer(body))
+	if err != nil {
+		return fmt.Errorf("create request failed, %v", err)
+	}
+
+	req.Header.Add("Accept", "application/json, text/plain, */*")
+	req.Header.Add("Content-Type", "application/json;charset=UTF-8")
+	resp, error := cli.client.Do(req)
+	if error != nil {
```
weekface: use `err` instead of `error`

pkg/util/annotationUtil.go
```diff
@@ -0,0 +1,151 @@
...
+	return prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "errorcount",
+		Help: "record error count",
+		ConstLabels: map[string]string{"fortest": "true"},
+	})
+}
+
+//IncreErrorCountWithAnno increments the errorcount by 1,
+//and add the annotation to grafanan.
+func (cli *client) IncreErrorCountWithAnno(annotation Annotation) error{
```
weekface: we don't need this method. only `addAnnotation` is needed.

pkg/util/annotationUtil.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2018 PingCAP, Inc.
```
weekface: place this package in `tests/metrics`

pkg/util/annotationUtil.go
qiffang: Updated

pkg/util/annotationUtil.go
qiffang: Split it as two functions
1.AddAnnotation
2.IncrErrorCount

pkg/util/annotationUtil.go
qiffang: Done

tests/metrics/annotationUtil.go
```diff
@@ -0,0 +1,150 @@
...
+	initFunc(prometheusExporterPort)
+	u.User = url.UserPassword(userName, password)
+	return &client{
+		baseUrl: *u,
+		client:  &http.Client{},
+	}, nil
+}
+
+func (annotation *Annotation) getBody() ([]byte, error) {
+	m := map[string]interface{}{
```
aylei: consider using struct annotation (`json:"field_name"`) to avoid creating an extra map?

tests/metrics/annotationUtil.go
```diff
@@ -0,0 +1,150 @@
...
+//initFunc is called with sync.Once, we use sync.Once to keep the thread safe.
+func initFunc(port int) {
+	initedOnce.Do(func() {
+		counterMetric = initErrorMetric()
+		prometheus.MustRegister(counterMetric)
+		mux := http.NewServeMux()
+
+		l, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+		if err != nil {
+			fmt.Fprint(os.Stderr, "listening port 8083 failed", err)
```
aylei:
```suggestion
			fmt.Fprintf(os.Stderr, "listening port %d failed, %v", port, err)
```

tests/metrics/annotationUtil.go
```diff
@@ -0,0 +1,150 @@
...
+
+		mux.Handle("/metrics", promhttp.Handler())
+		srv := &http.Server{Handler: mux}
+		go srv.Serve(l)
+	})
+}
+
+func initErrorMetric() prometheus.Counter {
+	return prometheus.NewCounter(prometheus.CounterOpts{
+		Name:        "errorcount",
```
aylei: prefer `error_count`

tests/metrics/annotationUtil.go
```diff
@@ -0,0 +1,150 @@
...
+
+//Client request grafana API on a set of resource paths.
+type client struct {
+	// base is the root URL for all invocations of the client
+	baseUrl url.URL
+	client  *http.Client
+}
+
+//Annotation is a specification of the desired behavior of adding annotation
+type Annotation struct {
```
aylei: Can we add a constructor `NewAnnotation(...)` for better struct defaults?

tests/metrics/annotationUtil.go
aylei: Wait... actually we should use log here

tests/metrics/annotationUtil.go
qiffang: It need more parameters , it is a bad writting.
Actually, tags filed may be empty

tests/metrics/annotationUtil.go
qiffang: accept

tests/metrics/annotationUtil.go
qiffang: Good idea and accept

tests/metrics/annotationUtil.go
aylei: Typically only required fields are added to constructor parameters, in this case I think the client code shall never omit `Timestamp` and `Text`(an annotation don't have proper message is worthless).
After create the struct, optional fields can be populated by simply assignment, e.g. `anno.DashboardId = 1` or by a setter method if validation logic is required.
Moreover, according to the annotation api, we may add `isRegion` and `timeEnd` fields in the struct, which is helpful when we perform some time consuming operations and want to record both the start and end time.

qiffang: tags field can be empty.
Actually, this object should created from out side, it is not duty for annotation_util class.
Each class should has its own responsibilities

tests/metrics/annotation_util.go
```diff
@@ -0,0 +1,140 @@
...
+type client struct {
+	// base is the root URL for all invocations of the client
+	baseUrl url.URL
+	client  *http.Client
+}
+
+//Annotation is a specification of the desired behavior of adding annotation
+type Annotation struct {
+	DashboardId         int      `json: "dashboardId"`
+	PanelId             int      `json: "panelId"`
```
aylei: `DashboardId` and `PanelId` are optional, so we should use `*int` here and add inform serializer to omit these field is empty (nil), the notation is `json:"dashboardId,omitempty"`

aylei: So do `tags`.

qiffang: dashboardId和pannelId明显是不能忽略的啊，要不然知道加哪张图上
tags accept

aylei: The scenario is that users may add the global annotation (without `DashboardId` and `PanelId`) to grafana and query it in multiple dashboards. To me it is the common case because when I am coding I cannot ensure the grafana dashboard id and panel id is consistent in different clusters and different e2e tests or stability tests.

weekface: can we post an annotation without  dashboardID and panelID?

aylei: Yes, https://grafana.com/docs/http_api/annotations/#create-annotation

weekface: We can leave them optional for a wider range of uses. @qiffang

qiffang: From the original requirement,  we want to add annotation in a graph(panel). 
Of course, we can support it.
Accept

tests/pkg/metrics/annotation_util.go
```diff
@@ -0,0 +1,142 @@
...
+
+//Annotation is a specification of the desired behavior of adding annotation
+type Annotation struct {
+	DashboardId         int      `json: "dashboardId, omitempty"`
+	PanelId             int      `json: "panelId, omitempty"`
+	TimestampInMilliSec int64    `json: "time""`
+	Tags                []string `json: "tags, omitempty""`
+	Text                string   `json: "text""`
+	IsRegin             bool     `json: "isRegion, omitempty""`
+	TimeEnd             int64    `json: "timeEnd, omitempty""`
```
aylei: Typo in above 4 lines, double quote not closed

qiffang: Ha， I just commit it for updating grafana pull request.
This change do not finished
Thanks for your help.

PR #383, opened 2019-04-10
Update local-pv documentation

aylei:

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -115,6 +115,16 @@ $ kubectl get po -n kube-system -l app=local-volume-provisioner
 $ kubectl get pv | grep local-storage
 ```
 
+### Remove a mount point
+
+If we want to remove a mount point, first we need to `umount` the mount point, and then delete the related directories. For example:
+
+```shell
+$ umount /mnt/disks/local-pv01
+$ rm -rf /mnt/disks/local-pv01 
+$ rm -rf /data/local-pv01
+```
+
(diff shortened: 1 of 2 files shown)


docs/setup.md
```diff
@@ -106,6 +115,8 @@ $ kubectl get po -n kube-system -l app=local-volume-provisioner
 $ kubectl get pv | grep local-storage
 ```
 
+> Note: The local-volume plugin expects paths to be stable, if you remove a previous mount-point in the discovery directory (default to `/mnt/disks/`), you should remove the PV manually to keep consistency.
```
weekface: First we need to `umount` the `/mnt/disks/local-pv01`, and then delete the two directories.

```
$ umount /mnt/disks/local-pv01
$ rm -rf /mnt/disks/local-pv01 
$ rm -rf /data/local-pv01
```

docs/setup.md
```diff
@@ -106,6 +115,18 @@ $ kubectl get po -n kube-system -l app=local-volume-provisioner
...
 
+### Remove a mount point
+
+If we want to remove a mount point, first we need to `umount` the mount point, and then delete the related directories. For example:
+
+```shell
+$ umount /mnt/disks/local-pv01
+$ rm -rf /mnt/disks/local-pv01 
+$ rm -rf /data/local-pv01
+```
```
onlymellb: You should also delete the related entries in /etc/fstab at the same time, otherwise the automount may cause problems when the machine restarts.

weekface: Do we need to add this sentence to the setup.md?

aylei: Of course, I will add it

PR #386, opened 2019-04-11
ensure test env normal and completed before cases run

xiaojingchen:
currently, stability test directly runs on k8s cluster without the test env check, this may cause invalid test and results. this pr add test env check and recover to ensure it is normal before the stability test cases start running; in the other hand if recover failed, the stability test will be failed,  need to recover  test environment manually.

### DIFF
--- a/tests/cmd/stability/main.go
+++ b/tests/cmd/stability/main.go
@@ -39,6 +39,11 @@ func main() {
 	conf := tests.ParseConfigOrDie()
 	cli, kubeCli := client.NewCliOrDie()
 	oa := tests.NewOperatorActions(cli, kubeCli, conf)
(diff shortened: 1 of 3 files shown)


tests/cmd/stability/main.go
```diff
@@ -39,6 +39,11 @@ func main() {
 	conf := tests.ParseConfigOrDie()
 	cli, kubeCli := client.NewCliOrDie()
 	oa := tests.NewOperatorActions(cli, kubeCli, conf)
+	fta := tests.NewFaultTriggerAction(cli, kubeCli, conf)
+	err := fta.CheckAndRecoverEnv()
```
weekface: CheckAndRecoverEnvOrDie()

PR #388, opened 2019-04-12
Use grafana provisioning to replace dashboard installer

qiffang:
**Function**
In the higher version of grafana,. it improve the experience by adding a new active provisioning system that uses config files. So we can remove dashboard installer logic by the feature.
But there is a problem the file of dashboard is too large exceed the limitation of ConfigMap's annotation field, we have to compress it by gzip. 
The logic  is here:
1. Make a ConfigMap which contains dashboard file named **monitor-dashboard-configmap.yaml**
    The content is binary data
2. Make a ConfigMap which contains dasource file and dashboard provisioning file(This file declare a dashboard path), the name of file is monitor-configmap.yaml
3. Update monitor-deployment file, the spec declare the ConfigMaps to the directory specified
4. Add a poststart logic to decompress the files to the path which dashboard provisioning declare
`postStart:
              exec:
                command:
                  - "/bin/sh"
                  - "-c"
                  - >
                    gzip -dc /tmp/dashboard-gz/tidb.json.gz > /grafana-dashboard-definitions/tidb/tidb.json;
                    gzip -dc /tmp/dashboard-gz/pd.json.gz > /grafana-dashboard-definitions/tidb/pd.json;
                    gzip -dc /tmp/dashboard-gz/tikv.json.gz > /grafana-dashboard-definitions/tidb/tikv.json;
                    gzip -dc /tmp/dashboard-gz/overview.json.gz > /grafana-dashboard-definitions/tidb/overview.json;
`
5. Delete the dashboad installer content

**Test**
1. There is not the Job pod to install dashboard and datasource by API
2. The dashboard works well
image


tennix:
There should be some documents about how the zipped dashboards come from, so other contributors know how to update it when there are newer dashboards in the upstream.

charts/tidb-cluster/templates/monitor-deployment.yaml
```diff
@@ -94,6 +94,17 @@ spec:
       - name: grafana
         image: {{ .Values.monitor.grafana.image }}
         imagePullPolicy: {{ .Values.monitor.grafana.imagePullPolicy | default "IfNotPresent" }}
+        lifecycle:
+            postStart:
```
aylei: I don't whether grafana can discover dashboard json dynamically at runtime, but there is a race condition of `postStart` hook: kubelet cannot ensure the execution order of `postStart` hook and container entrypoint.
So if grafana could discover dashboard jsons at runtime, this is safe. Otherwise we have to ensure the unzip logic executed before grafana runs by init-container or other mechanisms.

tennix: Yes, it is. The Grafana can discover and load dashboard json files at runtime.

qiffang: When Grafana starts, it will update/insert all dashboards available in the configured path. Then later on poll that path every updateIntervalSeconds and look for updated json files and update/insert those into the database
More detail in grafana document

qiffang:
> There should be some documents about how the zipped dashboards come from, so other contributors know how to update it when there are newer dashboards in the upstream.

Accept
1. Add description in monitor-dashboard-configmap.yaml file.
2. Delete invalid graphs from TIDB-Cluster-Overview dashboard("System Info" and "Services Port Status") and expanded pd relation graphs in overview dashboard.
3 . Will create another PR to push these dashboard files to tidb-docker-compose repository.

charts/tidb-cluster/templates/config/_grafana-dashboard.tpl
```diff
@@ -0,0 +1,14 @@
...
+            "folder": "",
+            "name": "0",
+            "options": {
+                "path": "/grafana-dashboard-definitions/tidb"
+            },
+            "orgId": 1,
+            "type": "file"
+        }
+    ]
+}
```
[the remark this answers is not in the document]
qiffang: Accept

xiaojingchen:
why not build an image which contains the config file, avoid to save config files to the configmap?

tennix:
> why not build an image which contains the config file, avoid to save config files to the configmap?

It's more convient to use configmap instead of a static docker image, especially when the dashboards needs to be updated.

charts/tidb-cluster/templates/monitor-deployment.yaml
```diff
@@ -94,6 +94,17 @@ spec:
...
+              exec:
+                command:
+                  - "/bin/sh"
+                  - "-c"
+                  - >
+                    gzip -dc /tmp/dashboard-gz/tidb.json.gz > /grafana-dashboard-definitions/tidb/tidb.json;
```
weekface: use `&&` instead of `;`?

weekface: @qiffang

PR #346, opened 2019-03-21
rewrite e2e

xiaojingchen:
this pr rewrite e2e test and refactor the test code. the test contains e2e test and stability test.

PR #392, opened 2019-04-15
Update e2e workflow in CONTRIBUTING.md

cofyc:
- use `[` instead of `[[` to be sh-compatible (more portable)
- fix e2e workflow in CONTRIBUTING.md
 
`make e2e-docker` builds e2e image from `tests/images/e2e` instead of `images/tidb-operator-e2e`, we should use new e2e yaml file `tests/manifests/e2e/e2e.yaml`.

xref: #346

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,9 +46,9 @@ e2e-docker-push: e2e-docker
 	docker push "${DOCKER_REGISTRY}/pingcap/tidb-operator-e2e:latest"
 
 e2e-docker: e2e-build
-	[[ -d tests/images/e2e/tidb-operator ]] && rm -r tests/images/e2e/tidb-operator || true
-	[[ -d tests/images/e2e/tidb-cluster ]] && rm -r tests/images/e2e/tidb-cluster || true
-	[[ -d tests/images/e2e/tidb-backup ]] && rm -r tests/images/e2e/tidb-backup || true
+	[ -d tests/images/e2e/tidb-operator ] && rm -r tests/images/e2e/tidb-operator || true
+	[ -d tests/images/e2e/tidb-cluster ] && rm -r tests/images/e2e/tidb-cluster || true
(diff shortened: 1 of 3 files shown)


tests/manifests/e2e/e2e.yaml
```diff
@@ -27,7 +27,7 @@ spec:
   serviceAccount: tidb-operator-e2e
   containers:
   - name: tidb-operator-e2e
-    image: ""
```
cofyc: @xiaojingchen 
Not sure why it is empty in your #346 PR, can we give it a default value, then it can be applied by `kubectl`?

weekface:
CI failed, because these is a gofmt issue caused by pr #386, you can fix it in this pr or just ignore it.

cofyc:
It has been fixed [here](https://github.com/pingcap/tidb-operator/pull/390/files#diff-983b6edfbe9124e0a016fc2cb1990c61L113), so I rebased on master and try again.

tests/manifests/e2e/e2e.yaml
xiaojingchen: this's ok. in fact the image will be replaced when e2e ci running.

Makefile
```diff
@@ -46,9 +46,9 @@ e2e-docker-push: e2e-docker
 	docker push "${DOCKER_REGISTRY}/pingcap/tidb-operator-e2e:latest"
 
 e2e-docker: e2e-build
-	[[ -d tests/images/e2e/tidb-operator ]] && rm -r tests/images/e2e/tidb-operator || true
-	[[ -d tests/images/e2e/tidb-cluster ]] && rm -r tests/images/e2e/tidb-cluster || true
-	[[ -d tests/images/e2e/tidb-backup ]] && rm -r tests/images/e2e/tidb-backup || true
+	[ -d tests/images/e2e/tidb-operator ] && rm -r tests/images/e2e/tidb-operator || true
```
xiaojingchen: @onlymellb  PTAL

PR #394, opened 2019-04-15
dashboard:update editable value from true to false

qiffang:
**Function**
Grafana allow to edit dashboard as default value.
For these dashboards which are provided with provisioning way, It is better not to let users edit by default.
So in this PR, just update the editable value from true to fase

**Test**
There is no delete button, customer can enable it in settings page.
image


xiaojingchen:
please run `make check` before commit pr.

PR #397, opened 2019-04-16
Support running stability test out of cluster

aylei:
Deploy & witness flow can be tedious when developing stability-test, this PR introduce the ability to run stability-test out of the cluster(your local machine, usually) while still operating the remote cluster.

**How-to**: generally we have three problems to solve: 

### TL;DR: 
```shell
$ telepresence --new-deployment ${POD_NAME}
$ go build -o stability ./tests/cmd/stability/main.go
$ ./stability --operator-repo-dir=${ABITRARY_EMPTY_DIR_TO_CLONE_OPERATOR_REPO}
```

### Explained
1. **Out of cluster client**: Now we try to load configs in the following order:
    * if `kubeconfig` command line option provided, use it
    * if `KUBECONFIG` env variable set, use it
    * try loading `InClusterConfig()`
    * if no `InClusterConfig()` provided, try loading kubeconfig file from default location (`~/.kube/config`)
so typically you will get right client configs if you have proper `~/.kube/config` in your local environment. 
2. **Privilege issue**: If you don't want to or cannot run stability test with root privilege, change the working dir or create it in advance:
    * git repo dir can be overridden by option `--git-repo-dir=xxxx`, but helm dir must be created manually. 
```shell
# helm dir
$ mkdir /charts
$ chmod 777 /charts
# git repo dir if you don't set command line option
$ mkdir /tidb-operator
$ chmod 777 /tidb-operator
```
3. **DNS and network issue**: Two-way proxy using Telepresence. We cannot resolve cluster dns name and access cluster ip easily, `telepresence` helps with that, it creates a proxy pod in the cluster and open a vpn connection to kubernetes cluster via this pod. Just run ([full documentations](https://www.telepresence.io/reference/install)):
```shell
$ brew cask install osxfuse
$ brew install datawire/blackbird/telepresence
$ telepresence --new-deployment ${POD_NAME}
``` 
**PS**: If you cannot resolve cluster dns names after set up, try clear DNS cache.
**PSS**: Typically you can't use telepresence VPN mode with other VPNs (of course SSR is ok).

### DIFF
--- a/tests/actions.go
+++ b/tests/actions.go
@@ -1319,7 +1319,7 @@ func releaseIsNotFound(err error) bool {
 }
 
 func (oa *operatorActions) cloneOperatorRepo() error {
-	cmd := fmt.Sprintf("git clone https://github.com/pingcap/tidb-operator.git %s", oa.cfg.GitRepoDir)
+	cmd := fmt.Sprintf("git clone https://github.com/pingcap/tidb-operator.git %s", oa.cfg.OperatorRepoDir)
 	glog.Info(cmd)
 	res, err := exec.Command("/bin/sh", "-c", cmd).CombinedOutput()
 	if err != nil && !strings.Contains(string(res), "already exists") {
@@ -1337,7 +1337,7 @@ func (oa *operatorActions) checkoutTag(tagName string) error {
 		cp -rf charts/tidb-operator /charts/%s/tidb-operator &&
 		cp -rf charts/tidb-cluster /charts/%s/tidb-cluster &&
 		cp -rf charts/tidb-backup /charts/%s/tidb-backup`,
-		oa.cfg.GitRepoDir, tagName, tagName, tagName, tagName, tagName)
+		oa.cfg.OperatorRepoDir, tagName, tagName, tagName, tagName, tagName)
 	glog.Info(cmd)
 	res, err := exec.Command("/bin/sh", "-c", cmd).CombinedOutput()
 	if err != nil {
--- a/tests/actions.go
+++ b/tests/actions.go
@@ -234,10 +234,10 @@ func (oi *OperatorConfig) OperatorHelmSetString(m map[string]string) string {
 
 func (oa *operatorActions) DeployOperator(info *OperatorConfig) error {
 	if info.Tag != "e2e" {
-		if err := cloneOperatorRepo(); err != nil {
+		if err := oa.cloneOperatorRepo(); err != nil {
 			return err
 		}
-		if err := checkoutTag(info.Tag); err != nil {
+		if err := oa.checkoutTag(info.Tag); err != nil {
 			return err
 		}
 	}
@@ -280,7 +280,7 @@ func (oa *operatorActions) CleanOperatorOrDie(info *OperatorConfig) {
 }
 
 func (oa *operatorActions) UpgradeOperator(info *OperatorConfig) error {
-	if err := checkoutTag(info.Tag); err != nil {
+	if err := oa.checkoutTag(info.Tag); err != nil {
 		return err
 	}
 
@@ -1329,8 +1329,8 @@ func releaseIsNotFound(err error) bool {
 	return strings.Contains(err.Error(), "not found")
 }
 
-func cloneOperatorRepo() error {
-	cmd := fmt.Sprintf("git clone https://github.com/pingcap/tidb-operator.git /tidb-operator")
+func (oa *operatorActions) cloneOperatorRepo() error {
+	cmd := fmt.Sprintf("git clone https://github.com/pingcap/tidb-operator.git %s", oa.cfg.GitRepoDir)
 	glog.Info(cmd)
 	res, err := exec.Command("/bin/sh", "-c", cmd).CombinedOutput()
 	if err != nil && !strings.Contains(string(res), "already exists") {
@@ -1340,15 +1340,15 @@ func cloneOperatorRepo() error {
 	return nil
 }
 
-func checkoutTag(tagName string) error {
-	cmd := fmt.Sprintf(`cd /tidb-operator &&
+func (oa *operatorActions) checkoutTag(tagName string) error {
+	cmd := fmt.Sprintf(`cd %s &&
 		git stash -u &&
 		git checkout %s &&
 		mkdir -p /charts/%s &&
 		cp -rf charts/tidb-operator /charts/%s/tidb-operator &&
 		cp -rf charts/tidb-cluster /charts/%s/tidb-cluster &&
 		cp -rf charts/tidb-backup /charts/%s/tidb-backup`,
-		tagName, tagName, tagName, tagName, tagName)
+		oa.cfg.GitRepoDir, tagName, tagName, tagName, tagName, tagName)
 	glog.Info(cmd)
 	res, err := exec.Command("/bin/sh", "-c", cmd).CombinedOutput()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


tests/config.go
```diff
@@ -39,6 +46,15 @@ func NewConfig() *Config {
...
 	flag.StringVar(&cfg.OperatorTag, "operator-tag", "master", "operator tag used to choose charts")
 	flag.StringVar(&cfg.OperatorImage, "operator-image", "pingcap/tidb-operator:latest", "operator image")
+	flag.StringVar(&cfg.GitRepoDir, "git-repo-dir", "/tidb-operator", "local directory to which tidb-operator cloned")
+	if home := homeDir(); home != "" {
+		flag.StringVar(&cfg.KubeConfigPath, "kubeconfig", filepath.Join(home, ".kube", "config"), "absolute path to the kubeconfig file")
+	} else {
+		// cannot get homedir, kube config file path must be provided explicitly
+		flag.StringVar(&cfg.KubeConfigPath, "kubeconfig", "", "absolute path to the kubeconfig file")
+	}
+	flag.BoolVar(&cfg.OutOfCluster, "out-of-cluster", false, "whether stability test runs out of cluster.")
```
cofyc: Maybe we can omit `out-of-cluster` flag,  an idiomatic way to load kube config is

- load config fro kubeconfig file if given
- load in-cluster config

aylei: How to omit kubeconfig option (default to `~/.kube/config`) in this case?

aylei: How about using `controller-runtime/pkg/client/config/config.go#GetConfig()`, seems that method is well designed.

tests/config.go
```diff
@@ -39,6 +46,15 @@ func NewConfig() *Config {
 	flag.StringVar(&cfg.TidbVersions, "tidb-versions", "v2.1.3,v2.1.4", "tidb versions")
 	flag.StringVar(&cfg.OperatorTag, "operator-tag", "master", "operator tag used to choose charts")
 	flag.StringVar(&cfg.OperatorImage, "operator-image", "pingcap/tidb-operator:latest", "operator image")
+	flag.StringVar(&cfg.GitRepoDir, "git-repo-dir", "/tidb-operator", "local directory to which tidb-operator cloned")
```
xiaojingchen: change `git-repo-dir` to `operator-repo-dir`

tests/config.go
cofyc: https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client/config#GetConfig seems very good. It helps to make our CLI behave as other k8s CLI commands.

tests/config.go
```diff
@@ -39,6 +46,15 @@ func NewConfig() *Config {
...
+	if home := homeDir(); home != "" {
+		flag.StringVar(&cfg.KubeConfigPath, "kubeconfig", filepath.Join(home, ".kube", "config"), "absolute path to the kubeconfig file")
+	} else {
+		// cannot get homedir, kube config file path must be provided explicitly
+		flag.StringVar(&cfg.KubeConfigPath, "kubeconfig", "", "absolute path to the kubeconfig file")
```
onlymellb: I think use kubeconfig like [this](https://github.com/pingcap/private-tidb-operator/blob/master/cmd/volume-manager/main.go#L80) is simpler and clearer.  If you specify kubeconfig, it means running outside k8s cluster, otherwise it means running in k8s cluster.

aylei:
@cofyc @onlymellb @xiaojingchen PTAL again

tests/config.go
```diff
@@ -22,6 +22,9 @@ type Config struct {
 	Nodes            []Nodes `yaml:"nodes" json:"nodes"`
 	ETCDs            []Nodes `yaml:"etcds" json:"etcds"`
 	APIServers       []Nodes `yaml:"apiservers" json:"apiservers"`
+
+	// For local test
+	OperatorRepoDir string `yaml:"git_repo_dir" json:"gir_repo_dir"`
```
xiaojingchen: typo: `git_repo_dir`

aylei: Good catch

aylei:
I've also updated the docs

tests/pkg/client/client.go
```diff
@@ -11,31 +16,52 @@ import (
...
+	}
+	// If no explicit location, try the in-cluster config
+	if c, err := rest.InClusterConfig(); err == nil {
+		return c, nil
+	}
+	// If no in-cluster config, try the default location in the user's home directory
+	if usr, err := user.Current(); err == nil {
+		if c, err := clientcmd.BuildConfigFromFlags(
+			"", filepath.Join(usr.HomeDir, ".kube", "config")); err == nil {
+			return c, nil
```
onlymellb: This operation is dangerous and may operate an unexpected k8s cluster, so we should avoid automatically finding the kubeconfig file in the user's home directory to create the k8s client.

tests/actions.go
```diff
@@ -1329,15 +1329,15 @@ func cloneOperatorRepo() error {
...
-func checkoutTag(tagName string) error {
-	cmd := fmt.Sprintf(`cd /tidb-operator &&
+func (oa *operatorActions) checkoutTag(tagName string) error {
+	cmd := fmt.Sprintf(`cd %s &&
 		git stash -u &&
 		git checkout %s &&
 		mkdir -p /charts/%s &&
 		cp -rf charts/tidb-operator /charts/%s/tidb-operator &&
 		cp -rf charts/tidb-cluster /charts/%s/tidb-cluster &&
 		cp -rf charts/tidb-backup /charts/%s/tidb-backup`,
```
cofyc: (do not block this PR) Is it possible to copy them to temporary directories which are always writable in the future? I guess the path does not matter here. Then we can simplify the workflow (do not need to prepare `/charts`).

aylei: Agreed for future work, I've tried but found that `/charts` in hard coded in many places😅

tests/pkg/client/client.go
aylei: This is same with someone execute `kubectl delete sth.` without confirming the context first, IMHO

cofyc: It makes sense to avoid operating on the unexpected k8s cluster, but even if users need to explicitly specify `--kubeconfig ~/.kube/config`, they may run against the wrong cluster because they forgot to switch context. Maybe we can ask for confirmation or/both add some preflight checks?

- show cluster information and ask the user for confirmation before proceeding (and add `--skip-confirm` to skip the confirmation explicitly in the batch job)
- add some preflight checks, e.g. check the version of tidb operator matches expected version

onlymellb: This is completely different from someone executes a dangerous command like `kubectl delete sth`, This is the user's own behavior, and your program may perform dangerous operations without the user's perception.

aylei: @cofyc @onlymellb I agree the safety concern, however, there is always a trade-off between convenient, safety and complexity. I think a reasonable solution for now is remind user of this in the documetation, how do you think?

aylei: After all, out-of-cluster stability test is for convenient "stability test developing" but not actually running.

onlymellb: I just don't think it's necessary to automatically find the kubeconfig file in the user's home directory to create the k8s client.  If the user does not specify kubeconfig explicitly and does not run the program in the k8s cluster, let it exit directly. What do you think？ @weekface @tennix @xiaojingchen

weekface: I agree with @onlymellb

aylei: emmm... I obey the majority then

aylei:
@cofyc @xiaojingchen @weekface @onlymellb I've updated the doc, PTAL again.

PR #398, opened 2019-04-16
update tidb secret docs and charts

LinuxGit:


onlymellb:
Can you separate the python script in [tidb-initializer-job.yaml](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/templates/tidb-initializer-job.yaml#L30-L52) by the way?

LinuxGit:
> Can you separate the python script in [tidb-initializer-job.yaml](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/templates/tidb-initializer-job.yaml#L30-L52) by the way?

OK

docs/operation-guide.md
```diff
@@ -34,10 +34,17 @@ By default TiDB service is exposed using [`NodePort`](https://kubernetes.io/docs
 $ kubectl get svc -n ${namespace} # check the available services
 ```
 
-By default the TiDB cluster has no password set. You can specify a password by setting `tidb.password` in `values.yaml` before deploying. You can retrieve the password from the initialization `Secret`:
+By default the TiDB cluster has no password set. You can create `Secret` with following command, and uncomment setting `tidb.passwordSecretName` in `values.yaml` before deploying:
```
gregwebs:
```suggestion
By default the TiDB cluster has no root password set. Setting a password in helm is insecure. Instead you can set the name of a K8s secret as `tidb.passwordSecretName` in `values.yaml`. Note that this is only used to initialize users: once your tidb cluster is initialized you may delete the secret. The format of the secret is `user=password`, so you can set the root user password with:
```

LinuxGit: good suggestion

LinuxGit:
@onlymellb PTAL

Issue #223, opened 2018-12-10
tidb-operator default image tag “latest” need “imagePullPolicy: Always”


malc0lm:
My cluster deploy tidb-operator 3 month ago. But when I use helm to deploy tidb-operator again, the tidb-controller-manager pod can't start again. The log output is that "this is no auto-failover flag". Then I found I use the outdated "pingcap/tidb-operator:latest" which was pulled 3 month ago.
So I think the "latest" tag need  "imagePullPolicy: Always" config unless you specify a tag name.

weekface:
We may use an explicit tag instead, for example: `pingcap/tidb-operator:v1.0.0-beta.0`

@tennix what do you think?

gregwebs:
we should be using release tags in the chart values so that the pull policy doesn't have an effect.

tennix:
Yeah, I think we should use the last release tag in chart. And update the chart every time we release a new version.

gregwebs:
@tennix do you want me to send a PR for this?

tennix:
Sure, thank you! @gregwebs

tennix:
Fixed with https://github.com/pingcap/tidb-operator/pull/478

PR #403, opened 2019-04-16
default to a released version of the operator

gregwebs:
This is the latest release of the operator and the version that I use.

### DIFF
--- a/charts/tidb-operator/values.yaml
+++ b/charts/tidb-operator/values.yaml
@@ -9,7 +9,7 @@ rbac:
   create: true
 
 # operatorImage is TiDB Operator image
-operatorImage: pingcap/tidb-operator:latest
+operatorImage: pingcap/tidb-operator:v1.0.0-beta.1-p2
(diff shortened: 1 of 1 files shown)


weekface:
There are many [fixes](https://github.com/pingcap/tidb-operator/releases) from `v1.0.0-beta.1-p2`, so it is better to use `latest` instead of `v1.0.0-beta.1-p2` before the next release.

cofyc:
I guess this is going to fix https://github.com/pingcap/tidb-operator/issues/223. 

It's best practice to use a version tag in production deployment, we should recommend users to use the version tag instead of `latest`.  Using `latest`, users may encounter an unexpected upgrade when pods are migrated to another node even if they did nothing. Because users may apply our default values without modification, so this LGTM.

If `v1.0.0-beta.1-p2` is not a recommended version, we can hold this PR until the next release.

gregwebs:
@weekface you need to create a new stable release then

weekface:
Yes, we will release a new version recently.

PR #405, opened 2019-04-17
use `ioutil.TempDir()` create charts and operator repo's directories 

xiaojingchen:
stability test needs some dirs to store charts files and tidb-operator repo, these dirs is useful only during case running, but if user run stability test on local these directories may require authorization to access.
this pr use `ioutil.TempDir()` to replace the directories which were specified in the code.


tests/cmd/e2e/main.go
```diff
@@ -32,14 +32,10 @@ func main() {
 	logs.InitLogs()
 	defer logs.FlushLogs()
 
-	conf := tests.NewConfig()
-	err := conf.Parse()
-	if err != nil {
-		glog.Fatalf("failed to parse config: %v", err)
-	}
+	conf := tests.ParseConfigOrDie()
+	conf.ChartDir = "/charts"
```
aylei: The default value should be set before configuration file load, otherwise this will override anything

tests/config.go
```diff
@@ -34,22 +36,35 @@ type Nodes struct {
...
+		return nil, err
+	}
+	cfg.OperatorRepoDir = operatorRepo
+
+	chartDir, err := ioutil.TempDir("", "charts")
+	if err != nil {
+		return nil, err
+	}
+	cfg.ChartDir = chartDir
+	return cfg, nil
```
cofyc: Where is the place to clean these temporary directories?

tests/cmd/e2e/main.go
xiaojingchen: it is right, e2e test must set `conf.ChartDir` as "/charts" becuase the charts file is add by CI.

tests/config.go
xiaojingchen: done

PR #407, opened 2019-04-18
Fix and speed up local DinD e2e testing

cofyc:
- allocate30 (3x10) PVs by default (at most we need 3 * (5 + 5) PVs)
- reduce request resources of e2e pods, make it possible to run on
low resources machine
  - IIUC, we don't care about this in e2e testing
- smaller poll interval to speed up e2e 
  - checking against API is low overhead, smaller interval speeds up a lot
  - suggested by @xiaojingchen, only the e2e poll interval is changed


cofyc:
pushed a new commit to resolve conflicts, @weekface @xiaojingchen PTAL.

PR #412, opened 2019-04-20
Support plugin start argument for tidb instance

sdojjy:
Support plugin start argument for tidb instance. The default value is disabled

Tests
- enable the plugin,  ARGS is appended as expected.
` ARGS="${ARGS}  --plugin-dir  /pulugins --plugin-load whitelist-1  "`
- disable the plugin, no plugin argument is added

### DIFF
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -194,10 +194,10 @@ tidb:
 
   # tidb plugin configuration
   plugin:
-    # enable pulgin or not
+    # enable plugin or not
     enable: false
     # the start argument to specify the folder containing
-    directory: /pulugins
+    directory: /plugins
     # the start argument to specify the plugin id (name "-" version) that needs to be loaded, e.g. 'conn_limit-1'.
     list: whitelist-1
 
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -199,7 +199,7 @@ tidb:
     # the start argument to specify the folder containing
     directory: /plugins
     # the start argument to specify the plugin id (name "-" version) that needs to be loaded, e.g. 'conn_limit-1'.
-    list: whitelist-1
+    list: ["whitelist-1"]
 
 # mysqlClient is used to set password for TiDB
 # it must has Python MySQL client installed
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -192,6 +192,15 @@ tidb:
         cpu: 20m
         memory: 5Mi
 
+  # tidb plugin configuration
+  plugin:
+    # enable pulgin or not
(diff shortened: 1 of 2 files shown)


charts/tidb-cluster/values.yaml
```diff
@@ -192,6 +192,15 @@ tidb:
         cpu: 20m
         memory: 5Mi
 
+  # tidb plugin configuration
+  plugin:
+    # enable pulgin or not
```
LinuxGit: s/pulgin/plugin

charts/tidb-cluster/values.yaml
```diff
@@ -192,6 +192,15 @@ tidb:
...
+    enable: false
+    # the start argument to specify the folder containing
+    directory: /pulugins
```
LinuxGit: s/pulugins//plugins

sdojjy: fixed

charts/tidb-cluster/values.yaml
sdojjy: fixed

charts/tidb-cluster/values.yaml
```diff
@@ -192,6 +192,15 @@ tidb:
...
         memory: 5Mi
 
+  # tidb plugin configuration
+  plugin:
+    # enable plugin or not
+    enable: false
+    # the start argument to specify the folder containing
+    directory: /plugins
+    # the start argument to specify the plugin id (name "-" version) that needs to be loaded, e.g. 'conn_limit-1'.
+    list: whitelist-1
```
LinuxGit: If there're multiple plugins, is list a comma separated string? I found a example, but not sure It works or not.https://github.com/helm/charts/blob/master/stable/opa/templates/deployment.yaml#L96

sdojjy: yes!
From the design document.
https://github.com/pingcap/tidb/blob/master/docs/design/2018-12-10-plugin-framework.md#Management

and tidb uage document 
https://github.com/pingcap/tidb/blob/11d7199fcfbb66b1205ffa00f770f90f7b1ee9c1/tidb-server/main.go#L114

and the code that load plugins
https://github.com/pingcap/tidb/blob/11d7199fcfbb66b1205ffa00f770f90f7b1ee9c1/session/session.go#L1305
```go
	if len(cfg.Plugin.Load) > 0 {
		err := plugin.Load(context.Background(), plugin.Config{
			Plugins:        strings.Split(cfg.Plugin.Load, ","),
			PluginDir:      cfg.Plugin.Dir,
			GlobalSysVar:   &variable.SysVars,
			PluginVarNames: &variable.PluginVarNames,
		})
		if err != nil {
			return nil, err
		}
	}
```

sdojjy: the plugin.list is changed to array

PR #432, opened 2019-04-26
no need to set ClusterIP when syncing headless service

LinuxGit:

remove set ClusterIP operation when syncing headless service

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

Related changes

### DIFF
--- a/pkg/manager/member/pd_member_manager.go
+++ b/pkg/manager/member/pd_member_manager.go
@@ -152,8 +152,6 @@ func (pmm *pdMemberManager) syncPDHeadlessServiceForTidbCluster(tc *v1alpha1.Tid
 	if !equal {
 		svc := *oldSvc
 		svc.Spec = newSvc.Spec
-		// TODO add unit test
-		svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP
 		err = SetServiceLastAppliedConfigAnnotation(newSvc)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


LinuxGit:
I added a commit to remove op in tidb_member_manager.go and dismissed your review.
PTAL again. @weekface

pkg/manager/member/pd_member_manager.go
```diff
@@ -152,8 +152,6 @@ func (pmm *pdMemberManager) syncPDHeadlessServiceForTidbCluster(tc *v1alpha1.Tid
 	if !equal {
 		svc := *oldSvc
 		svc.Spec = newSvc.Spec
-		// TODO add unit test
-		svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP
```
xiaojingchen: this line is to ensure that the `ClusterIP` remains unchanged when svc updated.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -115,8 +115,6 @@ func (tmm *tidbMemberManager) syncTiDBHeadlessServiceForTidbCluster(tc *v1alpha1
 	if !equal {
 		svc := *oldSvc
 		svc.Spec = newSvc.Spec
-		// TODO add unit test
-		svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP
```
xiaojingchen: ditto

pkg/manager/member/pd_member_manager.go
LinuxGit: Yes. syncPDServiceForTidbCluster should keep ClusterIP unchanged. But when syncPDHeadlessServiceForTidbCluster, there's no need to set it, it should be default value of  'None' from newSvc.Spec.

xiaojingchen: ok, that's right.

PR #435, opened 2019-04-29
Use TiDB v2.1.8 by default & remove pushgateway

tennix:
This PR removes pushgateway sidecar to embrace newer version (>=v2.1.3) of TiKV. And also upgrade default TiDB to v2.1.8.

* TiKV pod pushgateway sidecar is removed
* Default TiDB version upgraded to v2.1.8
* Remove unused privileged tidb

### Check List 

Tests 

1. Build new tidb-operator and install into DinD via `helm install charts/tidb-operator --name=tidb-operator --namespace=tidb-admin`
2. Install tidb-cluster via `helm install charts/tidb-cluster --name=tidb-cluster --namespace=tidb`

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

 - Breaking backward compatibility: old version of TiDB cluster (< v2.1.3) monitor would fail to load TiKV dashboard.

Related changes


gregwebs:
Is "Remove unused privileged tidb" related to this change? Can this be done in a separate PR? I think privileged tidb for password reset is still a useful concept.

charts/tidb-cluster/templates/config/_drainer-config.tpl
```diff
@@ -24,14 +27,18 @@ pd-urls = "http://{{ template "cluster.name" . }}-pd:2379"
 # syncer Configuration.
 [syncer]
 
-# disable sync these schema
-ignore-schemas = {{ .Values.binlog.drainer.ignoreSchemas | default "INFORMATION_SCHEMA,PERFORMANCE_SCHEMA,mysql" | quote }}
+# Assume the upstream sql-mode.
+# If this is setted , will use the same sql-mode to parse DDL statment, and set the same sql-mode at downstream when db-type is mysql.
```
gregwebs:
```suggestion
# If this is set, will use the same sql-mode to parse DDL statment, and set the same sql-mode at downstream when db-type is mysql.
```

charts/tidb-cluster/templates/config/_drainer-config.tpl
```diff
@@ -24,14 +27,18 @@ pd-urls = "http://{{ template "cluster.name" . }}-pd:2379"
...
+# If this is not setted, it will not set any sql-mode.
```
gregwebs:
```suggestion
# The default value will not set sql modes.
```

charts/tidb-cluster/templates/config/_grafana-prestart.tpl
```diff
@@ -0,0 +1,21 @@
...
+gzip -dc /tmp/dashboard-gz/pd.json.gz > /grafana-dashboard-definitions/tidb/pd.json
+gzip -dc /tmp/dashboard-gz/tikv.json.gz > /grafana-dashboard-definitions/tidb/tikv.json
+gzip -dc /tmp/dashboard-gz/overview.json.gz > /grafana-dashboard-definitions/tidb/overview.json
+
+#replace dashboard name using cluster name
+clusterName={{ template "cluster.name" . }}
+
+if [ $clusterName != "" ]
+then
+    sed -i 's/TIDB-Cluster-TiDB/'$clusterName'-TIDB/g'  /grafana-dashboard-definitions/tidb/tidb.json
```
weekface:
```suggestion
    sed -i 's/TIDB-Cluster-TiDB/'$clusterName'-TiDB/g'  /grafana-dashboard-definitions/tidb/tidb.json
```

charts/tidb-cluster/templates/config/_grafana-prestart.tpl
```diff
@@ -0,0 +1,21 @@
...
+    sed -i 's/TIDB-Cluster-PD/'$clusterName'-PD/g'  /grafana-dashboard-definitions/tidb/pd.json
+    sed -i 's/TIDB-Cluster-TiKV/'$clusterName'-TIKV/g'  /grafana-dashboard-definitions/tidb/tikv.json
```
weekface:
```suggestion
    sed -i 's/TIDB-Cluster-TiKV/'$clusterName'-TiKV/g'  /grafana-dashboard-definitions/tidb/tikv.json
```

charts/tidb-cluster/templates/config/_tikv-config.tpl
```diff
@@ -18,21 +18,12 @@ log-level = {{ .Values.tikv.logLevel | default "info" | quote }}
 # log-rotation-timespan = "24h"
 
 [readpool.storage]
-{{- if .Values.tikv.readpoolStorageConcurrency }}
```
weekface: delete `readpoolStorageConcurrency ` in the `values.yaml`

charts/tidb-cluster/templates/config/_tikv-config.tpl
```diff
@@ -120,11 +119,7 @@ low-concurrency = {{ .Values.tikv.readpoolCoprocessorConcurrency }}
 
 # scheduler's worker pool size, should increase it in heavy write cases,
 # also should less than total cpu cores.
-{{- if .Values.tikv.storageSchedulerWorkerPoolSize }}
```
weekface: delete `storageSchedulerWorkerPoolSize ` in the values.yaml

charts/tidb-cluster/values.yaml
```diff
@@ -155,10 +144,32 @@ tidb:
...
+  txnEntryCountLimit: "300000"
+  # The limitation of the size in byte for the entries in one transaction.
+  # If using TiKV as the storage, the entry represents a key/value pair.
+  # WARNING: Do not set the value too large, otherwise it will make a very large impact on the TiKV cluster.
+  # Please adjust this configuration carefully.
+  txnTotalSizeLimit: "104857600"
+  # enableBatchDml enables batch commit for the DMLs
+  enableBatchDml: false
+  # check mb4 value in utf8 is used to control whether to check the mb4 characters when the charset is utf8.
+  checkMb4VauleInUtf8: true
```
weekface: This var is not used

charts/tidb-cluster/templates/config/_grafana-config.tpl
```diff
@@ -204,13 +204,13 @@ root_url = %(protocol)s://%(domain)s:%(http_port)s/
 #################################### Anonymous Auth ##########################
 [auth.anonymous]
 # enable anonymous access
-;enabled = false
+enabled = {{ .Values.monitor.grafana.anonymousEnabled | default true }}
```
gregwebs: I really want us to avoid the busywork of eventually templating every variable here. I would prefer to back out grafana changes from this PR and separately change them as per #449. Or you can take the changes in #449 here. With #449 we wouldn't change things here and the values would have `"auth.anonymous.enabled": true`, etc.

tennix: OK, I'll remove the configmap.

gregwebs: I think its still okay to load the grafana.ini file if someone wants to configure grafana through that. Its just that we won't template the file from helm.

tennix: What if both grafana.ini and environment variables are set? Leaving two different configurations confuses users.

gregwebs: Grafana has already chosen to allow multiple ways to configure their application. Our goal should be to just transparently pass-through configuration to it. Grafana documents how different configuration options interact.

tennix: But the configmap is hidden in the template directory, users can hardly configure it. So it's unnecessary to leave it there.

charts/tidb-cluster/templates/monitor-configmap.yaml
```diff
@@ -19,7 +19,7 @@ data:
 {{ tuple "config/_grafana-datasource.tpl" . | include "helm-toolkit.utils.template" | indent 4 }}
   dashboard-config: |-
 {{ tuple "config/_grafana-dashboard.tpl" . | include "helm-toolkit.utils.template" | indent 4 }}
-  grafana-config: |-
-{{ tuple "config/_grafana-config.tpl" . | include "helm-toolkit.utils.template" | indent 4 }}
+  grafana-prestart: |
+{{- include "setup" . | indent 4 }}
```
onlymellb: Why not use the same way with other [configmaps](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/templates/pd-configmap.yaml#L13) that include scripts? The scripts you define in this way cannot be executed separately.

tennix:
@onlymellb I've addressed your comment, PTAL again.

charts/tidb-cluster/templates/config/_grafana-config.tpl
gregwebs: We have several configurations there including that for pd/tikv/tidb so it is important that we document for users how to modify these files. Alternatively we could document how they can add a config file.

I found the helm chart for grafana to be confusing because one generates it from helm values very similar to the env variables: ` --set 'grafana\.ini.auth\.anonymous.enabled'='true' `

Issue #404, opened 2019-04-17
Add instructions to access Grafana dashboard in GKE tutorial


cofyc:
Follow up from https://github.com/pingcap/tidb-operator/pull/400#discussion_r276067770.

gregwebs:
fixed by #448

PR #448, opened 2019-05-01
Adds instructions to access Grafana in GKE tutorial

jlerche:
#404 

Adds instruction in the GKE tutorial to access Grafana dashboard

### Check List 

Tests 

 - No code

Code changes

 - Has documents change

Side effects

Related changes

### DIFF
--- a/docs/google-kubernetes-tutorial.md
+++ b/docs/google-kubernetes-tutorial.md
@@ -157,7 +157,8 @@ To do so, use the following command:
 
     kubectl -n tidb port-forward svc/demo-grafana 3000:3000 &>/dev/null &
 
-Now, point a browser to `localhost:3000` to access the dashboards.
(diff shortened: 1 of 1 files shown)


tennix:
I think it's better to use a load balancer for the tutorial, so new users can easily access it via a public DNS name.
Adding extra value: `monitor.grafana.service.type=LoadBalancer` when installing TiDB cluster.
```shell
helm install ./charts/tidb-cluster -n demo --namespace=tidb --set pd.storageClassName=pd-ssd,tikv.storageClassName=pd-ssd,monitor.grafana.service.type=LoadBalancer
```
And then get grafana service external IP: `kubectl get svc -n tidb demo-grafana`

gregwebs:
@tennix that will create additional security concerns that I think we should avoid getting into during the tutorial. But your comment would make sense to add to the user guide.

gregwebs:
We can merge this now and keep improving it as needed.

PR #449, opened 2019-05-03
make grafana configuration extensible

gregwebs:
This way we don't have to do any additional grafana config templating

@tennix this is relevant to changes in #435

I have properly tested this change but not on this version of the charts I just pull-requested to master.

### DIFF
--- a/charts/tidb-cluster/templates/monitor-deployment.yaml
+++ b/charts/tidb-cluster/templates/monitor-deployment.yaml
@@ -114,18 +114,6 @@ spec:
           containerPort: 3000
           protocol: TCP
         env:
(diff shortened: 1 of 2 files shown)


charts/tidb-cluster/templates/monitor-deployment.yaml
```diff
@@ -138,6 +126,12 @@ spec:
             secretKeyRef:
               name: {{ template "cluster.name" . }}-monitor
               key: password
+         {{- if .Values.monitor.grafana.config }}
```
tennix: This looks nice!

gregwebs:
Something similar to this change is now in master

PR #462, opened 2019-05-07
Fix bug use shareinformer without copy

shuijing198799:
Because the data in shareinformer is modified instead of using copy,，When the TiDB statefulset rolling upgrade process is interrupted, it will not be retried.
use copy data instead.

Tests 

 - E2E test
 - Stability test

Code changes

 - Has Go code change

Side effects

 - effect the TiDB rolling update process

### DIFF
--- a/pkg/manager/member/tidb_member_manager.go
+++ b/pkg/manager/member/tidb_member_manager.go
@@ -131,7 +131,7 @@ func (tmm *tidbMemberManager) syncTiDBStatefulSetForTidbCluster(tc *v1alpha1.Tid
 	tcName := tc.GetName()
 
 	newTiDBSet := tmm.getNewTiDBSetForTidbCluster(tc)
-	oldTiDBSet, err := tmm.setLister.StatefulSets(ns).Get(controller.TiDBMemberName(tcName))
+	oldTiDBSetTemp, err := tmm.setLister.StatefulSets(ns).Get(controller.TiDBMemberName(tcName))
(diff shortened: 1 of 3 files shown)


shuijing198799:
@tennix @xiaojingchen @weekface PTAL

onlymellb:
All related logic that contains the SetLastAppliedConfigAnnotation function needs to be modified

pkg/manager/member/tidb_member_manager.go
```diff
@@ -131,7 +131,7 @@ func (tmm *tidbMemberManager) syncTiDBStatefulSetForTidbCluster(tc *v1alpha1.Tid
 	tcName := tc.GetName()
 
 	newTiDBSet := tmm.getNewTiDBSetForTidbCluster(tc)
-	oldTiDBSet, err := tmm.setLister.StatefulSets(ns).Get(controller.TiDBMemberName(tcName))
+	oldTiDBSetTemp, err := tmm.setLister.StatefulSets(ns).Get(controller.TiDBMemberName(tcName))
```
weekface: These two places need to be modified:

https://github.com/pingcap/tidb-operator/blob/b0e6cd32f9089542078f6887612be95e23b18717/pkg/manager/member/pd_member_manager.go#L175

https://github.com/pingcap/tidb-operator/blob/8bcb333e639ae4a76fc64f141acef6519931d091/pkg/manager/member/tikv_member_manager.go#L165

Issue #445, opened 2019-04-30
Update the AWS deployment Guide


gregwebs:
We need to update our AWS docs after #401 is merged.

jlerche:
Closed by #463

PR #463, opened 2019-05-07
AWS EKS tutorial change to new terraform script

### LINKED ISSUE #450: Reuse Terraform AWS deployment for AWS EKS tutorial
The current tutorial for AWS EKS is very well written, but ultimately requires a lot of steps and knowledge not directly related to TiDB, as well as not enough integration with TiDB, such as specific nodes for PD, TiKV, etc.

It would be useful to be able to re-use the terraform script with a configuration with instance types suitable for trying it out.

jlerche:
This will close #450 and #445 

Some extra variables were extracted to accommodate a lighter-weight deployment re-purposed for the AWS EKS tutorial. The values specific to the tutorial are stored in a separate terraform variables file, defaults were kept the same.

The pd and tikv userdata script was unified into a single script.

### Check List 

Tests 

Checked that tutorial deployment works as expected
Checked that default deployment works as expected

Code changes

 - Has documents change

Side effects

Related changes


deploy/aws/variables.tf
```diff
@@ -83,3 +83,24 @@ variable "tikv_count" {
 variable "tidb_count" {
   default = 2
 }
+
+variable "pd_instance_type" {
```
tennix: Maybe we should add a warning here, since it's fragile to change instance type which may break the user-data and local volume setup.

deploy/aws/aws-tutorial.tfvars
```diff
@@ -0,0 +1,11 @@
+pd_instance_type = "c5d.large"
+tikv_instance_type = "c5d.large"
+tidb_instance_type = "c5d.large"
+monitor_instance_type = "c5d.large"
```
aylei: Local SSD of monitor and tidb instance is not used, so I suppose `c5.large` would be better?

jlerche: This is a good point, default for TiDB is actually a `c4.4xlarge` so maybe just `c4.large` for TiDB and `c5.large` for monitor?

deploy/aws/variables.tf
jlerche: Added a comment in `variables.tf` and also in tutorial doc.

Issue #332, opened 2019-03-20
Proposal: Allow tidb pod to be scheduled on the previous node


tennix:
For `NodePort` type Service, the `spec.externalTrafficPolicy` may set to `Local` to enable client IP propagation which is needed for source IP based permission control. This way, not all the nodes and node port is accessible for the client connection. Users may choose only the node where the tidb-server pod is on to connect. But when performing a rolling upgrade, the tidb-server pod may be scheduled to other nodes, thus the client connection info needs to be updated. This is inconvenient for users.

So here we propose to allow tidb pod to be scheduled on the previous node. Currently, the tidb pod scheduling will bypass our customized scheduler. This can be done by adding a scheduling policy for the tidb pod. We can record the node IP of the tidb pod in `TidbCluster` status, and give higher priority of the previous node if multiple nodes match tidb pod requirements.

PR #466, opened 2019-05-08
Design proposal of stable scheduling in TiDB

cofyc:

This expands the design proposal of https://github.com/pingcap/tidb-operator/issues/332.

https://github.com/cofyc/tidb-operator/blob/fix332-doc/docs/design-proposals/tidb-stable-scheduling.md

### Check List 

Has documents change


docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,141 @@
...
+
+```
+type TiDBMember struct {
+  ...
+	// Node hosting pod of this TiDB member.
+	NodeName string `json:"node,omitempty"`
+}
+```
+
+In new predicate `StableScheduling`, we filter out other nodes for TiDB pod if
```
aylei: Can I specify scheduling policy (whether enable `StableScheduling`) for a `TidbCluster` instance?

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Implementation
+
+At first, we track assigned node of TiDB member in status of TiDBCluster.
+
+```
+type TiDBMember struct {
+  ...
+	// Node hosting pod of this TiDB member.
+	NodeName string `json:"node,omitempty"`
```
aylei: Suppose one of the nodes is down, is this possible for the TiDB `StatefulSet` to recover without deleting the `NodeName` in `.status` manually?

docs/design-proposals/tidb-stable-scheduling.md
cofyc: Yes, but I'd like to add a global feature switch to control this scheduling behavior for all TiDB cluster instances tidb-operator managed. Because this is a best-effort policy, no harm if it fails.

What I am thinking is should we fail the pod (with a switch to control) if it cannot be scheduled to its previous node?

docs/design-proposals/tidb-stable-scheduling.md
cofyc: yes, the predicate is executed in tidb-scheduler, if the node is down, it will not exist in candidate nodes sent by kube-scheduler.

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,145 @@
...
+
+In this scenario, we need to use `NodePort` service. By default, NodePort
+services are available cluster-wide on all nodes but cannot propagate client's
+IP address to the end pods and may cause a second hop. 
+
+For propagating client's IP address and better performance, we can
+set `externalTrafficPolicy` of service to `Local`. A side-effect is the service of
+TiDB will be accessible only on the nodes which a running TiDB pod. To avoid
+manual intervention to update IP addresses in load balancer when performing
+a rolling update, we prefer to schedule new pod of TiDB member to its previous node.
```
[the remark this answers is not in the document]
cofyc: The best solution is to implement an operator for the load balancer, but it's beyond the scope of this document. Adding all nodes into load balancer is a good alternative solution, but it depends on Load Balancer:

- maybe too heavy for LB if the Kubernetes cluster is large (described by @tennix)
  - NumberOfTiDBCluster x NumberOfNodes ports should be health checked
- need to add every new node into the backend of LB
- hard to monitor LB (not all failed backends must be fixed)

Try to schedule new pod of TiDB member to its previous node is the best we can do in tidb-operator and will work in certain circumstances. No harm if failed.

docs/design-proposals/tidb-stable-scheduling.md
cofyc: One drawback is its previous node may not be the best node for the new node, e.g. there is another node which has more available CPU/Memory resources. However, whether to use this scheduling policy depends on cluster setup (use NodePort with Local policy or not), I think to control this behavior globally is enough. Furthermore, we can bypass this policy if TiDB service does not use `local` externalTrafficPolicy intelligently.

docs/design-proposals/tidb-stable-scheduling.md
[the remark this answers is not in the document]
cofyc: However, I think we can improve this by restricting TiDB pods in a fixed set of nodes (by using NodeSelector/PodAffinity/Taints&Tolerations). Then, we only need to add part of the nodes into the load balancer and no need to update them in the future.

In my understanding, the proposed solution in this doc is like this except it does not require the user to pre-select nodes for TiDB pods and the fixed set of nodes are limited (equals to the number of TiDB pod instances) and may change if it's not possible to run the new pod in which scenario we need manual intervention.

docs/design-proposals/tidb-stable-scheduling.md
tennix: I agree to add a switch for this scheduling policy as this feature is not suitable for all users.

docs/design-proposals/tidb-stable-scheduling.md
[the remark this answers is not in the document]
tennix: Also, the stable scheduling policy makes it easier to understand for users not familiar with Kubernetes.

docs/design-proposals/tidb-stable-scheduling.md
cofyc: One way is to control globally: 

A flexible feature gate like this:

```
tidb-scheduler --features StableScheduling,FeatureA,FeatureB
```

It works like Kubernetes feature gate, but simpler (all disabled by default).

Or a dedicated flag

```
tidb-scheduler --enable-stable-scheduling
```

Another way is to control per TiDBCluster, add a field in TiDBCluster.

What do you prefer?

tennix: I think it should be a global feature controlled by operator not `TidbCluster` CRD.

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+cluster on every node is very inefficient and will consume too many resources.
+
+### Leverage load balancer health check mechanism
+
+Load balancers often have health checks on its backends, it will remove invalid
+backend automatically. We can add all nodes into load balancer. Drawbacks with
+this solution are:
+
+- maybe too heavy for LB if the Kubernetes cluster is large
+  - `NumberOfTiDBClusters` x `NumberOfNodess ports should be health checked
```
tennix: Please fix the markdown format.

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+When a TiDB cluster is running in a dedicated Kubernetes cluster or nodes of
+TiDB cluster are reserved for it. After a rolling update of TiDB cluster is
+done, new pods of TiDB member will be scheduled to theirs previous nodes. User
+does not need to update IP addresses in load balancer.
+
+## Proposal
+
+Currently, we have tidb-scheduler to schedule all pods of TiDB cluster. We can
+write a new predicate function for pods of TiDB server. In this new predicate
+function we can choose the its previous node of TiDB member it it exists in candidate
```
tennix: In this new predicate function we can choose the previous node of TiDB member if it exists in candidate nodes.

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+- Guarantee that new pod of TiDB member will be scheduler to its node if some
+  scheduling requirements changed
+
+## Use Cases
+
+### No need to update IP addresses of TiDB in load balancer outside of the Kubernetes cluster
+
+When a TiDB cluster is running in a dedicated Kubernetes cluster or nodes of
+TiDB cluster are reserved for it. After a rolling update of TiDB cluster is
+done, new pods of TiDB member will be scheduled to theirs previous nodes. User
```
tennix: s/theirs/their/

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+## Goals
+
+- Able to schedule new pod of TiDB member to its previous node
+
+### Non-Goals
+
+- Stable scheduling for TiKV/PD pods
+- Guarantee that new pod of TiDB member will be scheduled to its node if
+  there is another scheduler in cluster which may schedule pods to its node
+- Guarantee that new pod of TiDB member will be scheduler to its node if some
```
tennix: s/scheduler/scheduled/

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+backend automatically. We can add all nodes into load balancer. Drawbacks with
+this solution are:
+
+- maybe too heavy for LB if the Kubernetes cluster is large
+  - `NumberOfTiDBClusters` x `NumberOfNodess ports should be health checked
+- need to add every new node into the backend of LB
+- hard to monitor LB (not all failed backends must be fixed)
+
+Some of them can be alleviated by restricting TiDB members in a fixed set of
+nodes (by using NodeSelector/NodeSelector/Taints&Tolerations). This requires
```
tennix: Duplicated `NodeSelector`.

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+the user to pre-select the nodes to run TiDB pods.
+
+### Use local PV to bind TiDB member to a nod
```
tennix: Typo: nod -> node

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,167 @@
...
+#### Workarounds
+
+- Reserve nodes for TiDB members (e.g. by taints)
+
+### tidb-cluster cannot schedule new pod of TiDB member back to its node if it does not meet new requirements
+
+If we upgrade TiDB pods to request more resources, it is possible that its node node
+may not have enough resources for the new pod.
+
+It applies if some other scheduling requirements are changed, e.g.
```
tennix: This should be written as a warning in the document.

docs/design-proposals/tidb-stable-scheduling.md
aylei: I suppose both options are valuable:

* feature gate of scheduler: controlled by cluster admin, determine whether current cluster **is willing to support stable scheduling**.
* field of `TidbCluster` CRD: controlled by user who make the decision of whether using local mode and stable scheduling

gregwebs:
There is an issue here of loss of availabiliy during the rolling update. A Deployment RollingUpdate will "surge" and add an new pod, temporarily increasing the replica set replication by 1. Then an old pod will be taken down, and the replication will be at its original setting N. So the replication is always N or N + 1.
For a stateful set, I believe the replication will be N or N -1 during rollout, meaning a loss of availability.

Our operator could increase the statefulset capacity by 1 before rollout to maintain the same level of availability. However, the current proposal's goal is to avoids automatically updating a load balancer, so that would not help. To avoid a loss of availability we would need to convince Kubernetes to schedule the new TiDB pod onto the same node while the previous TiDB is still running and wait to cutover to the new TiDB when the new TiDB is ready. However, it doesn't seem possible to schedule two TiDB to the same node due to capacity limitations, and there could be OOM during the cutover.

I am wondering what the specific issue is with the scalability of the load balancer? Why can't we announce just the nodes running TiDB to the load balancer?

cofyc:
> Our operator could increase the statefulset capacity by 1 before rollout to maintain the same level of availability. However, the current proposal's goal is to avoids automatically updating a load balancer, so that would not help. To avoid a loss of availability we would need to convince Kubernetes to schedule the new TiDB pod onto the same node while the previous TiDB is still running and wait to cutover to the new TiDB when the new TiDB is ready. However, it doesn't seem possible to schedule two TiDB to the same node due to capacity limitations, and there could be OOM during the cutover.

In my testing, the operator does not increase TiDB statefulset capacity by 1. Purpose of this design is to add a functionality in our scheduler extender (tidb-scheduler) to schedule the new pod of TiDB member back to its original node if possible.

Workflow:

- demo-tidb-2 is running on the node kube-node-2
- the user performs an update (except scale out)
- after the pod of demo-tidb-2 is terminated, the new pod of demo-tidb-2 is created
- kube-scheduler sends feasible nodes which can run the pod `demo-tidb-2` to tidb-scheduler (scheduler extender)
- tidb-scheduler filters out other nodes if the original node exists in these nodes
- tidb-scheduler does nothing if the original node does not exist in these nodes (another pod is assigned to kube-node-2 and not enough resources left for demo-tidb-2), kube-scheduler will continue to prioritize these nodes to find the best match

> I am wondering what the specific issue is with the scalability of the load balancer? Why can't we announce just the nodes running TiDB to the load balancer?

We have a discussion about this [here](https://github.com/pingcap/tidb-operator/pull/466#discussion_r282034235). One of the clients needs to configure TiDB instances in an existing load balancer, and want to avoid manual intervention as much as possible.

docs/design-proposals/tidb-stable-scheduling.md
cofyc: I added a note about this and workflow in the Proposal section. All other comments are addressed by the way.

tennix:
@gregwebs @cofyc Yes, using `StatefulSet` means there will be a pod unavailable during the upgrade. We can change this behavior by using `Deployment`. The rolling update of `Deployment` ensures no service degradation.

However, `Deployment` doesn't make sure the upgrade order which is required for tidb-server. Tidb-server internally has the DDL owner, and this requires special upgrading order which cannot be fulfilled by `Deployment`.

Besides, upgrading tidb-server will definitely close the client connection which results a service degradation. So it's reasonable to use `StatefulSet` for tidb-server.

tennix:
For bare-metal deployment, there is usually no automatic load balancer to use. These users have to use `NodePort` type service. And TiDB has source IP based permission system, to use this feature we have to use `Local` mode of `NodePort` service. However, this mode has some drawbacks:

* imbalanced traffic spreading as documented [here](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip)
* it's too heavy for the external load balancer to do health check when there are too many k8s nodes and `NodePort` services.

So in these scenarioes, users would prefer to only announce the nodes tidb pods are running.

gregwebs:
> Besides, upgrading tidb-server will definitely close the client connection which results a service degradation. So it's reasonable to use StatefulSet for tidb-server.
We will eventually need to support connection draining. If someone is running a long analytic query, it will be interrupted, but we can guarantee that transactions will not be.

I still don't understand the health check overhead issue. But it seems the issue comes down to the load balancer not knowing which nodes have TiDB? I would think that it is possible for software to solve this, but maybe users don't want to configure api access to update the load balancer? Maybe it is possible for the load balancer to make an outward request for an updated configuration rather than do its own health checks?

tennix:
Users would think it's unreasonable to attach all the k8s nodes to the load balancer. Suppose there are a thousand nodes (of course there are no such users right now) and one of the tidb cluster only runs one tidb pod, if we announce all nodes, then the load balancer has to do health check on these thousand nodes every now and then. This sounds absurd to them.

Besides, not all users can or want to make kubernetes to automatically configure their external load balancer. For these users, they need to configure the load balancer backends manually according to where we put tidb pods on.

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,196 @@
...
+  created
+- kube-scheduler sends feasible nodes which can run the pod `demo-tidb-2` to
+  tidb-scheduler (scheduler extender)
+- tidb-scheduler filters out other nodes if the original node exists in these
+  nodes, kube-scheduler will choose `kube-node-2` to run `demo-tidb-2`
+  - note that if `kube-node-2` exist in the nodes sent from kube-scheduler, it
+    meets all criteria to `demo-tidb-2`
+- tidb-scheduler does nothing if the original node does not exist in these
+  nodes (e.g. not enough resources left for demo-tidb-2 if another pod is
+  assigned to kube-node-2 after kube-demo-2 is deleted), kube-scheduler will
```
LinuxGit: s/kube-demo-2/`demo-tidb-2` ?

docs/design-proposals/tidb-stable-scheduling.md
```diff
@@ -0,0 +1,196 @@
...
+When `tidb-scheduler` schedules new pod of TiDB member, its previous node may
+not fit.
+
+#### Workarounds
+
+- Reserve nodes for TiDB members (e.g. by taints)
+
+### Cannot schedule new pod of TiDB member back to its node if the node does not meet new requirements
+
+If we upgrade TiDB pods to request more resources, it is possible that its node node
```
LinuxGit: s/its node node/its node

docs/design-proposals/tidb-stable-scheduling.md
cofyc: Fixed, Thanks!

docs/design-proposals/tidb-stable-scheduling.md
cofyc: Fixed too.

cofyc:
hi, @weekface PR has been rebased on the master, PTAL again.

PR #473, opened 2019-05-09
support-multi-version-dashboard

qiffang:
- Support multi TiDB Cluster dashboard version, default use v2
- Remove gzip logic because it is too difficult to maintain
_Note
In Grafana, the symbol "{{" represent a grammar, it conflict with helm grammar, {{ should be replaced by {{`{{`}}_
- Support multi version binlog dashboard 
Based on the image version to decide dashboard version and binlog version
### Check List 

Tests 

- V3 TiDB Cluster - PASS
- V2 TiDB Cluster - PASS
- Binlog dashboard - PASS

Code changes

 - Has Helm charts change

Side effects

Related changes


qiffang:
> LGTM, it works for me. In the future, I think it's better to re-package grafana image to follow TiDB version.

Yes,  this approach is only a temporary workaround. After a while, we need to discuss a graceful solution.

PR #475, opened 2019-05-09
support the affinity feature of k8s which define the rule of assigning pods to nodes

xiaojingchen:
- Add `affinity` settings and change the `nodeSelector` meaning
- Add the `Configuration` subject for doc `operation-guide.md`
Changed the `TidbCluster` struct , add the following field
```
Affinity         *corev1.Affinity    `json:"affinity,omitempty"`
```
to `PDSpec/TiKVSpec/TiDBSpec` and remove `NodeSelectorRequired`.
tidb-operator directly assigns the above `Affinity` field to pd/tidb/tikv's StatefulSet,and then done.
### Check List 

Tests
 - Manual test [done]
 - E2E test and stability test will be added in the next PR

Code changes
The following files were changed:
- charts/tidb-cluster
- pkg/apis/pingcap.com/v1alpha1/types.go
- pkg/util/util.go
- pkg/manager/member
- docs/operation-guide.md

Side effects
Because this PR changed the `TidbcCluster` CRD, so it can not compatible with previous versions.

 ```release-note
ACTION REQUIRED: `nodeSelectorRequired` was remove from values.yaml. 
ACTION REQUIRED:  Comma-separated values support in `nodeSelector` has been dropped, please use new-added `affinity` field which has a more expressive syntax.
Support the affinity feature of k8s which can define the rule of assigning pods to nodes
 ```


xiaojingchen:
@tennix @weekface @cofyc @aylei PTAL

charts/tidb-cluster/values.yaml
```diff
@@ -66,16 +66,68 @@ pd:
...
+  ## The NodeAffinity setting of the example ensure that the PD pods can only be scheduled to nodes with label:[type="pd"],
+  # affinity:
+  #   podAntiAffinity:
+  #     preferredDuringSchedulingIgnoredDuringExecution:
+  #     # this term work when the nodes have the label named region
+  #     - weight: 10
+  #       podAffinityTerm:
+  #         labelSelector:
+  #           matchLabels:
+  #             app.kubernetes.io/instance: <cluster.name>
```
```suggestion
  #             app.kubernetes.io/instance: <helm release name>
```

?

charts/tidb-cluster/values.yaml
```diff
@@ -66,16 +66,68 @@ pd:
...
+  #             app.kubernetes.io/component: "pd"
+  #         topologyKey: "region"
+  #         namespaces:
+  #         - <namespace of tidb cluster>
```
```suggestion
  #         - <helm namespace>
```
?

charts/tidb-cluster/values.yaml
```diff
@@ -66,16 +66,68 @@ pd:
...
-  # if nodeSelectorRequired is true, all the following labels must be matched
-  nodeSelector: {}
-    # kind: pd
-    # # zone is comma separated availability zone list
-    # zone: cn-bj1-01,cn-bj1-02
-    # # region is comma separated region list
-    # region: cn-bj1
-  # Tolerations are applied to pods, and allow pods to schedule onto nodes with matching taints.
-  # refer to https://kubernetes.io/docs/concepts/configuration/taint-and-toleration
+
```
weekface: we should add a key:
```
locationLabels:
- region
- zone
- rack
- host

```

xiaojingchen: right，that will be added in other pr

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
 
 > **Note:** The rest of the document will use `values.yaml` to reference `charts/tidb-cluster/values.yaml`
 
+## Configuration
+
+TiDB Operator use `values.yaml` as TiDB cluster configuration file. It provides the default basic configuration which you can use directly for quick deployment, but if you have specific configuration requirements or for production deployment, you need to manually modify the variables in the `value.yaml` file.
```
weekface:
```suggestion
TiDB Operator uses `values.yaml` as TiDB cluster configuration file. It provides the default basic configuration which you can use directly for quick deployment, but if you have specific configuration requirements or for production deployment, you need to manually modify the variables in the `value.yaml` file.
```

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+
+* Resource setting
+
+    * CPU & Memory
+
+        The default deployment doesn't set CPU and memory requests or limits for any of the pods, these settings can make TiDB cluster run on a small Kubernetes cluster like DinD or the default GKE cluster for testing. But for production deployment, you would likely to adjust the cpu, memory and storage resources according to the [recommendations](https://github.com/pingcap/docs/blob/master/op-guide/recommendation.md).
```
weekface: https://github.com/pingcap/docs/blob/master/op-guide/recommendation.md

this link is broken.

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+        The resource limits should be equal or bigger than the resource requests, it is suggested to set limit and request equal to get [`Guaranteed` QoS]( https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed).
+
+    * Storage
+
+        The variables `pd.storageClassName` and `tikv.storageClassName` in `values.yaml` are used to set `StorageClass` of pd and tikv,their default setting are `local-storage` with minimal size.
```
weekface:
```suggestion
        The variables `pd.storageClassName` and `tikv.storageClassName` in `values.yaml` are used to set `StorageClass` of PD and TiKV,their default setting are `local-storage` with minimal size.
```

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+        If you don't want to use the default `StorageClass` or your Kubernetes cluster does not support `local-storage` class, please execute the following command to find an available `StorageClass` and select the ones you want to provide to TiDB cluster.
+
+        ```shell
+        $ kubectl get sc
+        ```
+
+* Disaster Tolerance setting
+
+    TiDB cluster is a distributed database. Its Disaster Tolerance means that when any physical node failed, not only to ensure TiDB server is available, but also ensure the data is complete and available.
```
weekface:
```suggestion
    TiDB is a distributed database. Its disaster tolerance means that when any physical node failed, not only to ensure TiDB server is available, but also ensure the data is complete and available.
```

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+
+    How to guarantee Disaster Tolerance of TiDB cluster on Kubernetes?
+
+    We mainly solve the problem from the scheduling of services and data.
+
+    * Disaster Tolerance of TiDB server
```
weekface:
```suggestion
    * Disaster Tolerance of TiDB instance
```

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+
+        TiDB Operator provides a external scheduler to guarantee PD/TiKV/TiDB server disaster tolerance on host level. TiDB Cluster have set the external scheduler as default scheduler, you will find the setting in the variable `schedulerName` of `values.yaml`.
```
weekface:
```suggestion
        TiDB Operator provides an extended scheduler to guarantee PD/TiKV/TiDB instance disaster tolerance on host level. TiDB Cluster has set the extended scheduler as default scheduler, you will find the setting in the variable `schedulerName` of `values.yaml`.
```

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+
+        In the other hand use `PodAntiAffinity` term of `affinity` to ensure disaster tolerance on the other topology levels (e.g. rack, zone, region). 
+        refer to the doc: [pod affnity & anti affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#inter-pod-affinity-and-anti-affinity-beta-feature), moreover `values.yaml` also provides a typical disaster tolerance setting example in the comments of `pd.affinity`.
+
+    * Disaster Tolerance of data
+
+        Disaster tolerance of data is guaranteed by TiDB Cluster itself. The only work Operator needs to do is that collects topology info from specific labels of Kubernetes nodes where TiKV Pod runs on and then PD will schedule data replicas auto according to the topology info.
+        Cause currently TiDB Operator can only recognize some specific labels, so you can only set nodes topology info with the following particular labels
```
weekface:
```suggestion
        Because current TiDB Operator can only recognize some specific labels, so you can only set nodes topology info with the following particular labels
```

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+
+        * `region`: region where node is located
+        * `zone`: zone where node is located
+        * `rack`: rack where node is located
+        * `kubernetes.io/hostname`: hostname of the node
+
+        you need label topology info to nodes of Kubernetes cluster use the following command
+        ```shell
+        # The labels are optional
```
weekface:
```suggestion
        # Not all tags are required
```

cofyc:
Why removing `nodeSelector`? I suggest keeping this same as pod spec. Users can specify `nodeSelector` or `affinity`.

pkg/util/util.go
```diff
@@ -15,135 +15,15 @@ package util
...
-	affinity.NodeAffinity = &corev1.NodeAffinity{
-		RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
-			NodeSelectorTerms: requiredTerms,
-		},
-		PreferredDuringSchedulingIgnoredDuringExecution: preferredTerms,
-	}
-
-	return affinity
-}
-
```
cofyc: If I understand correctly, `nodeSelector` equals to pod.spec.nodeSelector when `nodeselectorRequired` is true?

How about remove `nodeselectorRequired` only, but keep `nodeSelector` . It can be copied directly to pod spec, I guess. Then the user can configure scheduling policies of TiDB, PD, TiKV like normal pods.

weekface: vote for it

cofyc: oh, there is a difference, our `nodeSelector` accepts comma-separated values which is not supported by `pod.spec.nodeSelector`

weekface: we can change our `nodeSelector` to be the same as `pod.spec.nodeSelector`?

cofyc: however, instead of removing it, I think it's better to keep it even if we need to drop the comma-separated values support
comma-separated values support is not necessary because we are able to configure in affinity

xiaojingchen: The previous`nodeSelector` was self-defined, and it was converted to `affinity` in the background.
the suggestion is nice.

xiaojingchen: @tennix  have any opinion?

xiaojingchen:
I have added `NodeSelector` @weekface @cofyc  PTAL

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
 
 > **Note:** The rest of the document will use `values.yaml` to reference `charts/tidb-cluster/values.yaml`
 
+## Configuration
+
+TiDB Operator uses `values.yaml` as TiDB cluster configuration file. It provides the default basic configuration which you can use directly for quick deployment, but if you have specific configuration requirements or for production deployment, you need to manually modify the variables in the `value.yaml` file.
```
tennix:
```suggestion
TiDB Operator uses `values.yaml` as TiDB cluster configuration file. It provides the default basic configuration which you can use directly for quick deployment, but if you have specific configuration requirements or for production deployment, you need to manually modify the variables in the `values.yaml` file.
```

charts/tidb-cluster/templates/tidb-cluster.yaml
```diff
@@ -33,11 +33,10 @@ spec:
   {{- if .Values.pd.resources }}
 {{ toYaml .Values.pd.resources | indent 4 }}
   {{- end }}
-  {{- if .Values.pd.nodeSelector }}
```
weekface: why remove this condition?

charts/tidb-cluster/values.yaml
weekface: open an issue to track it?

weekface:
This is a breaking change. You should add these to `ACTION REQUIRED` of the release note.

- `nodeSelectorRequired` was removed
- `nodeSelector` usage changed

cofyc:
```
+ kubectl delete -f tests/manifests/e2e/e2e.yaml

service "webhook-service" deleted

clusterrolebinding.rbac.authorization.k8s.io "tidb-operator-e2e" deleted

serviceaccount "tidb-operator-e2e" deleted

Error from server: error when stopping "tests/manifests/e2e/e2e.yaml": admission webhook "check-pod-before-delete.k8s.io" denied the request without explanation
```

`ValidatingWebhookConfigurations` not cleaned in previous e2e tests?

cofyc:
`ACTION REQUIRED:  `nodeSelector` has a different meaning than before，the new `nodeSelector` just a simple key-value map and not support multiple values separated by commas.`

-> 

```
ACTION REQUIRED:  Comma-separated values support in `nodeSelector` has been dropped, please use new-added `affinity` field which has a more expressive syntax.
```

xiaojingchen:
@cofyc  `ValidatingWebhookConfigurations` should be cleaned in before e2e tests, the reason can not certain, I need to check e2e's clean code. 
It has nothing to do with this PR.

docs/operation-guide.md
```diff
@@ -11,21 +11,68 @@ $ namespace="tidb"
...
+## Configuration
+
+TiDB Operator uses `values.yaml` as TiDB cluster configuration file. It provides the default basic configuration which you can use directly for quick deployment, but if you have specific configuration requirements or for production deployment, you need to manually modify the variables in the `values.yaml` file.
+
+* Resource setting
+
+    * CPU & Memory
+
+        The default deployment doesn't set CPU and memory requests or limits for any of the pods, these settings can make TiDB cluster run on a small Kubernetes cluster like DinD or the default GKE cluster for testing. But for production deployment, you would likely to adjust the cpu, memory and storage resources according to the [recommendations](https://pingcap.com/docs/dev/how-to/deploy/hardware-recommendations/#software-and-hardware-recommendations).
+        The resource limits should be equal or bigger than the resource requests, it is suggested to set limit and request equal to get [`Guaranteed` QoS]( https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed).
```
aylei: Add a newline above if you want to separate these two paragraph

PR #479, opened 2019-05-09
Support updating configuraion on the fly

### LINKED ISSUE #225: How to change the configuration of TiDB/TiKV/Pd on the fly?
I notice there's no related information in TiDB Cluster Operation Guide

aylei:
fix #225 

At the server side, A set of special annotations are introduced, if these annotations present, the PD/TiKV/TiDB statefulset will use the annotation to concat configmap name. 

At the client side, if feature gate `EnableConfigMapRollout` enabled, the client will generate SHA256 digest for configmap data and append the digest in configmap name, while keep the annotations mentioned above in sync.

After a configuration change, the digest changed so the configmap name and annotations changed, the statefulsets will use the new configmap name due to the annotation change, a rolling-update will performed consequently.

### Check List

Tests

 - [x] Unit test
 - [x] E2E test
 - Stability test

The stability test case will be more complicated (which may include mis-configuring scenario) and need more discussion, which will be tracked in another issue.

Code changes

 - Has Helm charts change ✔️
 - Has Go code change ✔️
 - Has CI related scripts change ✔️
 - Has documents change ✔️

## Release Note

```
Introduce ConfigMap rollout management. With the feature gate open, configuration file changes will be automatically applied to the cluster via a rolling update.
```


aylei:
@weekface @cofyc @tennix @xiaojingchen PTAL

charts/tidb-cluster/templates/tidb-cluster.yaml
```diff
@@ -2,6 +2,12 @@ apiVersion: pingcap.com/v1alpha1
 kind: TidbCluster
 metadata:
   name: {{ template "cluster.name" . }}
+{{- if .Values.enableConfigMapRollout }}
+  annotations:
+    pingcap.com/pd.{{ template "cluster.name" . }}-pd.sha: {{ template "pd-configmap.data-digest" . }}
```
weekface:
```
pingcap.com/pd.{{ template "cluster.name" . }}-pd.sha
```

There are two duplicate`pd`s in the annotations,  can we have only one?

```
pingcap.com/{{ template "cluster.name" . }}-pd.sha
```

tests/actions.go
```diff
@@ -465,6 +483,12 @@ func (oa *operatorActions) CleanTidbCluster(info *TidbClusterConfig) error {
 		return fmt.Errorf("failed to delete jobs: %v, %s", err, string(res))
 	}
 
+	// delete all configmaps
+	allConfigMaps := label.Label{}.Instance(info.ClusterName).String()
```
weekface:
```suggestion
	allConfigMaps := label.New().Instance(info.ClusterName).String()
```

tests/cmd/e2e/main.go
```diff
@@ -193,6 +201,22 @@ func main() {
 		}
 	}
 
+	// update configuration on the fly
+	for _, clusterInfo := range clusterInfos {
```
weekface: Need to add this check to e2e?

tests/config.go
```diff
@@ -55,6 +59,11 @@ type Nodes struct {
 // NewConfig creates a new config.
 func NewConfig() (*Config, error) {
 	cfg := &Config{
+
+		PDMaxReplicas:       5,
```
weekface: You may set `PDMaxReplicas ` default to `3`. This is the number of replicas for each region.

tests/config.go
```diff
@@ -37,6 +37,10 @@ type Config struct {
 	CertFile         string
 	KeyFile          string
 
+	PDMaxReplicas       int `yaml:"pd_max_replicas" json:"pd_max_replicas"`
```
weekface: I think it doesn't need these flags, just set a literal value to https://github.com/pingcap/tidb-operator/pull/479/files#diff-f2f48876f13bab9311a47b1c1be4a1f9R207 is enough.

```
		clusterInfo = clusterInfo.
			UpdatePdMaxReplicas(5).
			UpdateTiKVGrpcConcurrency(1024).
			UpdateTiDBTokenLimit(8)
```

tests/cmd/e2e/main.go
aylei: This is a verification of the "ConfigMap Rollout" functionality

tests/config.go
aylei: This config is what we update the ConfigMap **to**, maybe the variable name should more concise.

tests/cmd/e2e/main.go
weekface: Sorry, my fault. I mean to add this functionality to `stability` tests.

charts/tidb-cluster/templates/tidb-cluster.yaml
aylei: I suppose no, the form `pd.{{ template "cluster.name" . }}-pd.sha` is consists of `{component}.{configmap-name}.sha` according to the design document, and in this case, the `{configmap-name}` is just happen to be `{{ template "cluster.name" . }}-pd`, which have not yet been a constrain for the configmap name.

tests/config.go
aylei: Yes, using literal value is simple and clear. Maybe I don't have to do this at the very beginning, but the motivation is that I want these parameters being configurable in favor of flexibility, I tend to keep it because it does no harm even if we won't configure it in the near future. How do you think?

tests/cmd/e2e/main.go
aylei: I'm working on the stability test case in another PR now, is that OK or I should merge them together?

aylei: #486 tracks this

tests/config.go
weekface: I am fine with this solution.

tests/cmd/e2e/main.go
weekface: another pr is more reasonable

aylei:
> Do we need to add the `Release notes` or `a user-facing change` to the PR and open another cherry-pick to `release-1.0` when it merged as k8s does?
> 
> [kubernetes/kubernetes#75900](https://github.com/kubernetes/kubernetes/pull/75900)

Good idea, yes I think this should be cherry-picked to the 1.0 release

aylei:
I've introduced the logLevel configuration parameter to validate the rolling-update of PD, PTAL @xiaojingchen @weekface

PR #484, opened 2019-05-13
stability: retry truncating sst files upon failure

zyguan:
sst files may be deleted after compaction, thus the `TruncateSSTFile` may fail unexpectly.

This pr closes: #501 
add retry logical.

### Check List 

Tests 

 - E2E test
 - Stability test

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
NONE
 ```


tests/pkg/ops/tikv.go
```diff
@@ -53,41 +56,52 @@ func (ops *TiKVOps) TruncateSSTFile(opts TruncateOptions) error {
...
-	if err != nil {
-		glog.Errorf("list sst files: stderr=%s err=%s", stderr, err.Error())
-		return errors.Annotate(err, "list sst files")
-	}
+	retryCount := 0
+	for ; retryCount < retryLimit; retryCount++ {
+		stdout, stderr, err := exec("find", "/var/lib/tikv/db", "-name", "*.sst", "-o", "-name", "*.save")
+		if err != nil {
+			glog.Errorf("list sst files: stderr=%s err=%s", stderr, err.Error())
+			return errors.Annotate(err, "list sst files")
```
weekface:
```suggestion
			continue
```

tests/pkg/ops/tikv.go
```diff
@@ -53,41 +56,52 @@ func (ops *TiKVOps) TruncateSSTFile(opts TruncateOptions) error {
...
-		if strings.HasSuffix(k, ".sst") && !sstCandidates[k+".save"] {
-			sst = k
+		sst := ""
+		for k := range sstCandidates {
+			if strings.HasSuffix(k, ".sst") && !sstCandidates[k+".save"] {
+				sst = k
+			}
+		}
+		if len(sst) == 0 {
+			return errors.New("cannot find a sst file")
```
weekface:
```suggestion
			glog.Error("cannot find a sst file")
			continue
```

tests/pkg/ops/tikv.go
```diff
@@ -53,41 +56,52 @@ func (ops *TiKVOps) TruncateSSTFile(opts TruncateOptions) error {
 		})
 	}
 
-	stdout, stderr, err := exec("find", "/var/lib/tikv/db", "-name", "*.sst", "-o", "-name", "*.save")
-	if err != nil {
-		glog.Errorf("list sst files: stderr=%s err=%s", stderr, err.Error())
-		return errors.Annotate(err, "list sst files")
-	}
+	retryCount := 0
+	for ; retryCount < retryLimit; retryCount++ {
```
weekface:
```suggestion
	for ; retryCount < retryLimit; retryCount++ {
	        time.Sleep(10*time.Second)
```

tests/pkg/ops/tikv.go
```diff
@@ -53,41 +57,54 @@ func (ops *TiKVOps) TruncateSSTFile(opts TruncateOptions) error {
...
+			time.Sleep(10 * time.Second)
+		}
+		stdout, stderr, err := exec("find", "/var/lib/tikv/db", "-name", "*.sst", "-o", "-name", "*.save")
+		if err != nil {
+			glog.Warningf("list sst files: stderr=%s err=%s", stderr, err.Error())
+			continue
+		}
 
-	sstCandidates := make(map[string]bool)
+		sstCandidates := make(map[string]bool)
```
xiaojingchen: delete the blanks

tests/pkg/ops/tikv.go
```diff
@@ -53,41 +57,54 @@ func (ops *TiKVOps) TruncateSSTFile(opts TruncateOptions) error {
...
+
+		break
 	}
 
-	_, stderr, err = exec("truncate", "-s", "0", sst)
-	if err != nil {
-		glog.Errorf("truncate sst file: stderr=%s err=%s", stderr, err.Error())
-		return errors.Annotate(err, "truncate sst file")
+	if retryCount == retryLimit {
+		return errors.New("failed to truncate sst file after " + strconv.Itoa(retryLimit) + " trials")
```
xiaojingchen: the error log needs  `opts.Namespace` and `opts.Cluster` field

tests/pkg/ops/tikv.go
zyguan: the one more indent because it's in retry loop now.

tests/pkg/ops/tikv.go
zyguan: I added annotations to log methods. however, it's caller who pass the `opts` arg, that is, the caller must know these info, so I don't think there is a need for adding these fields to returned error.

xiaojingchen: but the caller has not added these fields to error log too
ref: https://github.com/pingcap/tidb-operator/blob/master/tests/failover.go#L74-L81

zyguan: I added some error logs to the case func. If you think every error should be annotated with additional fields, it would be better to create an another PR to do it, because most of methods of `operatorActions` failed to do so.

weekface:
@zyguan PTAL

PR #493, opened 2019-05-14
Gke terraform and guide

### LINKED ISSUE #446: GKE terraform and updated deployment guide.
Similar to #401, we will provide a terraform deployment for GKE.

jlerche:
Closes #446 

Adds GCP terraform deploy and README

### Check List 

Tests 

Code changes

 - Adds new Terraform script
- Adds readme for script

___

Several changes were made since the last review:

* More settings were extracted into variables
* Workarounds added to deal with some bugs in the google Terraform provider
* DaemonSet with startup script to increase open file descriptors on PD and TiKV nodes, also installs Linux Guest Environment
* Outputs added to make things a bit more user friendly.


gregwebs:
We should have a firewall rule to allow SSH to the GKE nodes from the bastion.

deploy/gcp/main.tf
```diff
@@ -0,0 +1,290 @@
...
+resource "google_container_cluster" "cluster" {
+  name = "${var.cluster_name}"
+  network = "${google_compute_network.vpc_network.self_link}"
+  subnetwork = "${google_compute_subnetwork.private_subnet.self_link}"
+  location = "${var.GCP_REGION}"
+  project = "${var.GCP_PROJECT}"
+
+  master_auth {
+    username = ""
+    password = ""
```
tennix: What does the empty password mean? Disable password login or allow login with empty password?

deploy/gcp/main.tf
```diff
@@ -0,0 +1,290 @@
...
+      image = "ubuntu-os-cloud/ubuntu-1804-lts"
+    }
+  }
+  "network_interface" {
+    subnetwork = "${google_compute_subnetwork.public_subnet.self_link}"
+    access_config {}
+  }
+  tags = ["bastion"]
+
+  metadata_startup_script = "sudo apt-get install -y mysql-client && curl -s https://packagecloud.io/install/repositories/akopytov/sysbench/script.rpm.sh | bash && sudo apt-get -y install sysbench"
```
tennix: Ubuntu image should use `script.deb.sh`. See here https://github.com/akopytov/sysbench#linux

deploy/gcp/manifests/local-volume-provisioner.yaml
```diff
@@ -0,0 +1,128 @@
+apiVersion: storage.k8s.io/v1
```
tennix: Is this file the same as `../../../manifests/local-dind/local-volume-provisioner.yaml`? If so, I think it's better to symlink there.

deploy/gcp/outputs.tf
```diff
@@ -0,0 +1,19 @@
...
+  value = "${local.kubeconfig}"
+}
+
+output "tidb_version" {
+  value = "${var.tidb_version}"
+}
+
+output "tidb_ilb_ip" {
+  value = "${data.external.tidb_ilb_ip.result["ip"]}"
+}
```
tennix: I think we also need to expose the monitor endpoint to users.

deploy/gcp/templates/tidb-cluster-values.yaml.tpl
```diff
@@ -0,0 +1,407 @@
...
+pvReclaimPolicy: Retain
+
+# services is the service list to expose, default is ClusterIP
+# can be ClusterIP | NodePort | LoadBalancer
+services:
+  - name: pd
+    type: ClusterIP
+
+discovery:
+  image: pingcap/tidb-operator:latest
```
tennix: Use a specific version `pingcap/tidb-operator:v1.0.0-beta.2`

deploy/gcp/outputs.tf
aylei: It will be more clear for user if we expose `cluster_name` in output (user will use this value in many places according to the README)

deploy/gcp/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Upgrade TiDB cluster
+
+To upgrade TiDB cluster, modify `tidb_version` variable to a higher version in variables.tf and run `terraform apply`.
+
+> *Note*: The upgrading doesn't finish immediately. You can watch the upgrading process by `watch kubectl --kubeconfig credentials/kubeconfig_<cluster_name> get po -n tidb`
+
+## Scale TiDB cluster
+
+To scale TiDB cluster, modify `tikv_count` or `tidb_count` to your desired count, and then run `terraform apply`.
```
aylei: I suppose `tikv_replica_count` and `tidb_replica_count` should also be updated

deploy/gcp/README.md
```diff
@@ -0,0 +1,99 @@
...
+# Or setting KUBECONFIG environment variable
+export KUBECONFIG=$PWD/credentials/kubeconfig_<cluster_name>
+kubectl get po -n tidb
+helm ls
+```
+
+When done, the infrastructure can be torn down by running `terraform destroy`
+
+
+> *NOTE*: Any provisioned disks will have to be manually deleted after `terraform destroy`, assuming you do not need the data on the volumes anymore.
```
gregwebs: This is a really horrible workflow. The easiest fix is to have the user install into a separate namespace for tidb and then delete the namespace before running destroy. We should also do this for AWS.

deploy/gcp/manifests/local-volume-provisioner.yaml
jlerche: The local-dind one doesn't have the tolerations. Having said that, I did copy the one from the AWS deploy, maybe we should move it to a common directory and symlink from both?

deploy/gcp/main.tf
jlerche: It's for disabling, otherwise I believe GKE generates a password with username `admin`

deploy/gcp/main.tf
```diff
@@ -295,6 +295,7 @@ until kubectl get po -n tidb -lapp.kubernetes.io/component=tidb | grep Running;
   echo "Wait for TiDB pod running"
   sleep 5
 done
+kubectl get pv -l app.kubernetes.io/namespace=tidb -o name | xargs -I {} kubectl patch {} -p '{"spec":{"persistentVolumeReclaimPolicy":"Delete"}}'
```
gregwebs: This has to be done on teardown because new PVC will be created.

deploy/gcp/README.md
```diff
@@ -72,8 +72,6 @@ helm ls
 When done, the infrastructure can be torn down by running `terraform destroy`
```
gregwebs: I think we will need to put the reclaim command above `terraform destroy` here.

deploy/gcp/main.tf
tennix: The local storage will be deleted when the nodes are deleted. If you want to make persistent disk storage get deleted when pvc is deleted, you can just create the storage class with `Delete` reclaim policy.

tennix: Besides, why changing the reclaim policy here? I've seen this already done in the above https://github.com/pingcap/tidb-operator/pull/493/files#diff-309b818571e8d7f3366262580cdb73e5R257

jlerche: One of those lines was left over, after talking on slack with @gregwebs it made sense to change the reclaim policy before `terraform destroy`

tennix:
@DanielZhangQD deployed a testing cluster on GCP with this PR, the tikv pods cannot start due to small max-open-file descriptors. Besides, it's the default local ssd has poor performance without installing [Linux Guest Environment](https://cloud.google.com/compute/docs/instances/linux-guest-environment#in_place)

deploy/gcp/main.tf
gregwebs: I would like to solve this in the storageclass. But we haven't done that previously I believe to help prevent accidental data destruction. But I don't know if that is actually helpful since we already replicate data. This would only prevent accidental deletion in the case of TiKV losing 3+ pods.

deploy/gcp/README.md
```diff
@@ -0,0 +1,97 @@
...
+* [helm](https://github.com/helm/helm/blob/master/docs/install.md#installing-the-helm-client) >= 2.9.0
+* [jq](https://stedolan.github.io/jq/download/)
+
+## Configure gcloud
+
+https://cloud.google.com/sdk/docs/initializing
+
+## Setup
+
+The default setup will create a new VPC, two subnetworks, and an f1-micro instance as a bastion machine. The GKE cluster is created with the following instance types as worker nodes:
```
tennix: The default setup exceeds the default cpu quota of a GCP project. There should be a note about this.

jlerche: Added a note that addresses this.

deploy/gcp/main.tf
```diff
@@ -0,0 +1,336 @@
...
+    }
+  }
+}
+
+resource "null_resource" "deploy-tidb-cluster" {
+  depends_on = ["null_resource.setup-env", "local_file.tidb-cluster-values", "google_container_node_pool.pd_pool", "google_container_node_pool.tikv_pool", "google_container_node_pool.tidb_pool"]
+
+  provisioner "local-exec" {
+    command = <<EOS
+helm upgrade --install tidb-cluster ${path.module}/charts/tidb-cluster --namespace=tidb -f ${local.tidb_cluster_values_path}
```
DanielZhangQD: Terraform does not run "helm upgrade" when I change the "tidb_version" in variables.tf, add the triggers below after L318 and it works:
+  triggers {
+    values = "${data.template_file.tidb_cluster_values.rendered}"
+  }
+

deploy/gcp/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+> *NOTE*: The default setup, as listed above, will exceed the default CPU quota of a GCP project. To increase your project's quota, please follow the instructions [here](https://cloud.google.com/compute/quotas). The default setup will require at least 91 CPUs, more if you need to scale out.
+
+The terraform script expects three environment variables. You can let Terraform prompt you for them, or `export` them ahead of time. If you choose to export them, they are:
+
+* `TF_VAR_GCP_CREDENTIALS_PATH`: Path to a valid GCP credentials file
+* `TF_VAR_GCP_REGION`: The region to create the resources in, for example: `us-west1`
+* `TF_VAR_GCP_PROJECT`: The name of the GCP project
+
+It is generally considered a good idea to create a service account to be used by Terraform. See https://cloud.google.com/iam/docs/creating-managing-service-accounts for more information on how to manage them.
```
DanielZhangQD: Suggest moving the service account descriptions after the "TF_VAR_GCP_CREDENTIALS_PATH" part, since they describe how to get a credential file.
BTW, it does not make it clear how to get the credential file.

deploy/gcp/outputs.tf
```diff
@@ -0,0 +1,35 @@
...
+
+output "monitor_ilb_ip" {
+  value = "${data.external.monitor_ilb_ip.result["ip"]}"
```
DanielZhangQD: Suggest to add port

gregwebs:
Can you mention the startup-script in the GKE related docs? Maybe move it to the manifests folder?

deploy/gcp/README.md
```diff
@@ -0,0 +1,99 @@
...
+https://cloud.google.com/sdk/docs/initializing
+
+## Setup
+
+The default setup will create a new VPC, two subnetworks, and an f1-micro instance as a bastion machine. The GKE cluster is created with the following instance types as worker nodes:
+
+* 3 n1-standard-4 instances for PD
+* 3 n1-highmem-8 instances for TiKV
+* 3 n1-standard-16 instances for TiDB
+* 3 n1-standard-2 instances for monitor
```
gregwebs: This is because GKE creates 3 nodes even when you only want one? I think we should add the gcloud api requests needed to delete the unused nodes (the scaler does this automatically using the go library). It looks like there is an unused node for TiDB also. I think TiDB and monitor should get re-scheduled if their node is removed.

deploy/gcp/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+To upgrade TiDB cluster, modify `tidb_version` variable to a higher version in variables.tf and run `terraform apply`.
+
+> *Note*: The upgrading doesn't finish immediately. You can watch the upgrading process by `watch kubectl --kubeconfig credentials/kubeconfig_<cluster_name> get po -n tidb`
+
+## Scale TiDB cluster
+
+To scale TiDB cluster, modify `tikv_count`, `tikv_replica_count`, `tidb_count`, and `tidb_replica_count` to your desired count, and then run `terraform apply`.
+
+> *Note*: Currently, scaling in is not supported since we cannot determine which node to scale. Scaling out needs a few minutes to complete, you can watch the scaling out by `watch kubectl --kubeconfig credentials/kubeconfig_<cluster_name> get po -n tidb`
```
gregwebs:
```suggestion
> *Note*: Currently, scaling in is not supported since we cannot determine which node to remove. Scaling out needs a few minutes to complete, you can watch the scaling out by `watch kubectl --kubeconfig credentials/kubeconfig_<cluster_name> get po -n tidb`
```

jlerche:
Per slack conversation with @gregwebs, startup-script related docs changes will be done in a separate PR.

deploy/gcp/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Customize GCP resources
+
+GCP allows attaching a local SSD to any instance type that is `n1-standard-1` or greater. This allows for good customizability.
+
+### Customize TiDB Parameters
+
+Currently, there are not too many parameters exposed to be customized. However, you can modify `templates/tidb-cluster-values.yaml.tpl` before deploying. If you modify it after the cluster is created and then run `terraform apply`, it will not take effect unless the pod(s) is manually deleted.
+
+### Customizing node pools
```
gregwebs:
```suggestion
### Removing unused instances
```

PR #495, opened 2019-05-16
fix `make lint`

### LINKED ISSUE #481: `make lint` does not work
## Bug Report

**What did you do?**

make lint

lint runs on our packages

**What did you see instead?**

```
$ make lint
linting
CGO_ENABLED=0 retool do revive -formatter friendly -config revive.toml $($(go list ./... | grep -vE "pkg/client" | grep -vE "zz_generated"))
/bin/sh: 1: github.com/pingcap/tidb-operator/cmd/controller-manager: not found
```

cofyc:

fixes #481 

Note that a lot of places violates `rule.var-naming`, it makes the PR too complicated to fix all. 

I decide to change the error level of it to "warning" temporarily. We can fix it in a separate PR.

### Check List 

Tests 

Unit test

 ```release-note
NONE
 ```

### DIFF
--- a/pkg/tkctl/config/config.go
+++ b/pkg/tkctl/config/config.go
@@ -14,17 +14,18 @@
 package config
 
 import (
-	"github.com/golang/glog"
-	"gopkg.in/yaml.v2"
 	"io/ioutil"
-	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"k8s.io/cli-runtime/pkg/genericclioptions/resource"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 	"os"
 	"os/user"
 	"path/filepath"
 	"sync"
+
+	"github.com/golang/glog"
+	"gopkg.in/yaml.v2"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/genericclioptions/resource"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 
 	restclient "k8s.io/client-go/rest"
 )
@@ -189,10 +190,7 @@ func (c *TkcContext) SwitchTidbCluster(context, namespace, clusterName string) e
 	if err != nil {
 		return err
 	}
-	if err := ioutil.WriteFile(tcConfigFile, content, 0644); err != nil {
-		return err
-	}
-	return nil
+	return ioutil.WriteFile(tcConfigFile, content, 0644)
 }
 
 func (c *TkcContext) collectOverrides() *clientcmd.ConfigOverrides {
--- a/pkg/tkctl/executor/executor.go
+++ b/pkg/tkctl/executor/executor.go
@@ -16,6 +16,8 @@ package executor
 import (
 	"context"
 	"fmt"
+	"time"
+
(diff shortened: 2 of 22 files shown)


revive.toml
```diff
@@ -21,6 +20,8 @@ warningCode = 0
 [rule.modifies-parameter]
 
 # Add these once issues are fixed
+[rule.var-naming]
+  severity = "warning"
```
cofyc: A lot of places violates rule.var-naming, it makes the PR too complicated to fix all.

I decide to change the error level of it to "warning" temporarily. We can fix it in a separate PR.

pkg/tkctl/config/config.go
```diff
@@ -14,17 +14,18 @@
...
 	"gopkg.in/yaml.v2"
-	"io/ioutil"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/genericclioptions/resource"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"os"
-	"os/user"
-	"path/filepath"
-	"sync"
```
cofyc: imports changed because the code is formatted by `goimports` tool. It automatically puts the standard packages at the front of the import list.

tests/pkg/webhook/route.go
```diff
@@ -42,7 +42,7 @@ func serve(w http.ResponseWriter, r *http.Request, admit admitFunc) {
 			goto returnData
 		}
 	} else {
-		err := errors.New("request body is nil!")
+		err := errors.New("request body is nil")
 		responseAdmissionReview.Response = toAdmissionResponse(err)
```
cofyc: This is required by `rule.error-strings`.

tests/pkg/ops/common.go
```diff
@@ -10,6 +10,7 @@
 // distributed under the License is distributed on an "AS IS" BASIS,
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
 package ops
```
cofyc: This is required by `[rule.package-comments]`.

tests/failover.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/pingcap/tidb-operator/tests/slack"
 
+	// To register MySQL driver
 	_ "github.com/go-sql-driver/mysql"
```
cofyc: This is required by `[rule.blank-imports]`.

cofyc:
It seems that revive does not check test files. I'm looking at how to fix.

pkg/tkctl/executor/executor.go
```diff
@@ -30,7 +31,7 @@ import (
 	"k8s.io/kubernetes/pkg/kubectl"
 	"k8s.io/kubernetes/pkg/kubectl/cmd"
 	"k8s.io/kubernetes/pkg/util/interrupt"
-	"time"
+	v1 "k8s.io/kubernetes/staging/src/k8s.io/api/core/v1"
```
weekface: why use it instead of `"k8s.io/api/core/v1"`?

cofyc: oh, this is a mistake. I'll fix it.

cofyc: fixed

PR #499, opened 2019-05-17
New stability test case for ConfigMap rollout

### LINKED ISSUE #486: New stability test case for ConfigMap rollout
We need a stability test case for ConfigMap rollout, so that we can verify:

* ConfigMap rollout won't break even in chaos environments;
* Problematic ConfigMap won't take down the cluster (only one of the replica will down if the rolling-update logic is right)
* After fixing a problematic ConfigMap, the cluster can recover without extra human intervention;
* ConfigMap rollout won't introduce other problems when interacting with other cluster operations.

General design of this case:

* Action: Update the config of pd/tikv/tidb to a new valid value
  * Desired: rolling-update, new config applied, no down time
* Update the start-script of pd/tikv/tidb to a new valid value
  * Desired: rolling-update, new config applied, no down time
* Update the config or start-script of pd/tikv/tidb to an invalid value
  * Desired: rolling-update blocked on the pod with the largest ordinal, no down time
* Rollback the config or start-script in the previous step
  * Desired: cluster recovers, no down time

aylei:
fix #486 

New stability test case

### Check List 

Tests 
 - Stability test [Passed]

@weekface @xiaojingchen @tennix @onlymellb PTAL


charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -9,6 +9,9 @@
 #
 
 set -uo pipefail
+
+{{ .Values.pd.preStartScript }}
```
aylei: For fault-injection

tests/actions.go
```diff
@@ -328,6 +336,7 @@ func (oi *OperatorConfig) OperatorHelmSetString(m map[string]string) string {
 		"scheduler.logLevel":               "2",
 		"controllerManager.replicas":       "2",
 		"scheduler.replicas":               "2",
+		"imagePullPolicy":                  "Always",
```
aylei: Typically, we will use the `latest` tag of operator image. `Always` can avoid using stale image

tests/actions.go
```diff
@@ -1593,7 +1602,7 @@ func releaseIsNotFound(err error) bool {
 }
 
 func (oa *operatorActions) cloneOperatorRepo() error {
-	cmd := fmt.Sprintf("git clone https://github.com/pingcap/tidb-operator.git %s", oa.cfg.OperatorRepoDir)
+	cmd := fmt.Sprintf("git clone %s %s", oa.cfg.OperatorRepoUrl, oa.cfg.OperatorRepoDir)
```
aylei: If the developers modify the chart in their development branch, they have to configure the git url in order to test the chart modification

tests/cmd/e2e/main.go
```diff
@@ -51,6 +52,7 @@ func main() {
 		WebhookServiceName: "webhook-service",
 		WebhookSecretName:  "webhook-secret",
 		WebhookConfigName:  "webhook-config",
+		ImagePullPolicy:    v1.PullIfNotPresent,
```
onlymellb: The ImagePullPolicy can be omitted. It is set to `IfNotPresent` by default.

aylei: If I understand correctly, according to the communication in this morning, `IfNotPresent` is required for e2e test so I think it's better to enforce this policy explicitly.

onlymellb: ok

PR #500, opened 2019-05-17
Support adding additional pod annotations for PD/TiKV/TiDB

LinuxGit:

Support adding additional pod annotations for PD/TiKV/TiDB.
We could add pod annotation `fluentbit.io/parser: pingcap_log` for PD/TiKV/TiDB pod, so fluentbit knows that  the pod log should be processed using the pre-defined parser called `pingcap_log `.
refer: https://docs.fluentbit.io/manual/filter/kubernetes#kubernetes-annotations

At the server side, `Annotations` field was added in TidbClusterSpec's Pod Spec (PDSpec, TiKVSpec and TiDBSpec). If Annotations field is not empty, the PD/TiKV/TiDB statefulset will add the annotation to PodTemplateSpec's Annotations field.

At the client side, you could add additional annotations in Helm values file `charts/tidb-cluster/values.yaml`. 

### Check List

Tests

 - [] Unit test
 - [] E2E test

The Unit and E2E test cases will be tracked in  issue https://github.com/pingcap/tidb-operator/issues/505.

Code changes

 - Has Helm charts change ✔️
 - Has Go code change ✔️

 ```release-note
Support adding additional pod annotations for PD/TiKV/TiDB,  e.g. [fluentbit.io/parser](https://docs.fluentbit.io/manual/filter/kubernetes#kubernetes-annotations).
 ```

### DIFF
--- a/pkg/manager/member/utils.go
+++ b/pkg/manager/member/utils.go
@@ -206,12 +206,11 @@ func tidbPodName(tcName string, ordinal int32) string {
 
 // CombineAnnotations merges two annotations maps
 func CombineAnnotations(a, b map[string]string) map[string]string {
-	m := make(map[string]string)
-	for k, v := range a {
-		m[k] = v
+	if a == nil {
+		a = make(map[string]string)
 	}
 	for k, v := range b {
-		m[k] = v
+		a[k] = v
 	}
-	return m
+	return a
 }
--- a/pkg/manager/member/utils.go
+++ b/pkg/manager/member/utils.go
@@ -203,3 +203,11 @@ func pdPodName(tcName string, ordinal int32) string {
 func tidbPodName(tcName string, ordinal int32) string {
 	return fmt.Sprintf("%s-%d", controller.TiDBMemberName(tcName), ordinal)
 }
+
+// CombineAnnotations merges two annotations maps
+func CombineAnnotations(a map[string]string, b map[string]string) map[string]string {
+	for k, v := range b {
+		a[k] = v
+	}
+	return a
+}
--- a/pkg/manager/member/utils.go
+++ b/pkg/manager/member/utils.go
@@ -205,9 +205,13 @@ func tidbPodName(tcName string, ordinal int32) string {
 }
 
 // CombineAnnotations merges two annotations maps
-func CombineAnnotations(a map[string]string, b map[string]string) map[string]string {
-	for k, v := range b {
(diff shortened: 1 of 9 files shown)


pkg/manager/member/utils.go
```diff
@@ -203,3 +203,11 @@ func pdPodName(tcName string, ordinal int32) string {
 func tidbPodName(tcName string, ordinal int32) string {
 	return fmt.Sprintf("%s-%d", controller.TiDBMemberName(tcName), ordinal)
 }
+
+// CombineAnnotations merges two annotations maps
+func CombineAnnotations(a map[string]string, b map[string]string) map[string]string {
+	for k, v := range b {
+		a[k] = v
```
onlymellb: If map a is nil, it will cause panic

LinuxGit: fixed, Thanks.

weekface:
- you should run `./hack/codegen.sh` to generate the deepcopy files
- This pr need to add unit and e2e test cases. Now, you can open a new issue to track it and do it on another pr.

pkg/manager/member/utils.go
```diff
@@ -203,3 +203,15 @@ func pdPodName(tcName string, ordinal int32) string {
...
+func CombineAnnotations(a, b map[string]string) map[string]string {
+	m := make(map[string]string)
+	for k, v := range a {
+		m[k] = v
+	}
+	for k, v := range b {
+		m[k] = v
+	}
+	return m
+}
```
onlymellb: Maybe use the following code more refine？
```
if a == nil {
    a = make(map[string]string)
}
for k, v := range b {
    a[k] = v
}
return a
```

LinuxGit: I agree. updated.

LinuxGit:
> * you should run `./hack/codegen.sh` to generate the deepcopy files
> * This pr need to add unit and e2e test cases. Now, you can open a new issue to track it and do it on another pr.

OK. I've add generated deepcopy file, andd add a issue for tracing e2e  cases.https://github.com/pingcap/tidb-operator/issues/505.

weekface:
add a release note like this pr does: https://github.com/pingcap/tidb-operator/pull/479

LinuxGit:
> add a release note like this pr does: #479

OK. I've refined the PR description.

PR #503, opened 2019-05-20
Add case for kube proxy

onlymellb:

-  add stop kube-proxy case
- adjust some code

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
NONE
 ```


shuijing198799:
LGTM

tests/actions.go
```diff
@@ -2303,12 +2303,7 @@ func (oa *operatorActions) StartValidatingAdmissionWebhookServerOrDie(info *Oper
 	if err != nil {
 		err = fmt.Errorf("failed to start webhook server %v", err)
 		glog.Error(err)
-		sendErr := slack.SendErrMsg(err.Error())
-		if sendErr != nil {
-			glog.Error(sendErr)
-		}
-		// TODO use context instead
-		os.Exit(4)
+		slack.NotifyAndPanic(err)
```
weekface: `slack.NotifyAndPanic ` can't exit the process, because this is recovered by http server: https://github.com/golang/go/blob/master/src/net/http/server.go#L1764

we should use context instead. we can use os.Exit() now.

tests/cmd/stability/main.go
```diff
@@ -230,12 +230,18 @@ func main() {
 	// stop one etcd node and k8s/operator/tidbcluster is available
 	faultEtcd := tests.SelectNode(conf.ETCDs)
 	fta.StopETCDOrDie(faultEtcd)
-	defer fta.StartETCDOrDie(faultEtcd)
```
weekface: @xiaojingchen ptal

tests/failover.go
```diff
@@ -449,42 +449,72 @@ func (oa *operatorActions) CheckOneEtcdDownOrDie(operatorConfig *OperatorConfig,
 		if err != nil {
 			return err
 		}
-		glog.V(4).Infof("all clusters is available")
+		glog.V(4).Infof("all clusters are available")
+		return nil
```
weekface: all should be available within 10 minutes.  should not return here.

tests/failover.go
```diff
@@ -449,42 +449,72 @@ func (oa *operatorActions) CheckOneEtcdDownOrDie(operatorConfig *OperatorConfig,
...
+			return err
+
+		}
+		glog.V(4).Infof("k8s cluster is available.")
+		err = oa.CheckTidbClustersAvailable(clusters)
+		if err != nil {
+			return err
+		}
+		glog.V(4).Infof("all clusters are available.")
 		return nil
```
weekface: ditto

tests/failover.go
```diff
@@ -449,42 +449,72 @@ func (oa *operatorActions) CheckOneEtcdDownOrDie(operatorConfig *OperatorConfig,
...
 	}
 	if apiserverPod != nil {
 		affectedPods[apiserverPod.GetName()] = apiserverPod
 	}
-	controllerPod, err := GetControllerManagerPod(oa.kubeCli, faultNode)
+
+	controllerPod, err := GetKubeControllerManagerPod(oa.kubeCli, faultNode)
 	if err != nil {
-		glog.Infof("can't find controllerManager in node:%s", faultNode)
+		slack.NotifyAndPanic(fmt.Errorf("can't find kube-controller-manager in k8s cluster"))
```
weekface: @xiaojingchen  ptal

tests/failover.go
```diff
@@ -449,42 +449,72 @@ func (oa *operatorActions) CheckOneEtcdDownOrDie(operatorConfig *OperatorConfig,
...
 	}
 	if controllerPod != nil {
 		affectedPods[controllerPod.GetName()] = controllerPod
 	}
-	schedulerPod, err := GetSchedulerPod(oa.kubeCli, faultNode)
+
+	schedulerPod, err := GetKubeSchedulerPod(oa.kubeCli, faultNode)
 	if err != nil {
-		glog.Infof("can't find scheduler in node:%s", faultNode)
+		slack.NotifyAndPanic(fmt.Errorf("can't find kube-scheduler in k8s cluster"))
```
weekface: @xiaojingchen  ptal

tests/cmd/stability/main.go
xiaojingchen: it's ok, we recover all etcd when test begins

tests/failover.go
xiaojingchen: it's ok

tests/failover.go
xiaojingchen: @onlymellb  the controller pod may be not in the master, this case is normal

tests/failover.go
xiaojingchen: scheduler pod can run in work node

tests/failover.go
xiaojingchen: no need to panic

tests/actions.go
onlymellb: This panic out of http server's handler func, so this panic cannot recover by the http server

weekface:
@onlymellb please address @xiaojingchen comments

tests/failover.go
onlymellb: If the returned err is not nil, it means that search this component in kube-system namespace has failed. If the master node does not have this component, the returned err is nil. Therefore, it is necessary to send an alarm and panic when the returned error is not nil.

cofyc:
@onlymellb `go vet check` failed.

weekface:
fixed by: https://github.com/pingcap/tidb-operator/pull/556

PR #526, opened 2019-05-27
Remove helm hook annotation for initializer job

### LINKED ISSUE #523: Helm webhook blocks tidb-cluster installation
## Bug Report

v1.12.4

latest

**What did you do?**

`helm install`
`helm install` complete quickly

**What did you see instead?**
`helm install` blocked for 2 minutes waiting for the `tidb-initializr-job` complete

aylei:

close #523 

A fix for wrong conditional expression is also included (`and` & `or` are **functions** in golang template, refer to [template#functions](https://golang.org/pkg/text/template/#hdr-Functions))

Helm hook removed

### Check List 

Tests 

- Manual test

Code changes

 - Has Helm charts change

@LinuxGit @xiaojingchen @tennix PTAL


aylei:
After communicating with @weekface , make the original tidb-configmap stable is a better idea. I've updated this PR accordingly.

@xiaojingchen PTAL again😆

charts/tidb-cluster/templates/tidb-configmap.yaml
```diff
@@ -1,11 +1,10 @@
+{{/*
+tidb-initializer-job need a stable configmap name
+*/}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
-{{- if .Values.enableConfigMapRollout }}
-  name: {{ template "cluster.name" . }}-tidb-{{ template "tidb-configmap.data-digest" . }}
-{{- else }}
   name: {{ template "cluster.name" . }}-tidb
```
weekface: I think `pd-configmap.yaml` and `tikv-configmap.yaml` should also be modified.

aylei: What's the purpose of this? `pd-configmap` and `tikv-configmap` are only intended to be referenced by the corresponding statefulset and there's no compatibility issue.

BTW, actually there **is** an issue about configmap rollout: the old configmap will be deleted when the new configmap created, which will cause the old `controller-revision`'s reference to the old configmap is invalid. Although this issue is not a problem now because of the careful statefulset upgrade procedure, it should (and will) be fixed for better maintainability.

weekface: OK, it is not related to this issue. Another PR is more reasonable.

aylei:
> I think it's better to separate the initializer configurations with tidb itself configurations. So the initializer job can have a standalone stable configmap.

Yes, It **should** be if there's no backward compatibility issue. If we choose a different configmap name for initializer job, users who have tidb-clusters installed cannot upgrade to our new chart because it will change the configmap name of initializer job, which will cause an error because the podTemplate of `Jobs` is immutable.

tennix:
The job should be run once. Does helm try to upgrade the Job even if it's already completed?

aylei:
> The job should be run once. Does helm try to upgrade the Job even if it's already completed?

According to my previous experience, it is. That's why I added the helm hook annotation in the initializer job, but this was proven to be un-acceptable due to the wait time when install tidb-cluster.

PR #529, opened 2019-05-28
make webhook configurable

shuijing198799:
Make webhook configurable
Use shell script to patch namespace to webhook.yaml 
### Check List 

Tests 

 - E2E test
 - Stability test

Code changes

 - Has Go code change
 - Has scripts change

Related changes

 ```release-note

 ```
None

### DIFF
--- a/manifests/create-cert.sh
+++ b/manifests/create-cert.sh
@@ -37,13 +37,13 @@ while getopts "$optstring" opt; do
 					;;
 			esac
 			;;
-        n)
-            namespace="${2}"
-            ;;
-        *)
-            usage
-            ;;
-    esac
+		n)
+			namespace="${2}"
+			;;
+		*)
+			usage
+			;;
+	esac
 done
 
 echo $namespace
--- a/manifests/create-cert.sh
+++ b/manifests/create-cert.sh
@@ -14,11 +14,26 @@ detailed explantion and additional instructions.
 
 The server key/cert k8s CA cert are stored in a k8s secret.
 
+       --namespace        Namespace where webhook service and secret reside.
 EOF
     exit 1
 }
 
-namespace=default
+while [[ $# -gt 0 ]]; do
+    case ${1} in
+        --namespace)
+            namespace="$2"
+            shift
+            ;;
+        *)
+            usage
+            ;;
+    esac
+    shift
+done
+
+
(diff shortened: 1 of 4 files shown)


shuijing198799:
@xiaojingchen @onlymellb PTAL

manifests/create-cert.sh
```diff
@@ -14,11 +14,26 @@ detailed explantion and additional instructions.
 
 The server key/cert k8s CA cert are stored in a k8s secret.
 
+       --namespace        Namespace where webhook service and secret reside.
 EOF
     exit 1
 }
 
-namespace=default
+while [[ $# -gt 0 ]]; do
```
xiaojingchen:
```suggestion
while [[ $# -gt 1 ]]; do
```

manifests/create-cert.sh
```diff
@@ -14,11 +14,26 @@ detailed explantion and additional instructions.
...
+    case ${1} in
+        --namespace)
+            namespace="$2"
```
xiaojingchen:
```suggestion
            namespace="${2}"
```

shuijing198799: It seems the same ……

weekface:
@onlymellb please take a look.

manifests/create-cert.sh
onlymellb: Refer [here](https://github.com/heroku/heroku-buildpack-php/blob/master/bin/heroku-hhvm-nginx#L141) to implement shell parameter parsing

manifests/create-cert.sh
```diff
@@ -14,11 +14,26 @@ detailed explantion and additional instructions.
...
+            ;;
+        *)
+            usage
+            ;;
+    esac
+    shift
+done
+
+
+[ -z ${namespace} ] && namespace=default
```
onlymellb: Use `namespace=${namespace:-default}` instead.

tennix: I think we should default to `tidb-admin` namespace.

manifests/create-cert.sh
shuijing198799: done

manifests/create-cert.sh
shuijing198799: done

PR #530, opened 2019-05-28
deploy/aws: split public and private subnets when using existing vpc

### LINKED ISSUE #514: AWS EKS: unable to connect to the cluster with create_vpc=false and existing VPC
## Bug Report

AWS EKS 1.12 on us-east-2

**What did you do?**
1. Create a new tidb cluster on EKS following the [README](https://github.com/pingcap/tidb-operator/blob/master/deploy/aws/README.md) with create_vpc tf variable set to true
2. Verified that creation is successful by logging into tidb and executing queries
3. Create a second tidb cluster with a copy of the terraform in a separate directory (so that tfstate files don't interfere with each other), this time with create_vpc = false, i.e., my tfvar files looks like

```
pd_instance_type = "c5d.large"
tikv_instance_type = "c5d.large"
tidb_instance_type = "c4.large"
monitor_instance_type = "c5.large"

pd_count = 1
tikv_count = 1
tidb_count = 1

cluster_name = "rep_george"
tikv_root_volume_size = "50"

region = "us-east-2"
create_vpc=false
vpc_id="$VPC_ID_CREATED_IN_STEP_ONE"
subnets=$SUBNET_IDS_CREATED_IN_STEP_ONE
```

the second tidb cluster will be created inside the same VPC, but on a second EKS cluster

**What did you see instead?**
Unable to connect to the bastion host, because it went to the private subnet

From the tf code, seems that when create_vpc=false, we should use separate subnet id variables for bastion and EKS, so that bastion goes to the public subnet, EKS private?
https://github.com/pingcap/tidb-operator/blob/master/deploy/aws/main.tf#L72
https://github.com/pingcap/tidb-operator/blob/master/deploy/aws/main.tf#L89

AstroProfundis:
This should close #514 
- Split `subnets` to `private_subnet_ids`and `public_subnet_ids`, to correctly put resources into different network.
- Add `cluster_name` to helm release name, making pod names unique
### Check List 

Tests 

Code changes

 - Has Terraform change
 - Has documents change

Side effects

 - It may cause error when using `terraform` with new code to manage an cluster deployed by old versions.

Related changes

 - N/A

 ```release-note

 ```

### DIFF
--- a/deploy/aws/README.md
+++ b/deploy/aws/README.md
@@ -76,7 +76,7 @@ monitor_endpoint = http://abd299cc47af411e98aae02938da0762-1989524000.us-east-2.
 region = us-east-2
 tidb_dns = abd2e3f7c7af411e98aae02938da0762-17499b76b312be02.elb.us-east-2.amazonaws.com
 tidb_port = 4000
-tidb_version = v2.1.8
+tidb_version = v3.0.0-rc.1
 ```
 
 > **Note:** You can use the `terraform output` command to get the output again.
@@ -127,7 +127,7 @@ For example, to upgrade the cluster to version 2.1.10, modify the `tidb_version`
  }
 ```
 
-> *Note*: The upgrading doesn't finish immediately. You can watch the upgrading process by `kubectl --kubeconfig credentials/kubeconfig_<cluster_name> get po -n tidb --watch`.
+> **Note**: The upgrading doesn't finish immediately. You can watch the upgrading process by `kubectl --kubeconfig credentials/kubeconfig_<cluster_name> get po -n tidb --watch`.
 
 ## Scale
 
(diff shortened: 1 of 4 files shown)


deploy/aws/README.md
```diff
@@ -76,7 +76,7 @@ monitor_endpoint = http://abd299cc47af411e98aae02938da0762-1989524000.us-east-2.
 region = us-east-2
 tidb_dns = abd2e3f7c7af411e98aae02938da0762-17499b76b312be02.elb.us-east-2.amazonaws.com
 tidb_port = 4000
-tidb_version = v3.0.0-rc.1
+tidb_version = v2.1.8
```
tennix: Oh, I forgot to upgrade the tidb versions for aws and gcp deployment. I'll send a PR later to update the versions.

tennix: I've sent a PR to fix this #532

deploy/aws/variables.tf
```diff
@@ -30,14 +36,27 @@ variable "public_subnets" {
...
-variable "subnets" {
-  description = "subnet id list"
+# The subnet IDs of your private and public networks, if you want
+# to use the same subnets for both private and public usage, just
+# set their values identical.
+# These values will be ignored if `create_vpc=true`
+variable "private_subnet_ids" {
+  description = "private subnet id list"
+  type = "list"
+  default = ["subnet-899e79f3", "subnet-a72d80cf", "subnet-a76d34ea"]
```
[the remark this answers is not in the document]
AstroProfundis: These are samples and should be changed to user's values if they need to use existing ones, by default (with `create_vpc=true`) these values are ignored.

PR #531, opened 2019-05-28
stable scheduling: fix bug in reporting assigned nodes of tidb members

cofyc:

fix bug in reporting assigned nodes of tidb members

### Check List 

Code changes

 - Has Go code change

 ```release-note
NONE
 ```


pkg/manager/member/tidb_member_manager.go
```diff
@@ -389,6 +389,7 @@ func (tmm *tidbMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, se
 		oldTidbMember, exist := tc.Status.TiDB.Members[name]
 		if exist {
 			newTidbMember.LastTransitionTime = oldTidbMember.LastTransitionTime
+			newTidbMember.NodeName = oldTidbMember.NodeName
```
cofyc: copy the old value because we should not clear it in any case (e.g. new pod is in pending status)

Issue #544, opened 2019-06-02
E2E Test failure


tkanng:
## Bug Report

- Docker in Docker mode
- Commit id: https://github.com/pingcap/tidb-operator/commit/46695bec10da99da53021d70ce8663de4eb52160

**What version of Kubernetes are you using?**

```
root@iZhp37kmiszbkwzt5oh9csZ:~/k# kubectl version
Client Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.0-alpha.1.222+0b8566f3887a2d", GitCommit:"0b8566f3887a2d13baba623d88aeb64f6e637f46", GitTreeState:"clean", BuildDate:"2019-05-04T08:58:31Z", GoVersion:"go1.11.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.1", GitCommit:"4ed3216f3ec431b140b1d899130a69fc671678f4", GitTreeState:"clean", BuildDate:"2018-10-05T16:36:14Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}

```

**What version of TiDB Operator are you using?**

```
root@iZhp37kmiszbkwzt5oh9csZ:~/k# kubectl  exec tidb-controller-manager-6c4dc9fbcf-fwfs4  -n pingcap  -- tidb-controller-manager -V
TiDB Operator Version: version.Info{TiDBVersion:"2.1.0", GitVersion:"v1.0.0-beta.2.34+37d693366bc562", GitCommit:"37d693366bc5622c5975de08fbb4df6fc78a9000", GitTreeState:"clean", BuildDate:"2019-05-31T05:37:32Z", GoVersion:"go1.12", Compiler:"gc", Platform:"linux/amd64"}

```
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

```
root@iZhp37kmiszbkwzt5oh9csZ:~/k# kubectl get pvc -n e2e-cluster1
NAME                       STATUS   VOLUME              CAPACITY   ACCESS MODES   STORAGECLASS    AGE
pd-e2e-cluster1-pd-0       Bound    local-pv-9371219a   39Gi       RWO            local-storage   20m
pd-e2e-cluster1-pd-1       Bound    local-pv-c52702d3   39Gi       RWO            local-storage   20m
[...]

Issue #293, opened 2019-03-04
controller manager should check the PD svc/pod status.


gregwebs:
My cluster has a single PD pod stuck in status Pending. If I look at the controller manager logs, I see repeats of this:

```
pd_member_manager.go:204] failed to sync TidbCluster: [tidb54/demo]'s status, error: Get http://demo-pd.tidb54:2379/pd/health: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

tidb-operator should know already from Kubernetes that there are no pods running for the service. Either because there are no endpoints or because it looks at the status of the underlying pods.

tennix:
So you want to remove the annoying error logs in this case?

gregwebs:
It would be great to see "service tidb54/demo-pd has no endpoints" instead. Just knowing that something cannot connect within a timeout is a very vague failure mode both for the user of the log and tidb-operator.

zhaohaidao:
Is there anyone working on it?
I want to have a try

tennix:
@zhaohaidao Currently, no one working on it. So just go ahead. We really appreciate your help.

weekface:
close in favor of #545

Issue #429, opened 2019-04-25
add unit test to tidb_controller GetInfo Interface #406


shuijing198799:
In #406 added a new interface GetInfo, we need to add unit tests for this interface later

shuijing198799:
In #406 added a new interface GetInfo, we need to add unit tests for this interface later.

PR #545, opened 2019-06-02
Check PD enpoints status when it's unhealthy.

tkanng:

This PR check PD endpoints status when PD is unhealthy, according to https://github.com/pingcap/tidb-operator/issues/293.

- Add `endpointsLister` to `pdMemberManager`
- Chage the order `pdClient.GetHealth()` and `pdClient.GetCluster()`

### Check List 

Tests 

 - Unit test [Done]
 - E2E test: Something went wrong when I try to do E2E test. [e2e test failure](https://github.com/pingcap/tidb-operator/issues/544)

Code changes
 - Has Go code change

Related changes

 ```release-note
NONE
 ```


pkg/manager/member/pd_member_manager.go
```diff
@@ -40,6 +40,7 @@ type pdMemberManager struct {
 	setLister    v1beta1.StatefulSetLister
 	svcLister    corelisters.ServiceLister
 	podLister    corelisters.PodLister
+	epsLister     corelisters.EndpointsLister
```
weekface: use gofmt to format the source code.

weekface:
@gregwebs PTAL

tkanng:
@weekface 
> * Use `gofmt` to format the code, or the `make check` will be filled.

Hi, the code has been formatted in the latest commit, and passed `make check` after I remove `megacheck` tool. And there is something wrong when I execute `make check` command. Detail: https://github.com/pingcap/tidb-operator/issues/546

> * Can you add the unit tests or create an issue to track it, like: #429 does.

New issue: https://github.com/pingcap/tidb-operator/issues/547

weekface:
Yes, `megacheck ` has been removed, this is a fix pr: #548

tkanng:
Hi, the latest commit adds some unit test about endpoints. However, I'm not sure it's good enough, so please help check it again. Thanks!

pkg/manager/member/pd_member_manager.go
```diff
@@ -258,18 +261,27 @@ func (pmm *pdMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, set
 
 	pdClient := pmm.pdControl.GetPDClient(tc)
 
-	cluster, err := pdClient.GetCluster()
+	healthInfo, err := pdClient.GetHealth()
```
weekface: why do you chage the order `pdClient.GetHealth()` and `pdClient.GetCluster()`?

If there is no special need, it is better to stay in the old order.

weekface:
@xiaojingchen @aylei ptal

pkg/manager/member/pd_member_manager.go
tkanng: It maybe more intuitive to check pd's health first, in my opinion.  QAQ

Alright, I'll reset the order.

pkg/manager/member/pd_member_manager.go
```diff
@@ -268,6 +271,15 @@ func (pmm *pdMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, set
 	healthInfo, err := pdClient.GetHealth()
 	if err != nil {
 		tc.Status.PD.Synced = false
+		// get endpoints info
```
weekface: `GetCluster()` will also be failed if there are no endpoints.

so we should move these codes to https://github.com/pingcap/tidb-operator/pull/545/files#diff-006f391cd7cdae269e89bd77e21c1a6fR266

tkanng: Hi, I think that it may be better to check health state at first, and check endpoints' status when it's unhealthy, just like the previous code. Here is what I think:
- It makes no difference to the whole component whether we check health state first or not. 
- Unit tests based on `GetHealth` rather than `GetCluser` are easier to understand, such as: https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/pd_member_manager_test.go#L209. And current unit tests on endpoints are based on whether `GetHealth` fail  or not. 
- More intuitive.

weekface: OK, i am fine with both orders. But these codes must be moved to the first method.

tkanng: Thank you for your patience！

PR #553, opened 2019-06-04
docs: note helm versions in deployment docs

### LINKED ISSUE #552: helm v3.0.0-alpha.1 not work correctly
## Bug Report

**What did you do?**

Run `terraform apply` as documented.

**What did you see instead?**
屏幕快照 2019-06-04 下午2 07 18

AstroProfundis:
Fix #552 
`helm` v3.0.0+ are not supported yet.
### Check List 

Tests 

 - No code

Code changes

 - Has documents change

Side effects

 - N/A

Related changes

 - N/A

 ```release-note

 ```

### DIFF
--- a/deploy/aws/README.md
+++ b/deploy/aws/README.md
@@ -20,7 +20,7 @@ Before deploying a TiDB cluster on AWS EKS, make sure the following requirements
   > **Note:** The access key must have at least permissions to: create VPC, create EBS, create EC2 and create role
 * [terraform](https://learn.hashicorp.com/terraform/getting-started/install.html)
 * [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/#install-kubectl) >= 1.11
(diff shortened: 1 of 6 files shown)


weekface:
Do we need to note all these helm versions? https://github.com/pingcap/tidb-operator/search?l=Markdown&q=helm

lucklove:
> Do we need to note all these helm versions? https://github.com/pingcap/tidb-operator/search?l=Markdown&q=helm

My point is that if it's not tested, then the answer is no. And if it's tested and v3.x not work at all, then the answer is yes. (Of course we should test these cases in the end)

In the case of this pr, the v3.x not work, so we need to document it.

tennix:
Yes, I think all places when mentioning helm versions should be noted.

tennix:
Helm 3.0 is a huge refactor. I don't think we can run with helm v3.0 in any of the environment without modification of the charts, so all the documents need to be updated.

PR #577, opened 2019-06-12
Local SSD on GKE

gregwebs:
* Add GKE ssd provisioning scripts that fix performance and capacity issues
* Document in the operation guide, and link documentation
* Don't use by default for PD.
Benchmarking has now shown a significant improvement.

## Testing

The provisioner script is well tested. I haven't yet tested this contribution to tidb-operator. We need to test
- [ ] documentation links
- [x] terraform deployment


gregwebs:
This is working now. One thing that showed up in my testing is that when deleting a tidb cluster deployment (helm delete + namespace delete) and then deploying again, tikv will be stuck in pending. I will check to see if this is already the current behavior and if changing the namespace helps.

gregwebs:
I tested the previous terraform, and destroying and then deploying again also does not work, so there are no regressions here.

Issue #469, opened 2019-05-08
Create TiKV and TIDB pods immediately with initcontainers to delay startup


gregwebs:
## Feature Request

See #355. We can provision and scale capacity faster with autoscalers and autoprovisioners if the pods for TIKV and TIDB are created immediately when the PD pods are created.

We can do this and avoid downside by creating the pods with an init container. The init container will block the startup of the TiKV pod until PD is ready. Similarly it will block TiDB startup until TIKV and PD are ready.

There are at least a few approaches to signalling the init container to exit.

* It directly checks that the services are ready via API call
* It checks on the status from the readiness probe
* tidb-operator signals the init container

I am looking for feedback on the best approach here.

cofyc:
It seems better to look at pending pods instead of TidbCluster CRDs to provision new nodes. It will be universal and can be used for any kinds of pods in cluster management.

I agree to create the pods immediately and use init container to check is it ready to run. The idea is like the webhook that we are going to implement to prevent pods from being deleted until they are ready to exit.

But I'm not familiar with this, not sure if it is hard to implement in init containers. cc @weekface

weekface:
Create all pods and keep them pending is ok to me and it can provision and scale capacity faster with autoscalers and autoprovisioners.

But use the init container solution may not a good solution, it requires a lot of refactoring of existing code. For example, the operator will not create TiKV's statefulset before PD is [running](https://github.com/pingcap/tidb-operator/blob/c8648e7fe5f9fe1e547ef5a4e3a78399f0c628ff/pkg/manager/member/tikv_member_manager.go#L98) now. All the three components's operations are sequential, include create, upgrade, scale and any other operations.

I suggest:
- add another controller to watch the TidbCluster object, when it is created, increase the Kube nodes if necessary, this is easy to implement now.
- using validating admission webhooks, we can implement this feature easily by creating these three statefulsets at once, but it needs a refactor of operator.

gregwebs:
> add another controller to watch the TidbCluster object, when it is created, increase the Kube nodes if necessary, this is easy to implement now.

I can do this, but it won't be compatible with autoscalers and provisioners that are based on pending pods.

> it requires a lot of refactoring of existing code.

I think the code you pointed to would just change the check to be:

if tc.PDIsAvailable() { tc.stopTiKVInitContainer() }

weekface:
[...]

PR #581, opened 2019-06-13
create all statefulsets immediately

gregwebs:
# TODO

- [ ] run stability tests
- [ ] decide on renaming wait-for-pd

As per #469 this creates statefulsets and pending pods immediately to help autoscalers scale more quickly.

Design doc is [here](https://docs.google.com/document/d/1LFlGEy2NRwtmbwUb_hS9_B7POIseBFUsOSyhezf1RXg)

A blocking init container wait-for-pd is used to prevent TiKV and TiDB from immediately starting up. This code calls the PD API directly. For TiDB it will call the new status is_initialized API if it is available.

Error handling is also generally improved in the member managers: as a side effect this may fix bugs where the controller sync loop is not re-ran during errors.

The PD API was updated to 3.0 which seems to be backwards compatible for our usage (more testing needed).

### Testing notes

I have created 50+ clusters with the operator that has the changes from the first commit. It has the desired effect for autoscaling. The second commit is still being tested.

* When PD is not available for TiKV, or TiKV is not available for TiDB, go ahead and create the statefulset.
* Don't break from the update function immediately when a RequeueError is seen.

### Check List

Tests

 - Unit test: using existing tests
 - E2E test: using existing e2e tests
 - Stability test: ?
 - Manual test: create single clusters of various sizes (multiple cluster creation often runs into issues with the current operator). Test with pd/tikv/tidb version 3.0.0-rc1 and version 3.0.0

### Side effects

Increased image size
Originally this change meant shipping another 50M binary. However, overall image size is smaller since this PR was originally started. This work was merged in other Pull requests:
* #618 Code was re-arragned so that wait-for-pd minimized its imports to just the interaction with the PD API and logging without importing all of Kubernetes. This reduced the binary size by 50%. We could do the same for the discovery service.
* #592 Removal of debugger symbols reduced the size of all the binaries

Pending Pod warnings
* May surprise existing users that pre-provision capacity and monitor for pending pods. Alerts that alert for pending pods could get triggered.

Upgrade procedure
* When upgrading the tidb-operator image from before this change one must first upgrade the tidb-operator helm chart templates. This is expected, but it could surprise someone since it often isn't necessary.

The PD API was updated to 3.0 which seems to be backwards compatible for our usage. We will probably want to see some additional testing here.

### Related changes

### Does this PR introduce a user-facing change?

Yes, to the extent users look at pod creation (which we have in our tutorials).

 ```release-note
Create TiKV/TiDB pending pods immediately to help autoscalers scale more quickly. They wait for their dependencies (e.g. PD) to be ready with an Init container, so they will show up as in state Init while waiting.
 ```


gregwebs:
This is working well now. I just need to go through the tests. There are some additional things on this branch that I am getting merged in other PRs right now and then I will force push a rebased version of this.

* gcp terraform
* error message improvements
* binary build size reduction

gregwebs:
I saw an issue where TiKV continually fails:

```
[2019/06/20 23:49:42.564 +00:00] [FATAL] [tikv-server.rs:273] ["failed to start node: StringError(\"[src/server/node.rs:171]: cluster ID mismatch, local 6704759663323983770 != remote 6704764185849198648, you are trying to connect to another cluster, please reconnect to the correct PD\")"]
```

This probably means I need to change the wait-for-pd logic to be more like `PDIsAvailable`

tennix:
You are reusing old data for the new cluster.

gregwebs:
@tennix I was destroying and creating clusters so it would be expected for a node to be re-used by a pod of a new cluster. Is this expected behavior then? Note that I deleted the PVC. Am I supposed to manually wipe the PV?

tennix:
Did you use local ssd disk or persistent disk?
I suppose you are using the local ssd disk. So if you delete the PVC, I don't think the PV will be reused. If the PV reclaim policy is `Delete`, then the old PV and data will be deleted by the provisioner, after that a new PV will be provisioned. If the policy is `Retain`, then the PV will be in `Released` state, it will not be used.

gregwebs:
ah, it seems the problem is that I wasn't setting the reclaim policy to Delete.
Because this PR creates pods immediately, node rescheduling is now more possible in my setup where before an empty node would have been automatically deleted.

gregwebs:
I will fix a couple failing tests now. I am ready to start review and get suggestions on test cases to write.

pkg/apis/pdapi/pdapi.go
```diff
@@ -0,0 +1,797 @@
+// Copyright 2018 PingCAP, Inc.
```
cofyc: `pkg/apis` is a directory to store Kubernetes API types. IMO it's not a good idea to put PD controller client here.

pkg/controller/pd_control.go
```diff
@@ -14,802 +14,18 @@
 package controller
```
cofyc: how about focusing on design change in this PR and optimize later?

pkg/apis/pdapi/pdapi.go
gregwebs: Where would a good place be for this?
I don't think of this as a PD controller but instead just the PD API interaction. Logic to "control" PD is elsewhere. The code take types from this project, serializes them for the API, sends a request, and then convert the response into a type for this project. The file is renamed now to pdapi. There is an interface PDControlInterface that I would also rename, but I tried to minimize my changes to only what is required.

pkg/controller/pd_control.go
gregwebs: okay, here is the PD API change: https://github.com/pingcap/tidb-operator/pull/618

gregwebs:
The usage of PD APIs in wait-for-pd should be updated. For TiDB 3.0 the TiDB component should check the PD api cluster_status for the is_initialized flag as per #586

gregwebs:
One piece of feedback I got it is that the name "wait-for-pd" may not accommodate future changes well. "tidb-init" is a very generic name, but I think too generic. An example given is password initialization: we might move that to an init container, but I think it will be best off as a separate executable.
My suggestion is "wait-for-deps" would help accommodate the fact that we are actually waiting for TiKV replication. That would also work with the example of putting the waiting for the discovery service that PD does into the same binary. "deps" in "wait-for-deps" is not the most clear name. "services" might be better.

gregwebs:
Closes #586.

cmd/wait-for-pd/main.go
```diff
@@ -0,0 +1,149 @@
...
+		waitFunction = waitForLeaderFunc
+	} else {
+		glog.Fatalf("Expected either the flag --initialization or --leader")
+	}
+
+	for {
+		if waitFunction() {
+			break
+		}
+		time.Sleep(oneSecond)
```
cofyc: nit: `time.Second`

cmd/wait-for-pd/main.go
```diff
@@ -0,0 +1,149 @@
...
+	if printVersion {
+		version.PrintVersionInfo()
+		os.Exit(0)
+	}
+	version.LogVersionInfo()
+
+	logs.InitLogs()
+	defer logs.FlushLogs()
+
+	oneSecond, err := time.ParseDuration("1s")
```
cofyc: configurable in the future?

cofyc: I think we can keep it simple at first.

cofyc:
target 1.1 release, what do you think?

gregwebs:
@cofyc that decision doesn't seem to be up to me. However, this PR does fix some problematic error handling which can be seen in the test cases.

gregwebs:
I see an issue where the operator will expand the tidb statefulset when it is waiting in init for a long time

```
demo-discovery-5bb58f7bb5-wklmq   1/1     Running    0          18m
demo-monitor-6f9c9fdffb-59bkp     1/1     Running    0          18m
demo-pd-0                         1/1     Running    0          18m
demo-pd-1                         1/1     Running    0          18m
demo-pd-2                         1/1     Running    1          18m
demo-tidb-0                       0/2     Init:0/1   0          18m
demo-tidb-1                       0/2     Init:0/1   0          18m
demo-tidb-2                       0/2     Init:0/1   0          9m23s
demo-tidb-3                       0/2     Init:0/1   0          9m23s
demo-tidb-4                       0/2     Init:0/1   0          4m23s
demo-tikv-0                       0/1     Pending    0          18m
demo-tikv-1                       1/1     Running    0          18m
demo-tikv-2                       1/1     Running    0          18m
```

gregwebs:
When this came out of the init state the extra tidb pods did get terminated.

cmd/wait-for-pd/main.go
gregwebs: yes, configurable in the future

cmd/wait-for-pd/main.go
gregwebs: We may want the time to be configurable in the future

gregwebs:
okay, I added some code to handle that case

pkg/manager/member/tidb_member_manager.go
```diff
@@ -151,8 +154,12 @@ func (tmm *tidbMemberManager) syncTiDBStatefulSetForTidbCluster(tc *v1alpha1.Tid
 		tc.Status.TiDB.StatefulSet = &apps.StatefulSetStatus{}
 		return nil
 	}
-
-	if !tc.TiKVIsAvailable() {
+	initialized, err := controller.GetPDClient(tmm.pdControl, tc).GetClusterInitialized()
```
gregwebs: Doing an API call here may slow down the sync loop. Previously it checked the status. Perhaps this API call should be done elsewhere and set a status field that is checked here?

pkg/manager/member/member.go
```diff
@@ -0,0 +1,30 @@
...
+		},
+		{
+			Name:  "CLUSTER_NAME",
+			Value: tcName,
+		},
+	}
+
+	return corev1.Container{
+		Name:    "wait-for-pd",
+		Image:   operatorImage,
```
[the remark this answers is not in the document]
gregwebs: Good point: this function should only be used on create, but not update.

cmd/wait-for-pd/main.go
```diff
@@ -0,0 +1,140 @@
...
+
+package main
+
+import (
+	"github.com/golang/glog"
+	"github.com/pingcap/tidb-operator/pkg/pdapi"
+	"github.com/pingcap/tidb-operator/version"
+	"k8s.io/apiserver/pkg/util/logs"
+
+	"flag"
```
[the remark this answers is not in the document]
gregwebs: gofmt seems fine with that. I moved it above now though.

pkg/manager/member/member.go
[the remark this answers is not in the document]
gregwebs: I added the change only on create functionality

pkg/manager/member/tidb_member_manager.go
gregwebs: I moved this into the PD status: it will be easier to create a PD API that reduces our API calls there.

aylei:
@gregwebs Cut of `release-1.1` branch is targeted at December 31th, could you please confirm whether this feature is expected to be releases in 1.1 or should be re-targeted to 1.2? Thanks!

gregwebs:
Probably needs to be 1.2

aylei:
Okay, removed from v1.1 milestone for now, feel free to add it back if this is necessary to land in v1.1

PR #585, opened 2019-06-15
GKE terraform

gregwebs:
* use gcloud CLI to automate service account provisioning
* provide option to use small instances or prod instances
* local-exec scripts: add set -e and make idempotent
* use data.google_compute_zones.available.names to prevent bastion re-creation


deploy/gcp/prod.tfvars
```diff
@@ -0,0 +1,5 @@
+tidb_operator_version = "v1.0.0-beta.3.start-immediate"
+tidb_operator_registry = "gcr.io/pingcap-tidb-alpha"
```
tennix: How does this registry get updated?

deploy/gcp/main.tf
```diff
@@ -308,62 +308,77 @@ resource "null_resource" "setup-env" {
...
+  ]
 
-provisioner "local-exec" {
+  triggers = {
+    values = data.template_file.tidb_cluster_values.rendered
+  }
+
+  provisioner "local-exec" {
+    interpreter = ["bash", "-c"]
 command = <<EOS
```
tennix: Use `EOT` instead of `EOS` to make Terraform 0.12 format this correctly.

deploy/gcp/README.md
```diff
@@ -34,30 +34,44 @@ gcloud services enable container.googleapis.com
...
-# Replace the values with the path to the JSON file you have downloaded, the GCP region and your GCP project name.
-export TF_VAR_GCP_CREDENTIALS_PATH="/Path/to/my-project.json"
-export TF_VAR_GCP_REGION="us-west1"
-export TF_VAR_GCP_PROJECT="my-project"
+git clone --depth=1 https://github.com/pingcap/tidb-operator
+cd tidb-operator/deploy/gcp
 ```
 
-You can also append them in your `~/.bash_profile` so they will be exported automatically next time.
+You need to decide on instance types. If you just want to just get a feel for a TiDB deployment and lower your cost, you can use the small settings.
```
tennix:
```suggestion
You need to decide on instance types. If you just want to get a feel for a TiDB deployment and lower your cost, you can use the small settings.
```

deploy/gcp/main.tf
```diff
@@ -308,62 +308,77 @@ resource "null_resource" "setup-env" {
...
+set -euo pipefail
```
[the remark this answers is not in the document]
gregwebs: When you put a shell line in an if/until/when then the bash script does not exit on error but instead used the control flow.

deploy/gcp/prod.tfvars
gregwebs: these 2 variables are an accidental commit from my testing.

deploy/gcp/main.tf
gregwebs: any word will work here.

gregwebs:
Also added:
* change the order of pool creation to prioritize local SSD setup and tiller deployment
* local volume provisioner shell script improvements
* don't auto-upgrade/repair
* set a maitenance window

PR #591, opened 2019-06-18
improve error messages

gregwebs:
Improve error messages that I saw in the logs when debugging. This was tested on GKE.


weekface:
make test failed

PR #601, opened 2019-06-21
local SSD provision: use a single install command

gregwebs:
I tested this on GKE


gregwebs:
I also added a commit that reduces the number of repositories that are updated.

Issue #604, opened 2019-06-24
Enable scheduled backup to Amazon S3


shinnosuke-okada:
## Feature Request

**Is your feature request related to a problem? Please describe:**
With current set of helm charts and terraform scripts, mydumper backup data cannot be uploaded to Amazon S3.

**Describe the feature you'd like:**
Enable the option to upload backup data to Amazon S3 using `scheduledBackup`.

**Describe alternatives you've considered:**
N/A

**Teachability, Documentation, Adoption, Migration Strategy:**
N/A

gregwebs:
Have you tried using the [ceph configuration](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-backup/templates/backup-job.yaml#L49)? ceph uses the AWS API.

shinnosuke-okada:
`uploader` uses a ceph specific code:
https://github.com/pingcap/tidb-cloud-backup/blob/master/pkg/setup.go#L108
Made changes on my end to get it to work:
https://github.com/shinnosuke-okada/tidb-operator/tree/support-s3-backup

gregwebs:
It looks like you added support for specifying the region. Can you open a pull request?

shinnosuke-okada:
Will do once I pass all your test suites. Currently struggling to get tidb-operator-e2e to work locally.

gregwebs:
I don't think the existing test suite will execute this new template code, although I don't have much familiarity with the e2e testing.

shinnosuke-okada:
Okay. In that case, I'll perform a basic manual test tomorrow, and open a PR. Thanks :)

shinnosuke-okada:
PR merged. Closing the issue.

PR #606, opened 2019-06-25
Add Amazon S3 support to the backup/restore features

shinnosuke-okada:
https://github.com/pingcap/tidb-operator/issues/604

Respective commit comment should explain what's being changed. Added Amazon S3 support to the backup/restore feature.

The problem with the current design is that Ceph and S3 has to share the same set of environment variables (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`) and thus both cannot be enabled at the same time unless we make changes to `uploader` and `downloader`. Given that they're planned to be deprecated with `rclone`, I wouldn't address this limitation in this PR.

### Check List 

Tests 

 - Unit test
1. Enable `scheduledBackup`, and upload to Amazon S3 everyday - uploaded successfully to the specified S3 bucket
2. Enable `scheduledBackup`, and upload to Amazon S3 every 15 minutes - uploaded successfully to the specified S3 bucket
3. Restore manually according to the documentation - downloaded and loaded the data successfully from the specified S3 bucket

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

 - N/A

Related changes

 - N/A

 ```release-note
Full backup data can be uploaded to or downloaded from Amazon S3
 ```


images/tidb-operator-e2e/tidb-cluster-values.yaml
```diff
@@ -224,7 +224,7 @@ fullbackup:
   binlogImage: pingcap/tidb-binlog:v3.0.0-rc.1
   binlogImagePullPolicy: IfNotPresent
   # https://github.com/tennix/tidb-cloud-backup
-  mydumperImage: pingcap/tidb-cloud-backup:latest
+  mydumperImage: pingcap/tidb-cloud-backup:20190610
```
gregwebs: It looks like this change was made for testing and shouldn't be committed? Perhaps we need to improve our e2e testing workflow.

tests/actions.go
```diff
@@ -2032,7 +2032,7 @@ func (oa *operatorActions) getBackupDir(info *TidbClusterConfig) ([]string, erro
 			Containers: []corev1.Container{
 				{
 					Name:    getBackupDirPodName,
-					Image:   "pingcap/tidb-cloud-backup:latest",
+					Image:   "pingcap/tidb-cloud-backup:20190610",
```
gregwebs: It looks like this change was made for testing and shouldn't be committed?

images/tidb-operator-e2e/tidb-cluster-values.yaml
shinnosuke-okada: Okay to revert the changes to e2e testing. Just thought it'd make more sense to test with the same version of tidb-cloud-backup as the main code.

shinnosuke-okada: Another thing is - is this tag `20190610` appropriate to be used in tidb-operator? The only reason why I replaced this is because the `latest` tag is outdated, and doesn't support Amazon S3.

weekface: the `20190610` is the latest of `tidb-cloud-backup` image. use it instead of `latest`.

weekface:
@tennix PTAL

images/tidb-operator-e2e/tidb-cluster-values.yaml
tennix: @shinnosuke-okada Do we have to update the [tidb-cloud-backup](https://github.com/pingcap/tidb-cloud-backup) code to support S3 backup/restore?

shinnosuke-okada: @tennix Yes, the `latest` tag is outdated, and cannot be used to upload to/download from Amazon S3.

shinnosuke-okada: @tennix My bad! Just realized that I misread your message here last time. No need to change the *code*. The latest code is good enough to support S3 backup/restore, and it's available via `20190610`. If you plan to rebuild the `latest` tag or some other tag, we can use that as well.

weekface:
The e2e was broken now, I am working on it.

PR #612, opened 2019-06-27
GKE local ssd provisioner for COS

gregwebs:

The Ubuntu local SSD provisioner has a startup delay while provisioning packages. Make the local SSD provisioner work on COS.

On COS, remount with UUID and set nobarrier.
We do not combine multiple disk like on Ubuntu.
Therefore this is recommended if your data size is < 375 GiB.
Note that by default COS does not set the nobarrier option. This script sets the nobarrier option when it remounts.

### Check List 

Tests 

 - Manual test: create clusters with < 375 GiB of TiKV disk needed using the COS image. Also create a cluster with > 375 GiB of TiKV disk.

Code changes

- manifest changes
 - Has documents change

Side effects

- If you already have this deployed, Don't forget to run `kubectl delete -n kube-system local-volume-provisioner`.

 ```release-note
on GKE one can use COS for TiKV nodes with small data for faster startup
 ```


docs/operation-guide.md
```diff
@@ -73,14 +73,18 @@ On GKE, local SSD volumes by default are limited to 375 GiB size and perform wor
...
-* install the Linux guest environment, which can only be done on the Ubuntu image, not the COS image
+* install the Linux guest environment on the Ubuntu image or use a recent COS image
 * make sure SSD is mounted with the `nobarrier` option.
 
-We have a [daemonset which does the above performance fixes](../manifests/gke/local-ssd-optimize.yaml).
-We also have a [daemonset that fixes performance and combines all SSD disks together with lvm](../manifests/gke/local-ssd-provision.yaml).
+We also have a [daemonset](../manifests/gke/local-ssd-provision.yaml) that
+* fixes any performance issues
+* remounts local SSD disks with a UUID for safety
+* On Ubuntu combines all local SSD disks into one large disk with lvm tools.
```
cofyc: Should we let the user decide whether to combine the disks? In case users need to deploy multiple TiKV instances on one node.

gregwebs: In the cloud I don't think it makes sense to deploy multiple TiKV instances to one node. I am more concerned about a cluster that both runs tidb-operator and something else. But it is up to them to take these daemonsets and use them appropriately.

PR #613, opened 2019-06-28
Fix e2e problem

shuijing198799:
Fixed a bug that exec command to be submitted before the container was ready
Ensure that the pod status is running before exec a command
### Check List 

Tests 

 - E2E test
 - Stability test

Code changes

 - Has Go code change


tests/actions.go
```diff
@@ -2076,11 +2076,13 @@ func (oa *operatorActions) getBackupDir(info *TidbClusterConfig) ([]string, erro
 	}
 
 	fn = func() (bool, error) {
-		_, err := oa.kubeCli.CoreV1().Pods(info.Namespace).Get(getBackupDirPodName, metav1.GetOptions{})
-		if errors.IsNotFound(err) {
-			return false, nil
+		pod, err := oa.kubeCli.CoreV1().Pods(info.Namespace).Get(getBackupDirPodName, metav1.GetOptions{})
+		if err == nil && pod.Status.Phase == corev1.PodRunning {
```
xiaojingchen: the conditions: `pod.Status.Phase==corev1.PodRunning` cannot make sure the pod is running

cofyc: It is enough.

https://github.com/kubernetes/kubernetes/blob/04c1fa22ec0bf502f94992a79f86081ffb7f67f4/pkg/apis/core/types.go#L2188-L2190

But I suggest to reuses the Kubernetes e2e/framework code.

https://github.com/kubernetes/kubernetes/blob/04c1fa22ec0bf502f94992a79f86081ffb7f67f4/test/e2e/framework/framework.go#L443-L446

shuijing198799: OK, I will adjust the judgment conditions.

shuijing198799: Since we are using the k8s 12.5 version, I went looking for a description of k8s 12.5 for pod running.

// PodRunning means the pod has been bound to a node and all of the containers have been started.
// At least one container is still running or is in the process of being restarted.

I think this condition judgment is OK.

Issue #575, opened 2019-06-12
Terraform should use released helm charts


aylei:
## Feature Request

**Describe the feature you'd like:**

Currently, the terraform scripts using the chart inside this repository, which is may not stable. We should use the release chart instead and add a new terraform variable to specify the chart version:

```
variable "helm_chart_version" {
    default = "v1.0.0-beta.3"
}
```

- [x] AWS (merged): #616 
- [x] Alibaba Cloud: #658 
- [x] GKE #717 

@gregwebs @jlerche @tennix @weekface PTAL

jlerche:
I think that's a good idea :+1:

jlerche:
Closing now that #717 has been merged.

PR #616, opened 2019-06-29
Multiple clusters management in EKS

aylei:
Support multiple TiDB clusters management in EKS with some updates per #575 
 
The control plane (Kubernetes master and tidb-operator) and data plane (node pools and helm release) are separate into two different modules: `tidb-operator` and `tidb-cluster`, a top-level module compose one `tidb-operator` module and one or more `tidb-cluster` modules to support multiple cluster management. 

See the [README of top-level module](https://github.com/aylei/tidb-operator/blob/aws-multi-cluster/deploy/aws/README.md), [README of tidb-operator module](https://github.com/aylei/tidb-operator/tree/aws-multi-cluster/deploy/aws/tidb-operator) and [README of tidb-cluster module](https://github.com/aylei/tidb-operator/blob/aws-multi-cluster/deploy/aws/tidb-cluster/README.md) for detail.

### Check List 

Tests 

    - [x] 1 EKS + 1 TiDB cluster
    - [x] 1 EKS + 2 TiDB cluster
    - [x] 1 EKS + 3 TiDB cluster

Code changes

 - Has documents change

Related changes

 ```release-note
The terraform scripts support manage multiple TiDB clusters in one EKS cluster.
 ```

### Limitations

- Terraform do not support loop for modules, I cannot figure out how to define multiple TiDB clusters in `.tfvars`, so it is inevitable to edit the `cluster.tf` directly for multiple cluster management now; 

@tennix @jlerche @gregwebs PTAL, most of the works are done by @tennix in bd2343f, I did some clean up and re-organization of code in the follow up commits.

@jlerche I changed the `aws-tutorial.tfvars` to keep compatibility, hope this won't break your tutorial😅


deploy/aws/README.md
```diff
@@ -174,4 +247,14 @@ It may take some while to finish destroying the cluster.
...
 ```
 
-> **Note:** You have to manually delete the EBS volumes in AWS console after running `terraform destroy` if you do not need the data on the volumes anymore.
+> **Note:**
+>
+> This will destroy your EKS cluster along with all the TiDB clusters you deployed on it.
+
+> **Note:**
+>
+> If you specify service type `LoadBalancer` for the services like the default configuration do, you have to delete these services before destroy, otherwise they will block the subnets to be destroyed.
```
gregwebs:
```suggestion
> If you specify service type `LoadBalancer` for the services like the default configuration do, you have to delete these services before destroy, otherwise they will block the subnets from being destroyed.
```

deploy/aws/eks/local.tf
```diff
@@ -0,0 +1,245 @@
...
+    kubeconfig                  = data.template_file.kubeconfig.rendered
+    kubeconfig_file             = "${var.config_output_path}kubeconfig_${var.cluster_name}"
+    worker_iam_role             = aws_iam_role.workers
+    worker_security_group_id    = local.worker_security_group_id
+    worker_iam_instance_profile = aws_iam_instance_profile.workers
+    vpc_id                      = var.vpc_id
+  }
+
+  ebs_optimized = {
+    "c1.medium"    = false
```
gregwebs: would be good to find a data source for this that gets updated.

deploy/aws/eks/manifests/crd.yaml
```diff
@@ -0,0 +1,103 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
gregwebs: Can we symlink this instead of copying?

deploy/aws/eks/manifests/local-volume-provisioner.yaml
```diff
@@ -36,12 +36,7 @@ spec:
     spec:
       tolerations:
         - key: dedicated
-          operator: Equal
-          value: pd
-          effect: "NoSchedule"
-        - key: dedicated
-          operator: Equal
-          value: tikv
+          operator: Exists
```
gregwebs: does this mean it will run on TiDB nodes now?

deploy/aws/eks/manifests/tiller-rbac.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
```
gregwebs: can we symlink this?

deploy/aws/eks/templates/kubeconfig.tpl
```diff
@@ -0,0 +1,28 @@
+apiVersion: v1
```
gregwebs: I think we should avoid creating this file ourselves. There is an AWS command for this: https://docs.aws.amazon.com/cli/latest/reference/eks/update-kubeconfig.html

deploy/aws/main.tf
```diff
@@ -1,242 +1,50 @@
...
-  key_name = "${module.key-pair.key_name}"
-  associate_public_ip_address = true
-  monitoring = false
-  user_data = "${file("bastion-userdata")}"
-  vpc_security_group_ids = ["${aws_security_group.ssh.id}"]
-  subnet_ids = "${split(",", var.create_vpc ? join(",", module.vpc.public_subnets) : join(",", var.public_subnet_ids))}"
-
-  tags = {
-    app = "tidb"
+    "kubernetes.io/cluster/${var.eks_name}" = "shared"
```
gregwebs: why shared?

deploy/aws/tidb-cluster/pre_userdata
```diff
@@ -0,0 +1,30 @@
...
+    if [ -e "/dev/nvme${i}" ]; then # find the device
+	if [ -e "/dev/nvme${i}n1" ]; then
+	    if ls /dev/nvme${i}n1p* > /dev/null 2>&1; then
+		echo "disk /dev/nvme${i}n1 already parted, skipping"
+	    else
+		echo "disk /dev/nvme${i}n1 is not parted"
+		if ! blkid /dev/nvme${i}n1 > /dev/null; then
+		    echo "/dev/nvme${i}n1 not formatted"
+		    mkfs -t ext4 /dev/nvme${i}n1
+		    mkdir -p /mnt/local-ssd/ssd${i}
```
gregwebs: We should used the uuid here.

deploy/aws/eks/manifests/crd.yaml
aylei: No, this module is supposed to be workable outside this repository. But it is definitely not a good idea to keep a copy here.

How about `kubectl appy https://raw.githubusercontent.com/pingcap/tidb-operator/v1.0.0-beta.3/manifests/crd.yaml`?

deploy/aws/eks/manifests/local-volume-provisioner.yaml
aylei: Yes

deploy/aws/main.tf
aylei: Required for ELB

deploy/aws/eks/templates/kubeconfig.tpl
aylei: Thanks for the reference! However, this is behavior of the original EKS module https://github.com/terraform-aws-modules/terraform-aws-eks/blob/80b0bc79af5d1218490d550d1c63f483e4948ee1/data.tf#L46, and we hope that the changes can be minimized

deploy/aws/tidb-cluster/pre_userdata
tennix: The NVMe ssd disks have fixed device name as described here https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nvme-ebs-volumes.html. But yes it's more robust to use uuid.

aylei:
@tennix @gregwebs I've removed the customized eks module for better maintainability, and addressed all the review comments, PTAL again

deploy/aws/eks/manifests/local-volume-provisioner.yaml
gregwebs: TiDB does not need local SSD disk so that isn't desirable in this setup.

gregwebs: Why not leave it how it was?

deploy/aws/eks/templates/kubeconfig.tpl
gregwebs: :ok_hand:

deploy/aws/tidb-cluster/pre_userdata
gregwebs: Good point. It is possible that the local volume provisioner code could (in the future, even if it doesn't now) assume different things depending on whether it sees it mounted with a UUID.

deploy/aws/eks/manifests/local-volume-provisioner.yaml
aylei: Because the values of taints are updated to ${cluster_name}-tikv in favor of multiple cluster management.

And yes, TiDB nodes don't have to run the provisioner pod, I will add an extra label to address this problem

deploy/aws/eks/templates/kubeconfig.tpl
gregwebs: Created an issue for this: https://github.com/pingcap/tidb-operator/issues/621

deploy/aws/eks/manifests/crd.yaml
gregwebs: Why does this need to work outside this repository? I have been referencing manifests with symlinks in the GCP terraform. The deployment doesn't work without a good way to reference our charts and manifests.

The suggested update is problematic because it will fall out of sync. Better would be to have a variable "manifests" which could be set to `../../manifests` or `https://raw.githubusercontent.com/pingcap/tidb-operator/v1.0.0-beta.3`.

Does it help to move the manifests directory under the deploy directory?

aylei: > Why does this need to work outside this repository? I have been referencing manifests with symlinks in the GCP terraform. The deployment doesn't work without a good way to reference our charts and manifests.

Cases that use these modules outside this repository:
- terraform store `state` in the working dir, users may copy this module to manage different EKS instances (one of our colleges have tried this and got hurt by the symlink)
- for advanced users, it is possible to compose the `tidb-operator` module and `tidb-cluster` module into their own terraform scripts. A self-contained module will be easier to be composed in this case

> Does it help to move the manifests directory under the deploy directory?

Yes, to some extent. But the `manifest` directory is still necessary because the `ebs-gp2` storage class and `local-volume-provisioner.yaml` are dedicated to AWS. I've considered use a small overlay yaml to customize the base `local-volume-provisioner.yaml` via kusomize, but this has a low priority.

deploy/aws/eks/manifests/local-volume-provisioner.yaml
aylei: I've addressed this in the latest commit😜

deploy/aws/eks/templates/kubeconfig.tpl
aylei: I've removed the customized eks module, now the kubeconfig is generated by the canonical eks module... in the same way😂

deploy/aws/tidb-operator/manifests/local-volume-provisioner.yaml
```diff
@@ -38,6 +38,8 @@ spec:
         - key: dedicated
           operator: Exists
           effect: "NoSchedule"
+      nodeSelector:
+        localVolume: "true"
```
gregwebs: `local-ssd` would be a slightly better name. GKE adds this tag automatically now: `cloud.google.com/gke-local-ssd`

deploy/aws/aws-tutorial.tfvars
```diff
@@ -1,11 +1,10 @@
...
-tikv_count = 1
-tidb_count = 1
+default_cluster_pd_count   = 1
+default_cluster_tikv_count = 1
+default_cluster_tidb_count = 1
 
-cluster_name = "aws_tutorial"
-tikv_root_volume_size = "50"
\ No newline at end of file
+default_cluster_cluster_name = "aws_tutorial"
```
tennix: cluster name should not contain underscore.

deploy/aws/eks/manifests/crd.yaml
gregwebs: I used kustomize for GCP, it was quite easy: https://github.com/pingcap/tidb-operator/tree/master/deploy/gcp/manifests/local-ssd

gregwebs: With respect to state, it seems that users are solving a problem by copying and in response we are copying :)
I want to get away from copying as a solution to anything. The way I solve this is by having users instantiate a module, rather than just changing a terraform variable files. A user instantiates the module in a directory representing their environment (staging, production). The usability is about the same for one instantiation, but when you do multiple everything is much better.

gregwebs: I think in our case we are missing a nice mechanism to depend on our manifest files. The only way I can see how to do it is to use the `file` function to read the file in (and we can write it out to a new local file to avoid memory consumption). But then I think we still end up needing the entire git repo, so it doesn't really seem better than a symlink.

deploy/aws/README.md
```diff
@@ -159,12 +159,91 @@ The TiDB version and component count are also configurable in variables.tf, you
...
+You can customize the TiDB operator by specifying a helm values file through the `operator_values` variable. For example:
+
+```hcl
+variable "operator_values" {
+  description = "The helm values of TiDB Operator"
+  default     = file("operator_values.yaml")
+}
+```
+
+## Multiple Cluster Management
```
gregwebs:
```suggestion
## Multiple TiDB Cluster Management
```

aylei:
@tennix @gregwebs I've updated this PR according to your comments, PTAL, thanks!

gregwebs:
As per my comments above, I would like to demonstrate usage that is safe for multiple environments. The good news is we have already created a module, we just need to show how to use it as one!

The only change we need to make is that all the usages of `file` need to have `path.module`, so `file("${path.module}.`

Then the usage as module is just a new directory with a file

```
module "staging" {
  source = "../"
}
```


The problem with this is then the suggestion to create multiple TiDB clusters by editing clusters.tf. I suggest we have the user instantiate the tidb-cluster modules themselves.

We can move all our top-level terraform to a separate directory, perhaps called "vpc-setup". Then we have a file `/staging/main.tf` with contents:

```
module "staging" {
  source = "../vpc-setup"
}

module module "default-cluster" {
  source = "../tidb-cluster"
  ...
}
```

gregwebs:
Great work on this, the above comment shows how easy it is to start changing things around now that things have been properly modularized :)

aylei:
Yes, and (luckily) all the `file()` invocation and the reference of files in `local-exec` handles directory properly now😄

The example is great and I'm going to add it to the README👍

gregwebs:
Glad that makes sense. I think if we add it to the README but also explain how to modify the vpc-setup module directly it will lead to problems where someone will have already modified a module directly. I don't know how to fix the state file at that point to move it to the multiple environment setup.

aylei:
> I suggest we have the user instantiate the tidb-cluster modules themselves.

IMHO, this UX is more applicable for advanced users with hands-on terraform experience, because after all a little (maybe a lot) glue codes are necessary.   

> I think if we add it to the README but also explain how to modify the vpc-setup module directly it will lead to problems where someone will have already modified a module directly. I don't know how to fix the state file at that point to move it to the multiple environment setup.

We should document that users should avoid reusing the top-level module because it is not modularized. Then the state of the top-level module will be stored in `/deploy/aws/`, and the state of custom modules will be stored in, say, `/deploy/aws/staging`. And there should be no tf state in `/deploy/aws/tidb-operator` and `/deploy/aws/tidb-cluster` because they are not supposed to be applied   directly.

gregwebs:
> IMHO, this UX is more applicable for advanced users with hands-on terraform experience, because after all a little (maybe a lot) glue codes are necessary.

The glue code is almost entirely default variables. But those are all just used in clusters.tf so we can move the default values into the clusters module itself.

We can also provide a default environment to use. So the user only needs to `cd default`.

gregwebs:
> Then the state of the top-level module will be stored in /deploy/aws/, and the state of custom modules will be stored in, say, /deploy/aws/staging.

Yes, but the user has already made edits to the top-level module directly. This will end up affecting `deploy/aws/staging`, which will cause problems. The user now needs to move their top-level usage to the environment based approach, but I don't even know how to do this without destroying the existing top-level environment first.

aylei:
I mean, the top-level module is indeed an example of composing the sub modules. Whether the module located in `/deploy/aws` or `/deploy/aws/default` makes no difference in usage except that the top-level may seems to be special because it is located in the ... top-level.

aylei:
Let me demonstrate if I get you correct now:

Maybe a better UX is moving the VPC setup to a separate sub module, and moving the current top-level module to the `./default` directory. Then we will have the following directories layout:

```
./tidb-cluster
./tidb-operator
./vpc
./default
```

Then:

- if the users want to provision more TiDB clusters in the `default` module, they go to edit the tf scripts in this module;
- if the users want to provision a new EKS cluster and TiDB clusters in it, user create a new module like `staging`:

```
./tidb-cluster
./tidb-operator
./vpc
./default
./staging
```
Does this make sense?

aylei:
I'd like to move the discussion about UX and the recommended usage of modules to a new issue, @gregwebs @tennix how do you think?

tennix:
Agreed, I think we can merge this PR now and improve the UX in following PRs.

gregwebs:
@aylei yes, that is a good module organization.

PR #618, opened 2019-07-02
refactor so that using the PD API avoids unnecessary imports

gregwebs:
The PD API no longer imports Kubernetes and the CRD.

In the create-statefulset-immediately PR this reduced the new wait-for-pd binary
by ~20M

Tests 

- Existing tests pass

Code changes

 - Has Go code change

No


tennix:
image
I recompile all the binaries and didn't see the size reduce.

gregwebs:
@tennix the size reduction can only be seen on the create-statefulset-immediately PR for the wait-for-pd binary. It was requested that I make this a separate PR. We could use this same technique for the discovery service and see it there also.

tennix:
@weekface PTAL

weekface:
the `pkg/apis` is the directory for [K8s apis](https://github.com/kubernetes/kubernetes/tree/master/pkg/apis), so it is not a good idea to put pdapis here.

gregwebs:
What directory should it be placed it? @weekface

weekface:
Any other directories other than `pkg/apis` are fine, for example: `pkg/pdapi`.

gregwebs:
@weekface okay, directories are moved. If you come up with a better idea feel free to move or name them something else. `rg -l -i 'pkg/apis/pdapi' | xargs -i@ sed -i 's|pkg/apis/pdapi|pkg/pdapi|g' @`

weekface:
@onlymellb @cofyc @xiaojingchen PTAL

PR #638, opened 2019-07-05
change the way to config tidb/tikv/pd in charts

DanielZhangQD:
Do not provide config parameters for tidb/tikv/pd in values.yaml of charts any more to avoid the config mismatch between different releases of tidb/tikv/pd and also tidb-operator does not need to maintain the config templates for tidb/tikv/pd any more.
Instead, users should follow the config template in tidb/tikv/pd repo and provide customized configs by themselves.
Update the way to config tidb/tikv/pd in values.yaml of chart.
### Check List 

Tests 

    - Install TiDB cluster with new values.yaml and cluster runs normally, TiDB service is available to access.

Code changes

 - Has Helm charts change
 - Has documents change

Side effects

    - Users should migrate the configs in values.yaml of previous chart releases to the new values.yaml of the new chart, otherwise, the configurations will be ignored when upgrading TiDB cluster with the new chart.

Related changes

 ```release-note
Action required: Users should migrate the configs in values.yaml of previous chart releases to the new values.yaml of the new chart, otherwise, the configurations will be ignored when upgrading TiDB cluster with the new chart.
For example, configs in old values.yaml:
pd:
  logLevel: info
  maxStoreDownTime: 30m
  maxReplicas: 3
tikv:
  logLevel: info
  syncLog: true
  readpoolStorageConcurrency: 4
  readpoolCoprocessorConcurrency: 8
  storageSchedulerWorkerPoolSize: 4
tidb:
  logLevel: info
  preparedPlanCacheEnabled: false
  preparedPlanCacheCapacity: 100
  txnLocalLatchesEnabled: false
  txnLocalLatchesCapacity: "10240000"
  tokenLimit: "1000"
  memQuotaQuery: "34359738368"
  txnEntryCountLimit: "300000"
  txnTotalSizeLimit: "104857600"
  checkMb4ValueInUtf8: true
  treatOldVersionUtf8AsUtf8mb4: true
  lease: 45s
  maxProcs: 0

After migration, new values.yaml should be as below:
pd:
  config: |
    [log]
    level = "info"
    [schedule]
    max-store-down-time = "30m"
    [replication]
    max-replicas = 3
tikv:
  config: |
    log-level = "info"
    [raftstore]
    sync-log = true
    [readpool.storage]
    high-concurrency = 4
    normal-concurrency = 4
    low-concurrency = 4
    [readpool.coprocessor]
    high-concurrency = 8
    normal-concurrency = 8
    low-concurrency = 8
    [storage]
    scheduler-worker-pool-size = 4
tidb:
  config: |
    token-limit = 1000
    mem-quota-query = 34359738368
    check-mb4-value-in-utf8 = true
    treat-old-version-utf8-as-utf8mb4 = true
    lease = "45s"
    [log]
    level = "info"
    [prepared-plan-cache]
    enabled = false
    capacity = 100
    [txn-local-latches]
    enabled = false
    capacity = 10240000
    [performance]
    txn-entry-count-limit = 300000
    txn-total-size-limit = 104857600
    max-procs = 0

Please check [cluster configuration](https://pingcap.com/docs-cn/v3.0/reference/configuration/tidb-in-kubernetes/cluster-configuration/) for detailed configuration.
 ```


gregwebs:
I like this. What about _pd-config.tpl, etc?

DanielZhangQD:
> I like this. What about _pd-config.tpl, etc?

Actually, we can remove the config.tpl for tidb/tikv/pd in charts, but for now, I prefer to keep them as they are just in case, but no need to maintain or update them anymore.

DanielZhangQD:
@aylei Since GCP and Aliyun are under development, I just updated the config for AWS, please sync the change for GCP and Aliyun later.
@xiaojingchen PR https://github.com/pingcap/docs-cn/pull/1582 is not merged yet and also needs to update.

deploy/aws/tidb-cluster/values/default.yaml
```diff
@@ -2,19 +2,39 @@
...
+  separateSlowLog: true
+  slowLogTailer:
+    image: busybox:1.26.2
+    resources:
+      limits:
+        cpu: 100m
+        memory: 50Mi
+      requests:
+        cpu: 20m
+        memory: 5Mi
```
aylei: The modifications of this file are same as the default one in the chart's values.yaml, so it is unnecessary

gregwebs:
@DanielZhangQD it is really confusing to have config.tpl if they don't do anything. I don't know what the current behavior is when both are present from looking at this PR.

deploy/aws/tidb-cluster/values/default.yaml
DanielZhangQD: For the charts released till now, the configs for tidb/tikv/pd are different and separateSlowLog is commented out.

DanielZhangQD:
> @DanielZhangQD it is really confusing to have config.tpl if they don't do anything. I don't know what the current behavior is when both are present from looking at this PR.
@gregwebs 
It's in the _helper.tpl:
https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/templates/_helpers.tpl#L33
It will first check if Values.pd.config is configured, if yes, take it as configmap data, else, take the file config/_pd-config.tpl.
The tpl could be an alternative just in case, and users do not need to be aware of them.

charts/tidb-cluster/values.yaml
```diff
@@ -48,9 +48,20 @@ discovery:
...
+  # pd configurations (change to the tags of your pd version), 
+  # just follow the format in the file and configure in the 'config' section
+  # as below if you want to customize any configuration.
+  # Please refer to https://pingcap.com/docs-cn/v3.0/reference/configuration/pd-server/configuration-file/
+  # (choose the version matching your pd) for detailed explanation of each parameter.
+  config: |
+    [log]
+    level = "info"
+    [replication]
+    location-labels = ["region", "zone", "rack", "host"]
```
xiaojingchen: It should be reminded that `location-labels` should be consistent with `tikv.storeLabels`.

DanielZhangQD: Reminder was added in the tikv.storeLabels section.

deploy/aws/tidb-cluster/values/default.yaml
aylei: Yes, the old version `v1.0.0-beta.3` don't need this because the chart has a proper default, and if we release your change, the new version neither need this because the new version will have a proper default too.

Moreover, keep these configurations here will increase our maintenance burden.

DanielZhangQD: OK, just keep the "separateSlowLog: true" because it's false in old charts and it's better to set it to true for the slow log.

PR #670, opened 2019-07-18
document and improve HA algorithm

### LINKED ISSUE #652: Can't schedule 4 TiKV instances on 3 K8s nodes
We have a new algorithm [here](https://github.com/pingcap/tidb-operator/issues/235), there is a corner that we can't schedule 4 tikv instances on a 3 k8s nodes.

https://github.com/pingcap/tidb-operator/blob/93d5c068df096cf3e2e49c4f73eb23e0396b3d36/pkg/scheduler/predicates/ha.go#L135-L137

https://github.com/pingcap/tidb-operator/blob/93d5c068df096cf3e2e49c4f73eb23e0396b3d36/pkg/scheduler/predicates/ha_test.go#L754-L773

cofyc:

fixes #652 

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

 ```release-note
NONE
 ```


xiaojingchen:
LGTM

pkg/scheduler/predicates/ha.go
```diff
@@ -132,7 +132,14 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
 		}
 
 		podsCount := len(podNames)
-		if podsCount+1 >= int(replicas+1)/2 {
+		// When replicas equals 3, pods on each node should not be greater than 1.
+		maxPodsPerNode := 1
+		if replicas > 3 {
```
gregwebs: `3` is hard-coded, but it should be gathered from configuration (in case the user sets it to 5)

cofyc: I think `3` is the number of copies of data regions in TiKV. Is it possible to has more than 3 copies when the replicas of TiKV instances increases, e.g. replicas data regions into more than 3 nodes? cc @tennix @weekface

weekface: `3` is the tikv replicas.

weekface: This pr just fix the corner case when replicas is 4.

cofyc:
it has some conflicts, I will resolve it later

cofyc:
@xiaojingchen @weekface conflicts are resolved, PTAL again. Thanks!
@gregwebs I'm going to fix the bug first.

pkg/scheduler/predicates/ha.go
```diff
@@ -134,9 +134,16 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
...
 
 		podsCount := len(podNames)
-		if podsCount+1 >= int(replicas+1)/2 {
-			glog.Infof("node %s podsCount+1 is %d, int(replicas+1)/2 is %d, skipping",
-				nodeName, podsCount+1, int(replicas+1)/2)
+		// When replicas equals 3, pods on each node should not be greater than 1.
+		maxPodsPerNode := 1
+		if replicas > 3 {
+			// When replicas is greater than 3, we allow more than one pods on one node.
+			maxPodsPerNode = int((replicas + 1) / 2)
```
xiaojingchen: if `replicas =5` , the `maxPodsPerNode` would be `3`.  Is it right?

cofyc: it seems not right... `2, 3` will be possible when there are two nodes.

`int((replicas + 1) / 2)` should be `ceil(replicas / 3)`

replicas | maxPodsPerNode| in three nodes|
--------|--------|--------|
< 3 | not limited, because HA is not possible |
= 3         | 1 | 1, 1, 1| 
4         | 2 | 1, 1, 2|
5         | 2 | 1, 2, 2 |
6         | 2 | 2, 2, 2 |
7        | 3 |  2, 2, 3 |
8        | 3 |  2, 3, 3 |
8        | 3 |  3, 3, 3 |

what do you think? @weekface

weekface: In the table above, the `maxPodsPerNode` is `2` when `replicas` is `4`, this is not HA.

cofyc: I think the `maxPodsPerNode` is right when `replicas` is `4`, because in three nodes, `1, 1, 2` is ok, but when there are two instances in two nodes (`1, 1`), `maxPodsPerNode` must be `1` at this moment to force the scheduler to find a node which has no TiKV pod. 

I guess we need to update the HA algorithm to take more information into account.

cofyc: `maxPodsPerNode` is always `1` when scheduled TiKV pods less than `3`?

weekface: The current algorithm is ok for `pd`, but not `tikv`. We need a new HA algorithm for `tikv`.

cofyc:
@weekface @xiaojingchen PTAL again.

pkg/scheduler/predicates/ha.go
```diff
@@ -79,6 +87,11 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
 	component := pod.Labels[label.ComponentLabelKey]
 	tcName := getTCNameFromPod(pod, component)
 
+	if component != label.PDLabelVal && component != label.TiKVLabelVal {
+		glog.V(4).Infof("component %s is ignored in HA predicate", component)
+		return nodes, nil
+	}
```
cofyc: this is a safe guard and improves readability

pkg/scheduler/predicates/ha_test.go
```diff
@@ -752,16 +752,33 @@ func TestHAFilter(t *testing.T) {
 			},
 		},
 		{
-			name:          "three nodes, three pods scheduled on these three nodes, replicas is 4, can't scheduled",
-			podFn:         newHAPDPod,
+			name:          "three nodes, three pods scheduled on these three nodes, replicas is 4, return all the three nodes",
+			podFn:         newHATiKVPod,
```
cofyc: this case is for TiKV, we need a TiKV pod

pkg/scheduler/predicates/ha.go
```diff
@@ -126,23 +143,65 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
...
+		podsCount := len(podNames)
+		maxPodsPerNode := 0
+
+		if component == label.PDLabelVal {
+			/**
+			 * replicas		maxPodsPerNode
+			 * ---------------------------
+			 * 1			1
+			 * 2			1
+			 * 3		    1
```
xiaojingchen: please align

cofyc: oh, I mixed tab and spaces, fixed now.

cofyc:
@weekface should this be merged before 1.0 GA?

weekface:
Yes, merging it to master. 

But I will not cherry-pick it to the `release-1.0`. The `release-1.0` is frozen.

PR #676, opened 2019-07-22
add more logs

weekface:
### Check List 

Tests 

 - Unit test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

 ```release-note
NONE
 ```


pkg/manager/member/tidb_failover.go
```diff
@@ -40,6 +41,7 @@ func (tf *tidbFailover) Failover(tc *v1alpha1.TidbCluster) error {
 		_, exist := tc.Status.TiDB.FailureMembers[tidbMember.Name]
 		if exist && tidbMember.Health {
 			delete(tc.Status.TiDB.FailureMembers, tidbMember.Name)
+			glog.Errorf("tidb failover: delete %s from tidb failoverMembers", tidbMember.Name)
```
tennix: Is this an error log?

weekface:
all these logs are "state changes", they can help us diagnose problems

pkg/manager/member/utils.go
```diff
@@ -182,6 +182,7 @@ func serviceEqual(new, old *corev1.Service) (bool, error) {
 // setUpgradePartition set statefulSet's rolling update partition
 func setUpgradePartition(set *apps.StatefulSet, upgradeOrdinal int32) {
 	set.Spec.UpdateStrategy.RollingUpdate = &apps.RollingUpdateStatefulSetStrategy{Partition: &upgradeOrdinal}
+	glog.Infof("set %s/%s partition to %d successfully", set.GetNamespace(), set.GetName(), upgradeOrdinal)
```
cofyc: I think it's not a good idea to log in the utility function. log when the API operation has done?

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -90,8 +91,10 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
 
 		err = opc.podControl.DeletePod(tc, pod)
 		if err != nil {
+			glog.Errorf("orphan pods cleaner: failed to clean orphan pod: %s/%s, %v", ns, podName, err)
```
gregwebs: I generally prefer to annotate errors: that way there is a single error data structure with all the information that can be logged in one spot, and if there is an API request sent back in the response.

weekface: Yes, our error logs style need a lot of improvements. We may discuss it in a new issue?

This pr tries to add detailed logs when something changes to help us diagnose problems.

gregwebs: It is up to you. I think either way you should get a detailed log to help diagnose problems. If there is concurrent logging annotation will be simpler to understand.

pkg/scheduler/predicates/ha.go
```diff
@@ -128,18 +129,21 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
 		// replicas less than 3 cannot achieve high availability
 		if replicas < 3 {
 			minNodeNames = append(minNodeNames, nodeName)
+			glog.Infof("replicas is %d, add node %s to minNodeNames", replicas, nodeName)
```
cofyc: Logs print here are unnecessary, at line 109 we know the replicas of the component, if the replicas is smaller than 3, all nodes will be added into `minNodeNames`. 

How about making it simpler, we can skip the for loop if the replicas is smaller than 3 and print an informative log?

pkg/scheduler/predicates/ha.go
```diff
@@ -128,18 +129,21 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
...
 			continue
 		}
 
 		podsCount := len(podNames)
 		if podsCount+1 >= int(replicas+1)/2 {
+			glog.Infof("node %s podsCount is %d, skipping", nodeName, podsCount)
```
cofyc: print the `int(replicas+1)/2` too?

pkg/manager/member/orphan_pods_cleaner.go
weekface: what is "logging annotation"?

pkg/scheduler/predicates/ha.go
weekface: How to skip the `for` loop? we must range the `nodeMap` anyway

pkg/scheduler/predicates/ha.go
weekface: `replicas` has been printed in 109.

pkg/manager/member/utils.go
weekface: removed the `successfully ` word. ptal

pkg/scheduler/predicates/ha.go
weekface: addressed.

pkg/scheduler/predicates/ha.go
cofyc: hmm, you're right, we need to get node name

PR #687, opened 2019-07-24
limit test cover packages

xiaojingchen:
current unit test cover generated codes and client codes, that influencing real test coverage.
this PR limit the unit test cover packages.

* Run `make test`
* Run `go tool cover -V -func=coverage.txt -o coverage.out && cat coverage.out`

after limit the cover packages the test coverage increased from 61.4% to 74.3%

### Check List 

Code changes

Side effects

```
NONE
```

 ```release-note
NONE
 ```
@onlymellb  @tennix @weekface  PTAL


.gitignore
```diff
@@ -22,6 +22,7 @@ filenametags
 *.coverprofile
 
 coverage.txt
+coverage.out
```
weekface: where does this file come from?

xiaojingchen: run `go tool cover -V -func=coverage.txt -o coverage.out ` will create a file which named coverage.out

PR #692, opened 2019-07-24
GetHealth should return network errors

gregwebs:

GetHealth should return network errors.
Otherwise the health may be set to false if the pod cannot be reached.

GetHealth now returns errors that are not HTTP codes. The sync code will log that there was an error.
GetHealth was changed to operate on a single TiDB now instead of returning a map.
Some other related refactorings were made.

### Check List 

Tests 

- Using existing tests

Code changes

 - Has Go code change

Side effects

- None

Related changes

 ```release-note
NONE
 ```


pkg/manager/member/tidb_member_manager.go
```diff
@@ -380,30 +380,37 @@ func (tmm *tidbMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, se
...
 
 	tidbStatus := map[string]v1alpha1.TiDBMember{}
-	tidbHealth := tmm.tidbControl.GetHealth(tc)
-	for name, health := range tidbHealth {
-		newTidbMember := v1alpha1.TiDBMember{
-			Name:   name,
-			Health: health,
+	for _, hostName := range tmm.tidbControl.HostNames(tc) {
+		var nodeName string
+		health, err := tmm.tidbControl.GetHealth(tc, hostName)
```
xiaojingchen: when `GetHealth` return error, we can not return the error directly, otherwise, the TiDB status in TC would alway not be updated.

gregwebs: It makes sense to me to log the error and use the existing health instead.

gregwebs: see the next commit https://github.com/pingcap/tidb-operator/pull/692/commits/027aa0990f729a7c272ff5db8d505c1d9789a74c

xiaojingchen: when the connection failed, the TiDB is mostly unhealthy, we must update the `tc.Status.TiDB.Members`. 
so we can log the error but not return.

gregwebs: This is difficult: the connection could fail due to a network partition. I can check to see if the pod is running.

gregwebs: What do you think now? https://github.com/pingcap/tidb-operator/pull/692/commits/22e982c30a27e05d3562d6a6239b752e8eb565a2

xiaojingchen: It is hard to confirm the TiDB is normal, even pod status is running.
I think we avoid to handle the detailed error, we should always think of it as a failed simply state when the access its `GetHealth` API error. 
The only effect is to create an additional TiDB，and this is acceptable.
any opinion @weekface

gregwebs: Now I check to see if the Readiness probe has set ready to true. The readiness probe is the same as GetHealth. What do you think about removing GetHealth and just using the readiness probe?

gregwebs: oh, right the readiness probe only transitions state once, we would need a liveness probe.

gregwebs: okay, so I agree there isn't a good approach to this issue. Now this change just logs this case which may be useful for diagnosing problems.

PR #693, opened 2019-07-25
binlog: support `binlog.pump.config` configurations for pump and drainer 

weekface:

- support `binlog.pump.config` and `binlog.drainer.config` configurations for pump and drainer 
- support pump and drainer configuration rollout 

### Check List 

Tests 

 - E2E test
 - Stability test

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
support `binlog.pump.config` and `binlog.drainer.config` configurations for pump and drainer 
 ```


charts/tidb-cluster/templates/pump-statefulset.yaml
```diff
@@ -50,9 +50,13 @@ spec:
       volumes:
       - name: config
         configMap:
+        {{- if .Values.enableConfigMapRollout }}
+          name: {{ template "cluster.name" . }}-pump-{{ template "pump-configmap.data-digest" . }}
```
weekface: pump configration rollout @aylei PTAL

charts/tidb-cluster/templates/drainer-statefulset.yaml
```diff
@@ -50,9 +50,13 @@ spec:
       volumes:
       - name: config
         configMap:
+        {{- if .Values.enableConfigMapRollout }}
+          name: {{ template "cluster.name" . }}-drainer-{{ template "drainer-configmap.data-digest" . }}
```
weekface: drainer configration rollout @aylei PTAL

charts/tidb-cluster/values.yaml
```diff
@@ -398,13 +398,17 @@ binlog:
...
+    # Please refer to https://github.com/pingcap/tidb-binlog/blob/master/cmd/pump/pump.toml for the default
+    # pump configurations (change to the tags of your pump version),
+    # just follow the format in the file and configure in the 'config' section
+    # as below if you want to customize any configuration.
+    config: |
+      gc = 7
+      heartbeat-interval = 2
+      [storage]
+      sync-log = true
+      stop-write-at-available-space = "100 gib"
```
aylei:
```
# stop write when disk available space less then the configured size
# 42 MB -> 42000000, 42 mib -> 44040192
# default: 10 gib
# stop-write-at-available-space = "10 gib"
```
I think it should be 1 Gigabytes or less

charts/tidb-cluster/values.yaml
```diff
@@ -417,39 +421,27 @@ binlog:
...
-    kafka: {}
-      # only need config one of zookeeper-addrs and kafka-addrs, will get kafka address if zookeeper-addrs is configed.
-      # zookeeperAddrs: "127.0.0.1:2181"
-      # kafkaAddrs: "127.0.0.1:9092"
-      # kafkaVersion: "0.8.2.0"
+    # Please refer to https://github.com/pingcap/tidb-binlog/blob/master/cmd/drainer/drainer.toml for the default
+    # drainer configurations (change to the tags of your drainer version),
+    # just follow the format in the file and configure in the 'config' section
+    # as below if you want to customize any configuration.
+    config: |
```
aylei: This causes the old configuration key invalid, which is a in-compatible change

aylei:
I think rollout the binlog configuration is not a small change, can we put this to 1.1

charts/tidb-cluster/values.yaml
weekface: yes, this is an in-compatible change like: https://github.com/pingcap/tidb-operator/pull/638, i think we should do it before GA. @tennix

charts/tidb-cluster/values.yaml
weekface: Yes, this is available space size, using default.

charts/tidb-cluster/templates/scripts/_start_drainer.sh.tpl
```diff
@@ -26,6 +26,7 @@ done
 
 /drainer \
 -L={{ .Values.binlog.drainer.logLevel | default "info" }} \
+-pd-urls={{ template "cluster.name" . }}-pd:2379 \
```
xiaojingchen: add `http://` scheme

charts/tidb-cluster/templates/scripts/_start_pump.sh.tpl
```diff
@@ -1,5 +1,6 @@
 set -euo pipefail
 /pump \
+-pd-urls={{ template "cluster.name" . }}-pd:2379 \
```
xiaojingchen: ditto

charts/tidb-cluster/values.yaml
DanielZhangQD: What about we keep these two ways, but make it clear in the document that we do not maintain the config for the original way and users should migrate to the new way if they want to configure any new parameters that are not exposed in values.yaml.

weekface:
@aylei @xiaojingchen @tennix PTAL again

charts/tidb-cluster/values.yaml
DanielZhangQD: Have we decided to make this change incompatible?

weekface: What do you think? @tennix

weekface: Agree with @DanielZhangQD we should not make this incompatible change.

tennix: I think we can keep both for binlog configurations to keep compatible.

weekface:
@aylei @DanielZhangQD @tennix 

This PR changed to support both configurations. PTAL again.

charts/tidb-cluster/values.yaml
weekface: comments addressed.

charts/tidb-cluster/values.yaml
```diff
@@ -423,6 +423,17 @@ binlog:
...
+    # Please refer to https://github.com/pingcap/tidb-binlog/blob/master/cmd/pump/pump.toml for the default
+    # pump configurations (change to the tags of your pump version),
+    # just follow the format in the file and configure in the 'config' section
+    # as below if you want to customize any configuration.
+    # config: |
+    #   gc = 7
+    #   heartbeat-interval = 2
+    #   [storage]
+    #   sync-log = true
+    #   stop-write-at-available-space = "10 gib"
```
tennix:
```suggestion
    #   stop-write-at-available-space = "10Gi"
```

PR #699, opened 2019-07-27
support kubeconfig env

yeya24:

Add support for KUBECONFIG env when starting components.
This env enables running and debugging this operator in local development environment easier.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Helm charts change
 - Has Go code change yes
 - Has documents change

Side effects

Related changes

 ```release-note

 ```

### DIFF
--- a/cmd/discovery/main.go
+++ b/cmd/discovery/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"github.com/pingcap/tidb-operator/pkg/util"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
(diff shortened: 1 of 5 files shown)


yeya24:
Looks like the ci met the problem of `failed to watch file "/var/log/pods/15de0014-b112-11e9-b349-02427567777a/tidb-operator-e2e/0.log": no space left on device++`

weekface:
Yes, I rebuild the CI env, and will try again later.

cmd/discovery/main.go
```diff
@@ -38,10 +39,23 @@ func main() {
...
+		cfg, err = clientcmd.BuildConfigFromFlags("", kubeConfigFile)
+		if err != nil {
+			glog.Fatalf("failed to create config from file: %s %v\n", kubeConfigFile, err)
+		}
+	} else {
+		cfg, err = rest.InClusterConfig()
+		if err != nil {
+			glog.Fatalf("failed to get config: %v", err)
+		}
 	}
```
onlymellb: Can you extract this logic into a method `NewKubeCli(kubeCfg string) kubernetes.Interface` and then reference this method to initialize the k8s client? In addition, the client of the following CR `versioned.NewForConfig` should be initialized in the same way.

yeya24: I have get the logic into a new method which return *rest.config

yeya24:
Sorry for the delay. Could take a look? @weekface @onlymellb Thanks!

weekface:
@aylei This PR just add the kubeconf env support to operator. I think we can merge it and extract these logics in another PR.

yeya24:
@weekface @aylei  Should I extract the logics from this PR and these initialization in tests? If I need to, I think I can work on it this weekend.

tennix:
I don't think we can run the controller outside of the cluster, the controller needs to access PD by service name, the service name is not resolvable if not in a pod.

yeya24:
> I don't think we can run the controller outside of the cluster, the controller needs to access PD by service name, the service name is not resolvable if not in a pod. But other components should work IIRC.
Make sense. Maybe it is better to close this PR.

Issue #408, opened 2019-04-18
autoFailover leaves pod in error state


gregwebs:
Steps to reproduce:
1) install operator with autoFailover=true and a cluster
2) gcloud compute instances delete <tikv-node>

Actual behavior:
The same number of pods exist (3). The pod that loses its node is re-scheduled to a new node. However, it continually fails with a duplicated store error: 

```
2019/04/17 23:11:08.233 ERRO util.rs:336: fail to request: Grpc(RpcFailure(RpcStatus { status: Unknown, details: Some("duplicated store address: id:64 address:\"demo-tikv-1.demo-tikv-peer.tidb120002.svc:20160\" version:\"2.1.3\" , already registered by id:4 address:\"demo-tikv-1.demo-tikv-peer.tidb120002.svc:20160\" ...
```

So the capacity of TiKV is now 2.

Expected behavior:
* There should be 3 TiKV pods that run without error

tennix:
The expected behavior is there will be a pending pod and a new pod. That's 4 tikv pods.

tennix:
The deleted node came back again with the same PV, but data in the PV was lost. So the tikv pod is scheduled to the node again with the PV. This causes tikv to print that error message.

cofyc:
> The deleted node came back again with the same PV, but data in the PV was lost. So the tikv pod is scheduled to the node again with the PV. This causes tikv to print that error message.

Yes, this is what happened. There is one issue we need to fix and one situation we can improve:

1) First, we need to use the unique path of volume in PV object to prevent using new volume (e.g. newly created when the node gets recreated) with old PV object which is referenced by existing Pod.

Currently, we use [`--local-ssd-count` to create node pools](https://github.com/pingcap/dbaas/blob/master/operations/scaler/gke/autoscale.go#L239) which will mount local SSDs at non-unique paths (e.g. `/mnt/disks/ssd0`, `/mnt/disks/ssd1`). PV objects will have `/mnt/disks/ssdX` in its `spec.path` field.

But when nodes get recreated, local SSDs will be recreated and mounted at the same paths (note: autoDelete cannot be set to false for local SSD, `--keep-disks` cannot be used to keep local SSDs). 

But the underlying volume of local PV object is deleted, pods which use old volume data should not start with another volume data (empty here).

[...]

PR #703, opened 2019-07-30
Enable useNodeNameOnly in local-volume-provisioner deployment

### LINKED ISSUE #421: Enable useNodeNameOnly in local-volume-provisioner deployment
If by accident kubernetes node object has been recreated, new node object with the same name for the same node will get a new UID. By default, provisioner includes node.UID in its name, it will filter out local PVs that were not created with this name and [do not reclaim these PVs](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/blob/master/docs/faqs.md#pv-with-delete-reclaimpolicy-is-released-but-not-going-to-be-reclaimed).

cofyc:

fix #421

### Check List 

Tests 

 - Unit test

Side effects

Related changes

 ```release-note
ACTION REQUIRED: In cloud platforms, a node may be recreated with the same name but different UID, we enable `useNodeNameOnly` in local-volume-provisioner to avoid issues. However, you must fix existing PVs by removing UUID suffixes in annotation `pv.kubernetes.io/provisioned-by` when you upgrade your current local-volume-provisioner.
 ```

### DIFF
--- a/deploy/modules/aliyun/tidb-operator/templates/local-volume-provisioner.yaml.tpl
+++ b/deploy/modules/aliyun/tidb-operator/templates/local-volume-provisioner.yaml.tpl
@@ -4,6 +4,9 @@ metadata:
   name: local-provisioner-config
   namespace: kube-system
 data:
+  useNodeNameOnly: "true"
+  nodeLabelsForPV: |
+    - kubernetes.io/hostname
   storageClassMap: |
     local-volume:
        vendor: alibabacloud
--- a/deploy/modules/aws/tidb-operator/manifests/local-volume-provisioner.yaml
+++ b/deploy/modules/aws/tidb-operator/manifests/local-volume-provisioner.yaml
@@ -12,6 +12,9 @@ metadata:
   name: local-provisioner-config
   namespace: kube-system
 data:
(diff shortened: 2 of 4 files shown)


deploy/modules/aliyun/tidb-operator/templates/local-volume-provisioner.yaml.tpl
```diff
@@ -4,6 +4,9 @@ metadata:
   name: local-provisioner-config
   namespace: kube-system
 data:
+  useNodeNameOnly: "true"
+  nodeLabelsForPV: |
```
cofyc: `nodeLabelsForPV` is used to sync node labels to PV.

gregwebs:
as per #408 I think this may make things worse. It seems that the node name is not reliable for local SSD and we do need the node UUID.

cofyc:
Since GKE 1.11 the node object is no longer recreated on upgrade/repair even though the underlying instance is still recreated (and local SSDs are wiped), see discussion [here](https://github.com/kubernetes-incubator/external-storage/issues/945#issuecomment-414393266). This is not necessary for now.

PR #711, opened 2019-07-30
Terraform gke scale out fix

### LINKED ISSUE #708: GCP Terraform scripts destroy and create a new one when increasing the cluster size
## Bug Report

When increasing TiKV count in variables and run `terraform apply` again, the cluster is destroyed and a new one is created.

jlerche:
Closes #708 

This incorporates changes from @aylei's' commit https://github.com/aylei/tidb-operator/commit/47be0a7b388b5023f926571446b9a3390a98ff7c in his fork. This prevents node pools from being recreated.

In #673 the `initial_node_count` variable in the cluster resource was changed to depend on the tidb/monitor/tikv/pd count variables. Because of this, when a count variable is changed to scale out, the cluster resource is recreated by terraform. To get around this, a literal value of `5` is used. According to https://kubernetes.io/docs/setup/best-practices/cluster-large/#size-of-master-and-master-components this will cause a master node to be created that can accommodate up to 100 nodes.

Additional ordering in the resource creation to ensure that the tidb-cluster helm release is installed/upgraded only after the node pools are created/updated to prevent an issue where pods might be scheduled on old nodes.

### Check List 

Tests 

* terraform apply
* wait for finish
* check master node is not upgrading
* increase `tikv_count` `tikv_replica_count` (availabillity zones means `tikv_count=1` will have 3 nodes, 2 -> 6, etc)
* terraform apply
* check plan to make sure cluster is not recreated
* type `yes`
* check that cluster is not recreated, tidb node pool is created, then tidb-cluster release is upgraded
* check tidb pods are running properly
* terraform destroy

Code changes

 - Terraform changes

Side effects

Related changes

 ```release-note
NONE
 ```

### DIFF
--- a/deploy/gcp/main.tf
+++ b/deploy/gcp/main.tf
@@ -99,7 +99,7 @@ resource "google_container_cluster" "cluster" {
     use_ip_aliases = true
   }
 
-  // see https://github.com/terraform-providers/terraform-provider-google/issues/3385 for why initial_node_count is sum of node counts
+  // See https://kubernetes.io/docs/setup/best-practices/cluster-large/#size-of-master-and-master-components for why initial_node_count is 5. The master node should accommodate up to 100 nodes like this
   remove_default_node_pool = true
   initial_node_count       = 5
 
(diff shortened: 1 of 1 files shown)


deploy/gcp/main.tf
```diff
@@ -99,9 +99,9 @@ resource "google_container_cluster" "cluster" {
     use_ip_aliases = true
   }
 
+  // See https://kubernetes.io/docs/setup/best-practices/cluster-large/#size-of-master-and-master-components for why initial_node_count is 5. The master node should accommodate up to 100 nodes like this
   remove_default_node_pool = true
-  // see https://github.com/terraform-providers/terraform-provider-google/issues/3385 for why initial_node_count is sum of node counts
-  initial_node_count       = var.pd_count + var.tikv_count + var.tidb_count + var.monitor_count
+  initial_node_count       = 5
```
tennix: Will the master node get resized when setting different node count for the initial cluster? Previously, we observed that when other node pools created, the cluster is in a blocking state, the master node is upgraded during creation time.

PR #716, opened 2019-07-31
scheduler: inject a failure by pod annotation

gregwebs:
This can be useful for testing scheduling failures.
You can for example add it to monitor-deployment
along with changing the scheduler to the custom scheduler.

### Check List 

Tests

I tested as mentioned in the PR. I observed that no other pods had failed scheduling, including after deploying a second TiDB cluster.

Code changes

 - Has Go code change

Side effects

Related changes

 ```release-note
NONE
 ```


pkg/scheduler/scheduler.go
```diff
@@ -118,6 +123,15 @@ func (s *scheduler) Filter(args *schedulerapiv1.ExtenderArgs) (*schedulerapiv1.E
 	}, nil
 }
 
+// FailureError is returne when the FailTiDBSchedulerLabelKey is seen
```
tennix:
```suggestion
// FailureError is returned when the FailTiDBSchedulerLabelKey is seen
```

pkg/label/label.go
```diff
@@ -65,6 +65,9 @@ const (
 	TiDBLabelVal string = "tidb"
 	// TiKVLabelVal is TiKV label value
 	TiKVLabelVal string = "tikv"
+	// FailTiDBSchedulerLabelKey is for injecting a failure into the TiDB custom scheduler
+	// A pod with this label will produce an error when scheduled.
+	FailTiDBSchedulerLabelKey string = "tidb.pingcap.com/fail-scheduler"
```
aylei: How about using annotation?

tennix: Yeah, I think annotation is better.

gregwebs: :+1: Done

PR #719, opened 2019-08-01
tests: change cluster clean logic

xiaojingchen:

The problem is that the e2e test was blocked when e2e was deleting PVCs.  
Because then the Jobs have not deleted cause the PVC that was bound to Job is blocking in `Terminating` status. 

change e2e test 's delete cluster logic, make delete jobs before the delete pvc to avoid block.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

 ```release-note
`NONE`
 ```


xiaojingchen:
@cofyc @shuijing198799 PTAL

PR #726, opened 2019-08-01
Fix tikv scale in failure in some cases

### LINKED ISSUE #725: tikv failover may cause scale in failure in some cases
## Bug Report

v1.12

the latest release version

**What did you do?**

1. Prepare a k8s cluster that can only deploy three tikv instances
2. Simulate a tikv instance failure and let the operator perform a failover operation
3. At this point, the fourth tikv instance added by failover is the pending status
4. Restore the previous tikv instance and then set the tidbcluster's  `.Status.TiKV.FailureStores` to empty

When set the `.Status.TiKV.FailureStores` to empty, tidb-operator should scale in the tikv instances to the original value set by a user

**What did you see instead?**
 The number of tikv instances is still four, not scale in three

onlymellb:
This PR fixes #725 
### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

Related changes

 ```release-note
Fix tikv scale in failure in some cases after tikv failover
 ```

### DIFF
--- a/pkg/manager/member/tikv_scaler.go
+++ b/pkg/manager/member/tikv_scaler.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pingcap/tidb-operator/pkg/pdapi"
 	apps "k8s.io/api/apps/v1beta1"
 	corelisters "k8s.io/client-go/listers/core/v1"
+	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 )
 
 type tikvScaler struct {
@@ -78,6 +79,7 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
 		resetReplicas(newSet, oldSet)
 		return err
 	}
+
 	for _, store := range tc.Status.TiKV.Stores {
 		if store.PodName == podName {
 			state := store.State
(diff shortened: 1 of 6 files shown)


pkg/manager/member/tikv_scaler.go
```diff
@@ -135,9 +137,48 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
-	// store not found in TidbCluster status,
-	// this can happen when TiKV joins cluster but we haven't synced its status
-	// so return error to wait another round for safety
+	// When store not found in TidbCluster status, there are two situations as follows:
+	// 1. This can happen when TiKV joins cluster but we haven't synced its status.
+	//    In this situation return error to wait another round for safety.
+	//
+	// 2. This can happen when TiKV pod has not been successfully registered in the cluster, such as always pending.
+	//    In this situation we should delete this TiKV pod immediately to avoid blocking the subsequent operations.
+	if !podutil.IsPodReady(pod) {
```
weekface: I think we should only handle `Pending` pods other than all not ready pods.

gregwebs:
This logic is not so much Scale-In as it is updating the status of TiKV pods (and then when in Scale-In mode deciding to do something with that status). One reason the logic is placed here is to avoid returning an error and halting the sync process. However, in my PR #581 by raising a `RequeueError` one [does not stop the sync loop](https://github.com/pingcap/tidb-operator/pull/581/files#diff-dc05f71f0bdaa4c65bb300310b46c7e2R88). It is possible to do something similar here.

Note that my PR incidentally fixed a few minor bugs showing up in test cases due to dealing with halting sync at the right time.

pkg/manager/member/tikv_scaler.go
onlymellb: Not only pending, for example, considering this situation, the newly launched pod has been crashing and never really joined the tidb cluster. In this case, we can also scale in safely.

onlymellb:
Here we need to scale in immediately when it can be reduced safely, instead of generating a `RequeueError` waiting for the next round of sync (consider this situation, if the tikv pod has been pending or unhealthy，it has never really joined the tidb cluster during this period.), this situation can lead to an inability to scale in, no matter how many sync loops have passed. refer to issue #725

PR #750, opened 2019-08-09
Add basic TLS support for TiDB cluster

AstroProfundis:
Add basic TLS support for TiDB cluster.

When creating new TiDB cluster, set `enableTLSServer` to `true` will enable the TLS encryption feature in PD/TiDB/TiKV.

Setting `enableTLSClient` to `true` will enable support of TLS encrypted connection from MySQL client.

Server certs are placed in Secrets with the same namespace as the cluster, and named `<release-name>-{pd, tidb, tikv}`, client certs are placed in Secrets with the same namespace as the cluster, and named `client-tls`, another client cert is placed in TiDB Operator's namespace and named `client-tls`.

All certs are signed by the CA of the Kubernetes cluster.

All client requests (from operator) are modified to support TLS encrypted connections.

### Check List 

Tests 

Manually validated on internal testing k8s cluster.

Known Issues

 - Does not support of switching `enableTLSServer` on/off after creation of cluster
 - gRPC uses unencrypted connections between multiple PDs, this might be an issue of cert generation and is still under investigation

Code changes

 - Has Helm charts change
 - Has Go code change

Related changes

 - Need to update the documentation after all works are done

 ```release-note
Add basic support of TLS encrypted connections in TiDB cluster
 ```


charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -56,12 +56,19 @@ while true; do
     fi
 done
 
+if [[ {{ .Values.enableTLSServer }} == "true" ]]
+then
+    SCHEMA="https"
```
gregwebs: This should be "scheme" not "schema"

charts/tidb-operator/templates/controller-manager-deployment.yaml
```diff
@@ -50,3 +50,12 @@ spec:
...
             value: {{ .Values.timezone | default "UTC" }}
+        volumeMounts:
+        - mountPath: /var/lib/tls
+          name: tls
+          readOnly: true
+      volumes:
+      - name: tls
+        secret:
+          defaultMode: 420
+          secretName: client-tls
```
cofyc: what if more than one TiDB clusters are managed by the operator with different client certs?

AstroProfundis:
@gregwebs As of this PR, certs are generated manually, I'm working on an auto-generating feature of necessary certs

@cofyc It doesn't matter, PD accepts any valid certs that signed by the k8s CA, so controller can use one client cert for all TiDB clusters

gregwebs:
@AstroProfundis cert-manager is widely used for certificate generation and management and is working well for me now: https://docs.cert-manager.io/en/latest/tasks/issuers/index.html

AstroProfundis:
@tennix @cofyc @weekface PTAL

The basic process is working with this PR with some limitations:
 - Secrets that store the certs need to be manually added (and certs are manually issued with k8s' certificate management facility as well)
 - Only 1 PD replica is working at present, no limitations for TiDB and TiKV

Automatically management of certs should be added in another PR (still WIP), and the known multi-PD connectivity issue is still under investigation, not likely to be fixed in this PR.

charts/tidb-cluster/values.yaml
```diff
@@ -47,9 +47,20 @@ discovery:
...
+# Whether enable TLS connections between server nodes.
+# When enabled, PD/TiDB/TiKV will use TLS encrypted connections to transfer data between each node,
+# certificates will be generated automatically (if not already present).
+enableTLSServer: false
+
+# Whether enable TLS connection between TiDB server and MySQL client.
+# When enabled, TiDB will accept TLS encrypted connections from MySQL client, certificates will be generated
+# automatically.
+# Note: TLS connection is not forced on the server side, plain connections are also accepted after enableing.
+enableTLSClient: false
```
cofyc: `enableTLSClient` adds TLS support between tidb-server and MySQL clients and does not affect other components. Actually, it enables tidb-server server-side TLS support, `client` in its name is misleading. How about moving it to `tidb.enableTLS` field? What do you think? cc @weekface @tennix

charts/tidb-cluster/values.yaml
```diff
@@ -47,9 +47,20 @@ discovery:
 # if the ConfigMap was not changed.
 enableConfigMapRollout: true
 
+# Whether enable TLS connections between server nodes.
+# When enabled, PD/TiDB/TiKV will use TLS encrypted connections to transfer data between each node,
+# certificates will be generated automatically (if not already present).
+enableTLSServer: false
```
cofyc: This enables TLS encrypted connections in TiDB cluster. For users, no clients are involved. I prefer `cluster` over `server` here.

charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -37,6 +37,13 @@ config-file: |-
     {{- if .Values.pd.config }}
 {{ .Values.pd.config | indent 2 }}
     {{- end -}}
+    {{- if .Values.enableTLSServer }}
+  [security]
+  cacert-path = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+  cert-path = "/var/lib/pd-tls/pd.crt"
+  key-path = "/var/lib/pd-tls/pd.key"
```
cofyc: Does PD use the same cert/key both for peer communication and API server?

charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -73,6 +87,20 @@ config-file: |-
     {{- if .Values.tidb.config }}
 {{ .Values.tidb.config | indent 2 }}
     {{- end -}}
+    {{- if or .Values.enableTLSServer .Values.enableTLSClient }}
+  [security]
+    {{- end -}}
+    {{- if .Values.enableTLSServer }}
+  cluster-ssl-ca = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+  cluster-ssl-cert = "/var/lib/tidb-tls/tidb.crt"
+  cluster-ssl-key = "/var/lib/tidb-tls/tidb.key"
```
cofyc: For TiDB, these certs are used to communicate with PD or TiKV or both?

charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -73,6 +87,20 @@ config-file: |-
...
+    {{- end -}}
+    {{- if .Values.enableTLSClient }}
+  ssl-ca = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+  ssl-cert = "/var/lib/tidb-tls/tidb.crt"
+  ssl-key = "/var/lib/tidb-tls/tidb.key"
```
cofyc: Should we support users to configure their own MySQL certificates?

charts/tidb-cluster/templates/_helpers.tpl
cofyc: ignore me, [this document](https://pingcap.com/docs/v3.0/how-to/secure/enable-tls-between-components/) clarifies it.

charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -56,12 +56,19 @@ while true; do
     fi
 done
 
+if [[ {{ .Values.enableTLSServer }} == "true" ]]
```
tennix:
```
SCHEME={{ if .Values.enableTLSServer }}"https"{{ else }}"http"{{ end }}
```

charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
```diff
@@ -28,7 +28,14 @@ then
 	tail -f /dev/null
 fi
 
-ARGS="--pd=${CLUSTER_NAME}-pd:2379 \
+if [[ {{ .Values.enableTLSServer }} == "true" ]]
```
tennix:
```
SCHEME={{ if .Values.enableTLSServer }}"https"{{ else }}"http"{{ end }}

File: pkg/controller/tidb_control.go [hunk 10]
```diff
@@ -63,10 +64,20 @@ func (tdc *defaultTiDBControl) GetHealth(tc *v1alpha1.TidbCluster) map[string]bo
 	tcName := tc.GetName()
 	ns := tc.GetNamespace()
 
+	scheme := "http"
+	if tc.Spec.EnableTLSServer {
+		scheme = "https"
+		rootCAs, _ := httputil.ReadCACerts()
```
tennix: Don't ignore errors in Go. if you're definitely sure the error would never happen, then add a comment for annotation.

pkg/controller/tidb_control.go
```diff
@@ -81,8 +92,18 @@ func (tdc *defaultTiDBControl) ResignDDLOwner(tc *v1alpha1.TidbCluster, ordinal
...
```
tennix: Are the certs the same for different components? If so, then this can be done once when tidb-controller-manager starts and use when needed.

pkg/discovery/discovery.go
```diff
@@ -92,12 +92,17 @@ func (td *tidbDiscovery) Discover(advertisePeerUrl string) (string, error) {
 	currentCluster = td.clusters[keyName]
 	currentCluster.peers[podName] = struct{}{}
 
+	scheme := "http"
+	if tc.Spec.EnableTLSServer {
```
tennix: This can be abstracted as a function.

pkg/httputil/httputil.go
```diff
@@ -44,3 +52,30 @@ func GetBodyOK(httpClient *http.Client, apiURL string) ([]byte, error) {
...
+		return nil, err
+	}
+	if ok := rootCAs.AppendCertsFromPEM(caCert); !ok {
+		glog.Warningf("fail to append CA file to pool, using system CAs only")
+	}
+	return rootCAs, nil
+}
+
+func ReadCerts() (*x509.CertPool, tls.Certificate, error) {
+	rootCAs, err := ReadCACerts()
```
tennix: Check the error.

pkg/httputil/httputil.go
```diff
@@ -44,3 +52,30 @@ func GetBodyOK(httpClient *http.Client, apiURL string) ([]byte, error) {
 	}
 	return body, err
 }
+
+func ReadCACerts() (*x509.CertPool, error) {
+	// try to load system CA certs
+	rootCAs, _ := x509.SystemCertPool()
```
tennix: ditto.

tests/failover.go
```diff
@@ -38,7 +38,7 @@ func (oa *operatorActions) TruncateSSTFileThenCheckFailover(info *TidbClusterCon
 	}
 
 	// checkout pd config
-	pdCfg, err := oa.pdControl.GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName()).GetConfig()
+	pdCfg, err := oa.pdControl.GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), false).GetConfig()
```
tennix:
```suggestion
	pdCfg, err := oa.pdControl.GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.EnableTLSServer).GetConfig()
```

tests/pkg/webhook/pods.go
```diff
@@ -53,7 +53,7 @@ func admitPods(ar v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 		return &reviewResponse
 	}
 
-	pdClient := pdapi.NewDefaultPDControl().GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName())
+	pdClient := pdapi.NewDefaultPDControl().GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), false)
```
tennix:
```suggestion
	pdClient := pdapi.NewDefaultPDControl().GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.EnableTLSServer)
```

tests/actions.go
```diff
@@ -994,7 +994,7 @@ func (oa *operatorActions) CheckUpgrade(ctx context.Context, info *TidbClusterCo
 			glog.Errorf("failed to get tidbcluster: %s/%s, %v", ns, tcName, err)
 			continue
 		}
-		pdClient := pdapi.NewDefaultPDControl().GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName())
+		pdClient := pdapi.NewDefaultPDControl().GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), false)
```
tennix:
```suggestion
		pdClient := pdapi.NewDefaultPDControl().GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.EnableTLSServer)
```

charts/tidb-cluster/values.yaml
tennix: Yes, I agree. I prefer `tidb.enableTLS` because `tidb.tls` should be reserved if we want to add more tls related configurations into it.

weekface: `tidb.enableTLS` is not clear, i prefer `tidb. enableTLSClient`

cofyc: but `tidb-server` does not act as a client. It enables server-side TLS support, then MySQL clients can connect to it. It's better not to introduce any new term when not necessary, no `client` term in tidb raw configuration too.

charts/tidb-cluster/templates/_helpers.tpl
cofyc: cc @tennix @weekface What do you think?
One case is when a user migrates a MySQL instance which already has TLS enabled but doesn't want to update application certificates. The user must configure exiting certificates into tidb-server.
If we're going to support this, do we need to support users to configure secret with an arbitrary name?

charts/tidb-cluster/values.yaml
weekface: I think the `client` word means the tidb server should validate the request's client cert.

charts/tidb-cluster/templates/_helpers.tpl
AstroProfundis: I'm thinking about the same thing too, but AFAIK we don't have any plan to support custom certs yet, if we do, custom CA importing would also need to be added, and we need to validate user uploaded certs to ensure they are signed by the uploaded CA, that would need more code changes and seems better to be another PR.

cofyc: agree. but it seems not right to use cluster TLS certs as tidb-server TLS server for MySQL clients, focusing on cluster TLS support only in this PR?

onlymellb: Since we did not support the custom certs now, I don't think it is necessary to expose these parameters to the user, which will make the user confused or misconfigured.

charts/tidb-cluster/templates/discovery-deployment.yaml
```diff
@@ -45,4 +45,14 @@ spec:
...
+        volumeMounts:
+        - mountPath: /var/lib/tls
+          name: tls
+          readOnly: true
+      volumes:
+      - name: tls
+        secret:
+          defaultMode: 420
+          secretName: client-tls
+{{- end -}}
```
onlymellb: IIRC, have you changed the way the certs was injected, by injecting certs in the program instead of mount certs into pod?

charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -58,12 +58,14 @@ while true; do
...
 ARGS="--data-dir=/var/lib/pd \
 --name=${POD_NAME} \
---peer-urls=http://0.0.0.0:2380 \
---advertise-peer-urls=http://${domain}:2380 \
---client-urls=http://0.0.0.0:2379 \
---advertise-client-urls=http://${domain}:2379 \
+--peer-urls=$SCHEME://0.0.0.0:2380 \
+--advertise-peer-urls=$SCHEME://${domain}:2380 \
+--client-urls=$SCHEME://0.0.0.0:2379 \
+--advertise-client-urls=$SCHEME://${domain}:2379 \
```
onlymellb: Use `${SCHEME}` instead,  keep consistent with other variables

charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
```diff
@@ -28,9 +28,11 @@ then
 	tail -f /dev/null
 fi
 
+SCHEME={{ if .Values.enableTLSCluster }}"https"{{ else }}"http"{{ end }}
+
 # Use HOSTNAME if POD_NAME is unset for backward compatibility.
 POD_NAME=${POD_NAME:-$HOSTNAME}
-ARGS="--pd=${CLUSTER_NAME}-pd:2379 \
+ARGS="--pd=$SCHEME://${CLUSTER_NAME}-pd:2379 \
```
onlymellb: ditto

charts/tidb-operator/templates/controller-manager-deployment.yaml
```diff
@@ -52,3 +52,12 @@ spec:
...
```
weekface: Do we need to create the `client-tls ` secret manually?

charts/tidb-cluster/templates/discovery-deployment.yaml
AstroProfundis: Yes, it's been changed in #782

charts/tidb-operator/templates/controller-manager-deployment.yaml
AstroProfundis: As of this PR, Yes, and the cert need to be signed by the k8s CA by `kubectl certificate approve`, as described in the PR OP.

PR #782 implements the automatic process of generating certs.

pkg/controller/tidb_control.go
```diff
@@ -51,22 +54,40 @@ type TiDBControlInterface interface {
...
 	httpClient := &http.Client{Timeout: timeout}
-	return &defaultTiDBControl{httpClient: httpClient}
+
+	rootCAs, err := httputil.ReadCACerts()
+	if err != nil {
+		glog.Errorf("fail to load CA certs, use plain connections only, TLS enabled clusters may not work")
+	}
+
+	return &defaultTiDBControl{httpClient: httpClient, rootCAs: rootCAs}
+}
```
onlymellb: This func  `NewDefaultTiDBControl ` should return `error` or put the logic to get the rootCA outside of this function.

pkg/manager/member/pd_member_manager.go
```diff
@@ -460,6 +466,15 @@ func (pmm *pdMemberManager) getNewPDSetForTidbCluster(tc *v1alpha1.TidbCluster)
 			},
 		},
 	}
+	if tc.Spec.EnableTLSCluster {
+		vols = append(vols, corev1.Volume{
+			Name: "pd-tls", VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: fmt.Sprintf("%s-pd", tcName),
```
onlymellb: Maybe you can use `controller.PDMemberName(tcName)` instead.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -248,6 +254,15 @@ func (tmm *tidbMemberManager) getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbClust
 			}},
 		},
 	}
+	if tc.Spec.EnableTLSCluster {
+		vols = append(vols, corev1.Volume{
+			Name: "tidb-tls", VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: fmt.Sprintf("%s-tidb", tcName),
```
onlymellb: ditto

pkg/manager/member/tikv_member_manager.go
```diff
@@ -298,6 +304,15 @@ func (tkmm *tikvMemberManager) getNewSetForTidbCluster(tc *v1alpha1.TidbCluster)
 			}},
 		},
 	}
+	if tc.Spec.EnableTLSCluster {
+		vols = append(vols, corev1.Volume{
+			Name: "tikv-tls", VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: fmt.Sprintf("%s-tikv", tcName),
```
onlymellb: ditto

pkg/pdapi/pdapi.go
```diff
@@ -134,10 +140,26 @@ type pdClient struct {
 }
 
 // NewPDClient returns a new PDClient
-func NewPDClient(url string, timeout time.Duration) PDClient {
+func NewPDClient(url string, timeout time.Duration, tlsEnabled bool) PDClient {
+	httpClient := &http.Client{Timeout: timeout}
+	if tlsEnabled {
+		rootCAs, cert, err := httputil.ReadCerts()
+		if err != nil {
+			glog.Errorf("fail to load certs, fallback to plain connection, err: %s", err)
```
onlymellb: Some problem with func `NewDefaultTiDBControl`

AstroProfundis: The logic of `NewPDClient` is changed in #782 and does not have this issue anymore

onlymellb: But func `NewDefaultTiDBControl `  still has this problem

AstroProfundis: `NewDefaultTiDBControl` has been changed in the latest commit  7d9d03b

charts/tidb-operator/templates/controller-manager-deployment.yaml
cofyc: make this optional, otherwise, it will fail when users don't need to enable TLS

PR #769, opened 2019-08-14
pump: prevent pump progress from an exit with 0 if the pump become `offline`

weekface:

prevent pump progress from an exit with 0 if the pump become `offline`
, because it will be `online` again if the pod get restarted by `statefulset`.

### Check List 

Tests 

Code changes

 - Has Helm charts change

Side effects

Related changes

 ```release-note
prevent pump progress from an exit with 0 if the pump become `offline`
 ```

### DIFF
--- a/charts/tidb-cluster/templates/scripts/_start_pump.sh.tpl
+++ b/charts/tidb-cluster/templates/scripts/_start_pump.sh.tpl
@@ -5,3 +5,8 @@ set -euo pipefail
 -advertise-addr=`echo ${HOSTNAME}`.{{ template "cluster.name" . }}-pump:8250 \
 -config=/etc/pump/pump.toml \
 -log-file=
+
+if [ $? == 0 ]; then
+    echo $(date -u +"[%Y/%m/%d %H:%M:%S.%3N %:z]") "pump offline, please delete my pod"
+    tail -f /dev/null
+fi


charts/tidb-cluster/templates/scripts/_start_pump.sh.tpl
```diff
@@ -5,3 +5,8 @@ set -euo pipefail
 -advertise-addr=`echo ${HOSTNAME}`.{{ template "cluster.name" . }}-pump:8250 \
 -config=/etc/pump/pump.toml \
 -log-file=
+
+if [ $? == 0 ]; then
+    echo $(date -u +"[%Y/%m/%d %H:%M:%S.%3N %:z]") "pump offline, please delete my pod"
+    tail -f /dev/null
+fi
```
onlymellb: Can you be sure that if the exit code is 0, it must be pump offline?

weekface: This script ensures that the `pump` container would not exit normally, it must be killed by `kubelet`.

tennix: The `pause` state also ends with exit code 0, but anyway the pod needs to be running.

aylei: What if the container get restarted?

aylei: For example, due to node failure or eviction

tennix: If container restarted, it will be running again, users have to issue `binlogctl` again to offline the pump. And if the pod is evicted for disk or node failure, then that data may be lost because pump data is not replicated.

PR #772, opened 2019-08-15
deploy, manifests: enable cloud storage volume expansion & label local volume

tennix:

Cloud storage volume expansion is enabled by default. Hostname label will be attached to local PV.

### Check List 

Tests 

 - Manual test
    1. Cloud provider storage class `allowVolumeExpansion` attribute can be updated in-place by `kubectl edit sc ebs-gp2` to add `allowVolumeExpansion: true`
    2. After adding `allowVolumeExpansion: true`, updating PVC with larger storage size, the corresponding PV will be expanded to the desired size.
    3. The storage class `reclaimPolicy` can not be changed dynamically. So old gke persistent disk storage class cannot migrate to `Retain`. 

Code changes
 - Has volume setup change

Side effects

 None

Related changes

 - Need to cherry-pick to the release-1.0 branch

 ```release-note
Users can now expand cloud storage PV dynamically by increasing PVC storage size.
 ```


deploy/modules/aws/tidb-operator/manifests/local-volume-provisioner.yaml
```diff
@@ -12,6 +12,8 @@ metadata:
   name: local-provisioner-config
   namespace: kube-system
 data:
+  nodeLabelsForPV: |
```
gregwebs: This copies the label from the node to the PV, but what is the purpose of that?

tennix: Users can simply know the PV belongs to which node or a node contains which PV by `kubectl get pv --show-labels | grep <node-name>`.

gregwebs:
> The storage class reclaimPolicy can not be changed dynamically. So old gke persistent disk storage class cannot migrate to Retain

Currently we migrate from Retain to Delete. Can we still do this?

tennix:
> > The storage class reclaimPolicy can not be changed dynamically. So old gke persistent disk storage class cannot migrate to Retain
> 
> Currently we migrate from Retain to Delete. Can we still do this?

No, but for tidb-operator managed PVC, the reclaim policy was reset to `Retain` by default no matter what is the storage class's default reclaim policy. And the existing PV can be patched to use `Retain`.

tennix:
> LGTM
> BTW, I suppose we should document the extra works if the user want to expand the volume size of PVCs managed by StatefulSet.

I'll send a PR to add the document to docs and docs-cn later.

PR #774, opened 2019-08-15
Support hostNetwork

### LINKED ISSUE #759: Able to run TiDB components in host network
## Feature Request

In GKE TiDB sysbench benchmark, performance in the host network is slightly better (10%) than the pod network (GKE cluster network is VPC-native).

Able to run TiDB components (pd/tidb/tikv) in host network.

One issue to run pods in host network is hostname is not pod name anymore. In Kubernetes, the hostname is set to the host hostname when the pod is running in host network (xref: https://github.com/kubernetes/kubernetes/issues/12893). We need to update our start scripts to use pod name to construct the component address.

cofyc:

This adds a new filed in PDSpec/TiKVSpec/TiDBSpec to make it possible to run TiDB components (pd, tikv, tidb) in the host network.

fix #759 

### Check List 

Tests 

Code changes

 - Has Helm charts change
 - Has Go code change

 ```release-note
Add `hostNetwork` field in pd/tikv/tidb spec to make it possible to run tidb components in host network.
 ```


pkg/apis/pingcap.com/v1alpha1/types.go
```diff
@@ -103,26 +103,28 @@ type TidbClusterStatus struct {
...
 
+// PodSpec repreresents shared pod fields between PD/TiDB/TiKV
+type PodSpec struct {
+	Replicas     int32               `json:"replicas"`
+	Affinity     *corev1.Affinity    `json:"affinity,omitempty"`
+	NodeSelector map[string]string   `json:"nodeSelector,omitempty"`
+	Tolerations  []corev1.Toleration `json:"tolerations,omitempty"`
+	Annotations  map[string]string   `json:"annotations,omitempty"`
+	HostNetwork  bool                `json:"hostNetwork,omitempty"`
+}
```
cofyc: `PodSpec` is created to simplify code only. Only one field `hostNetwork` is introduced.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -320,6 +320,11 @@ func (tkmm *tikvMemberManager) getNewSetForTidbCluster(tc *v1alpha1.TidbCluster)
 		storageClassName = controller.DefaultStorageClassName
 	}
 
+	dnsPolicy := corev1.DNSClusterFirst // same as k8s defaults
+	if tc.Spec.PD.HostNetwork {
+		dnsPolicy = corev1.DNSClusterFirstWithHostNet
```
cofyc: `DNSClusterFirstWithHostNet` is required in the host network, see https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy. Otherwise, the DNS resolver is the same as the host. Our components cannot resolve Kubernetes DNS names.

charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -30,7 +30,7 @@ fi
 
 # the general form of variable PEER_SERVICE_NAME is: "<clusterName>-pd-peer"
 cluster_name=`echo ${PEER_SERVICE_NAME} | sed 's/-pd-peer//'`
-domain="${HOSTNAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc"
+domain="${POD_NAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc"
```
cofyc: We should use `POD_NAME` instead of `HOSTNAME` environment. In Kubernetes, the hostname in the pod is implicitly set to the host hostname when the pod is running in host network (xref: https://github.com/kubernetes/kubernetes/issues/12893). Currently, there is no way to change this behavior.

weekface:
Add a case in stability test?

cofyc:
Sure, I'll add a stability test.

cofyc:
I removed the `needs-cherry-pick-1.0` label because it's a new feature, not a bugfix or improvement. We can target 1.1 release.

cofyc:
The stability test is under development, I'll update this PR soon.

pkg/apis/pingcap.com/v1alpha1/types.go
cofyc: This has been reverted because it cannot pass `go vet` check.

```
pkg/apis/pingcap.com/v1alpha1/tidbcluster_test.go:180:5: cannot use promoted field PodSpec.Replicas in struct literal of type PDSpec
pkg/apis/pingcap.com/v1alpha1/tidbcluster_test.go:183:5: cannot use promoted field PodSpec.Replicas in struct literal of type TiKVSpec
pkg/apis/pingcap.com/v1alpha1/tidbcluster_test.go:186:5: cannot use promoted field PodSpec.Replicas in struct literal of type TiDBSpec
...
```

cofyc:
I've added a test in e2e. @weekface @tennix @xiaojingchen @gregwebs PTAL.
It's safe to switch back and forth between pod network and host network (if no port conflicts). I also manually tested in GKE.

tennix:
In this mode, the pod should be set with anti-affinity to disable two pods scheduled on the same node. This should be added to dev document.

charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
tennix: `POD_NAME` is undefined if using old tidb-operator. To be compatible with old tidb-operator, set `POD_NAME` to `$(hostname)` or `HOSTNAME` if `POD_NAME` is undefined.

PR #776, opened 2019-08-15
terraform/gcp: Able to configure image type

cofyc:

Sometimes users may favor non-COS node images. This adds variables to configure image type for TiDB node pools.

### Check List 

Tests 

 ```release-note
terraform/gcp: Able to configure node image types for PD/TiDB/TiKV node pools.
 ```

### DIFF
--- a/deploy/modules/gcp/tidb-cluster/main.tf
+++ b/deploy/modules/gcp/tidb-cluster/main.tf
@@ -15,6 +15,7 @@ resource "google_container_node_pool" "pd_pool" {
 
   node_config {
     machine_type    = var.pd_instance_type
+    image_type      = var.pd_image_type
     local_ssd_count = 0
 
     taint {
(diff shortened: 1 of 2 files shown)


deploy/gcp/variables.tf
```diff
@@ -80,6 +80,21 @@ variable "tikv_instance_type" {}
...
+
+variable "tidb_image_type" {
+  description = "TiDB image type, avaiable: UBUNTU/COS"
+  default = "COS"
+}
+
+variable "tikv_image_type" {
+  description = "TiKV image type, avaiable: UBUNTU/COS"
+  default = "COS"
+}
```
aylei: These variables should be added to `deploy/modules/gcp/tidb-cluster/variables.tf`

cofyc: OK, I'll fix it.

cofyc: fixed

PR #777, opened 2019-08-15
Monitoring: Integrated kubernetes cluster monitoring in public cloud platform

qiffang:
Signed-off-by: qiffang <947321353@qq.com>

Integrated k8s cluster monitor to terraform for public cloud platform.

All major changes is in terraform files. It will apply monitoring files after you execute the command `terraform apply`.

### Check List 

Tests 

 - Manual test 
There are monitoring pods in monitoring namespaces

Code changes

 - Has terraform files change

### DIFF
--- a/deploy/modules/aliyun/tidb-operator/operator.tf
+++ b/deploy/modules/aliyun/tidb-operator/operator.tf
@@ -36,7 +36,12 @@ mkdir monitor
 tar -zxvf prometheus-operator.tar.gz -C monitor/
 tar -zxvf prometheus.tar.gz -C monitor/
 kubectl apply -f monitor/manifests/prometheus-operator
+sed -i'.bak' 's/local-storage/alicloud-disk-available/g' monitor/manifests/prometheus/grafana-pvc.yaml
+sed -i'.bak' 's/local-storage/alicloud-disk-available/g' monitor/manifests/prometheus/prometheus-prometheus.yaml
+rm -rf monitor/manifests/prometheus/*.bak
 kubectl apply -f monitor/manifests/prometheus
+rm -rf prometheus*.tar.gz
+rm -rf monitor/
 
 helm init
 until helm ls; do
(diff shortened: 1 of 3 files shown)


gregwebs:
Should the downloaded content be deleted?
How do we ensure that there are resources on nodes that are available for this?
Is there a related change to the docs to tell the user about this?

gregwebs:
This is not added in a modular way: a user would have to delete/comment out theses lines in terraform and maybe manually delete the operator. Also, I am not a fan of helm in general, but it is nice to know that I can delete a helm install with `helm del --purge`. How does one delete this?

I think this is available as a helm chart and the user can install it with a single command? It may be better to just add that one line command to the end of the docs for now until we develop a more custom integration for this.

qiffang:
> Should the downloaded content be deleted?
> How do we ensure that there are resources on nodes that are available for this?
> Is there a related change to the docs to tell the user about this?

The downloaded repo is maintained by us, so we can ensure that the resources always exist.
Currently, terraform also use the same way to apply CRD.
```
kubectl apply -f https://raw.githubusercontent.com/pingcap/tidb-operator/${var.operator_version}/manifests/crd.yaml
```

gregwebs:
The new terraform is designed to be modular and users should be able to bring their own K8s cluster. So they may already have Prometheus running and this could create a duplicate installation.

I would really like to just add this suggestion to the end of our docs rather than deal with all the complications from including it.

jlerche:
I will second @gregwebs suggestion. At the very least, if it will be added in to the terraform script, it should be included as an optional module. @qiffang @aylei what do you think?

qiffang:
> I will second @gregwebs suggestion. At the very least, if it will be added in to the terraform script, it should be included as an optional module. @qiffang @aylei what do you think?

@tennix, what do you think?

tennix:
Yes, I agree it should be optional, users can enable or disable it.

PR #782, opened 2019-08-16
Auto generate and sign certificates for TLS enabled cluster

AstroProfundis:
Automatically generate and sign certificates for TLS enabled TiDB clusters, if:
 - No `Secret` matching the name and labels, and
 - No `CertificateSigningRequest` matching the name that not matching the labels (so that they are not created by us)

After certificates get automatically approved, they are stored as Secrets and mount to Pods.

The procedure is based on the [manual for ansible deployment](https://github.com/pingcap/docs-cn/blob/master/v3.0/how-to/secure/enable-tls-between-components.md).

The corresponding documents are here:
- https://pingcap.com/docs/stable/how-to/secure/enable-tls-between-components/
- https://pingcap.com/docs/v3.1/how-to/secure/enable-tls-clients/

This PR includes and close #750 and added controllers to handle certificate managements.

### Check List 

Tests 

Manually validated on internal testing k8s cluster.

Code changes

 - Has Helm charts change
 - Has Go code change

Known Issues

 - [pd#1682](https://github.com/pingcap/pd/issues/1682) (fixed)
    - Only support PD version >= `v3.0.5`
 - TiKV and TiDB/PD have different behavior on metrics endpoints, TiKV does not require client certs so the monitoring metrics of TiKV are collected via plain HTTP as a workaround
 - The certificates are signed with 1 year expiration time, but due to the limit of servers, there is no clean way (yet) to renew the certs without stopping the TiDB cluster

Related changes

#750 

 ```release-note
Enable auto generate certificates for TiDB cluster
 ```


gregwebs:
It can be problematic to automatically create the cert if the secret is not found: the deployer may have forgotten that step. It may be better to more explicitly configure the cert source.

AstroProfundis:
@gregwebs We can check if a csr already exist and try to approve that, but we don't have private key in that situation, I haven't came up with any idea good enough to check if user had forgotten to create secret (or mis-configured the secret), or there is no pre-confiugred key pair at all. Any suggestions?

BTW, I think our current procedure don't support user generated certs, supporting custom cert would need to support external CA as well, this won't work with this PR at present.

gregwebs:
The approach would be to explicitly configure whether a secret is expected. If it is expected and does not exist, than that produces an error.

pkg/controller/cert_control.go
```diff
@@ -0,0 +1,289 @@
...
+
+func (rcc *realCertControl) Create(tc *v1alpha1.TidbCluster, commonName string,
+	hostList []string, IPList []string, suffix string) error {
+	ns := tc.GetNamespace()
+	tcName := tc.GetName()
+	csrName := fmt.Sprintf("%s-%s", tcName, suffix)
+
+	// generate certificate if not exist
+	_, _, err := rcc.LoadFromSecret(ns, csrName)
+	if !apierrors.IsNotFound(err) {
```
gregwebs: The above comment seems wrong now. I would annotate this error with "secret csrName not found". `err != nil` is not handled now

AstroProfundis:
TiKV has different behavior on metrics page with TiDB/PD after TLS enabled [#5340](https://github.com/tikv/tikv/issues/5340), I'll see if any workaround could be implemented.

charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -37,6 +37,13 @@ config-file: |-
     {{- if .Values.pd.config }}
 {{ .Values.pd.config | indent 2 }}
     {{- end -}}
+    {{- if .Values.enableTLSCluster }}
+  [security]
+  cacert-path = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+  cert-path = "/var/lib/pd-tls/cert"
+  key-path = "/var/lib/pd-tls/key"
+    {{- end -}}
+
```
onlymellb: refer https://github.com/pingcap/tidb-operator/pull/750#discussion_r317957297

charts/tidb-cluster/templates/scripts/_start_drainer.sh.tpl
```diff
@@ -24,9 +24,11 @@ while true; do
     fi
 done
 
+SCHEME={{ if .Values.enableTLSCluster }}"https"{{ else }}"http"{{ end }}
+
 /drainer \
 -L={{ .Values.binlog.drainer.logLevel | default "info" }} \
--pd-urls=http://{{ template "cluster.name" . }}-pd:2379 \
+-pd-urls=$SCHEME://{{ template "cluster.name" . }}-pd:2379 \
```
onlymellb: Use `${SCHEME}` instead

charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl
```diff
@@ -58,12 +58,14 @@ while true; do
...
 --name=${POD_NAME} \
---peer-urls=http://0.0.0.0:2380 \
---advertise-peer-urls=http://${domain}:2380 \
---client-urls=http://0.0.0.0:2379 \
---advertise-client-urls=http://${domain}:2379 \
+--peer-urls=$SCHEME://0.0.0.0:2380 \
+--advertise-peer-urls=$SCHEME://${domain}:2380 \
+--client-urls=$SCHEME://0.0.0.0:2379 \
+--advertise-client-urls=$SCHEME://${domain}:2379 \
 --config=/etc/pd/pd.toml \
```
onlymellb: ditto

charts/tidb-cluster/templates/scripts/_start_pump.sh.tpl
```diff
@@ -1,6 +1,9 @@
 set -euo pipefail
+
+SCHEME={{ if .Values.enableTLSCluster }}"https"{{ else }}"http"{{ end }}
+
 /pump \
--pd-urls=http://{{ template "cluster.name" . }}-pd:2379 \
+-pd-urls=$SCHEME://{{ template "cluster.name" . }}-pd:2379 \
```
onlymellb: ditto

charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
```diff
@@ -28,9 +28,11 @@ then
 	tail -f /dev/null
 fi
 
+SCHEME={{ if .Values.enableTLSCluster }}"https"{{ else }}"http"{{ end }}
+
 # Use HOSTNAME if POD_NAME is unset for backward compatibility.
 POD_NAME=${POD_NAME:-$HOSTNAME}
-ARGS="--pd=${CLUSTER_NAME}-pd:2379 \
+ARGS="--pd=$SCHEME://${CLUSTER_NAME}-pd:2379 \
```
onlymellb: ditto

pkg/controller/cert_control.go
```diff
@@ -0,0 +1,323 @@
...
+*/
+/*
+func (rcc *realCertControl) RenewCert() error {
+	return nil
+}
+*/
+
+// LoadFromSecret loads cert and key from Secret matching the name
+func (rcc *realCertControl) LoadFromSecret(ns string, secretName string) ([]byte, []byte, error) {
+	secret, err := rcc.kubeCli.CoreV1().Secrets(ns).Get(secretName, types.GetOptions{})
```
onlymellb: We should try to avoid access to apiserver directly, use the `SecretLister` to access the local cache instead. Refer https://github.com/pingcap/tidb-operator/blob/master/pkg/backup/util/util.go#L79

pkg/controller/cert_control.go
```diff
@@ -0,0 +1,323 @@
...
+				return err
+			}
+			continue
+		}
+	}
+}
+
+func (rcc *realCertControl) getCSR(ns string, instance string, csrName string) (*capi.CertificateSigningRequest, error) {
+	getOpts := types.GetOptions{TypeMeta: types.TypeMeta{Kind: "CertificateSigningRequest"}}
+	csr, err := rcc.kubeCli.CertificatesV1beta1().CertificateSigningRequests().Get(csrName, getOpts)
```
onlymellb: Use `v1beta1.CertificateSigningRequestLister` instead

pkg/controller/cert_control.go
```diff
@@ -0,0 +1,323 @@
...
+		},
+	}
+
+	labelTemp := label.New()
+	secret.Labels[label.NamespaceLabelKey] = ns
+	secret.Labels[label.ManagedByLabelKey] = labelTemp[label.ManagedByLabelKey]
+	secret.Labels[label.InstanceLabelKey] = instance
+	secret.Labels[label.ComponentLabelKey] = component
+
+	_, err := rcc.kubeCli.CoreV1().Secrets(ns).Create(secret)
```
onlymellb: In order to keep the code style consistent, I think need to create a `secret_control.go` to manage the secret-related operations like [service_control.go](https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/service_control.go) does.

pkg/controller/cert_control.go
```diff
@@ -0,0 +1,323 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	types "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/kubernetes"
+)
+
+// CertControlInterface manages certificates used by TiDB clusters
+type CertControlInterface interface {
+	Create(ns string, instance string, commonName string, hostList []string, IPList []string, component string, suffix string) error
```
onlymellb: Pass too many parameters，I think we can pass `*v1alpha1.TidbCluster` instead of `ns` and `instance`，and define a `tidbClusterCert` struct to store certificate related content，like [kubeadm](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/certs/certlist.go#L34-L43) does.

charts/tidb-cluster/templates/config/_drainer-config.tpl
```diff
@@ -11,7 +11,7 @@ detect-interval = {{ .Values.binlog.drainer.detectInterval | default 10 }}
 data-dir = "/data"
 
 # a comma separated list of PD endpoints
-pd-urls = "http://{{ template "cluster.name" . }}-pd:2379"
+pd-urls = "{{ if .Values.enableTLSCluster }}https{{ else }}http{{ end }}://{{ template "cluster.name" . }}-pd:2379"
```
weekface: define scheme as a var like: https://github.com/pingcap/tidb-operator/pull/782/files#diff-d636dd7eac2fffa04bb4a9d148d705afR27

charts/tidb-cluster/templates/config/_pump-config.tpl
```diff
@@ -17,7 +17,7 @@ data-dir = "/data"
 heartbeat-interval = {{ .Values.binlog.pump.heartbeatInterval | default 2 }}
 
 # a comma separated list of PD endpoints
-pd-urls = "http://{{ template "cluster.name" . }}-pd:2379"
+pd-urls = "{{ if .Values.enableTLSCluster }}https{{ else }}http{{ end }}://{{ template "cluster.name" . }}-pd:2379"
```
weekface: ditto

charts/tidb-cluster/values.yaml
```diff
@@ -47,6 +47,11 @@ discovery:
 # if the ConfigMap was not changed.
 enableConfigMapRollout: true
 
+# Whether enable TLS connections between server nodes.
+# When enabled, PD/TiDB/TiKV will use TLS encrypted connections to transfer data between each node,
+# certificates will be generated automatically (if not already present).
+enableTLSCluster: false
```
weekface: how to supply custom certificates?

cmd/controller-manager/main.go
```diff
@@ -136,6 +136,10 @@ func main() {
 	}
 
 	tcController := tidbcluster.NewController(kubeCli, cli, informerFactory, kubeInformerFactory, autoFailover, pdFailoverPeriod, tikvFailoverPeriod, tidbFailoverPeriod)
+	secControl := controller.NewRealSecretControl(kubeCli, kubeInformerFactory.Core().V1().Secrets().Lister())
+	certController := controller.NewRealCertControl(kubeCli,
```
weekface:
```suggestion
	certControl := controller.NewRealCertControl(kubeCli,
```

pkg/controller/secret_control.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, nil, err
+	}
+
+	return secret.Data["cert"], secret.Data["key"], nil
+}
+
+// Check returns true if the secret already exist
+func (rsc *realSecretControl) Check(ns string, secretName string) bool {
+	certBytes, keyBytes, err := rsc.Load(ns, secretName)
+	if err != nil {
```
weekface: use `glog.Errorf` to log the err message.

cmd/controller-manager/main.go
```diff
@@ -145,6 +149,9 @@ func main() {
 	go kubeInformerFactory.Start(controllerCtx.Done())
 
 	onStarted := func(ctx context.Context) {
+		if err := generateClientCert(ns, "tidb-operator", certController); err != nil {
```
weekface: I recommend creating the operator client certificate in the tidb-operator installation.

pkg/controller/cert_control.go
```diff
@@ -0,0 +1,256 @@
...
+	// wait at most 5min for the cert to be signed
+	timeout := int64(time.Minute.Seconds() * 5)
+	tick := time.After(time.Second * 10)
+	watchReq := types.ListOptions{
+		Watch:          true,
+		TimeoutSeconds: &timeout,
+		FieldSelector:  fields.OneTermEqualSelector("metadata.name", csrName).String(),
+	}
+
+	csrCh, err := rcc.kubeCli.Certificates().CertificateSigningRequests().Watch(watchReq)
```
weekface: this will block the loop 5 minutes

pkg/controller/secret_control.go
```diff
@@ -0,0 +1,140 @@
...
+		return false
+	}
+
+	// validate if the certificate is valid
+	block, _ := pem.Decode(certBytes)
```
weekface: don't omit `err`

pkg/manager/member/pd_member_manager.go
```diff
@@ -239,6 +252,58 @@ func (pmm *pdMemberManager) syncPDStatefulSetForTidbCluster(tc *v1alpha1.TidbClu
...
+	certOpts := &controller.TiDBClusterCertOptions{
+		Namespace:  ns,
+		Instance:   tcName,
+		CommonName: commonName,
+		HostList:   hostList,
+		Component:  "pd",
+		Suffix:     "pd-client",
+	}
+
+	return pmm.certControl.Create(certOpts)
```
weekface: this will create the client cert every loop

charts/tidb-cluster/values.yaml
AstroProfundis: Custom certificates are not supported at present

pkg/controller/secret_control.go
AstroProfundis: It's not an error `func Decode(data []byte) (p *Block, rest []byte)`

pkg/controller/cert_control.go
AstroProfundis: If everything work normal, the loop will exit at line 142 (usually within few seconds), the 5 min timeout is the worst case that CSR is not properly signed by k8s.

pkg/manager/member/pd_member_manager.go
AstroProfundis: `syncPDServerCerts()` and `syncPDClientCerts()` are only called when the StatefulSet doesn't exist (line 188)

charts/tidb-cluster/templates/config/_drainer-config.tpl
AstroProfundis: It's an YAML template, not a shell script here

charts/tidb-cluster/templates/_helpers.tpl
tennix: Adding `security` config section here is fragile, users may provide this section in the `.Values.pd.config` too thus causing invalid configuration file.

tennix:
The `make tidy` fails in GitHub Action CI, please fix it @AstroProfundis

charts/tidb-cluster/templates/config/_prometheus-config.tpl
```diff
@@ -46,11 +53,63 @@ scrape_configs:
     - source_labels: [__meta_kubernetes_pod_ip]
       action: replace
       target_label: kubernetes_pod_ip
+    {{- if .Values.enableTLSCluster }}
```
tennix: Add a TODO here that when TiKV supports metrics API over https, this workaround should be removed.

charts/tidb-cluster/templates/config/_drainer-config.tpl
tennix: You can define a `SCHEME` macro in `_helpers.tpl` to avoid repeating the `if-else` block again and again.

charts/tidb-cluster/templates/monitor-deployment.yaml
```diff
@@ -134,6 +134,11 @@ spec:
           - name: prometheus-rules
             mountPath: /prometheus-rules
             readOnly: false
+          {{- if .Values.enableTLSCluster }}
+          - name: tls-pd-client
```
tennix: What about tidb?

AstroProfundis: The same cert pair can be used for every endpoint, as long as the cert is signed by the same CA as server, using any one of our valid secret would work.

tennix: OK, got it.

PR #788, opened 2019-08-19
Adds workaround for data.google_compute_zones.available weirdness

jlerche:
```hcl
length(data.google_compute_zones.available)
```
returns `5` for `us-west1`.

In `deploy/modules/gcp/tidb-cluster/main.tf` switches to 
```hcl
length(data.google_compute_zones.available.names)
```
which correctly reteurns `3`

### Check List 

Tests 

 - Manual test (add detailed scripts or steps below):

    - `terraform apply`
    - When complete, note that number of pods is equal to the number of availability zones in the region (3 for us-west1).
    - `terraform destroy`

Code changes

 - Has Terraform changes

Related changes

 ```release-note
NONE
 ```

### DIFF
--- a/deploy/modules/gcp/tidb-cluster/main.tf
+++ b/deploy/modules/gcp/tidb-cluster/main.tf
@@ -125,7 +125,7 @@ resource "google_container_node_pool" "monitor_pool" {
 }
 
 locals {
-  num_availability_zones = length(data.google_compute_zones.available)
+  num_availability_zones = length(data.google_compute_zones.available.names)
 }
 
 module "tidb-cluster" {


cofyc:
have you tested it in region `us-central1`?  it has [four zones](https://cloud.google.com/compute/docs/regions-zones/#locations).

jlerche:
> have you tested it in region `us-central1`? it has [four zones](https://cloud.google.com/compute/docs/regions-zones/#locations). Does it have a problem?

It correctly returns `4`. Here's a quick script to demonstrate:

```bash
$ cat main.tf 
variable "region" {
  default = "us-west1"
}

variable "project" { default = "jacob-operator" }

provider "google" {
  project = var.project
  region  = var.region
  credentials = file("/path/to/credentials/file.json")
}
data "google_compute_zones" "available" {}

output "zones" {
  value = "${data.google_compute_zones.available.names}"
}

output "length" {value=length(data.google_compute_zones.available.names)}
$ terraform apply -var 'region=us-central1'
data.google_compute_zones.available: Refreshing state...

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.

Outputs:

length = 4
zones = [
  "us-central1-a",
  "us-central1-b",
  "us-central1-c",
  "us-central1-f",
]
```

cofyc:
this place may not be right

https://github.com/pingcap/tidb-operator/blob/89d41bdb7f9edc1151e7f2de2adbccfedf4c274b/deploy/modules/gcp/tidb-cluster/main.tf#L134-L136

[the default number of zones of the regional cluster is three](https://cloud.google.com/kubernetes-engine/docs/concepts/regional-clusters#overview) but we will create `node*4` instances.

I guess we need to use the number of `cluster.locations`. It's not available `google_container_cluster` resource attributes. One hack method is to run this command:

```
gcloud container clusters list --region us-central1 --filter="name='$name'" --format='csv[no-heading](locations)'
```

jlerche:
So even in us-central1, a regional cluster will still have 3 zones by default? Will the `gcloud` command also work for regions like `us-west1`?

cofyc:
yes, by default it will choose 3 from all available zones (4 in `us-central1`). This behavior can be changed by specifying the desired node locations with `--node-locations` flag. I'm working on a PR to support the zonal/multi-zonal cluster.

> Will the gcloud command also work for regions like us-west1?

It works too. This command retrieves the `locations` field from the cluster object. It's cluster's attribute (I think we can send a PR to google terraform provider for this).

PR #789, opened 2019-08-19
use mdadm instead of LVM if it is available

gregwebs:
local-volume-provisioner for GKE. Use mdadm in preference to LVM. Initial tests indicate this may perform better. This only works on Ubuntu. COS for GKE 1.14 is suppose to support RAID, but the mdadm binary is not there yet.

This has been tested on GKE to work on both COS and Ubuntu.

 ```release-note
On GKE, use mdadm and RAID rather than LVM when it is available
 ```


manifests/gke/local-ssd-provision/local-ssd-provision.yaml
```diff
@@ -88,21 +93,34 @@ spec:
                   fi
                 done
 
-                if ! /sbin/pvs | grep volume_all_ssds ; then
-                  # Don't combine with lvm if there is 1 disk or the environment variable is set.
-                  # lvm does have overhead, so if there is just 1 disk do not use lvm.
+                devs=$(echo "$ssd_mounts" | awk '{print $2}')
+                raid_dev=/dev/md0
+
+                # If RAID or LVM is already in use, this may have been re-deployed.
```
cofyc: When `USE_LVM` is changed from `1` to `0`, will combined disks with mdadm be wiped?

gregwebs: The disks would never be wiped again if the write to `/var/dev_wiped/` is persisted. But this comment is to indicate that if we detect an existing LVM or RAID setup then don't go down this path where we wipe the disks.

The design of the script is that it to detect if it already ran and to never change what it has already done. So if you change to USE_LVM=1 that will effectively be ignored if the raid device has already been created.

Issue #792, opened 2019-08-20
tidb-drainer cannot be deployed successfully out-of-box


shinnosuke-okada:
## Bug Report

**What version of Kubernetes are you using?**
1.13

**What version of TiDB Operator are you using?**
1.0.0-Release1

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**
N/A

**What's the status of the TiDB cluster pods?**
tidb-drainer pod crashes right after container start

**What did you do?**
Install TiDB cluster v2.1.16 using GA helm charts (v1.0.0)

**What did you expect to see?**
Successful drainer deployment out-of-box

**What did you see instead?**
Error due to incompatible changes made in drainer from v2.1.15 to v2.1.16
```
2019/08/20 13:38:10 main.go:26: [fatal] verifying flags error, See 'drainer --help'. component drainer's config file /etc/drainer/drainer.toml contained unknown configuration options: syncer.detect-interval, syncer.to.compression
github.com/pingcap/tidb-binlog/pkg/util.StrictDecodeFile
    /home/jenkins/workspace/release_tidb_2.1/go/src/github.com/pingcap/tidb-binlog/pkg/util/util.go:159
github.com/pingcap/tidb-binlog/drainer.(*Config).configFromFile
    /home/jenkins/workspace/release_tidb_2.1/go/src/github.com/pingcap/tidb-binlog/drainer/config.go:211
github.com/pingcap/tidb-binlog/drainer.(*Config).Parse
    /home/jenkins/workspace/release_tidb_2.1/go/src/github.com/pingcap/tidb-binlog/drainer/config.go:156
main.main
    /home/jenkins/workspace/release_tidb_2.1/go/src/github.com/pingcap/tidb-binlog/cmd/drainer/main.go:25
runtime.main
    /usr/local/go/src/runtime/proc.go:200
runtime.goexit
    /usr/local/go/src/runtime/asm_amd64.s:1337
```

aylei:
What's your installation command?

shinnosuke-okada:
@aylei it was like this:
`helm install http://download.pingcap.org/tidb-drainer-latest.tgz --name tidb-drainer --namespace ${CLUSTER_NAME}`

Sorry, I had a fix already but had to run to get dinner 😆I made a PR just now - please check. It was due to incompatible config format changes from drainer.

aylei:
@shinnosuke-okada Thanks!

aylei:
closed via #793

PR #793, opened 2019-08-20
Fix default tidb-drainer config

shinnosuke-okada:
https://github.com/pingcap/tidb-operator/issues/792

Default tidb-drainer config is adjusted to be compatible with the latest drainer config format.
To be more specific:
detect-interval is moved out from [syncer]
db-type value is changed from pb to file
compression is moved out from [syncer] and renamed to compressor

### Check List 

Tests 

1. Deploy tidb-drainer chart out-of-box to AWS
2. Drainer is successfully deployed and running

Code changes

 - Has Helm charts change

Side effects

Related changes

 - N/A

 ```release-note
NONE
 ```

### DIFF
--- a/charts/tidb-drainer/values.yaml
+++ b/charts/tidb-drainer/values.yaml
@@ -24,7 +24,7 @@ initialCommitTs: 0
 # Refer to https://github.com/pingcap/tidb-binlog/blob/master/cmd/drainer/drainer.toml
 config: |
   detect-interval = 10
-  compressor = "gzip"
+  compressor = ""
   [syncer]
   worker-count = 16
   disable-dispatch = false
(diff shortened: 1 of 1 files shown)


aylei:
@tennix @weekface PTAL

aylei:
@WangXiangUSTC Could you please take a look?

july2993:
start from 2.1.16, pump/drainer will return error when loading some undefined configs.

- detect-interval should be  moved out from [syncer] for all drainer version, i think it's wrongly set here before and will take no any effect for drainer version before 2.1.16
- compressor means Use the specified compressor to compress payload between pump and drainer, I will suggest set it as empty as default. 
- `Syncer.compression` is a wrongly used config by the older version, it's removed since v2.1.9 
and change 'pb' to 'file'(but pb is still accepted by drainer for compatibility)


I will suggest compressor as "" default, rest LGTM

shinnosuke-okada:
@july2993 Addressed your review comment.

To all reviewers: PTAL again. Thanks.

PR #802, opened 2019-08-21
cleanup of bash scripts

kolbe:
This PR resolves several bash mistakes and brings the scripts in-line with best practices, while adding some new features.

Variable expansions are quotes, lowercase variable names are used, tests are cleaned up.

### Check List 

Tests 

I manually tested a number of permutations, including changes to project, removing credentials files, etc.

 ```release-note
NONE
 ```


deploy/gcp/create-service-account.sh
```diff
@@ -7,45 +7,72 @@
...
-  if cat "$cred_file" | awk -F'=' '/GCP_CREDENTIALS/ {print $2}' >/dev/null ; then
-    echo "GCP_CREDENTAILS_PATH already set in $cred_file"
+if ! cd "$(dirname "$0")"; then
+    printf "Could not change to base directory of script." >&2
     exit 1
-  fi
 fi
 
-GCLOUD="gcloud --project $PROJECT"
+if ! [[ -d .terraform ]]; then
```
gregwebs: The way the docs are written now, someone would run this before running `terraform init`. https://pingcap.com/docs/v3.0/tidb-in-kubernetes/deploy/gcp-gke/

gregwebs: This new approach uses `terraform console` which is much more reliable. We can update the code snippet in the docs to have `terraform init` before running this.

deploy/gcp/create-service-account.sh
```diff
@@ -7,45 +7,72 @@
...
+    echo "no .terraform directory, perhaps you need to run ''terraform init''?" >&2
```
gregwebs: All the quoting looked a little odd to me, but no big deal.

gregwebs:
I ran this myself. I already have the service account and it properly recognized that.

PR #809, opened 2019-08-22
deploy/gcp: support zonal and multi-zonal cluster

cofyc:

- support zonal/multi-zonal cluster
- more examples

### Check List 

Tests 

Side effects

Related changes

 ```release-note
deploy/gcp: support zonal and multi-zonal clusters
 ```

### DIFF
--- a/deploy/gcp/variables.tf
+++ b/deploy/gcp/variables.tf
@@ -18,7 +18,7 @@ variable "location" {
 variable "node_locations" {
   description = "The list of zones in which the cluster's nodes should be located. These must be in the same region as the cluster zone for zonal clusters, or in the region of a regional cluster. In a multi-zonal cluster, the number of nodes specified in initial_node_count is created in all specified zones as well as the primary zone. If specified for a regional cluster, nodes will be created in only these zones."
   type        = list(string)
-  default = []
+  default     = []
 }
 
 variable "tidb_version" {
@@ -108,7 +108,7 @@ variable "tikv_image_type" {
 
 variable "tikv_local_ssd_count" {
   description = "TiKV node pool local ssd count (cannot be changed after the node pool is created)"
-  default = 1
+  default     = 1
 }
 
 variable "monitor_instance_type" {
--- a/deploy/gcp/variables.tf
+++ b/deploy/gcp/variables.tf
@@ -10,6 +10,17 @@ variable "GCP_PROJECT" {
   description = "The GCP project in which to create the necessary resources"
 }
 
+variable "location" {
+  description = "The GKE cluster location. If you specify a zone (such as us-central1-a), the cluster will be a zonal cluster with a single cluster master. If you specify a region (such as us-west1), the cluster will be a regional cluster with multiple masters spread across zones in the region. If not specified, the cluster will be a regional cluster in GCP_REGION."
+  type        = string
+}
+
+variable "node_locations" {
+  description = "The list of zones in which the cluster's nodes should be located. These must be in the same region as the cluster zone for zonal clusters, or in the region of a regional cluster. In a multi-zonal cluster, the number of nodes specified in initial_node_count is created in all specified zones as well as the primary zone. If specified for a regional cluster, nodes will be created in only these zones."
+  type        = list(string)
+  default = []
+}
+
 variable "tidb_version" {
   description = "TiDB version"
   default     = "v3.0.1"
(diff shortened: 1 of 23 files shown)


cofyc:
@gregwebs @tennix @jlerche @aylei 
I'd run a few tests manually. However, I'm not terraform expert. Can you help test this when you have time?

Known issues:

- when `var.location` is given a region different than GCE_REGION. VPC/bastion will be created in the wrong region
  - terraform is not able to raise error in 0.12. There are some workaround ways in https://github.com/hashicorp/terraform/issues/15469 we can do in the future.

deploy/gcp/variables.tf
```diff
@@ -24,6 +24,12 @@ variable "tidb_operator_chart_version" {
   default     = ""
 }
 
+variable "operator_helm_values" {
+  description = "Operator helm values"
+  type        = string
+  default     = ""
+}
```
cofyc: declare it in root module to make it configurable

deploy/gcp/variables.tf
```diff
@@ -33,6 +50,12 @@ variable "gke_name" {
   default     = "tidb-cluster"
 }
 
+variable "gke_version" {
+  description = "Kubernetes version to use for the GKE cluster"
+  type        = string
+  default     = "latest"
+}
```
cofyc: declare it in root module to make it configurable

deploy/gcp/variables.tf
```diff
@@ -43,18 +66,6 @@ variable "vpc_name" {
...
-  default = 3
-}
-
-variable "tikv_replica_count" {
-  default = 3
-}
-
-variable "tidb_replica_count" {
-  default = 3
-}
```
cofyc: these three variables are not used anymore

deploy/gcp/variables.tf
```diff
@@ -80,6 +91,26 @@ variable "tikv_instance_type" {}
...
+
+variable "tikv_image_type" {
+  description = "TiKV image type, avaiable: UBUNTU/COS"
+  default     = "COS"
+}
+
+variable "tikv_local_ssd_count" {
+  description = "TiKV node pool local ssd count (cannot be changed after the node pool is created)"
+  default     = 1
+}
```
cofyc: declare these in root module to make it configurable

deploy/modules/gcp/tidb-operator/variables.tf
```diff
@@ -11,14 +11,20 @@ variable "subnetwork_name" {
   description = "The name of the subnetwork in which to place the cluster"
 }
 
-variable "gcp_region" {
-  description = "The GCP region"
-}
-
```
cofyc: replaced by `location`

deploy/modules/gcp/tidb-operator/variables.tf
```diff
@@ -11,14 +11,20 @@ variable "subnetwork_name" {
...
   description = "The GCP project name"
 }
 
+variable "location" {
+  description = "The GKE cluster location. If you specify a zone (such as us-central1-a), the cluster will be a zonal cluster with a single cluster master. If you specify a region (such as us-west1), the cluster will be a regional cluster with multiple masters spread across zones in the region."
+  type        = string
+}
+
+variable "node_locations" {
+  description = "The list of zones in which the cluster's nodes should be located. These must be in the same region as the cluster zone for zonal clusters, or in the region of a regional cluster. In a multi-zonal cluster, the number of nodes specified in initial_node_count is created in all specified zones as well as the primary zone. If specified for a regional cluster, nodes will be created in only these zones."
```
gregwebs: What is the purpose of this for a single-zone cluster? Doesn't the above location variable already specify this properly?

gregwebs: oh, I see, there are zonal clusters that are multi-zone clusters. I wonder why not just use the regional cluster though?

cofyc: yes, if you need to run a multi-zone cluster, it's better to use a regional cluster. However, if users want a small cluster for testing or for low network latency between work nodes, they can use a zonal cluster and run nodes in a single zone. 

`node_locations` can also be used to choose node locations for the regional cluster too (at least two locations for a regional cluster).

Currently, in GCP, some types of machines are only available in a few zones, e.g. in region `us-central1` [c2-standard machines](https://cloud.google.com/compute/docs/machine-types#c2_machine_types) are only available in `us-central1-b` and `us-central1-c` zones.

If users want to use these compute-optimized machines, they must specify `node_locations` to deploy nodes in these zones.

cofyc:
@gregwebs @aylei 
I pushed a [new commit](https://github.com/pingcap/tidb-operator/pull/809/commits/c5ed1a496a9d673d53796e2e9e8bf128a6934454) to set default value for `var.location`. PTAL.

PR #814, opened 2019-08-22
Monitoring: Integrated kubernetes cluster monitoring in public cloud platform

qiffang:
Integrated k8s cluster monitor to terraform for public cloud platform. It is included as an optional module, we will install prometheus operator and k8s monitor as default action.
You can update variables.tf file.
All major changes is in terraform files. It will apply monitoring files after you execute the command `terraform apply`.
### Check List 
Tests

- Manual test
- There are monitoring pods in monitoring namespaces

Code changes

 - Has terraform files change


deploy/modules/aliyun/kubernetes-monitor/main.tf
```diff
@@ -0,0 +1,38 @@
...
+resource "null_resource" "setup-env" {
+  depends_on = [data.template_file.kubeconfig_filename]
+
+  provisioner "local-exec" {
+    working_dir = path.cwd
+    command     = <<EOS
+mkdir monitor
+echo "${local_file.kubeconfig.sensitive_content}" > monitor/config.yaml
+# Kubernetes cluster monitor
+if var.install_prometheus_operator; then
```
gregwebs: this looks like it should be template inserted. All bash should have `set -euo pipefail` so errors are not ignored.

deploy/modules/gcp/tidb-operator/main.tf
```diff
@@ -100,7 +100,6 @@ kubectl apply -f https://raw.githubusercontent.com/pingcap/tidb-operator/${var.t
 kubectl apply -f https://raw.githubusercontent.com/pingcap/tidb-operator/${var.tidb_operator_version}/manifests/tiller-rbac.yaml
 kubectl apply -k manifests/local-ssd
 kubectl apply -f manifests/gke/persistent-disk.yaml
-
```
gregwebs: why remove this space?

deploy/modules/gcp/kubernetes-monitor/main.tf
```diff
@@ -0,0 +1,34 @@
...
+if var.install_prometheus_operator; then
+    wget https://raw.githubusercontent.com/pingcap/monitoring/master/k8s-cluster-monitor/manifests/archive/prometheus-operator.tar.gz
+    tar -zxvf prometheus-operator.tar.gz -C monitor/
+    kubectl apply -f monitor/manifests/prometheus-operator
+fi
+
+if var.install_kubernetes_monitor; then
+    wget https://raw.githubusercontent.com/pingcap/monitoring/master/k8s-cluster-monitor/manifests/archive/prometheus.tar.gz
+    tar -zxvf prometheus.tar.gz -C monitor/
+    sed -i'.bak' 's/local-storage/ebs-gp2/g' monitor/manifests/prometheus/grafana-pvc.yaml
```
gregwebs: ebs-gp2 doesn't seem right here. We should pass in a variable for the storage class and then this module should work for all cloud providers.

gregwebs:
I don't think this is a good addition because there is no ability to uninstall.

qiffang:
> I don't think this is a good addition because there is no ability to uninstall.

What's your advice? 
Actually, EKS is installed in operator.tf files, and I still think it is ok by adding monitor in operator.tf like [PR](https://github.com/pingcap/tidb-operator/pull/777)

Issue #347, opened 2019-03-21
lightning backup import


gregwebs:
The lightning importer is becoming a mature tool now. We should consider supporting it in our backup chart.

weekface:
@gregwebs we may not support lightning recently, can we close this issue?

gregwebs:
I would really like lightning support so I can restore quickly from a large backup.

tennix:
I've heard that lightning will be refactored to integrate into TiDB itself. So I would suggest we wait until it has a clear roadmap. Before that, we can just provide some YAML files to avoid writing CRD controller.

gregwebs:
If lightning is available to use directly in TiDB that is a non-optimal deployment to increase user convenience. Its current form should still be available to use the way it is now, and that is the way that I would always use it in the cloud.

PR #817, opened 2019-08-23
charts: Add tidb-lightning support

tennix:

Integrate tidb-lightning on k8s, this closes https://github.com/pingcap/tidb-operator/issues/347

Add tidb-lightning chart. The tikv-importer is inside the tidb-cluster chart and tidb-lightning is in a separate chart.

### Check List 

Tests 

Manual test (add detailed scripts or steps below)
Both the local and remote (S3) data sources are tested. The compressed sysbench dataset is 15GB, and the uncompressed dataset is 31GB. The restore time is `18m` while the sysbench prepare average time is `1h`

Code changes

 - Has Helm charts change

Side effects

 No

Related changes

 ```release-note
Add tidb-lightning support
 ```

### DIFF
--- a/charts/tidb-cluster/templates/_helpers.tpl
+++ b/charts/tidb-cluster/templates/_helpers.tpl
@@ -110,3 +110,18 @@ drainer-config: |-
 {{- define "drainer-configmap.data-digest" -}}
 {{ include "drainer-configmap.data" . | sha256sum | trunc 8 }}
 {{- end -}}
+
+
+{{/*
+Encapsulate tikv-importer configmap data for consistent digest calculation
+*/}}
+{{- define "importer-configmap.data" -}}
+config-file: |-
+    {{- if .Values.importer.config }}
+{{ .Values.importer.config | indent 2 }}
+    {{- end -}}
+{{- end -}}
+
+{{- define "importer-configmap.data-digest" -}}
+{{ include "importer-configmap.data" . | sha256sum | trunc 8 }}
+{{- end -}}
--- a/charts/tidb-cluster/templates/_helpers.tpl
+++ b/charts/tidb-cluster/templates/_helpers.tpl
@@ -111,7 +111,6 @@ drainer-config: |-
 {{ include "drainer-configmap.data" . | sha256sum | trunc 8 }}
 {{- end -}}
 
-
 {{/*
 Encapsulate tikv-importer configmap data for consistent digest calculation
 */}}
--- a/charts/tidb-cluster/templates/tikv-importer-service.yaml
+++ b/charts/tidb-cluster/templates/tikv-importer-service.yaml
@@ -1,0 +1,21 @@
+{{- if .Values.importer.create }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "cluster.name" . }}-importer
+  labels:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: importer
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  clusterIP: None
+  ports:
+  - name: importer
+    port: 20170
+  selector:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: importer
+{{- end }}
--- a/charts/tidb-cluster/templates/tikv-importer-service.yaml
+++ b/charts/tidb-cluster/templates/tikv-importer-service.yaml
@@ -13,7 +13,7 @@ spec:
   clusterIP: None
   ports:
   - name: importer
-    port: 20170
+    port: 8287
   selector:
     app.kubernetes.io/name: {{ template "chart.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
--- a/charts/tidb-cluster/templates/tikv-importer-statefulset.yaml
+++ b/charts/tidb-cluster/templates/tikv-importer-statefulset.yaml
@@ -14,7 +14,6 @@ spec:
     matchLabels:
       app.kubernetes.io/name: {{ template "chart.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
-      app.kubernetes.io/managed-by: tidb-operator
       app.kubernetes.io/component: importer
   serviceName: {{ template "cluster.name" . }}-importer
   replicas: {{ .Values.importer.replicas }}
@@ -23,20 +22,19 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/path: "/metrics"
-        prometheus.io/port: "20170"
+        prometheus.io/port: "9091"
       labels:
         app.kubernetes.io/name: {{ template "chart.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-        app.kubernetes.io/managed-by: tidb-operator
         app.kubernetes.io/component: importer
     spec:
     {{- if .Values.importer.affinity }}
       affinity:
-{{ toYaml .Values.importer.affinity | indent 8 }}
+{{ toYaml .Values.importer.affinity | indent 6 }}
     {{- end }}
     {{- if .Values.importer.tolerations }}
       tolerations:
-{{ toYaml .Values.importer.tolerations | indent 8 }}
+{{ toYaml .Values.importer.tolerations | indent 6 }}
     {{- end }}
       containers:
       - name: importer
@@ -45,8 +43,8 @@ spec:
         command:
         - /tikv-importer
         # tikv-importer does not support domain name: https://github.com/tikv/importer/issues/16
-        # - --addr=${MY_POD_NAME}.tikv-importer:20170
-        - --addr=$(MY_POD_IP):20170
+        # - --addr=${MY_POD_NAME}.tikv-importer:8287
+        - --addr=$(MY_POD_IP):8287
         - --config=/etc/tikv-importer/tikv-importer.toml
         - --import-dir=/var/lib/tikv-importer
         env:
@@ -65,6 +63,13 @@ spec:
           mountPath: /var/lib/tikv-importer
         - name: config
           mountPath: /etc/tikv-importer
+        {{- if .Values.importer.resources }}
+        resources:
+{{ toYaml .Values.importer.resources | indent 10 }}
+        {{- end }}
+      - name: pushgateway
+        image: {{ .Values.importer.pushgatewayImage }}
+        imagePullPolicy: {{ .Values.importer.pushgatewayImagePullPolicy | default "IfNotPresent" }}
       volumes:
       - name: config
         configMap:
--- a/charts/tidb-cluster/templates/tikv-importer-statefulset.yaml
+++ b/charts/tidb-cluster/templates/tikv-importer-statefulset.yaml
@@ -1,0 +1,84 @@
+{{- if .Values.importer.create }}
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ template "cluster.name" . }}-importer
+  labels:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: importer
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ template "chart.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+      app.kubernetes.io/managed-by: tidb-operator
+      app.kubernetes.io/component: importer
+  serviceName: {{ template "cluster.name" . }}-importer
+  replicas: {{ .Values.importer.replicas }}
+  template:
+    metadata:
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "20170"
+      labels:
+        app.kubernetes.io/name: {{ template "chart.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/managed-by: tidb-operator
+        app.kubernetes.io/component: importer
+    spec:
+    {{- if .Values.importer.affinity }}
+      affinity:
+{{ toYaml .Values.importer.affinity | indent 8 }}
+    {{- end }}
+    {{- if .Values.importer.tolerations }}
+      tolerations:
+{{ toYaml .Values.importer.tolerations | indent 8 }}
+    {{- end }}
+      containers:
+      - name: importer
+        image: {{ .Values.importer.image }}
+        imagePullPolicy: {{ .Values.importer.imagePullPolicy | default "IfNotPresent"}}
+        command:
+        - /tikv-importer
+        # tikv-importer does not support domain name: https://github.com/tikv/importer/issues/16
+        # - --addr=${MY_POD_NAME}.tikv-importer:20170
+        - --addr=$(MY_POD_IP):20170
+        - --config=/etc/tikv-importer/tikv-importer.toml
+        - --import-dir=/var/lib/tikv-importer
+        env:
+        - name: MY_POD_NAME
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.name
+        - name: MY_POD_IP
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: status.podIP
+        volumeMounts:
+        - name: data
+          mountPath: /var/lib/tikv-importer
+        - name: config
+          mountPath: /etc/tikv-importer
+      volumes:
+      - name: config
+        configMap:
+          name: {{ template "cluster.name" . }}-importer-{{ template "importer-configmap.data-digest" . }}
+          items:
+          - key: config-file
+            path: tikv-importer.toml
+  volumeClaimTemplates:
+  - metadata:
+      name: data
+    spec:
+      accessModes: [ "ReadWriteOnce" ]
+      storageClassName: {{ .Values.importer.storageClassName }}
+      resources:
+        requests:
+          storage: {{ .Values.importer.storage }}
+{{- end }}
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -49,7 +49,7 @@ enableConfigMapRollout: true
 
 pd:
   # Please refer to https://github.com/pingcap/pd/blob/master/conf/config.toml for the default
-  # pd configurations (change to the tags of your pd version), 
+  # pd configurations (change to the tags of your pd version),
   # just follow the format in the file and configure in the 'config' section
   # as below if you want to customize any configuration.
   # Please refer to https://pingcap.com/docs-cn/v3.0/reference/configuration/pd-server/configuration-file/
@@ -59,7 +59,7 @@ pd:
     level = "info"
     [replication]
     location-labels = ["region", "zone", "rack", "host"]
-  
+
   replicas: 3
   image: pingcap/pd:v3.0.1
   # storageClassName is a StorageClass provides a way for administrators to describe the "classes" of storage they offer.
@@ -553,6 +553,17 @@ scheduledBackup:
   # You can create the secret by:
   # kubectl create secret generic s3-backup-secret --from-literal=access_key=<access-key> --from-literal=secret_key=<secret-key>
   # secretName: s3-backup-secret
+
+importer:
+  create: false
+  image: pingcap/tidb-lightning:v3.0.1
+  imagePullPolicy: IfNotPresent
+  storageClassName: local-storage
+  storage: 20Gi
+  affinity: {}
+  tolerations: []
+  config: |
+    log-level = "info"
 
 metaInstance: "{{ $labels.instance }}"
 metaType: "{{ $labels.type }}"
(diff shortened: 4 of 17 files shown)


charts/tidb-lightning/templates/job.yaml
```diff
@@ -0,0 +1,98 @@
...
+          --server-mode=false \
+          --tidb-user=root \
+          --tidb-host={{ .Values.targetTidbCluster.name }}-tidb.{{ .Values.targetTidbCluster.namespace }} \
+          {{ if and .Values.dataSource.local.hostPath .Values.dataSource.local.nodeName -}}
+          --d={{ .Values.dataSource.local.hostPath }} \
+          {{ else -}}
+          --d=${data_dir} \
+          {{ end -}}
+          --config=/etc/tidb-lightning/tidb-lightning.toml
+          if [ $? != 0 ]; then
```
tennix: Currently, there is an error even if the restore succeeds.
```
[2019/08/22 17:58:51.675 +00:00] [INFO] [restore.go:921] ["skip full compaction"]                                                                             
[2019/08/22 17:58:51.675 +00:00] [INFO] [restore.go:435] ["everything imported, stopping periodic actions"]                                                   
[2019/08/22 17:58:51.754 +00:00] [INFO] [restore.go:1073] ["skip clean checkpoints"]                                                                          
[2019/08/22 17:58:51.754 +00:00] [INFO] [restore.go:259] ["the whole procedure completed"] [takeTime=18m48.515158344s] []                                     
[2019/08/22 17:58:51.754 +00:00] [INFO] [main.go:61] ["tidb lightning exit"]
sync log failed sync /dev/stdout: invalid argument
[2019/08/22 17:58:51. tidb-lightning exits abnormally, please exec into my container to do manual intervention
```
So the restore job will not be completed, users have to check the log to verify the status of the restore job.

charts/tidb-lightning/templates/job.yaml
```diff
@@ -0,0 +1,98 @@
...
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: tidb-lightning
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "tidb-lightning.name" . }}
+        app.kubernetes.io/instance:  {{ .Release.Name }}
```
aylei: This label should be consistent with the source cluster to get the metrics collected.

Alternatively, we add extra labels to tag the cluster name and update the prometheus rule to use these new labels.

charts/tidb-lightning/templates/job.yaml
```diff
@@ -0,0 +1,98 @@
...
+        - name: data
+          mountPath: /data
+      {{ end -}}
+      restartPolicy: Never      # if lightning fails, manual intervention is required so no restart
+      containers:
+      - name: tidb-lightning
+        image: {{ .Values.image }}
+        imagePullPolicy: {{ .Values.imagePullPolicy | default "IfNotPresent" }}
+        command:
+        - /bin/sh
```
aylei: Can we move these to a dedicated start script template for better maintainability?

charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -110,3 +110,18 @@ drainer-config: |-
...
+{{- define "importer-configmap.data" -}}
+config-file: |-
+    {{- if .Values.importer.config }}
+{{ .Values.importer.config | indent 2 }}
+    {{- end -}}
+{{- end -}}
+
+{{- define "importer-configmap.data-digest" -}}
+{{ include "importer-configmap.data" . | sha256sum | trunc 8 }}
+{{- end -}}
```
weekface: newline

charts/tidb-cluster/templates/tikv-importer-statefulset.yaml
```diff
@@ -0,0 +1,84 @@
...
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: importer
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ template "chart.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+      app.kubernetes.io/managed-by: tidb-operator
```
weekface: managed-by: tidb-operator?

charts/tidb-cluster/templates/tikv-importer-statefulset.yaml
```diff
@@ -0,0 +1,84 @@
...
+  template:
+    metadata:
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "20170"
+      labels:
+        app.kubernetes.io/name: {{ template "chart.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/managed-by: tidb-operator
```
weekface: ditto

charts/tidb-lightning/secret.yaml
```diff
@@ -0,0 +1,31 @@
...
+    endpoint = <ceph-object-store-endpoint>
+
+    [gcs]
+    type = google cloud storage
+    client_id =
+    client_secret =
+    token = {"AccessToken":"xxxx.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx","RefreshToken":"x/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx_xxxxxxxxx","Expiry":"2014-07-17T20:49:14.929208288+01:00","Extra":null}
+    project_number = 12345678
+    object_acl = private
+    bucket_acl = private
```
aylei: Why we need this file?

charts/tidb-lightning/values.yaml
```diff
@@ -0,0 +1,41 @@
+# Default values for tidb-lightning.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image: pingcap/tidb-lightning:v3.0.1
+imagePullPolicy: IfNotPresent
+service:
+  type: NodePort
```
aylei: Does lightning need external accessing?

charts/tidb-lightning/templates/job.yaml
aylei: Can we check this in the script?

charts/tidb-lightning/templates/job.yaml
```diff
@@ -0,0 +1,98 @@
...
+      initContainers:
+      - name: data-retriever
+        image: {{ .Values.dataSource.remote.rcloneImage }}
+        command:
+        - /bin/sh
+        - -c
+        - |
+          set -euo pipefail
+          filename=$(basename {{ .Values.dataSource.remote.path }})
+          rclone --config /etc/rclone/rclone.conf copy -P {{ .Values.dataSource.remote.path }} /data
```
aylei: The data should be downloaded to the directory backed by a PV so that to not full the system disk of the node.

charts/tidb-lightning/templates/job.yaml
```diff
@@ -0,0 +1,98 @@
...
+      labels:
+        app.kubernetes.io/name: {{ include "tidb-lightning.name" . }}
+        app.kubernetes.io/instance:  {{ .Release.Name }}
+        app.kubernetes.io/component: tidb-lightning
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: "/metrics"
+        prometheus.io/port: "8289"
+    spec:
+      {{ if .Values.dataSource.local.nodeName -}}
```
aylei: The condition of whether using hostPath mode should be consistent

(subsequent one uses `{{ if and .Values.dataSource.local.hostPath .Values.dataSource.local.nodeName }}`)

charts/tidb-lightning/templates/job.yaml
gregwebs: Can we fix tidb-lightning?

charts/tidb-lightning/templates/job.yaml
```diff
@@ -0,0 +1,98 @@
...
+      restartPolicy: Never      # if lightning fails, manual intervention is required so no restart
+      containers:
+      - name: tidb-lightning
+        image: {{ .Values.image }}
+        imagePullPolicy: {{ .Values.imagePullPolicy | default "IfNotPresent" }}
+        command:
+        - /bin/sh
+        - -c
+        - |
+          data_dir=$(dirname $(find /var/lib/tidb-lightning -name metadata 2>/dev/null) 2>/dev/null)
```
gregwebs: Why would `dirname` need `2>/dev/null` ?
We should always have `set -euo pipefail` for our shell scripts.
Although it seems this is only a script to find the metadata and I think we should already know where it is?

charts/tidb-lightning/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+targetTidbCluster:
+  name: tidb-cluster
+  namespace: tidb
+
+resources: {}
+  # limits:
+  #  cpu: 100m
+  #  memory: 128Mi
+  # requests:
```
gregwebs: Shouldn't we have a minimum here?

gregwebs:
We need to know how to estimate resource requirements.

charts/tidb-lightning/secret.yaml
```diff
@@ -0,0 +1,31 @@
...
+    secret_access_key = <my-secret-key>
+    region = us-east-1
+
+    [ceph]
+    type = s3
+    provider = Ceph
+    env_auth = false
+    access_key_id = <my-access-key
+    secret_access_key = <my-secret-key>
+    endpoint = <ceph-object-store-endpoint>
```
onlymellb: add `region = :default-placement` for ceph

charts/tidb-lightning/secret.yaml
tennix: This is a secret example to tell user how to create the secret.

charts/tidb-lightning/templates/job.yaml
tennix: The remote data is archived from a tarball file, and the extracted directory is unknown.

charts/tidb-lightning/templates/job.yaml
tennix: From @kennytm, it's a known issue https://github.com/uber-go/zap/issues/328

charts/tidb-lightning/values.yaml
tennix: There is a webpage to show current job progress.

charts/tidb-cluster/templates/tikv-importer-service.yaml
```diff
@@ -0,0 +1,21 @@
...
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: importer
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  clusterIP: None
+  ports:
+  - name: importer
+    port: 20170
```
kennytm: We use port 8287 for importer.

charts/tidb-cluster/values.yaml
```diff
@@ -554,6 +554,17 @@ scheduledBackup:
   # kubectl create secret generic s3-backup-secret --from-literal=access_key=<access-key> --from-literal=secret_key=<secret-key>
   # secretName: s3-backup-secret
 
+importer:
+  create: false
+  image: pingcap/tidb-lightning:v3.0.1
+  imagePullPolicy: IfNotPresent
+  storageClassName: local-storage
+  storage: 20Gi
```
kennytm: 20 GB is likely not enough. It should be big enough to store the top 8 largest tables.

charts/tidb-lightning/templates/job.yaml
kennytm: Filed pingcap/tidb-lightning#230 to ignore the sync log error.

charts/tidb-cluster/templates/tikv-importer-service.yaml
tennix: Then what is this port for? https://github.com/tikv/importer/blob/master/etc/tikv-importer.toml#L10

charts/tidb-lightning/templates/job.yaml
tennix: Yes, the `/data` directory is under the PV.

charts/tidb-cluster/templates/tikv-importer-service.yaml
kennytm: The sample config file is not yet updated 🤣

charts/tidb-lightning/secret.yaml
tennix: I don't think this field is required.

charts/tidb-lightning/values.yaml
tennix: @kennytm Seems this webpage is broken.

charts/tidb-lightning/templates/job.yaml
```diff
@@ -0,0 +1,110 @@
...
+          filename=$(basename {{ .Values.dataSource.remote.path }})
+          rclone --config /etc/rclone/rclone.conf copy -P {{ .Values.dataSource.remote.path }} /data
+          cd /data && tar xzvf ${filename}
+        volumeMounts:
+        - name: credentials
+          mountPath: /etc/rclone
+        - name: data
+          mountPath: /data
+      {{ end -}}
+      restartPolicy: Never      # if lightning fails, manual intervention is required so no restart
```
tennix: From Kubernetes [Job documentation](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#handling-pod-and-container-failures):
> Note that even if you specify `.spec.parallelism = 1` and `.spec.completions = 1` and `.spec.template.spec.restartPolicy = "Never"`, the same program may sometimes be started twice.

@kennytm Are there any ways to ensure there is only one tidb-lightning from tidb-lightning itself?

charts/tidb-lightning/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+imagePullPolicy: IfNotPresent
+service:
+  type: NodePort
+
+dataSource:
+  local:
+    # nodeName: kind-worker3
+    hostPath: /data/export-20190820
+  remote:
+    rcloneImage: tynor88/rclone
```
tennix: Here we use the latest `rclone` image because other versions don't support `-P` option which shows the copying progress.

charts/tidb-lightning/secret.yaml
onlymellb: No, this field is required. Refer rook/rook#2945

charts/tidb-lightning/templates/job.yaml
gregwebs: Why is the directory unknown? Is it named after data given by the user?

Why would `dirname` need `2>/dev/null` ?

It also looks like this should only be ran if certain options are set below.

tennix: We don't know if the tarball filename is `dump_dir.tar.gz`, because the backup tarball file is created by users, they can name it whatever they want. So we can not extract the dir name from the tarball file. Also, the dumped files may be broken or even not mydumper files. So this line may fail in different ways.

charts/tidb-lightning/values.yaml
kennytm: (Tracking issue: pingcap/tidb-lightning#213)

tennix: @kennytm What does this mean? Did you mean it gets fixed in that PR? I've tried v3.0.2 which contains this PR, but still got the same broken page.

charts/tidb-lightning/templates/job.yaml
kennytm: @tennix No `tidb-lightning` doesn't register itself on anything. Furthermore, by design multiple `tidb-lightning` can run in parallel. 

Perhaps we could investigate if `tidb-lightning` could be changed into a (long-running) StatefulSets with `--server-mode` and the schedule import tasks via the HTTP interface?

charts/tidb-lightning/values.yaml
kennytm: Sorry just a typo.

charts/tidb-lightning/templates/job.yaml
tennix: For one time task I think BatchJob is more suitable. Besides, the server mode makes it a bit tedious to set up.

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
```diff
@@ -0,0 +1,22 @@
+data_dir=$(dirname $(find /var/lib/tidb-lightning -name metadata 2>/dev/null) 2>/dev/null)
+if [ -z $data_dir]; then
+    echo "No mydumper files are found, please exec into my container to diagnose"
+    tail -f /dev/null
```
gregwebs: The echo should be to stderr `>&2`.

Instead of holding this container open by default, maybe we can have an environment variable or some kind of debug switch to turn that behavior on?

tennix: The backup data is usually very large, it's costly to restart from scratch.
With the container open, users can exec into the container to diagnose and run the tidb-lightning manually.

charts/tidb-lightning/templates/job.yaml
kennytm: Okay.

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
gregwebs: Yes, but how will we know when it fails?

For large data in the cloud we need to stream the data directly from a storage bucket (but download the metadata first).

tennix: I don't think tidb-lightning supports streaming data from cloud storage right now.

tennix: I've added an option to allow fail fast, the job status will indicate the restore status when `failFast` is enabled.

tennix:
@qiffang The tidb-lightning dashboard needs to be included in the `tidb-monitor-initializer` docker image.

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
gregwebs: failFast is good, but still how will someone know that lightning has failed? It seems like the only thing they can do is tail the log.
And then how does one fix the situation?
If we store all data in PVC that are retained, then shouldn't it be easy to launch a new job with a shell for the user to complete the process?

tennix: If `failFast` is enabled, the job fails means the lightning fails, no need to tail the log (After https://github.com/pingcap/tidb-lightning/pull/230 is merged).
The recommended workflow is:
1. Deploy tidb-lightning with `failFast` enabled.
2. If the job fails (indicated by job status), delete the failed job with `kubectl delete job -n <namespace> <release-name-tidb-lightning>`
3. Use helm and kubectl to launch the job with `failFast` disabled by `helm template pingcap/tidb-lightning --name <release-name> | kubectl apply -n <namespace> -f -`. This will keep the pod running when fails.

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
```diff
@@ -0,0 +1,31 @@
...
+{{ end }}
+/tidb-lightning \
+    --pd-urls={{ .Values.targetTidbCluster.name }}-pd.{{ .Release.Namespace }}:2379 \
+    --status-addr=0.0.0.0:8289 \
+    --importer={{ .Values.targetTidbCluster.name }}-importer.{{ .Release.Namespace }}:8287 \
+    --server-mode=false \
+    --tidb-user={{ .Values.targetTidbCluster.user | default "root" }} \
+    --tidb-host={{ .Values.targetTidbCluster.name }}-tidb.{{ .Release.Namespace }} \
+    --d=${data_dir} \
+    --config=/etc/tidb-lightning/tidb-lightning.toml
```
zyguan: For some old version of tidb-lightning, `tikv_gc_life_time` cannot be restored even when the procedure completed successfully. Should we handle this explicitly in the script?

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
tennix: I leave the default `failFast` to `false` to make it handy for users to handle when there are errors.
If the first try fails, the second running should quickly get the debug shell. The failure should happen instantly. So no need to add "debug statements and then sleep", the current setup should be enough.

charts/tidb-lightning/templates/persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
```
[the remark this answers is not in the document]
tennix: Yeah, you are right, this is not needed when using local data source. I'll fix it in https://github.com/pingcap/tidb-operator/pull/827

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
tennix: Which versions of tidb-lightning have this issue? If only too old versions, then I don't think it's worth to do it.

zyguan: Actually it has been fixed only recently by https://github.com/pingcap/tidb-lightning/pull/218 . According to the PR, it's a concurrency problem. However, I remember that I can almost always reproduce it by tidb-ansible, but I failed to do it in k8s. Maybe it's because of difference of config.

PR #820, opened 2019-08-23
WIP: deploy/gcp: able to refresh helm resource in refresh state

### LINKED ISSUE #815: deploy/gcp: helm resources cannot be refreshed
## Bug Report

After resources are created with `terraform apply -var-file small.tf` command.

`terraform destory -var-file small.tf` will fail because it cannot refresh `helm_release` resources.

I guess the problem is `helm_release.tidb-cluster` use the output value `kubeconfig_path` of `tidb-operator` module. However, in refreshing this value is nil.

BTW, `terraform destory -var-file small.tf -refresh=false` will work.

cc @gregwebs @jlerche

cofyc:

fixes #815

### Check List 

Tests 

 ```release-note
NONE
 ```

### DIFF
--- a/deploy/gcp/tidbclusters.tf
+++ b/deploy/gcp/tidbclusters.tf
@@ -8,7 +8,7 @@ provider "helm" {
   insecure       = true
   install_tiller = false
   kubernetes {
-    config_path = local.kubeconfig
+    config_path = data.local_file.kubeconfig.filename
   }
 }
 module "default-tidb-cluster" {
@@ -21,7 +21,7 @@ module "default-tidb-cluster" {
   gke_cluster_name           = module.tidb-operator.gke_cluster_name
   cluster_name               = var.default_tidb_cluster_name
   cluster_version            = var.tidb_version
-  kubeconfig_path            = local.kubeconfig
+  kubeconfig_path            = data.local_file.kubeconfig.filename
   tidb_cluster_chart_version = coalesce(var.tidb_operator_chart_version, var.tidb_operator_version)
   pd_instance_type           = var.pd_instance_type
   tikv_instance_type         = var.tikv_instance_type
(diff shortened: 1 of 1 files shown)


deploy/gcp/tidbclusters.tf
```diff
@@ -1,19 +1,14 @@
 data "local_file" "kubeconfig" {
   depends_on = [module.tidb-operator.cluster_id]
-  filename   = module.tidb-operator.kubeconfig_path
-}
-resource "local_file" "kubeconfig" {
-  depends_on = [module.tidb-operator.cluster_id]
-  content    = data.local_file.kubeconfig.content
-  filename   = module.tidb-operator.kubeconfig_path
+  filename   = local.kubeconfig
```
cofyc: resource `local_file.kubeconfig` is not needed because it simply creates the same file again with the contents in it

deploy/gcp/tidbclusters.tf
```diff
@@ -26,7 +21,7 @@ module "default-tidb-cluster" {
   gke_cluster_name           = module.tidb-operator.gke_cluster_name
   cluster_name               = var.default_tidb_cluster_name
   cluster_version            = var.tidb_version
-  kubeconfig_path            = module.tidb-operator.kubeconfig_path
+  kubeconfig_path            = data.local_file.kubeconfig.filename
```
cofyc: in helm provider, and this module we can reference data source `local_file.kubeconfig` which presents kubeconfig local file data source in root module.

deploy/gcp/tidbclusters.tf
```diff
@@ -26,7 +21,7 @@ module "default-tidb-cluster" {
   gke_cluster_name           = module.tidb-operator.gke_cluster_name
```
cofyc: in my testing these output variables are not available too, I guess we need to use variables in root module.

cofyc: it seems we should follow this pattern: https://www.terraform.io/docs/modules/composition.html#dependency-inversion

cofyc:
I decided to this in https://github.com/pingcap/tidb-operator/pull/819 too, these are all issues because of these two problems:

- no `depends_on` keyword in the module, we need to use output variables to setup dependency relationship (tracking issue in terraform: https://github.com/hashicorp/terraform/issues/10462)
- however, module output variables are not available in the fresh state

PR #824, opened 2019-08-26
Update mount options in local volume provisioner

cofyc:

- `defaults` includes `rw,data=ordered`
- `noatime` and `nodelalloc` are recommneded in https://pingcap.com/docs/v3.0/how-to/deploy/orchestrated/ansible/
- share mount options by using a variable

### Check List 

Tests 

 ```release-note
NONE
 ```


cofyc:
I did a simple fio test, the difference is trivial. You can run some more benchmarks.

PR #828, opened 2019-08-27
add recommendations options to mydumper

weekface:

add recommendations options to `mydumper`:

https://pingcap.com/docs/v3.0/how-to/maintain/backup-and-restore/#best-practices-of-full-backup-and-restoration-using-mydumper-loader

- `-t 16`: means 16 threads are used to export the data.
- `-F 64`: means a table is partitioned into chunks and one chunk is 64MB.
- `--skip-tz-utc`: the purpose of adding this parameter is to ignore the inconsistency of time zone setting between MySQL and the data exporting machine and to disable automatic conversion.

### Check List 

Tests 

Code changes

 - Has Helm charts change

Side effects

Related changes

 ```release-note
add recommendations options to mydumper: `-t 16 -F 64 --skip-tz-utc`
 ```

### DIFF
--- a/charts/tidb-backup/values.yaml
+++ b/charts/tidb-backup/values.yaml
@@ -42,7 +42,7 @@ storage:
   size: 100Gi
 
 # backupOptions is the options of mydumper https://github.com/maxbube/mydumper/blob/master/docs/mydumper_usage.rst#options
-backupOptions: "--verbose=3"
+backupOptions: "-t 16 -F 256 --skip-tz-utc --verbose=3"
 # Set the tidb_snapshot to be used for the backup
 # Use `show master status` to get the ts:
 #   MySQL [(none)]> show master status;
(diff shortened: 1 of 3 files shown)


charts/tidb-backup/values.yaml
```diff
@@ -42,7 +42,7 @@ storage:
   size: 100Gi
 
 # backupOptions is the options of mydumper https://github.com/maxbube/mydumper/blob/master/docs/mydumper_usage.rst#options
-backupOptions: "--verbose=3"
+backupOptions: "-t 16 -F 256 --skip-tz-utc --verbose=3"
```
tennix: Add simple descriptions for these default options.
- `-t` is thread count, larger thread count will speed up the backup, but may impact the performance of the upstream TiDB.
- `-F` is the chunk size, a big table is partitioned into many chunks.

Other useful options are `-B` for database, and `-T` for tables. See https://github.com/maxbube/mydumper/blob/master/docs/mydumper_usage.rst#options for more options.

weekface: done

PR #836, opened 2019-08-27
fix ad-hoc full backup failure due to incorrect claimName

DanielZhangQD:
ad-hoc backup will fail if use default backup name which is not rendered during helm install in backup-job.yaml.
Make helm render the backup name.
### Check List 

Tests 

   - ad-hoc backup with the default name
   - ad-hoc backup with the specified name

Code changes

 - Has Helm charts change

Related changes

 ```release-note
The ad-hoc backup will fail if use default backup name which is not rendered during helm install in backup-job.yaml.
 ```

### DIFF
--- a/charts/tidb-backup/templates/backup-job.yaml
+++ b/charts/tidb-backup/templates/backup-job.yaml
@@ -45,7 +45,7 @@ spec:
       {{- end }}
         env:
         - name: BACKUP_NAME
-          value: {{ .Values.name | quote }}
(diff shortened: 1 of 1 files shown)


onlymellb:
The `restore.yaml` file also needs to be changed

weekface:
This is a bug, add a release note.

DanielZhangQD:
> The `restore.yaml` file also needs to be changed

Users need to specify the name during the restore, the auto-generated name will not be used.

PR #851, opened 2019-08-29
Set cache binlog count for drainer to reduce memory footprint

aylei:

According to @july2993 , the default cache binlog count of drainer is 64k, which may cause the drainer OOM if the size of each binlog record is large. 

Make the cache binlog count configurable and set it default to 16
### Check List 

Tests 

 - E2E test

Code changes

 - Has Helm charts change

Side effects

Related changes

 ```release-note
Set cache binlog count for drainer to reduce the memory footprint.
 ```


aylei:
@weekface @tennix @july2993 PTAL

PR #854, opened 2019-08-30
charts: allow cross namespace metrics scraping

tennix:

Allow Prometheus to scrape cross namespace pods. The tidb-lightning chart may be deployed in the upstream tidb cluster to use the adhoc backup PVC. So Prometheus needs to scrape metrics cross namespace.

The rbac rule of Prometheus changed to use cluster role and cluster role binding in order scrape cross namespace.

### Check List 

Tests 

Manual test

Code changes

 - Has Helm charts change

Side effects

None

Related changes

 - Need to cherry-pick to the release branch release-1.1

 ```release-note
None
 ```

### DIFF
--- a/charts/tidb-cluster/templates/config/_prometheus-config.tpl
+++ b/charts/tidb-cluster/templates/config/_prometheus-config.tpl
@@ -14,9 +14,6 @@ scrape_configs:
     honor_labels: true
     kubernetes_sd_configs:
     - role: pod
-      namespaces:
-        names:
-        - {{ .Release.Namespace }}
     tls_config:
       insecure_skip_verify: true
     relabel_configs:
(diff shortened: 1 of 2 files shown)


gregwebs:
All of K8s security comes down to namespaces, so this is a big grant.
For the marketplace install this would fail because cluster roles must be passed in.
Note that for tidb-operator/scheduler we have the cluster scoped option so that it doesn't require cross-namespace permission.

I think we should come up with some alternative models of deployment here. I think we should definitely have an option to not support cross-namespace requests here (and thus this type of lightning deployment). In the cloud one will want to stream to and from object storage and not retain a backup on the PVC.
Another approach to this issue is to have an option to pass in the ClusterRole instead of defining it here, but I think we can avoid this for now.

PR #856, opened 2019-09-02
compatible with the old controller manager

onlymellb:
Make the latest charts compatible with the old controller manager

### Check List 

Tests 

 - Manual test

Code changes

 - Has Helm charts change

Side effects

Related changes

 ```release-note
make the latest  charts compatible with the old controller manager
 ```

### DIFF
--- a/charts/tidb-operator/values.yaml
+++ b/charts/tidb-operator/values.yaml
@@ -15,8 +15,8 @@ imagePullPolicy: IfNotPresent
 defaultStorageClassName: local-storage
 
 # tidbBackupManagerImage is tidb backup manager image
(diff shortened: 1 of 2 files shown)


charts/tidb-operator/values.yaml
```diff
@@ -15,8 +15,8 @@ imagePullPolicy: IfNotPresent
 defaultStorageClassName: local-storage
 
 # tidbBackupManagerImage is tidb backup manager image
-tidbBackupManagerImage: pingcap/tidb-backup-manager:latest
-defaultBackupStorageClassName: local-storage
+# tidbBackupManagerImage: pingcap/tidb-backup-manager:latest
```
tennix: This does not break compatibility, so just keep it.

PR #858, opened 2019-09-02
add docker registry for kind Kubernetes cluster

xiaojingchen:
currently, `hack/kind-cluster-build.sh` has not provided inner docker registry for user
add docker registry for kind Kubernetes cluster
### Check List 

Tests 
 - No code

Code changes

 - Has documents change

Side effects

Related changes

 ```release-note
NONE
 ```


hack/kind-cluster-build.sh
```diff
@@ -119,6 +125,137 @@ echo "start to create k8s cluster"
...
+                operator: NotIn
+                values:
+                  - ${registryNode}
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        operator: "Equal"
+        effect: "NoSchedule"
+      containers:
+        - name: nginx
+          image: nginx:1.15-alpine
```
cofyc: register:2 has builtin http server, what's the purpose of nginx proxy?

xiaojingchen: Because the kubelets use localhost:5000 to get images, so need to deploy `nginx-proxy` on every node.

tennix: Only the localhost is trusted by Docker or containerd. Otherwise, we have to configure TLS for the registry.

cofyc: I see, this makes sense

cofyc: BTW, is a plain TCP proxy tunneling enough here? if it is, use socat (docker [image](https://hub.docker.com/r/alpine/socat/)) is much simpler, e.g. `socat tcp-listen:5000,fork,reuseaddr tcp-connect:<registry2-ip>:5000`.

PR #868, opened 2019-09-03
Update values.yaml

lhy1024:

Add note about tikv config to avoid parameters not taking effect

### Check List 

Tests
 - Unit test


lhy1024:
@onlymellb PTAL

PR #870, opened 2019-09-03
docs: benchmark TiDB in Kubernetes

cofyc:

This provides a general way to run sysbench in Kubernetes to benchmark TiDB cluster.

 ```release-note
NONE
 ```


gregwebs:
It looks like sysbench is not requesting resources. Users may not choose the dedicated node approach, so I think that is important.

gregwebs:
It looks like this is assuming a root user without a password. The dealing with the password should at least be mentioned in the docs.

cofyc:
> It looks like this is assuming a root user without a password. The dealing with the password should at least be mentioned in the docs.

marked this WIP, I'm going to document how to customize this with kustomize

manifests/sysbench/prepare.yaml
```diff
@@ -0,0 +1,99 @@
...
+        - /bin/sh
+        - -c
+        - |-
+          MYSQL_OPTS="-h${TIDB_HOST} -P${TIDB_PORT} -u${TIDB_USER}"
+          if [ -n "$TIDB_PASSWORD" ]; then
+            MYSQL_OPTS+=" -p${TIDB_PASSWORD}"
+          fi
+          mysql $MYSQL_OPTS -Nse 'set @@global.tidb_hashagg_final_concurrency=1;'
+          mysql $MYSQL_OPTS -Nse 'set @@global.tidb_hashagg_partial_concurrency=1;'
+          mysql $MYSQL_OPTS -Nse 'set @@global.tidb_disable_txn_auto_retry=0;'
```
gregwebs: This is a side-effecting operation? After sysbench is run the configuration will be changed.
The approach I try to take is to alter the TiDB configuration to the desired production settings and then to not make any special configuration changes for the purpose of the benchmark.

cofyc: I guess it is, it is taken from this benchmark report: https://pingcap.com/docs/v3.0/benchmark/sysbench-v4/

it seems not good to alter database in these general benchmark jobs, I'll remove it

gregwebs: The documentation here can point to advice for configuration. It is also possible to add an opt-in option to set this configuration.

docs/BenchmarkTiDB.md
```diff
@@ -0,0 +1,109 @@
+# Benchmark TiDB
```
tennix: It's better to rename it as `benchmark-tidb.md`

manifests/sysbench/bench.yaml
```diff
@@ -0,0 +1,83 @@
...
+  parallelism: 3
+  template:
+    metadata:
+      labels:
+        app: sysbench-bench
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: sysbench
+        image: uhub.ucloud.cn/pingcap/sysbench:latest
```
tennix: Can this image be customized?

docs/BenchmarkTiDB.md
cofyc: done

manifests/sysbench/bench.yaml
cofyc: yes, add a section "More customizations" for this

PR #873, opened 2019-09-04
tls: fix temp volume configs for controller-manager

AstroProfundis:
This breaks CI and is fixed in later PR, so not necessary to keep in master

Fix broken CI on master after #750 is merged

The CI doesn't test TLS enabled cluster yet, so no need to keep that config in master

### Check List 

Tests 

 - No code

Code changes

 - Has Helm charts change

Side effects

 - Manual added secret for TLS server won't work, but that doesn't matter as later PR fix it

Related changes

 - #750 

 ```release-note

 ```


cofyc:
lint errors are introduced in #872, which will be fixed in a separate PR (@shuijing198799)
I'm going to merge this PR because it fixes a deployment issue in tidb-controller-manager.

PR #878, opened 2019-09-05
fix some orphan pods cleaner bugs

cofyc:
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
NONE
 ```


cmd/controller-manager/main.go
```diff
@@ -141,8 +141,23 @@ func main() {
 	bsController := backupschedule.NewController(kubeCli, cli, informerFactory, kubeInformerFactory)
 	controllerCtx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	go informerFactory.Start(controllerCtx.Done())
-	go kubeInformerFactory.Start(controllerCtx.Done())
+
+	// Start informer factories after all controller are initializated.
+	informerFactory.Start(controllerCtx.Done())
+	kubeInformerFactory.Start(controllerCtx.Done())
```
cofyc: `Start` will start a new goroutine, so no need to use `go ...` here.

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +118,9 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
 			return skipReason, err
 		}
 
+		// if the PVC is not found in apiserver (also informer cache) and the
+		// phrase of the Pod is Pending, delete it and let the stateful
+		// controller to create the pod and its PVC(s) again
```
cofyc: sorry, I forgot add pod phrase check, will add soon

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +118,9 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
...
 		err = opc.podControl.DeletePod(tc, pod)
```
weekface: only delete pod if the pod status is `Pending`

cofyc: sorry, I forgot it...added in https://github.com/pingcap/tidb-operator/pull/878/commits/979117825df3ddd85232fd8a3f1354a67b80cfd3

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -18,17 +18,33 @@ import (
...
 
 // OrphanPodsCleaner implements the logic for cleaning the orphan pods(has no pvc)
+//
+// In scaling out and failover, we will try to delete the old PVC to prevent it
+// from being used by the new pod. However, the PVC might not be deleted
+// immediately in the apiserver because of finalizers (e.g.
+// kubernetes.io/pvc-protection) and the statefulset controller may not have
+// received PVC delete event when it tries to create the new replica.
+// Statefulset controller will create the new pod which will be pending forever
+// because no PVC to use. We need to clean these orphan pods and let the
```
onlymellb: I think this description is not very accurate. When sts judges whether to create a replica, it will first determine whether the PVC of the pod already exists. If PVC exists, it will create a pod. But there is a case when sts judges that PVC already exists and this PVC actually is being deleted. When the pod is created, PVC has been deleted, causing the pod pending forever.

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +110,23 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
 			return skipReason, err
 		}
 
+		// check api server
```
tennix: If PVC not found in cache, re-check from apiserver directly to make sure the PVC really not exist.

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +110,23 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
...
+		if err == nil {
+			skipReason[podName] = skipReasonOrphanPodsCleanerPVCIsFound
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			return skipReason, err
+		}
+
+		// if the PVC is not found in apiserver (also informer cache) and the
+		// phrase of the Pod is Pending, delete it and let the stateful
```
tennix:
```suggestion
		// phase of the Pod is Pending, delete it and let the stateful
```

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +110,23 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
...
+		// controller to create the pod and its PVC(s) again
+		if pod.Status.Phase != v1.PodPending {
```
tennix: The pod status may also be changed during this time from pending to running.

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +110,23 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
...
+			skipReason[podName] = skipReasonOrphanPodsCleanerPodIsNotPending
+			continue
```
onlymellb: Since this is a very lightweight operation, so I think this judgment should be placed at the top to get back quickly.

pkg/manager/member/orphan_pods_cleaner.go
cofyc: >  But there is a case when sts judges that PVC already exists and this PVC actually is being deleted

I think `the statefulset controller may not have received PVC delete event when it tries to create the new replica` covers this case. The Delete event here is apiserver's object Delete event, not `deletionTimestamp=<time>` update event.

pkg/manager/member/orphan_pods_cleaner.go
cofyc: good suggestion, fixed in https://github.com/pingcap/tidb-operator/pull/878/commits/1a352865c2bf31f9747fb7092a4dfc0f3d28eedf.

pkg/manager/member/orphan_pods_cleaner.go
cofyc: done in https://github.com/pingcap/tidb-operator/pull/878/commits/7a3caf0fd6fe28106ffc5c67afad53bf438f51a1

pkg/manager/member/orphan_pods_cleaner.go
cofyc: done in https://github.com/pingcap/tidb-operator/pull/878/commits/7a3caf0fd6fe28106ffc5c67afad53bf438f51a1.

pkg/manager/member/orphan_pods_cleaner.go
cofyc: fixed

cmd/controller-manager/main.go
```diff
@@ -141,8 +141,23 @@ func main() {
...
-	go kubeInformerFactory.Start(controllerCtx.Done())
+
+	// Start informer factories after all controller are initialized.
+	informerFactory.Start(controllerCtx.Done())
+	kubeInformerFactory.Start(controllerCtx.Done())
+
+	// Wait for all started informers' cache were synced.
+	for v, synced := range informerFactory.WaitForCacheSync(wait.NeverStop) {
+		if !synced {
+			glog.Fatalf("error syncing informer for %v", v)
```
tennix: When does this happen?

cofyc: e.g. apiserver errors, the controller manager should restart to reconnect again

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -80,7 +104,19 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
...
+		if err == nil {
+			skipReason[podName] = skipReasonOrphanPodsCleanerPVCIsFound
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			return skipReason, err
+		}
+
+		// if PVC not found in cache, re-check from apiserver directly to make sure the PVC really not exist
+		_, err = opc.kubeCli.CoreV1().PersistentVolumeClaims(ns).Get(pvcName, metav1.GetOptions{})
```
DanielZhangQD: Should the default QPS and Burst be increased for the kubeclient?

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -68,6 +86,12 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
 			continue
 		}
 
+		if pod.Status.Phase != v1.PodPending {
+			skipReason[podName] = skipReasonOrphanPodsCleanerPodIsNotPending
```
DanielZhangQD: For non Pending pod without PVC, should we print any error log?

cmd/controller-manager/main.go
tennix: Okay, sgtm.

pkg/manager/member/orphan_pods_cleaner.go
cofyc: we do now, because if there is a pending pod without PVC, we will delete it and print logs whether it succeeds or not

cmd/controller-manager/main.go
DanielZhangQD: This logic can only be triggered during tidb-controller-manager start, if it starts successfully, the restart will happen due to leader lost with kube-apiserver errors, right?

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +117,33 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
...
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			return skipReason, err
+		}
+
+		// if the PVC is not found in apiserver (also informer cache) and the
+		// phase of the Pod is Pending, delete it and let the stateful
+		// controller to create the pod and its PVC(s) again
+		apiPod, err := opc.kubeCli.CoreV1().Pods(pod.Namespace).Get(pod.Name, metav1.GetOptions{})
```
onlymellb: Use `ns` and `podName` instead of `pod.Namespace` and `pod.Name`

pkg/manager/member/orphan_pods_cleaner.go
cofyc: we can see perf result in stability test and production, if it has a problem (operates too slow because QPS limit), we should try to optimize our code to avoid unnecessary API requests before increasing client-go QPS/Burst

pkg/manager/member/orphan_pods_cleaner.go
cofyc: oh, sorry, I misunderstood...

cofyc: we can print error logs in this scenario (non-Pending pod without PVC), but this rarely happens because 

- PVC cannot be deleted if it is used by a pod
- A pod cannot run if its PVC does not exist

cmd/controller-manager/main.go
cofyc: yes, and of course other fatal errors or panics in the code.

pkg/manager/member/orphan_pods_cleaner.go
cofyc: fixed

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -89,6 +117,33 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
...
+		apiPod, err := opc.kubeCli.CoreV1().Pods(ns).Get(podName, metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			skipReason[podName] = skipReasonOrphanPodsCleanerPodIsNotFound
+			continue
+		}
+		if err != nil {
+			return skipReason, err
+		}
+		// try our best to avoid deleting wrong object in apiserver
+		// TODO upgrade to use deleteOption.Preconditions.ResourceVersion in client-go 1.14+
```
cofyc: xref: https://github.com/kubernetes/kubernetes/pull/74040

PR #890, opened 2019-09-09
hack/kind: fix some issues in kind e2e environment

cofyc:

fix some issues in kind e2e environment

 ```release-note
NONE
 ```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -148,7 +148,8 @@ $ make e2e-docker-push
 After Docker images are pushed to the inner Docker registry, run e2e tests:
 
 ```sh
-$ kubectl apply -f tests/manifests/e2e/e2e.yaml
+# replace 127.0.0.1:5000 with localhost:5000 to make it work with containerd
+$ set 's/127.0.0.1:5000/localhost:5000/g' tests/manifests/e2e/e2e.yaml | kubectl apply -f -
 ```
 
 You can get the e2e test report from the log of testing pod: 
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -148,8 +148,7 @@ $ make e2e-docker-push
 After Docker images are pushed to the inner Docker registry, run e2e tests:
 
 ```sh
-# replace 127.0.0.1:5000 with localhost:5000 to make it work with containerd
-$ set 's/127.0.0.1:5000/localhost:5000/g' tests/manifests/e2e/e2e.yaml | kubectl apply -f -
+$ kubectl apply -f tests/manifests/e2e/e2e.yaml
 ```
 
 You can get the e2e test report from the log of testing pod: 
--- a/hack/kind-cluster-build.sh
+++ b/hack/kind-cluster-build.sh
@@ -259,9 +259,9 @@ EOF
 kubectl apply -f ${registryFile}
 
 echo "init tidb-operator env"
-kubectl apply -f manifests/local-dind/local-volume-provisioner.yaml
-kubectl apply -f manifests/tiller-rbac.yaml
-kubectl apply -f manifests/crd.yaml
+kubectl apply -f ${ROOT}/manifests/local-dind/local-volume-provisioner.yaml
+kubectl apply -f ${ROOT}/manifests/tiller-rbac.yaml
+kubectl apply -f ${ROOT}/manifests/crd.yaml
 kubectl create ns tidb-operator-e2e
 helm init --service-account=tiller --wait
 
--- a/hack/kind-cluster-build.sh
+++ b/hack/kind-cluster-build.sh
@@ -94,8 +94,6 @@ configFile=${workDir}/kind-config.yaml
 cat <<EOF > ${configFile}
 kind: Cluster
 apiVersion: kind.sigs.k8s.io/v1alpha3
-networking:
-  disableDefaultCNI: true
 nodes:
 - role: control-plane
   extraPortMappings:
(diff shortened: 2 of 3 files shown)


hack/kind-cluster-build.sh
```diff
@@ -94,8 +94,6 @@ configFile=${workDir}/kind-config.yaml
 cat <<EOF > ${configFile}
 kind: Cluster
 apiVersion: kind.sigs.k8s.io/v1alpha3
-networking:
-  disableDefaultCNI: true
```
cofyc: default CNI works, so I don't think we need to install flannel.

hack/kind-cluster-build.sh
```diff
@@ -258,14 +256,29 @@ EOF
...
 kubectl apply -f manifests/tiller-rbac.yaml
 kubectl apply -f manifests/crd.yaml
 kubectl create ns tidb-operator-e2e
 helm init --service-account=tiller --wait
 
+# This is required because current tidb-operator-e2e use a DIND-only image `mirantis/hypokube:final`
+# FIXME: remove this
+docker pull gcr.io/google-containers/kube-scheduler:v1.12.2
+docker tag gcr.io/google-containers/kube-scheduler:v1.12.2 mirantis/hypokube:final
+kind load docker-image --name=${clusterName} mirantis/hypokube:final
```
cofyc: we can make it better, however, to make it work, we can add this temporarily.

hack/kind-cluster-build.sh
```diff
@@ -258,14 +256,29 @@ EOF
...
 echo "    export KUBECONFIG=$(kind get kubeconfig-path --name=${clusterName})"
+echo ""
+echo <<EOF
+NOTE: In kind, nodes run docker network and cannot access host network.
+If you configured local HTTP proxy in your docker, images may cannot be pulled
+because http proxy is inaccessible.
+
+If you cannot remove http proxy settings, you can either whitelist image
+domains in NO_PROXY environment or use 'kind load docker-image' command to load
+images into nodes.
```
cofyc: a note about potential http proxy problem

docs/CONTRIBUTING.md
```diff
@@ -148,7 +148,8 @@ $ make e2e-docker-push
 After Docker images are pushed to the inner Docker registry, run e2e tests:
 
 ```sh
-$ kubectl apply -f tests/manifests/e2e/e2e.yaml
+# replace 127.0.0.1:5000 with localhost:5000 to make it work with containerd
+$ set 's/127.0.0.1:5000/localhost:5000/g' tests/manifests/e2e/e2e.yaml | kubectl apply -f -
```
cofyc: kind is planning to support insecure registries in the future. for now, we need to use `localhost:5000`.

tennix: Maybe we should replace `127.0.0.1` to `localhost` in our CI scripts instead of here.

cofyc: you mean `tests/manifests/e2e/e2e.yaml`? that would be better
I am going to update it to make it work both in local kind and IDC e2e environments.

weekface: This doesn't work as expected:
```
cjc:tidb-operator changjunchang$ set 's/127.0.0.1:5000/localhost:5000/g' tests/manifests/e2e/e2e.yaml | kubectl apply -f -
error: no objects passed to apply
```

cofyc: oh, I made a typo here, `set` should be `sed`
however, I think it's better to update e2e.yaml directly. I think it will not break e2e job.

weekface: what about change `127.0.0.1` to `localhost`directly

cofyc: yes, updated in https://github.com/pingcap/tidb-operator/pull/890/commits/28e8f03c29958c970316de65569e1a12ed9ddb4c

hack/kind-cluster-build.sh
```diff
@@ -258,14 +256,29 @@ EOF
 kubectl apply -f ${registryFile}
 
 echo "init tidb-operator env"
-kubectl apply -f manifests/local-dind/kube-flannel.yaml
 kubectl apply -f manifests/local-dind/local-volume-provisioner.yaml
```
onlymellb:
```
error: the path "manifests/local-dind/local-volume-provisioner.yaml" does not exist
```
The execution result of this script is related to the execution location. We should define a `WORKSPACE` variable that is location-agnostic. Then use this variable and relative path to assemble into an absolute path.  For example:
```bash
WORKSPACE=$(cd $(dirname $0)/..; pwd)
kubectl apply -f ${WORKSPACE}/manifests/local-dind/local-volume-provisioner.yaml
...
```

cofyc: good suggestion!

cofyc: fixed in https://github.com/pingcap/tidb-operator/pull/890/commits/50e369bbf48de084a6c0a07b3103c762155effab
cd to the root of the repo, then the script can be executed in any working directroy

hack/kind-cluster-build.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -e
 
+ROOT=$(unset CDPATH && cd $(dirname "${BASH_SOURCE[0]}")/.. && pwd)
+cd $ROOT
```
onlymellb: I think we should try not to change the working directory in the shell script, which will make the script readability worse. Users can't know exactly where the file is when they see the relative path. Because he doesn't know if the other place of this script has changed the working directory too.

cofyc: this is the script of this repository, we can assume working directory is at the root of the repo.

cofyc: paths are updated to be absolute: https://github.com/pingcap/tidb-operator/pull/890/commits/972b9f91529dd905fea206f49047b60054c0cb37 

I think changing the working directory is fine because the script belongs to this repository, even if it's execute outside of the repo, it's working directory should be in the repository.

cofyc:
@xiaojingchen @weekface @onlymellb @tennix PTAL again.

PR #892, opened 2019-09-09
discovery: use port 2379 for cmdline parameters

AstroProfundis:
Change port in the URL returned by discovery for `--join` parameters from `2380` to `2379`.

As per comment in [#1682](https://github.com/pingcap/pd/issues/1682), `2379` is the correct port for `--join`.

While `--init-cluster` uses `2380` as the port, and discovery reads member list from PD, which also uses port `2380`, the only change is to replace port `2380` with `2379` when building `--join` parameters for PD instances trying to join the cluster.

### Check List 

Tests 

Code changes

 - Has Go code change

Related changes

The current approach of using `2380` for `--join` works when PD instances use plain connections, but when enable TLS encrypted connections, it fails to start etcd server. See comments in [#1682](https://github.com/pingcap/pd/issues/1682) for details.

 ```release-note
NONE
 ```

### DIFF
--- a/pkg/discovery/discovery.go
+++ b/pkg/discovery/discovery.go
@@ -105,7 +105,8 @@ func (td *tidbDiscovery) Discover(advertisePeerUrl string) (string, error) {
 
 	membersArr := make([]string, 0)
 	for _, member := range membersInfo.Members {
-		membersArr = append(membersArr, member.PeerUrls[0])
+		memberURL := strings.ReplaceAll(member.PeerUrls[0], ":2380", ":2379")
+		membersArr = append(membersArr, fmt.Sprintf("%s", memberURL))
 	}
 	delete(currentCluster.peers, podName)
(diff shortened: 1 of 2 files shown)


pkg/discovery/discovery.go
```diff
@@ -105,7 +105,8 @@ func (td *tidbDiscovery) Discover(advertisePeerUrl string) (string, error) {
 
 	membersArr := make([]string, 0)
 	for _, member := range membersInfo.Members {
-		membersArr = append(membersArr, member.PeerUrls[0])
+		memberURL := strings.ReplaceAll(member.PeerUrls[0], ":2380", ":2379")
+		membersArr = append(membersArr, fmt.Sprintf("%s", memberURL))
```
weekface:
```suggestion
		membersArr = append(membersArr, memberURL))
```

memberURL is `string` already.

PR #898, opened 2019-09-10
Monitoring: enable alert rules persistence

qiffang:
Previously, The new alert rules added by monitoring-reloader will be loss if prometheus restart.
It is unreasonable and we update the logic following rules

- TiDB Version is not changed, the updated rules should be persistence even though prometheus restart 

- TiDB Version is changed, the updated rules will be loss (It is also stored PV).

- The Cluster is deleted, the updated rules will be loss (It is also stored PV).

It is related the [PR](https://github.com/pingcap/monitoring/pull/13)

### Check List 

- helm chart

Tests 

- Upgrade TIDB Cluster but do not change the TIDB version , the updated files will be persisted

- Upgrade TIDB Version, Updated rules is loss

Code changes

 - Has Helm charts change

 ```release-note
change tidb-monitor-reloader image to pingcap/tidb-monitor-reloader:v1.0.1
 ```


charts/tidb-cluster/templates/monitor-deployment.yaml
```diff
@@ -121,6 +125,8 @@ spec:
         imagePullPolicy: {{ .Values.monitor.reloader.imagePullPolicy | default "IfNotPresent" }}
         command:
         - /bin/reload
+        - --root-store-path=/data
+        - --sub-store-path={{ .Values.tidb.image }}
```
tennix: This command-line argument looks odd, can we rename it to make the name and value match?

qiffang: reloader 想做成和 TIDB 没什么关联独立存在的 prometheus ui reload 工具，所以取了和 TIDB 无关的名字

tennix: Yes, this is OK. But maybe we can convert the `.Values.tidb.image` to the expected path in Helm template.

PR #900, opened 2019-09-11
[DNM] cmd: add tkctl check subcommand to check tidb cluster configuration

tennix:

We currently don't have any environment checking tools like [tidb-ansible](https://github.com/pingcap/tidb-ansible). The OS kernel parameters, k8s environment and tidb cluster configuration need to be tuned for production deployment. Though we have some [document](https://pingcap.com/docs/v3.0/tidb-in-kubernetes/deploy/prerequisites/) for environment requirement, checking all these is tedious and some settings are tightly related which makes the checking more difficult manually. So we need to provide a similar tool for environment checking.

This PR adds a `check` subcommand for `tkctl` which can verify and check the TiDB cluster configurations for production deployment.

The `tkctl check` subcommand will render the tidb-cluster chart with the provided values. And checks the validation of the PD/TiKV/TiDB configuration file, it also makes some suggestions for some settings like log level, pd max-replicas, tikv block-cache size, tidb prepared-plan-cache, txn-local-latches etc.

*Note:* Though newer versions of PD/TiKV/TiDB support configuration validation, multiple versions of PD/TiDB `config` package can not be used in the same version of tidb-operator due to Go module limitation. Besides TiKV is written in Rust, we can not call the `config` library directly in Go. So I copied and converted the PD/TiKV/TiDB configuration definition, this adds extra maintenance burden to us, we may need to find a better way.

### Check List 

Tests 

 - Manual test
```
tkctl check --chart pingcap/tidb-cluster --values tennix-cluster-values.yaml --chart-version v1.1.0-alpha.1 --cloud aws
```

Code changes

 - Has Go code change

Side effects

No

Related changes

 - Need to cherry-pick to the release-1.1 branch

 ```release-note
add `tkctl check` subcommand for TiDB cluster configuration check
 ```


pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+		fmt.Println(cfg)
+		infoPrintln("#### PD Config ####\n")
+		_, err = toml.Decode(cfg, &pdcfg)
+		if err != nil {
+			return fmt.Errorf("failed to decode pd config: %v", err)
+		}
+		if pdcfg.Log.Level != "info" {
+			warnPrintln("PD log-level:", pdcfg.Log.Level)
+		}
+		if pdcfg.Replication.MaxReplicas != 3 {
```
DanielZhangQD: Also, exclude 5? and warn about even numbers?

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+	pdResources := util.ResourceRequirement(tc.Spec.PD.ContainerSpec)
+	tikvResources := util.ResourceRequirement(tc.Spec.TiKV.ContainerSpec)
+	tidbResources := util.ResourceRequirement(tc.Spec.TiDB.ContainerSpec)
+	if err := compareResources(pdResources); err != nil {
+		errorPrintf("PD resources settings (%v) error: %v\n", pdResources, err)
+	}
+	if err := compareResources(tikvResources); err != nil {
+		errorPrintf("TiKV resources settings (%v) error: %v\n", tikvResources, err)
+	}
+	if err := compareResources(tidbResources); err != nil {
```
DanielZhangQD: The slowlog container should also be checked.

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+		policy := "Cluster"
+		if svc.Spec.ExternalTrafficPolicy == "Local" {
+			policy = "Local"
+		}
+		warnPrintf("External Traffic Policy: %s\n", policy)
+	}
+	annotations := svc.Annotations
+	if o.Cloud != "" {
+		if svc.Spec.Type != "LoadBalancer" || annotations == nil {
+			errorPrintln("TiDB Service not exposed as load balancer on", o.Cloud)
```
DanielZhangQD: Miss `%s`?

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+	checkExample = `
+      # check config from local chart
+      tkctl check -t demo --chart charts/tidb-cluster --values charts/tidb-cluster/values.yaml
+
+      # check config from remote chart
+      tkctl check -t demo --chart pingcap/tidb-cluster --values charts/tidb-cluster/values.yaml --version v1.1.0-alpha.2
+`
+	// AWS load balancer annotations
+	awsILBAnnotationKey    = "service.beta.kubernetes.io/aws-load-balancer-internal"
+	awsILBAnnotationVal    = "0.0.0.0/0"
```
DanielZhangQD: The value for internal LB is "true" on [k8s doc](https://kubernetes.io/docs/concepts/services-networking/#loadbalancer)
[...]
metadata:
    name: my-service
    annotations:
        service.beta.kubernetes.io/aws-load-balancer-internal: "true"
[...]

pkg/tkctl/cmd/check/check.go
tennix: This is `errorPrintln`, no need to add `%s`.

pkg/tkctl/cmd/check/check.go
tennix: From AWS [document](https://docs.aws.amazon.com/eks/latest/userguide/load-balancing.html), its value is `0.0.0.0/0`. And we've already tested and used this configuration in benchmark environment.

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+	case -1:
+		warnPrintln("CPU request < CPU limit, not Guaranteed QoS")
+	case 0:
+	case 1:
+		return fmt.Errorf("CPU request > CPU limit")
+	}
+	_, b1 := cpuRequest.AsInt64()
+	_, b2 := cpuLimit.AsInt64()
+	if !b1 || !b2 {
+		fmt.Errorf("CPU request or limit cannot convert as integer, taskset will not set")
```
DanielZhangQD: Warnning or error here?

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+	if container := getContainerFromSts(drainer, "drainer"); container != nil {
+		drainerVersion = getImageVersion(container.Image)
+	}
+	if container := getContainerFromSts(pump, "pump"); container != nil {
+		pumpVersion = getImageVersion(container.Image)
+	}
+	if pdVersion != "" && pdVersion == tikvVersion &&
+		pdVersion == tidbVersion {
+		if pump == nil || pdVersion == pumpVersion {
+			return true
```
DanielZhangQD: What if `pdVersion != drainerVersion` here?

pkg/tkctl/cmd/check/check.go
DanielZhangQD: For `svc.Spec.Type == "LoadBalancer" && annotations == nil` case, the error log does not make sense.

DanielZhangQD: Should return here?

pkg/config/v2_1_16/tidb.go
```diff
@@ -0,0 +1,319 @@
...
+		GrpcKeepAliveTimeout: 3,
+		CommitTimeout:        "41s",
+		MaxTxnTimeUse:        590,
+	},
+	Binlog: Binlog{
+		WriteTimeout: "15s",
+	},
+}
+
+var globalConf = defaultTidbConf
```
weekface: This var is not used by any package

pkg/config/v2_1_16/tidb.go
```diff
@@ -0,0 +1,319 @@
...
+	DefaultQueryLogMaxLen = 2048
+)
+
+// FileLogConfig serializes file log related config in toml/json.
+type FileLogConfig struct {
+	zaplog.FileLogConfig
+}
+
+// NewFileLogConfig creates a FileLogConfig.
+func NewFileLogConfig(rotate bool, maxSize uint) FileLogConfig {
```
weekface: ditto

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+)
+
+const (
+	kubeVersion   = "v1.12.8"
+	checkLongDesc = `
+                  Preflight check for TiDB cluster configuration.
+`
+	checkExample = `
+      # check config from local chart
+      tkctl check -t demo --chart charts/tidb-cluster --values charts/tidb-cluster/values.yaml
```
weekface: what does `-t` means, `target`?

pkg/config/v2_1_16/tidb.go
tennix: The configuration file provided by users usually are part of the configuration, other values are using default values defined by PD/TiKV/TiDB itself. Some of default values may still be important. So this variable is kept for future use for more detailed check.

pkg/tkctl/cmd/check/check.go
tennix: The `tkctl` uses `-t` flag for tidb-cluster name, it's a global flag. Actually, the tidb-cluster name is not used by the check command right now. So this can be omitted.
I'll remove it and re-add when we support online check which connects to the k8s cluster to retrieve the node/pv and existing tidb cluster info.

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+	cpuLimit := requirement.Limits[corev1.ResourceCPU]
+	memoryRequest := requirement.Requests[corev1.ResourceMemory]
+	memoryLimit := requirement.Limits[corev1.ResourceMemory]
+
+	if cpuRequest.IsZero() || cpuLimit.IsZero() {
+		return fmt.Errorf("cpu request or limit is zero")
+	}
+	switch cpuRequest.Cmp(cpuLimit) {
+	case -1:
+		warnPrintln("CPU request < CPU limit, not Guaranteed QoS")
```
aylei: Messages printed in this method should print which component fail the check.

Alternatively we can print a message when entering this method to declare which component are being checked.

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+			return fmt.Errorf("failed to decode pd config: %v", err)
+		}
+		if pdcfg.Log.Level != "info" {
+			warnPrintln("PD log-level:", pdcfg.Log.Level)
+		}
+		if pdcfg.Replication.MaxReplicas != 3 {
+			warnPrintln("TiKV data replication factor is", pdcfg.Replication.MaxReplicas)
+		}
+	}
+	return nil
```
aylei: For better CI integration, I suggest this command exit with non-zero code upon any errors by default(some errors are just printed), and could be configured to also exit with non-zero code upon warnings.

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+
+	o.checkTidbService(tidbSvc)
+	checkResourceSettings(&tidbcluster, pumpSts, drainerSts)
+
+	var tidbcfg *config_v2_1_16.TidbConfig
+	var pdcfg *config_v2_1_16.PdConfig
+	var tikvcfg *config_v2_1_16.TikvConfig
+	if cfg, exist := tidbConfigMap.Data["config-file"]; exist {
+		infoPrintln("#### TiDB Config ####")
+		fmt.Println(cfg)
```
aylei: These configs can be quite long. It would be useful if we can control what will be printed via a verbose level or other flags.

pkg/config/v2_1_16/pd.go
```diff
@@ -0,0 +1,335 @@
...
+
+	"github.com/coreos/go-semver/semver"
+	gh "github.com/dustin/go-humanize"
+	"github.com/pingcap/log"
+	"github.com/pkg/errors"
+	"go.uber.org/zap"
+)
+
+// Config is the pd server configuration.
+type PdConfig struct {
```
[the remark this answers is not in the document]
tennix: I can't find a proper way to vendor different versions of the same package. And also the import path is the same. This seems impossible in Go.

pkg/tkctl/cmd/check/check.go
tennix: Good point! I'll try to use `verbose` flag.

gregwebs:
TiDB now has a flag to validate on startup. We could make sure all our components have a validate and exit flag. Then we can actually run the components with this flag. The pod would request only a small amount of cpu/memory needed for this task).
This full execution approach is required for a full validation unless we have all our components make their validation logic available externally (using something like cue lang, which is easy to also execute  from go code, but not from TiKV).

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,460 @@
...
+			return fmt.Errorf("failed to decode tikv config: %v", err)
+		}
+		tikvResources := util.ResourceRequirement(tidbcluster.Spec.TiKV.ContainerSpec)
+		memoryLimit := tikvResources.Limits[corev1.ResourceMemory]
+		m, b1 := memoryLimit.AsInt64()
+		if b1 {
+			// rocksdb.writecf.block-cache-size 10%~30% of memory limit
+			// rocksdb.defaultcf.block-cache-size 30%~50% of memory limit
+			if int64(tikvcfg.Rocksdb.Writecf.BlockCacheSize) < m/10 || int64(tikvcfg.Rocksdb.Writecf.BlockCacheSize) > 3*m/10 {
+				errorPrintf("TiKV rocksdb.writecf.block-cache-size should be 10%% ~ 30%% of memory limit, but got %d/%d\n", tikvcfg.Rocksdb.Writecf.BlockCacheSize, m)
```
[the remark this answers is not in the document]
tennix: Now TiKV configuration uses a separate type (`ReadableSize`) for `size`, it will fail to parse `KiB`, `MiB`, `GiB` and `TiB`. This is for `2.1.16`, for `3.0.0` and later, we'll use `ByteSize` just like PD.

tennix:
> TiDB now has a flag to validate on startup. We could make sure all our components have a validate and exit flag. Then we can actually run the components with this flag. The pod would request only a small amount of cpu/memory needed for this task).
> This full execution approach is required for a full validation unless we have all our components make their validation logic available externally (using something like cue lang, which is easy to also execute from go code, but not from TiKV).

@gregwebs There are some specific validation logic, TiDB component itself configuration validation is not enough. For example, tikv `sync-log`, tidb experimental features like `prepared-plan-cache`, `binlog.ignore-error`, pd `replication.max-replicas`. Besides the validation is not available for old versions of TiDB. And we have plan to do configuration management, especially for auto-upgrading which will not be implemented in TiDB.

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,479 @@
...
+	kubeVersion   = "v1.12.8"
+	checkLongDesc = `
+                  Preflight check for TiDB cluster configuration.
+`
+	checkExample = `
+      # check config from local chart
+      tkctl check --chart charts/tidb-cluster --values charts/tidb-cluster/values.yaml
+
+      # check config from remote chart
+      tkctl check --chart pingcap/tidb-cluster --values charts/tidb-cluster/values.yaml --version v1.1.0-alpha.2
```
weekface:
```suggestion
      tkctl check --chart pingcap/tidb-cluster --values charts/tidb-cluster/values.yaml --chart-version v1.1.0-alpha.2
```

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,479 @@
...
+		}
+		_, err = toml.Decode(cfg, &pdcfg)
+		if err != nil {
+			return fmt.Errorf("failed to decode pd config: %v", err)
+		}
+		if pdcfg.Log.Level != "info" {
+			warnPrintf("The recommended PD log-level is info, the current is %s\n", pdcfg.Log.Level)
+		}
+		if pdcfg.Replication.MaxReplicas != 3 && pdcfg.Replication.MaxReplicas != 5 {
+			errorPrintf("Production cluster must set TiKV data replication factor to 3 or 5, the current is %d\n", pdcfg.Replication.MaxReplicas)
```
weekface: should we need add the `max-replicas` to [pd config](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/values.yaml#L67-L68)? or the check will failed with:

```
Production cluster must set TiKV data replication factor to 3 or 5, the current is 0
```

pkg/tkctl/cmd/check/check.go
```diff
@@ -0,0 +1,479 @@
...
+			if int64(tikvcfg.Rocksdb.Writecf.BlockCacheSize) < m/10 || int64(tikvcfg.Rocksdb.Writecf.BlockCacheSize) > 3*m/10 {
+				errorPrintf("TiKV rocksdb.writecf.block-cache-size should be 10%% ~ 30%% of memory limit, but got %d/%d\n", tikvcfg.Rocksdb.Writecf.BlockCacheSize, m)
+			}
+			if int64(tikvcfg.Rocksdb.Defaultcf.BlockCacheSize) < 3*m/10 || int64(tikvcfg.Rocksdb.Defaultcf.BlockCacheSize) > m/2 {
+				errorPrintf("TiKV rocksdb.defaultcf.block-cache-size should be 30%% ~ 50%% of memory limit, but got %d/%d\n", tikvcfg.Rocksdb.Defaultcf.BlockCacheSize, m)
+			}
+			infoPrintf("TiKV rocksdb.lockcf.block-cache-size: %dMiB\n", tikvcfg.Rocksdb.Lockcf.BlockCacheSize/1024/1024)
+		}
+		if !tikvcfg.RaftStore.SyncLog {
+			errorPrintln("Raftstore sync-log is disabled")
```
weekface: do we need to add `sync-log` to tikv default config? or check will failed:

```
Raftstore sync-log is disabled
```

pkg/tkctl/cmd/check/check.go
tennix: All the components have default configuration, I'll merge user-provided configuration with the default value and then do the check.

tennix:
The server side needs validation anyway, so it's better to do the configuration check on the server side. We plan to provide an aggregated apiserver to do the validation. So I modified the title as `DNM`.

PR #915, opened 2019-09-15
Support specifying SecurityContext for Pods and enable tcp keepalive for AWS

### LINKED ISSUE #795: Feature request: support using sysctls for tidb pods
## Feature Request

Operator should support configure and use kernel parameters within a Kubernetes cluster using the sysctl interface: https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/

### LINKED ISSUE #880: Terraform: TiDB nodes should configure tcp_keepalive_time
## Feature Request

Progress:

- [x] AWS
- [ ] GCP
- [ ] Alibaba Cloud

**Edit:**

**Because the `net.*` kernel parameters are namespaced and will not inherit from parent namespace, we have to change the keepalive parameters in the Pod level, which requires change to the `TidbCluster` CRD definition (a `securityContext` filed of TiKV/TiDB/PD pods will be introduced). This will be addressed in #795. Also note that much of the parameters are considered unsafe and should be enabled via the kubelet args explicitly, this issue will track the kubelet one.** 

Typically, the load balancers provided by public cloud have a fixed idle-timeout or a limited configurable range:

* AWS ELB: 1~4000s
* AWS NLB: 350s
* GCP TCP load balancer: 600s ([ref](https://cloud.google.com/load-balancing/docs/https/))
* Alibaba Cloud SLB: not verified yet

When the idle-timeout reaches, the LB either explicitly close the connection (send `FIN` to both sides, e.g. AWS ELB) or close the connection silently (response `RST` when the client send packets again). Both cases cause long time queries to fail.

One can simulate an idle timeout by the following command:

```
mysql -u <user> -p -e "select sleep(400); select tidb_version();"
``` 
Where 400 should be adjusted to more than the idle-timeout of LB.

To tackle this problem, one can enable tcp_keepalive in the client-side or the server-side, can a the `tcp_keepalive_time` that shorter than the LB idle-timeout.

I propose to enable tcp_keepalive for the TiDB servers ~~and adjust the `tcp_alive_time` kernel parameter of **TiDB nodes** in our terraform deployment.~~  Because:

* Generally speaking, we know the infracture provisioned by our scripts better than the users. Tackling this problem in scripts is more user-friendly than warning this in docs;
* We provision dedicated nodes for TiDB Pods, so the scope of influence is controllable;

cc @tennix @cofyc @AstroProfundis @jlerche @gregwebs @kolbe 

ref:
The tidb tcp_keepalive configuration: https://pingcap.com/docs/v3.0/reference/configuration/tidb-server/configuration-file/#performance
tcp keep alive: http://tldp.org/HOWTO/TCP-Keepalive-HOWTO/overview.html

aylei:

<!--
Thank you for contributing to TiDB Operator! Please read TiDB Operator's [CONTRIBUTING](https://github.com/pingcap/tidb-operator/blob/master/CONTRIBUTING.md) document **BEFORE** filing this PR.
-->

### What problem does this PR solve? <!--add and issue link with summary if exists-->
close #880 
close #795 

A new field `podSecurityContext` is introduced for TiKV/TiDB/PD's spec which can specify sysctls for Pods. Only the securityContext of TiDB is used now, but users can freely customize these fields as needed.
In terraform, enable configuration of `net.*` sysctls in kubelet args, and set proper defaults for AWS.

### Check List <!--REMOVE the items that are not applicable-->

Tests <!-- At least one of them must be included. -->

Tested upon AWS NLB with 350s idle timeout:
```
$ mysql -h <elb-host> -P 4000 -u root
MySQL [(none)]> select sleep(360); select tidb_version();
+------------+
| sleep(360) |
+------------+
|          0 |
+------------+
1 row in set (6 min 0.00 sec)

+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| tidb_version()                                                                                                                                                                                                                                                                                              |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v3.0.1
Git Commit Hash: 9e4e8da3c58c65123db5f26409759fe1847529f8
Git Branch: HEAD
UTC Build Time: 2019-07-16 01:03:40
GoVersion: go version go1.12 linux/amd64
Race Enabled: false
TiKV Min Version: 2.1.0-alpha.1-ff3dd160846b7d1aed9079c389fc188f7f5ea13e
Check Table Before Drop: false |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)
```

Verify the sysctls are properly set:

```
kubectl -n my-cluster exec -it my-cluster-tidb-1 -c tidb -- sh
/ # sysctl -a | grep keepalive
net.ipv4.tcp_keepalive_intvl = 300
net.ipv4.tcp_keepalive_probes = 9
net.ipv4.tcp_keepalive_time = 300
```

Code changes

 - Has Helm charts change
 - Has Go code change

Related changes

 <!--
 If no, just write "NONE" in the release-note block below.
 If yes, a release note is required:
 Enter your extended release note in the block below. If the PR requires additional action from users switching to the new release, include the string "action required".
 -->
 ```release-note
Support specifying SecurityContext for PD, TiKV and TiDB Pods and enable tcp keepalive for AWS.
 ```


deploy/modules/aws/tidb-cluster/values/default.yaml
```diff
@@ -14,6 +14,17 @@ tidb:
...
+    [log]
+    level = "info"
+    [performance]
+    tcp-keep-alive = true
+  podSecurityContext:
+    sysctls:
+    - name: net.ipv4.tcp_keepalive_time
+      value: "300"
+    - name: net.ipv4.tcp_keepalive_intvl
+      value: "300"
```
aylei: send keepalive packet every 300s to survive the 350s fixed idle timeout of AWS NLB.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -359,7 +359,7 @@ func (tkmm *tikvMemberManager) getNewSetForTidbCluster(tc *v1alpha1.TidbCluster)
 					SchedulerName: tc.Spec.SchedulerName,
 					Affinity:      tc.Spec.TiKV.Affinity,
 					NodeSelector:  tc.Spec.TiKV.NodeSelector,
-					HostNetwork:   tc.Spec.PD.HostNetwork,
```
aylei: This was a typo I suppose, is it?
@cofyc

cofyc: yes, thanks!

pkg/apis/pingcap.com/v1alpha1/types.go
```diff
@@ -165,6 +153,16 @@ type ContainerSpec struct {
...
 
+// PodAttributesControlSpec is a spec of some general attributes of TiKV, TiDB and PD Pods
+type PodAttributesSpec struct {
+	Affinity           *corev1.Affinity           `json:"affinity,omitempty"`
+	NodeSelector       map[string]string          `json:"nodeSelector,omitempty"`
+	Tolerations        []corev1.Toleration        `json:"tolerations,omitempty"`
+	Annotations        map[string]string          `json:"annotations,omitempty"`
+	HostNetwork        bool                       `json:"hostNetwork,omitempty"`
+	PodSecurityContext *corev1.PodSecurityContext `json:"podSecurityContext,omitempty"`
+}
```
cofyc: `go vet` will fail on this, because it does not allow using promoted fields in `PodAttributesSpec` outside of this package.

deploy/modules/aws/tidb-cluster/values/default.yaml
cofyc: `net.ipv4.tcp_keepalive_intvl` defaults to 75 seconds, do you think it's necessary to increase it?

charts/tidb-cluster/values.yaml
```diff
@@ -242,6 +246,10 @@ tikv:
   # Default to false.
   hostNetwork: false
 
+  # Specify the security context of PD Pod.
```
```suggestion
  # Specify the security context of TiKV Pod.
```

charts/tidb-cluster/values.yaml
```diff
@@ -304,6 +312,10 @@ tidb:
...
```
```suggestion
  # Specify the security context of TiDB Pod.
```

pkg/apis/pingcap.com/v1alpha1/types.go
aylei: I see... but `make check` passed in my local environment, is there no `go vet` check in `make check`?

cofyc: we can see the results of e2e test

deploy/modules/aws/tidb-cluster/values/default.yaml
aylei: I have no preference actually, just want to make sure the heartbeat packet interval is less 350 despite the information from kernel compiling time.

cofyc: I prefer not to change it. To prevent the connection from being closed by the load balancer which has shorter timeout, setting `net.ipv4.tcp_keepalive_time` is enough. `net.ipv4.tcp_keepalive_intvl` determines when the unresponding connection will be aborted. Increase it will increase the time the connection is kept on the server-side.

aylei: Good point! But `net.ipv4.tcp_keepalive_intvl` determines the interval of subsequent probes, so it should be less than 350s too. I would like to set it to 75s explicitly (as the well-known defaults), how do you think?

cofyc: it's ok

pkg/apis/pingcap.com/v1alpha1/types.go
cofyc: `go vet` does not fail on your PR, because there is no place to access fields in `PodAttributesSpec` in this way: `TiDBSpec.Affinity`.
I'd like to abstract these fields into one struct. What about using a shorter name, `PodSpec`. Because this struct name must be used to access internal fields, e.g `TiDBSpec.PodSpec.Affinity`.

aylei: I've considered `PodSpec`, but IMHO it's a little ambiguous with the actual specification of **Pod**, which contains important fields like `containers`, `command`, etc.

cofyc: IMO it's fine, we already have `ContainerSpec`.

aylei: Besides, access promoted fields is OK for go vet, the check is for struct literal:

```

type t1 struct {
	t1f1 int
	t1f2 int
}
type t2 struct {
	t2f1 int
	t2f2 int
	t1
}

var (
	_ = t2{t1f1: 600} // ERROR "cannot use promoted field t1.t1f1 in struct literal of type t2"
)

func main() {
    a := t2{
        t1: t1{t1f1: 2,t1f2: 3,},
        t2f1: 4,
        t2f2: 5,
    }
    _ = a.t1f1
    _ = a.t1f2

}
```

```
go vet
# _/Users/alei/repo/alei/vet [_/Users/alei/repo/alei/vet.test]
./main.go:14:9: cannot use promoted field t1.t1f1 in struct literal of type t2
```

aylei: Having this, I think it's better to use a longer but more concise name

cofyc: you're right, it passes because there are no struct literals in test files.

deploy/modules/aws/tidb-cluster/local.tf
```diff
@@ -76,14 +76,15 @@ locals {
...
-      kubelet_extra_args   = join(" ",
+      name             = "${var.cluster_name}-tidb"
+      key_name         = var.ssh_key_name
+      instance_type    = var.tidb_instance_type
+      root_volume_type = "gp2"
+      root_volume_size = "50"
+      public_ip        = false
+      kubelet_extra_args = join(" ",
         [
+          "--allowed-unsafe-sysctls=\\\"net.*\\\"",
```
cofyc: is PodSecurityPolicy required? https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#podsecuritypolicy

aylei: No, by default all the sysctls are allowed by `PodSecurityPolicy`

pkg/apis/pingcap.com/v1alpha1/types.go
tennix: I I think it's better to remove the Pod prefix since it's already inside the podattributespec.

pkg/apis/pingcap.com/v1alpha1/types.go
```diff
@@ -142,14 +134,10 @@ type TiDBSlowLogTailerSpec struct {
...
-	Affinity         *corev1.Affinity    `json:"affinity,omitempty"`
-	NodeSelector     map[string]string   `json:"nodeSelector,omitempty"`
-	Tolerations      []corev1.Toleration `json:"tolerations,omitempty"`
-	Annotations      map[string]string   `json:"annotations,omitempty"`
-	HostNetwork      bool                `json:"hostNetwork,omitempty"`
-	Privileged       bool                `json:"privileged,omitempty"`
-	StorageClassName string              `json:"storageClassName,omitempty"`
+	PodAttributesSpec
+	Replicas         int32  `json:"replicas"`
+	Privileged       bool   `json:"privileged,omitempty"`
```
tennix: Is `Privileged` a pod attribute?

deploy/modules/aws/tidb-cluster/local.tf
```diff
@@ -76,14 +76,15 @@ locals {
...
-      root_volume_size     = "50"
-      public_ip            = false
-      kubelet_extra_args   = join(" ",
+      name             = "${var.cluster_name}-tidb"
+      key_name         = var.ssh_key_name
+      instance_type    = var.tidb_instance_type
+      root_volume_type = "gp2"
+      root_volume_size = "50"
+      public_ip        = false
+      kubelet_extra_args = join(" ",
```
DanielZhangQD: What about GKE and Ali Cloud?

deploy/modules/aws/tidb-cluster/values/default.yaml
```diff
@@ -14,6 +14,17 @@ tidb:
...
+  config: |
+    [log]
+    level = "info"
+    [performance]
+    tcp-keep-alive = true
+  podSecurityContext:
+    sysctls:
+    - name: net.ipv4.tcp_keepalive_time
+      value: "300"
+    - name: net.ipv4.tcp_keepalive_intvl
```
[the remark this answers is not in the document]
DanielZhangQD: Should we put these as default configs in values.yaml of tidb-cluster chart?

charts/tidb-cluster/values.yaml
```diff
@@ -242,6 +246,10 @@ tikv:
   # Default to false.
   hostNetwork: false
 
+  # Specify the security context of TiKV Pod.
+  # refer to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
```
DanielZhangQD: There is not sysctl config example in this page, is it OK to add below link?
https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/

pkg/apis/pingcap.com/v1alpha1/types.go
aylei: It's an attribute of PD/TiKV/TiDB container

deploy/modules/aws/tidb-cluster/values/default.yaml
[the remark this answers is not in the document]
aylei: No, these configurations are specific to AWS.

For `net.core.somaxconn`, I think it's another problem and can be addressed in an separate issue.

deploy/modules/aws/tidb-cluster/local.tf
aylei: They are not included in this issue

charts/tidb-cluster/values.yaml
aylei: This field is not only used to specifying sysctls but the entire pod security context

pkg/apis/pingcap.com/v1alpha1/types.go
aylei: Because we use `PodAttributesSpec` for composition, I think `PodSecurityContext` is more concise when access the field of PD/TiDB/TiKVSpec. e.g.: `TiDBSpec.PodSecurityContext`

deploy/modules/aws/tidb-cluster/values/default.yaml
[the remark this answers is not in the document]
aylei: BTW, what's the proper value of `net.core.somaxconn`?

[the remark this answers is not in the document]
weekface: 32768

ref: https://pingcap.com/docs/v3.0/tidb-in-kubernetes/deploy/prerequisites/

pkg/apis/pingcap.com/v1alpha1/types.go
tennix: OK, got it.

aylei:
@tennix @DanielZhangQD PTAL again

deploy/modules/aws/tidb-cluster/values/default.yaml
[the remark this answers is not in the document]
tennix: The `net.core.somaxconn` is a general issue, so I think we can set this in the tidb-cluster chart values.yaml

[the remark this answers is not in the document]
cofyc: not possible, it's marked unsafe because of kernel memory accounting issue, so it must be whitelisted via kubelet flag, otherwise the pod will fail to start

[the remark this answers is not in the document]
cofyc: oh, I misunderstood here, it's ok to add in deploy/aws values.yaml file.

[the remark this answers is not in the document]
aylei: Most of the kernel parameters in the prerequisites document are namespaced, seems like we should configure the safe part for users by default and add document about how to configure these parameters via pod security context. Tracked in: #924

Issue #899, opened 2019-09-11
initSQL from configMap


cofyc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Some users think it's hard to maintain multiple and complex SQL statements in `values.yaml` file.

**Describe the feature you'd like:**

They want to write multiple SQL statements in a file and let the job to read the file and execute SQL statements.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

The job pod needs to mount the file into the container. We can use ConfigMap. In `tidb-cluster` chart, we can add a new field, e.g. `initSQlFromConfigMap`, which is like `envFromConfigMap`.

In the deploy process, the user creates a ConfigMap first (`kubectl create configmap my-config --from-file=path/to/bar`) and then configures it in the values.yaml.

Yisaer:
How about the initSql in `values.yaml` when we support this feature?

cofyc:
mutually exclusive with each other. because there is no validation, we can give one of them a higher priority.

Yisaer:
supported in https://github.com/pingcap/tidb-operator/pull/917

cofyc:
closed via #917

PR #917, opened 2019-09-16
add option to initSql from Configmap

Yisaer:
https://github.com/pingcap/tidb-operator/issues/899
It will be user-friendly if the sql of tidb initialization is supported to be read from configmap. and the previous way to init sql is still work.

Make initSql from configmap in `tidb-cluster` chart have a higher priority than iniSql in `values.yaml`

### Check List 

Tests 

Code changes

 - Has Helm charts change

Side effects

 - N/A

Related changes

 - N/A

NONE

### DIFF
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -267,6 +267,10 @@ tidb:
   # If unset, defaults to '%' which means allow any host to connect to the TiDB.
   # permitHost: 127.0.0.1
   # initSql is the SQL statements executed after the TiDB cluster is bootstrapped.
+  # The configmap name of init sql, you can create configmap with following command:
+  # kubectl create configmap tidb-initsql --from-file=init-sql --namespace=<namespace>
+  # If both set the initSqlConfigMapName and initSql, the initSqlConfigMapName will have the higher priority and both are mutually exclusive.
+  # initSqlConfigMapName: tidb-initsql
   # initSql: |-
   #   create database app;
   image: pingcap/tidb:v3.0.1
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -267,8 +267,8 @@ tidb:
   # If unset, defaults to '%' which means allow any host to connect to the TiDB.
   # permitHost: 127.0.0.1
   # initSql is the SQL statements executed after the TiDB cluster is bootstrapped.
-  # The configmap name of init sql, you can create configmap with following command:
-  # kubectl create configmap tidb-initsql --from-file=init-sql --namespace=<namespace>
+  # The configmap name of init sql and the name of file must be init-sql with no file extension, you can create configmap with following command:
+  # kubectl create configmap tidb-initsql --from-file=init-sql=/path/to/your/init-sql --namespace=<namespace>
   # If both set the initSqlConfigMapName and initSql, the initSqlConfigMapName will have the higher priority and both are mutually exclusive.
   # initSqlConfigMapName: tidb-initsql
   # initSql: |-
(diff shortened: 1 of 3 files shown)


charts/tidb-cluster/values.yaml
```diff
@@ -267,6 +267,10 @@ tidb:
   # If unset, defaults to '%' which means allow any host to connect to the TiDB.
   # permitHost: 127.0.0.1
   # initSql is the SQL statements executed after the TiDB cluster is bootstrapped.
+  # The configmap name of init sql, you can create configmap with following command:
+  # kubectl create configmap tidb-initsql --from-file=init-sql --namespace=<namespace>
+  # If both set the initSqlConfigMapName and initSql, the initSqlConfigMapName will have the higher priority and both are mutually exclusive.
+  # initSqlConfigMapName: tidb-initsql
```
cofyc: add a note that the key name must be `init-sql`?

charts/tidb-cluster/values.yaml
```diff
@@ -267,6 +267,10 @@ tidb:
   # If unset, defaults to '%' which means allow any host to connect to the TiDB.
   # permitHost: 127.0.0.1
   # initSql is the SQL statements executed after the TiDB cluster is bootstrapped.
+  # The configmap name of init sql, you can create configmap with following command:
+  # kubectl create configmap tidb-initsql --from-file=init-sql --namespace=<namespace>
```
cofyc: `--from-file=init-sql=/path/to/your/init-sql` is better, because it set the key as `init-sql` explicitly

charts/tidb-cluster/values.yaml
```diff
@@ -267,6 +267,10 @@ tidb:
   # If unset, defaults to '%' which means allow any host to connect to the TiDB.
   # permitHost: 127.0.0.1
   # initSql is the SQL statements executed after the TiDB cluster is bootstrapped.
+  # The configmap name of init sql and the name of file must be init-sql with no file extension, you can create configmap with following command:
+  # kubectl create configmap tidb-initsql --from-file=init-sql=/path/to/your/init-sql --namespace=<namespace>
+  # If both set the initSqlConfigMapName and initSql, the initSqlConfigMapName will have the higher priority and both are mutually exclusive.
```
cofyc:
```suggestion
  # If both initSqlConfigMapName and initSql are set, the initSqlConfigMapName will be used.
```

cofyc:
https://github.com/pingcap/tidb-operator/blob/f2232a056cb83f9510b41e07db0c9be45290a6bc/charts/tidb-cluster/templates/tidb-initializer-job.yaml#L1

when only the `initSqlFromConfigMap` is set, the job should be created too

Yisaer:
> https://github.com/pingcap/tidb-operator/blob/f2232a056cb83f9510b41e07db0c9be45290a6bc/charts/tidb-cluster/templates/tidb-initializer-job.yaml#L1
> 
> when only the `initSqlFromConfigMap` is set, the job should be created too

I found that the `.Values.tidb.passwordSecretName` must be set if trying to initialize, or the helm installation will be failed which didn't match the judge condition ` {{- if or .Values.tidb.passwordSecretName .Values.tidb.permitHost .Values.tidb.initSql }} `

cofyc:
- better to make it clear that when initSqlFromConfigMap is set, tidb-initializer-job must be created to run init SQLs
- I guess it's a bug that password must be set for now. In its comment, the password can be unset and it makes sense not to change the password in initializer job

Yisaer:
I've test all 8 cases composed of `passwordSecretName` ,`initSqlConfigMapName` and `initSql` and the result is expected.

aylei:
> I've test all 8 cases composed of `passwordSecretName` ,`initSqlConfigMapName` and `initSql` and the result is expected.

Maybe we should consider adopting an unit test framework for helm

Yisaer:
@cofyc Please merge it if it's all right.

cofyc:
Great work!

PR #919, opened 2019-09-16
helm chart: fix grafana.create=false

gregwebs:
I tested cluster creation with monitor.grafana.create=false


charts/tidb-cluster/templates/monitor-deployment.yaml
```diff
@@ -207,11 +207,11 @@ spec:
           - key: dashboard-config
             path: dashboards.yaml
         name: dashboards-provisioning
+      {{- end }}
```
tennix: I think it should include `prometheus-rules` when not creating Grafana.

gregwebs: it does. The helm chart is broken now: I would suggest merging this and creating a follow up issue for the changes you would like to make.

tennix: This should solve the problem: 
```yaml
      - emptyDir: {}
        name: grafana-dashboard
      {{- end }}
      - emptyDir: {}
        name: prometheus-rules
```

gregwebs: That would break the initializer job.

Honestly I am afraid to make any extra changes to our helm charts now that there are so many options that are unverified.

gregwebs:
@tennix can we get this merged? We can have @qiffang make any additional configuration changes you want since he committed the original changes here. I am just trying to unbreak this option somehow.

qiffang:
> @tennix can we get this merged? We can have @qiffang make any additional configuration changes you want since he committed the original changes here. I am just trying to unbreak this option somehow.

I will fix it now. For this pull request, i think it should be closed because it does not solve the problem completely.
I will link my pull request in here.
Do you think it is ok?

gregwebs:
okay, I will merge this now to fix this option which is currently broken. We can follow up with additional improvements.

PR #935, opened 2019-09-20
make the discovery service independent of tidb-operator

gregwebs:

We can now make a version of this that does not require K8s. Easier service discovery is a frequent request. This would help with the part of bootstrapping PD.

Move the K8s code into a separate K8s module. Create a GetCluster function which provides the data that from the k8s TidbCluster that we need.

### Check List

Tests

 - Unit test

Code changes

 - Has Go code change

 ```release-note
NONE
 ```


cofyc:
make it possible to run outside of k8s?

gregwebs:
> make it possible to run outside of k8s?

Yes.

pkg/discovery/k8s/k8s.go
```diff
@@ -0,0 +1,73 @@
...
+				Replicas:        tc.Spec.PD.Replicas,
+				ResourceVersion: tc.ResourceVersion,
+				Scheme:          tc.Scheme(),
+			}, nil
+		}
+	}
+}
+
+func makeRealTCGetFn(cli versioned.Interface) func(ns, tcName string) (*v1alpha1.TidbCluster, error) {
+	return func(ns, tcName string) (*v1alpha1.TidbCluster, error) {
```
weekface: discovery still needs `v1alpha1.TidbCluster` object, it may not run outside of k8s.

gregwebs: I don't understand this comment, but will make some assumptions and try to respond.
This file is a separate go module called "k8s" which requires v1alpha1.TidbCluster. The go module "discovery" only requires `discovery.Cluster`. Now I can write a new go module in place of this k8s module to use the discovery code.

weekface: I mean the `tidb-discovery` service may not run **without** k8s. It must get the `TidbCluster` object from `kube-apiserver`.

gregwebs: Yes, the executable that is generated here uses K8s. However, we now have a go package that does not depend on K8s. I can now easily write a different executable that does not rely on K8s since the go package clearly defines what it is getting from K8s. Does that make sense?

weekface: okey

pkg/discovery/k8s/k8s.go
```diff
@@ -0,0 +1,73 @@
...
+	tcGetFn := makeRealTCGetFn(cli)
+	makeGetCluster := makeGetTidbCluster(tcGetFn)
+	go wait.Forever(func() {
+		server.StartServer(makeGetCluster, port)
+	}, 5*time.Second)
+	glog.Fatal(http.ListenAndServe(":6060", nil))
+
+}
+
+func makeGetTidbCluster(tcGetFn func(ns, tcName string) (*v1alpha1.TidbCluster, error)) discovery.MakeGetCluster {
```
tennix: Though Go supports closure, this pattern still looks very weird. Could you make this function more idiomatic?

gregwebs: okay, I created an interface.

gregwebs:
There is now a discover-pd binary that I have manually tested with the new POST & GET routes.

cmd/discover-pd/main.go
```diff
@@ -0,0 +1,63 @@
...
+// Errors in this function are fatal
+func RunDiscoveryService(port int) {
+	logs.InitLogs()
+	defer logs.FlushLogs()
+
+	td := discovery.NewTiDBDiscoveryImmediate(getCluster{})
+	go wait.Forever(func() {
+		server.StartServer(td, port)
+	}, 5*time.Second)
+	glog.Fatal(http.ListenAndServe(":6060", nil))
```
tennix: I guess this is for pprof, if so you need to import `net/http/pprof` package.

pkg/discovery/k8s/k8s.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/pingcap/tidb-operator/pkg/discovery/server"
+	"github.com/pingcap/tidb-operator/pkg/pdapi"
+	"github.com/pingcap/tidb-operator/pkg/version"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apiserver/pkg/util/logs"
+	"k8s.io/client-go/rest"
+)
+
+// PrintVerstionInfo calls version.PrintVersionInfo
```
tennix: Why wraps these two functions? The `version` package is already pretty small and standalone.

tennix:
The CI failed. Can this be used somewhere else besides k8s? For binary deployment, I think the deployment topology is static and known before deploying. Also if the topology changes, users know that and will manage manually.

pkg/discovery/discovery.go
```diff
@@ -15,103 +15,334 @@ package discovery
...
+	addresses, err := td.GetClientAddresses(clusterID)
 	if err != nil {
 		return "", err
 	}
+	delete(currentCluster.peers, pdName)
+	return fmt.Sprintf("--join=%s", strings.Join(addresses, ",")), nil
+}
+
+// ParseURL calls url.Parse but parses according to user expecdtation when there is no protocol or leading "//" or host brackets "[]"
+func ParseURL(inputURL string) (url.URL, error) {
```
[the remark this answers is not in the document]
gregwebs: I agree that typing `nil` is a little easier. But if I wanted to type less I would stop programming in golang. I am concerned about safety, and in the function both alternatives are equally safe. My larger concern is instead on how to make life easier for the caller. A pointer is always harder to deal with unless one is specifically needed.
This function is used and tested independently of `ParseK8sURL`.

pkg/discovery/discovery.go
```diff
@@ -15,103 +15,334 @@ package discovery
...
+	parsedURL, err := url.Parse(inputURL)
```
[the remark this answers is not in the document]
gregwebs: I agree that normalization would be easier. I should document that this function preserves the input rather than normalizing it so that when one does `url.String()` they get back the original url. Currently this behavior is desirable since TiKV doesn't want a protocol prefix. However, I should probably fix this issue elsewhere.

pkg/discovery/discovery.go
```diff
@@ -15,103 +15,334 @@ package discovery
...
 	"github.com/pingcap/tidb-operator/pkg/pdapi"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// PDName is a stable identifier of a PD process
+type PDName string
+
+// ClusterID is a stable identifier of a TIDB cluster
+// A PDName is unique when scoped to a ClusterID
+type ClusterID string
```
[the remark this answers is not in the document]
gregwebs: ok

pkg/discovery/discovery.go
```diff
@@ -15,103 +15,334 @@ package discovery
...
+
+type Address struct {
```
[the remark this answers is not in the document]
gregwebs: ok

gregwebs:
>  Can this be used somewhere else besides k8s?
Yes! I am using it outside of K8s right now.
I think we could add it to our docker-compose setup now to get rid of hard-coded things there?

> For binary deployment, I think the deployment topology is static and known before deploying. Also if the topology changes, users know that and will manage manually.
I think our goal is to avoid manual management. These binary deployments when done in the cloud should use an instance group/autoscaling group for resilience

pkg/discovery/k8s/k8s.go
```diff
@@ -0,0 +1,106 @@
...
+func (tgc tiDBGetCluster) GetMembers(clusterID string) (*pdapi.MembersInfo, error) {
+	ns, tcName, tc, err := tgc.getTC(clusterID)
+	if err != nil {
+		return nil, err
+	}
+	pdClient := tgc.pdControl.GetPDClient(pdapi.Namespace(ns), tcName, tc.Spec.EnableTLSCluster)
+	return pdClient.GetMembers()
+}
+
+func (tgc tiDBGetCluster) GetCluster(clusterID string) (discovery.Cluster, error) {
```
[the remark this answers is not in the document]
gregwebs: I disagree that this is better. On success, I want to return to the caller a non-pointer.

tennix:
tidb-docker-compose is on a single machine, PD does not need dynamic configuration. The PD component in tidb-docker-compose almost never dynamically added or deleted.

For cloud discovery, maybe we should use cloud metadata service for discovery as with [consul](https://www.consul.io/docs/agent/cloud-auto-join.html) instead of our homemade discovery service.

gregwebs:
I agree that there are alternative service discovery mechanisms that it can be used. And if one is already available it may be preferred. The downside is that it can be a heavyweight mechanism for our simple need to bootstrap and locate PD.

The lack of persistence of the current implementation is a big problem in a non-K8s environment wher there is no PD service to rely on after bootstrap. Without a similar PD service one still needs a population mechanism in the discovery service for when it is restarted or to be able to persist its data to storage. But fortunately either is a simple addition.

cmd/discover-pd/main.go
gregwebs: hmm, I copied this from the original code. I think it is preferable to just call StartServer and not run pprof.

gregwebs: okay, I looked into pprof more, it looks like it doesn't create much overhead so we can run it in a goroutine. I don't understand though why we need `wait.Forever`.

gregwebs: I am adding pprof in the main as a goroutine and getting rid of wait.Forever.

pkg/discovery/discovery.go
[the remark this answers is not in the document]
gregwebs: Actually, only the old code path requires that this be a PD node. The new code has no such requirement and it is just an address.

pkg/discovery/discovery.go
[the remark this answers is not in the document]
gregwebs: The current behavior is much better though if one doesn't want to assume http. I am changing the name to `ParseAddress`.

gregwebs:
@tennix the PR is updated according to your review.

PR #942, opened 2019-09-23
Convert capacity into MiB for tikv-server when necessary

### LINKED ISSUE #941: Convert capacity into megabytes for tikv-server when necessary
## Bug Report

If the given capacity is less than 1 GB or not a multiple of GB. We can convert it to megabytes instead.

cofyc:

fixes #941 

### Check List 

Tests 

 - Unit test

 ```release-note
Convert tikv capacity into MiB when it's not a multiple of GiB.
 ```


pkg/controller/controller_utils_test.go
```diff
@@ -125,10 +125,29 @@ func TestTiKVCapacity(t *testing.T) {
 		{
 			name: "100GiB",
 			limit: &v1alpha1.ResourceRequirement{
-				Storage: "100Gi",
+				Storage: "100GiB",
```
cofyc: `Storage: "100Gi"` has already been tested in previous test case. I guess this should be `100GiB`.

cofyc:
Perhaps, rounding down is better than rounding up here, because it’s capacity limit.

Issue #879, opened 2019-09-06
externalTrafficPolicy of tidb service should default to local in terraform deployment 


aylei:
## Bug Report

When the `externalTrafficPolicy` is set to cluster (the default value), every Kubernetes node is a valid backend for the load balancer, a single machine failure will impact part of new connections. More seriously, for immature clients that has a 0 connect-timeout (no timeout), the new connections that routed to the failure machine hangs forever.

With `externalTrafficPolicy` set to `local`, the nodes without backend pods route the traffic to blackhole, therefore, LB can mark these nodes as unhealthy when performs the initial check. `local` mode may cause [unbalanced traffic](https://www.asykim.com/blog/deep-dive-into-kubernetes-external-traffic-policies), but not for our terraform deployment which uses dedicated node.   

**What did you do?**

1. Deploy TiDB cluster via the AWS terraform scripts
2. Simulate inserts using mysqlcli
3. Terminate a PD instance from the AWS console

**What did you expect to see?**
No impact on the workload

**What did you see instead?**
mysqlcli hangs forever

cc @tennix @weekface 

ref:
https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/
https://www.asykim.com/blog/deep-dive-into-kubernetes-external-traffic-policies

shinnosuke-okada:
I'll quickly fix this one as it's related to our use case.

shinnosuke-okada:
@aylei i can't close the ticket - can you?

aylei:
Yes.
closed via #947

PR #947, opened 2019-09-24
Set the default externalTrafficPolicy to be Local for TiDB service

shinnosuke-okada:
https://github.com/pingcap/tidb-operator/issues/879

The default externalTrafficPolicy is now set to be Local

### Check List 

Tests 

1. Deploy to AWS from /tidb-operator/deploy/aws out-of-box
2. Run `kubectl describe service my-cluster-tidb -n my-cluster` and verify the value of externalTrafficPolicy.

Code changes

 - Has Helm charts change

Side effects

Related changes

 - N/A

 ```release-note
NONE
 ```

### DIFF
--- a/deploy/modules/aliyun/tidb-cluster/values/default.yaml
+++ b/deploy/modules/aliyun/tidb-cluster/values/default.yaml
@@ -15,6 +15,7 @@ tidb:
   logLevel: info
   service:
     type: LoadBalancer
+    externalTrafficPolicy: Local
     exposeStatus: true
     annotations:
       service.beta.kubernetes.io/alicloud-loadbalancer-address-type: intranet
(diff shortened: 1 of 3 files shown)


weekface:
Can we do this in the terraform deployment script? This looks like a breaking change to the operator.

shinnosuke-okada:
@weekface: which part/file do you have in mind for "the terraform deployment script"? `/tidb-operator/deploy/modules`?

tennix:
@shinnosuke-okada The default values.yaml for terraform module https://github.com/pingcap/tidb-operator/blob/master/deploy/modules/aws/tidb-cluster/values/default.yaml

weekface:
@aylei PTAL

shinnosuke-okada:
@weekface @tennix: Sure! Let me move the change to `/tidb-operator/deploy/modules`.
Although it's a backward incompatible change, I modified the helm chart default as I couldn't think of a use case where multiple TiDB pods should be scheduled per node.

tennix:
Users may deploy tidb-cluster in their IDC where a node can have multiple pods scheduled.

shinnosuke-okada:
Addressed the review comment. PTAL.

PR #950, opened 2019-09-25
Should use GB/MB rather than GiB/MiB for backward compatibility with old TiKV versions

cofyc:

fix bug in https://github.com/pingcap/tidb-operator/pull/942.

We should use GB/MB rather than GiB/MiB for backward compatibility with old TiKV versions.

### Check List 

Tests 

Unit test
E2E test

 ```release-note

 ```


weekface:
Do we need to cherry-pick this PR together with https://github.com/pingcap/tidb-operator/pull/942 to `release-1.0`?

cofyc:
> Do we need to cherry-pick this PR together with #942 to release-1.0?
I don't think it is necessary. It's not a critical bug.

PR #951, opened 2019-09-25
WIP: Do not set --capacity to tikv-server if resources.limits.storage is absent

### LINKED ISSUE #944: Do not set --capacity to tikv-server if resources.limits.storage is not given
## Feature Request

Currently, we cannot set `raftstore.capacity` in TiKV configuration.

https://github.com/pingcap/tidb-operator/blob/761daab5900d98959cd58762690bdfcce2432193/charts/tidb-cluster/values.yaml#L209

Do not pass TiKV default value if `resources.limits.storage` is absent. This will make it possible for users to set capacity in TiKV configuration.

xref: https://github.com/pingcap/tidb-operator/pull/868 https://github.com/pingcap/tidb-operator/pull/942

cofyc:

fixes #944

This makes it possible for users to set capacity in TiKV configuration.

Do not set --capacity to tikv-server if resources.limits.storage is absent.

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
Fix a bug that `raftstore.capacity` is ignored even if  `limits.storage` is absent.
 ```

### DIFF
--- a/charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
+++ b/charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
@@ -37,9 +37,12 @@ ARGS="--pd=${SCHEME}://${CLUSTER_NAME}-pd:2379 \
 --addr=0.0.0.0:20160 \
 --status-addr=0.0.0.0:20180 \
 --data-dir=/var/lib/tikv \
---capacity=${CAPACITY} \
 --config=/etc/tikv/tikv.toml
 "
+
+if [ -n "${CAPACITY:-}" ]; then
+    ARGS="${ARGS} --capacity=${CAPACITY}"
+fi
 
 echo "starting tikv-server ..."
 echo "/tikv-server ${ARGS}"
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -206,8 +206,6 @@ tikv:
   #
   #   # Normally it should be tuned to 30%-50% of `tikv.resources.limits.memory`, for example: 32Gi -> 16GB
   #   capacity = "1GB"
-  # Note that we can't set raftstore.capacity in config because it will be overridden by the command line parameter, 
-  # we can only set capacity in tikv.resources.limits.storage.
 
   replicas: 3
   image: pingcap/tikv:v3.0.1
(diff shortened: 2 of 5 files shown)


tennix:
The `0` capacity means unlimited for TiKV, this is the same meaning that users do not set the limit of the storage.

cofyc:
> The 0 capacity means unlimited for TiKV, this is the same meaning that users do not set the limit of the storage.

The problem is users cannot specify capacity via `raftstore.capacity` configuration because it will be overridden by the `--capacity 0` even if they didn't specify `resources.limits.storage`.

tennix:
> > The 0 capacity means unlimited for TiKV, this is the same meaning that users do not set the limit of the storage.
> 
> The problem is users cannot specify capacity via `raftstore.capacity` configuration because it will be overridden by the `--capacity 0` even if they didn't specify `resources.limits.storage`.

In this case, if users want to set `raftstore.capacity`, can we guide users to configure the storage limit? Since semantically this two should have the same meaning.

cofyc:
You mean we disable `raftstore.capacity` configuration and ask users to configure `resources.limits.storage`?

I don't think it's a good idea, it might confuse users that part of configuration does not take effect. It will be convenient to view all TiKV configuration in one place, especially we plan to manage component configurations via CRDs.

cofyc:
@weekface @aylei @onlymellb 
What do you think?

onlymellb:
I think this modification is reasonable, but introduce this PR will lead to the existing clusters rolling upgrade. So in order to avoid this problem, we should not modify the sts's pod spec but make a judgment in the tikv startup script, If CAPACITY is "-1" we do not specify the `--capacity` for tikv command-line arguments, otherwise specify `--capacity =${CAPACITY}` command-line arguments 
 as before. This is a compatible modification, what do you think? @cofyc @weekface @tennix @aylei @xiaojingchen

cofyc:
you're right, it's an incompatible change.

but I don't like using a special `-1` because it requires users to act if they want to use `raftstore.capacity` in TiKV configuration.

A general solution is to fix this kind of bugs in the next version of TiDBCluster CRD. When the user upgrades tidb-operator, it will set `CAPACITY` environment for the old version of TiDBCluster but don't for the new version of TiDBCluster.

This can extend to other incompatible changes. We can hold this until we are going to release the next version of TiDBCluster.

onlymellb:
If users want to use `raftstore.capacity` in TiKV configuration, they just don't set `tikv.resources.limits.storage`, if users want to specify command-line arguments just need to set 
 `tikv.resources.limits.storage`.  Set `CAPACITY` to "-1"(of course, you can use any special tag to distinguish whether the user wants to use TiKV configuration or TiKV command-line.)  is automatically by tidb-operator based on the user's behavior and is transparent to the user.

cofyc:
IIUC, the problem exists when the tidb-operator is upgraded, desired StatefulSet spec from the same CRD will change. The original value for `CAPACITY` is `0`, so setting `CAPACITY` to  `-1` is the same as skipping it.

onlymellb:
Yes, this is what I think.

PR #990, opened 2019-10-10
add unit tests for bks control

onlymellb:
Add unit test for bks control 

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

 ```release-note
None
 ```


pkg/controller/backupschedule/backup_schedule_control_test.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		{
+			name: "normal",
+			update: func(bs *v1alpha1.BackupSchedule) {
+				bs.Status.LastBackupTime = &metav1.Time{Time: time.Now()}
+			},
+			syncBsManagerErr: false,
+			updateStatusErr:  false,
+			errExpectFn: func(g *GomegaWithT, err error) {
+				g.Expect(err).NotTo(HaveOccurred())
```
weekface: assert the `LastBackupTime ` is the value you set?

onlymellb: I think it is not necessary to check it. The main purpose here is to simulate the status modification, and don't care what value is finally modified, just need to make sure it has been modified.

Issue #996, opened 2019-10-11
Extensions API failed to install in Kubernetes 1.16


tennix:
## Bug Report

Kubernetes 1.16 removed the `extensions` and `v1beta1`, `v1beta2` APIs [by default](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#deprecations-and-removals), but in tidb-operator, there are still some places using these removed APIs. Thus causing TiDB cluster unable to be deployed in Kubernetes 1.16. #974 has upgraded the local volume provisioner daemonset api version to `apps/v1`. Other places also need to be upgraded.

PR #1012, opened 2019-10-15
Replace obsolete versions into apps/v1

Yisaer:
https://github.com/pingcap/tidb-operator/issues/996
Replacing `extensions`,`apps/v1beta1`,`apps/v1beta2` into `apps/v1`  according to [kubernetes 1.16 changelog](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#deprecations-and-removals)

Fixing api versions which will be obsolete in kubernetes 1.16.

Replacing api version.

Tests 
 - No code

Code changes
 - Has Helm charts change

 ```release-note
None
 ```


tennix:
The Go code using deprecated API group also needs to be updated.

pkg/manager/member/utils.go
```diff
@@ -55,13 +55,13 @@ func annotationsMountVolume() (corev1.VolumeMount, corev1.Volume) {
 
 // statefulSetIsUpgrading confirms whether the statefulSet is upgrading phase
 func statefulSetIsUpgrading(set *apps.StatefulSet) bool {
-	if set.Status.ObservedGeneration == nil {
+	if &(set.Status.ObservedGeneration) == nil {
```
tennix: This looks weird, why change this?

pkg/apis/pingcap.com/v1alpha1/types.go
```diff
@@ -292,9 +290,6 @@ type TiKVFailureStore struct {
 
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-
-// +k8s:openapi-gen=true
-// Backup is a backup of tidb cluster.
```
tennix: Why remove this?

Yisaer: Mistake.

pkg/manager/member/utils.go
Yisaer: The `ObservedGeneration` property defined in `StatefulSetStatus` in `v1beta1` package was a pointer type for`int64` while it was `int64` type in `v1` package.
```golang
package v1beta1

type StatefulSetStatus struct {
	// observedGeneration is the most recent generation observed for this StatefulSet. It corresponds to the
	// StatefulSet's generation, which is updated on mutation by the API Server.
	// +optional
	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,1,opt,name=observedGeneration"`
        ....
}
```
```golang
package v1

// StatefulSetStatus represents the current state of a StatefulSet.
type StatefulSetStatus struct {
	// observedGeneration is the most recent generation observed for this StatefulSet. It corresponds to the
	// StatefulSet's generation, which is updated on mutation by the API Server.
	// +optional
	ObservedGeneration int64 `json:"observedGeneration,omitempty" protobuf:"varint,1,opt,name=observedGeneration"`
        ....
}
```

cofyc:
LGTM

`apps/v1` is [introduced in `1.9`](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.9.md#workloads-api-appsv1). I guess no one will use Kubernetes below 1.9 now, however I think it's better to add a note in the `user-facing change` that we migrate to use `app/v1` and does not support Kubernetes before 1.9 anymore.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -26,14 +26,14 @@ import (
 	"github.com/pingcap/tidb-operator/pkg/manager"
 	"github.com/pingcap/tidb-operator/pkg/pdapi"
 	"github.com/pingcap/tidb-operator/pkg/util"
-	apps "k8s.io/api/apps/v1beta1"
+	apps "k8s.io/api/apps/v1"
```
weekface: upgrade from `v1beta1` to `v1` here can cause the pods to recreate or not? do you have a test manually?

cofyc: oh, it's might be a problem to upgrade an existing operator in the cluster.

I think it will fail here because the API version does not match. I guess we need to do API objects migration at the startup of our new operator.

cofyc: made a mistake, in a cluster which supports both v1beta1 & v1 of stateful sets, kube-apiserver will do conversion between them, the client can CRUD different versions of an object with the same name.

Anyway, we should verify it if we can update the operator gracefully from the old operator.

Yisaer: > upgrade from `v1beta1` to `v1` here can cause the pods(pd/tikv/tidb) to recreate or not? do you have a test manually?

In stability test, `UpgradeOperatorOrDie` will ensure whether the TiDB cluster will be recreated or not.
I also test it mannuly,and the upgrading won't cause TiDB cluster recreated.

pkg/manager/member/utils.go
cofyc: this statement can be removed because it's a noop operation if ObservedGeneration is an int64 type.

pkg/manager/member/pd_member_manager_test.go
```diff
@@ -500,7 +500,7 @@ func TestPDMemberManagerPdStatefulSetIsUpgrading(t *testing.T) {
 			setUpdate: func(set *apps.StatefulSet) {
 				set.Status.CurrentRevision = "v1"
 				set.Status.UpdateRevision = "v2"
-				set.Status.ObservedGeneration = func() *int64 { var i int64; i = 1000; return &i }()
+				set.Status.ObservedGeneration = func() int64 { var i int64; i = 1000; return i }()
```
cofyc: `= 1000`

pkg/manager/member/tidb_member_manager_test.go
```diff
@@ -285,7 +285,7 @@ func TestTiDBMemberManagerTiDBStatefulSetIsUpgrading(t *testing.T) {
 			setUpdate: func(set *apps.StatefulSet) {
 				set.Status.CurrentRevision = "v1"
 				set.Status.UpdateRevision = "v2"
-				set.Status.ObservedGeneration = func() *int64 { var i int64; i = 1000; return &i }()
+				set.Status.ObservedGeneration = func() int64 { var i int64; i = 1000; return i }()
```
cofyc: `= 1000`

pkg/manager/member/tikv_member_manager_test.go
```diff
@@ -447,7 +447,7 @@ func TestTiKVMemberManagerTiKVStatefulSetIsUpgrading(t *testing.T) {
 			setUpdate: func(set *apps.StatefulSet) {
 				set.Status.CurrentRevision = "v1"
 				set.Status.UpdateRevision = "v2"
-				set.Status.ObservedGeneration = func() *int64 { var i int64; i = 1000; return &i }()
+				set.Status.ObservedGeneration = func() int64 { var i int64; i = 1000; return i }()
```
cofyc: `= 1000`

pkg/manager/member/utils_test.go
```diff
@@ -58,7 +58,7 @@ func TestStatefulSetIsUpgrading(t *testing.T) {
 		{
 			name: "CurrentRevision not equal UpdateRevision",
 			update: func(set *apps.StatefulSet) {
-				set.Status.ObservedGeneration = func() *int64 { var i int64; i = 1000; return &i }()
+				set.Status.ObservedGeneration = func() int64 { var i int64; i = 1000; return i }()
```
cofyc: same in this file

pkg/manager/member/utils.go
cofyc: @Yisaer this should be cleaned too

DanielZhangQD: The &() will never be nil.

cofyc:
/run-e2e-in-kind

e2e cluster is being rebuilt just now

pkg/manager/member/utils.go
Yisaer: fix

PR #1021, opened 2019-10-16
pin k8s staging repos to the same version as k8s.io/kubernetes

### LINKED ISSUE #1019: pin all k8s dependencies at the same version and install binaries without modifying `go.mod/sum`
## Feature Request

We didn't lock all k8s.io pkgs at 1.12.5. For example, the code-generator in `go.mod` is not 1.12.5.

The root directory of code-generator does not contain go files, so `go mod` cannot vendor it, we should import its commands instead before 1.15.

`go get/install k8s.io/code-generator/cmd/openapi-gen@kubernetes-1.12.5` will modify `go.mod` file because of the required version in `go.mod` is not 1.12.5.

Pin dependencies of k8s.

cofyc:

fixes #1019 

### Check List 

Tests 

 - Unit test
 - E2E test

 ```release-note

 ```


tools/tools.go
```diff
@@ -18,8 +18,18 @@
 package tools
 
 import (
-	_ "github.com/dnephin/govet"
-	_ "k8s.io/code-generator"
-	_ "k8s.io/gengo"
-	_ "k8s.io/klog"
+	// Because in 1.12.x there is no go files udner k8s.io/code-generator, so
+	// we need to import these commands explicitly.
+	// TODO import k8s.io/code-generator only in k8s.io 1.15+
```
cofyc: - `github.com/dnephin/govet` is not used anymore
- no need to vendor `k8s.io/klog` and `k8s.io/gengo`
- in 1.12.x, we should vendor k8s.cio/code-generator/cmd/xxx directly

hack/pin-deps.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+#
+# Pin all k8s.io dependencies to a specified version.
+#
+
+VERSION=1.12.5
```
cofyc: to pin to new k8s version, we can upgrade this VARIABLE and run this script to pin dependencies

hack/pin-deps.sh
```diff
@@ -0,0 +1,33 @@
...
+# Usage: version_ge "$v1" "$v2"
+#
+function version_ge() {
+    local a="$1"
+    local b="$2"
+    [[ "${a}" == $(echo -e "${a}\n${b}" | sort -s -t. -k 1,1n -k 2,2n -k3,3n | head -n1) ]]
+}
+
+if version_ge "1.15.0" $VERSION; then
+    STAGING_REPOS=($(curl -sS https://raw.githubusercontent.com/kubernetes/kubernetes/v${VERSION}/go.mod | sed -n 's|.*k8s.io/\(.*\) => ./staging/src/k8s.io/.*|k8s.io/\1|p'))
```
cofyc: `go.mod` exists only in k8s.io/kubernetes 1.15+

PR #1025, opened 2019-10-18
e2e: free resources on successful runs

cofyc:

because we share a machine to run e2e tests, if we don't free resources, a lot of processes will run in the machine even if no e2e tests are running.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note

 ```

### DIFF
--- a/tests/cmd/e2e/main.go
+++ b/tests/cmd/e2e/main.go
@@ -188,22 +188,27 @@ func main() {
 		testBasic(&wg, cluster1)
 		testHostNetwork(&wg, cluster1)
 		oa.CheckDataRegionDisasterToleranceOrDie(cluster1)
+		oa.CleanTidbClusterOrDie(cluster1)
 	}()
 	go func() {
 		defer wg.Done()
 		testBasic(&wg, cluster5)
 		oa.CheckDataRegionDisasterToleranceOrDie(cluster5)
+		oa.CleanTidbClusterOrDie(cluster5)
 	}()
 	go func() {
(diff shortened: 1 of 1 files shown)


tests/cmd/e2e/main.go
```diff
@@ -188,22 +188,27 @@ func main() {
 		testBasic(&wg, cluster1)
 		testHostNetwork(&wg, cluster1)
 		oa.CheckDataRegionDisasterToleranceOrDie(cluster1)
+		oa.CleanTidbClusterOrDie(cluster1)
```
weekface:
```suggestion
		go oa.CleanTidbClusterOrDie(cluster1)
```

clean a cluster may need a few minutes. Can do it in a goroutine.

cofyc: it's already in the goroutine of this task (see previous code hidden by GitHub)

cofyc: what about deleting the `e2e` namespace at the end of main function and wait all resources are freed?

weekface: Yes, but it increases the time of that goroutine. Failed to clean is not a failure of the whole e2e test.

weekface: Just a suggestion, not a necessity.

cofyc:
/run-e2e-in-kind
rerun after rebase

PR #1037, opened 2019-10-22
Add switch to reclaim pv when scale in tikv or pd

### LINKED ISSUE #908: tikv-scaler need to auto delete old PVCs after scale in.
## Feature Request

There are two tidb-clusters A and B sharing one k8s cluster.  When we scale-in cluster A tikv, the old PVCs aren't deleted. And cluster B  can't scale-out because it's short of PVs.

When scaling-in tikv, delete the old PVCs. 

Add an option, let user choice defer deleting PVC policy.

https://github.com/pingcap/tidb-operator/blob/799930d59dd86c5d60f39bd42f5ed46761cf0639/pkg/manager/member/tikv_scaler.go#L44-L184

onlymellb:
Resolve #908 

Add  `enablePVReclaim` option in `charts/tidbcluster/values.yaml`，set this option to `true` to enable PV reclaim function.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Manual test

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Related changes

 ```release-note
add switch `enablePVReclaim` option to reclaim pv when scale in tikv or pd
 ```


pkg/manager/member/pvc_cleaner.go
```diff
@@ -40,37 +47,144 @@ type PVCCleanerInterface interface {
...
+		}
+
+		_, err := rpc.podLister.Pods(ns).Get(podName)
+		if err == nil {
+			// PVC is still referenced by this pod, can't reclaim PV
+			skipReason[pvcName] = skipReasonPVCCleanerPVCeferencedByPod
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			return skipReason, fmt.Errorf("cluster %s/%s get pvc %s pod %s from local cache failed, err: %v", ns, tcName, pvcName, podName, err)
```
DanielZhangQD: Why `return` for this one error for one PVC instead of `continue`? Same for the following returns.

pkg/manager/member/pvc_cleaner.go
```diff
@@ -40,37 +47,144 @@ type PVCCleanerInterface interface {
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				skipReason[pvcName] = skipReasonPVCCleanerNotFoundPV
+				continue
+			}
+			return skipReason, fmt.Errorf("cluster %s/%s get pvc %s pv %s failed, err: %v", ns, tcName, pvcName, pvName, err)
+		}
+
+		if pv.Spec.PersistentVolumeReclaimPolicy == corev1.PersistentVolumeReclaimDelete {
+			skipReason[pvcName] = skipReasonPVCCleanerPVWithDeletePolicy
```
aylei: Should we skip here? If the reclaim policy of the PV was patched to `delete` and the deletion to PVC was failed in the last round of reconcile, the cleaner would never be able to reclaim the PV.

pkg/manager/member/pvc_cleaner.go
```diff
@@ -98,8 +212,8 @@ func (rpc *realPVCCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string, e
 
 		podName, exist := pvc.Annotations[label.AnnPodNameKey]
 		if !exist {
-			// waiting for pvc's meta info to be synced
-			skipReason[pvcName] = skipReasonPVCCleanerWaitingForPVCSync
+			// PVC has not pod name annotation, this is an unexpected PVC, skip it
```
aylei:
```suggestion
			// PVC has no pod name annotation, this is an unexpected PVC, skip it
```

pkg/manager/member/pvc_cleaner.go
```diff
@@ -40,37 +47,144 @@ type PVCCleanerInterface interface {
...
+			skipReason[pvcName] = skipReasonPVCCleanerPVWithDeletePolicy
+			continue
+		}
+
+		err = rpc.pvControl.PatchPVReclaimPolicy(tc, pv, corev1.PersistentVolumeReclaimDelete)
+		if err != nil {
+			return skipReason, fmt.Errorf("cluster %s/%s patch pv %s to %s failed, err: %v", ns, tcName, pvName, corev1.PersistentVolumeReclaimDelete, err)
+		}
+
+		err = rpc.pvcControl.DeletePVC(tc, pvc)
```
aylei: The `DeletePVC` func should be updated to check the UID in preconditions.

pkg/manager/member/pvc_cleaner.go
onlymellb: The logic here is that if this error occurs, then we have reason to believe that the next operation is likely to fail, so we let it fail quickly and wait for the next sync cycle.

pkg/manager/member/pvc_cleaner.go
onlymellb: Thank you for pointing this out, this is a bug.

pkg/manager/member/pvc_cleaner.go
```diff
@@ -31,7 +33,14 @@ const (
 	skipReasonPVCCleanerPVCNotHasLock            = "pvc cleaner: pvc not has schedule lock"
 	skipReasonPVCCleanerPodWaitingForScheduling  = "pvc cleaner: waiting for pod scheduling"
 	skipReasonPVCCleanerPodNotFound              = "pvc cleaner: the corresponding pod of pvc has not been found"
-	skipReasonPVCCleanerWaitingForPVCSync        = "pvc cleaner: waiting for pvc's meta info to be synced"
+	skipReasonPVCCleanerPVCNotBound              = "pvc cleaner: the pvc is not bound"
+	skipReasonPVCCleanerPVCNotHasPodNameAnn      = "pvc cleaner: pvc not has pod name annotation"
```
DanielZhangQD:
```suggestion
	skipReasonPVCCleanerPVCNotHasPodNameAnn      = "pvc cleaner: pvc has no pod name annotation"
```

pkg/manager/member/pvc_cleaner.go
```diff
@@ -40,37 +49,160 @@ type PVCCleanerInterface interface {
...
+			continue
+		}
+
+		if pvc.DeletionTimestamp != nil {
+			// PVC has been deleted, skip it
+			skipReason[pvcName] = skipReasonPVCCleanerPVCHasBeenDeleted
+			continue
+		}
+
+		if len(pvc.Annotations[label.AnnPVCDeferDeleting]) == 0 {
```
DanielZhangQD: Suggest to check nil for pvc.Annotations.

onlymellb: `len(pvc.Annotations[label.AnnPVCDeferDeleting]) == 0` including this situation.

weekface:
Can we add this feature to `tkctl` to reclaim pv manually?

pkg/manager/member/pvc_cleaner.go
```diff
@@ -40,37 +49,161 @@ type PVCCleanerInterface interface {
...
+			// PVC is still referenced by this pod, can't reclaim PV
+			skipReason[pvcName] = skipReasonPVCCleanerPVCeferencedByPod
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			return skipReason, fmt.Errorf("cluster %s/%s get pvc %s pod %s from local cache failed, err: %v", ns, tcName, pvcName, podName, err)
+		}
+
+		// if pod not found in cache, re-check from apiserver directly to make sure the pod really not exist
+		_, err = rpc.kubeCli.CoreV1().Pods(ns).Get(podName, metav1.GetOptions{})
```
Yisaer: Why not find Pod from apiServer directly? It seems we would alway to chek whether pod exist or not from apiServer.

onlymellb: The object may have been modified during the first query to the second query, so when you need to make some decisions based on this object, we need to query the object from apiserver again to confirm this object has not been modified during this time, as much as possible to ensure that our decision is accurate.

.gitignore
```diff
@@ -10,6 +10,7 @@ tests/images/e2e/tidb-cluster/
 tests/images/e2e/tidb-backup/
 tests/images/e2e/tidb-operator/
 tests/images/e2e/manifests/
+manifests/crd-verify.yaml
```
cofyc: @Yisaer better to update verify script to use a temporary file

pkg/manager/member/pvc_cleaner.go
cofyc: > Why not find Pod from apiServer directly?
This avoids API request as possible as we can because we don't need to clean the PVC in time, so we can skip if the pod exist in the cache (even if the pod may do not exist in the apiserver). However, if the pod does not exist in the cache, we must re-check the apiserver before deleting the PVC because a new pod may be just created by the statefulset controller to use this PVC.

pkg/manager/member/pvc_cleaner.go
```diff
@@ -40,37 +49,161 @@ type PVCCleanerInterface interface {
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				skipReason[pvcName] = skipReasonPVCCleanerNotFoundPV
+				continue
+			}
+			return skipReason, fmt.Errorf("cluster %s/%s get pvc %s pv %s failed, err: %v", ns, tcName, pvcName, pvName, err)
+		}
+
+		if pv.Spec.PersistentVolumeReclaimPolicy != corev1.PersistentVolumeReclaimDelete {
+			err := rpc.pvControl.PatchPVReclaimPolicy(tc, pv, corev1.PersistentVolumeReclaimDelete)
```
cofyc: if `pvReclaimPolicy` is set retain or user's storage class use Retain policy, should we delete the PV?

onlymellb: I think it is necessary, no matter what reclaim policy PV starts with, we should reclaim it when it can be reclaimed, and this reclaim logic does not destroy the data security of normal PV. Just for those PVs that can be reclaimed.

cofyc: add a note in the comments of this option?

onlymellb: Done

PR #1046, opened 2019-10-23
export varibles replicas to values.yaml

onlymellb:
fix undefined variables `replicas` 

### Check List 

Tests 

 - Manual test 

Code changes

 - Has Helm charts change

Side effects

Related changes

 ```release-note

 ```


charts/tidb-cluster/values.yaml
```diff
@@ -706,6 +706,7 @@ scheduledBackup:
 
 importer:
   create: false
+  replicas: 1
```
tennix: IIRC, TiDB Lightning cannot use mulitple tikv-importer to import data right now.

onlymellb: Okay，then we can use number 1 instead of `{{ .Values.importer.replicas }}`

onlymellb: In fact, we can also use multiple sets of `tidb-lighting` and `tikv-importer` at the same time.

Issue #1002, opened 2019-10-11
Adopt aggregated apiserver


aylei:
## Feature Request

**Is your feature request related to a problem? Please describe:**
According to #986 #981 #936 , there are more and more features requires a feasible way to validate, defaults, converse and intercept custom resources.

CRD has been adopted in tidb-operator to define custom resources, but the resources defined by CRD are eventually processed by kube-apiserver. So custom logics upon the CRUD time of resources can only be implemented via hooks (typically webhook), which are tedious to setup and bounded by the extension point provided by kube-apiserver. For example, resource conversion hook is only available in k8s 1.15 or higher.

**Describe the feature you'd like:**
I propose to adopt aggregated apiserver to host custom resources that requires custom logic that hard to implement for CRD.

## Progress

- [x] Skeleton and toolchain #1048 
- [x] ~SharedInformer based watch~ Tracked separately in #1059 
- [x] Installation within tidb-operator #1060 
- [x] Unit tests #1048 
- [x] E2E tests #1058 
- [x] Tests in managed kubernetes
  - [x] ACK (Aliyun)
  - [x] AWS
  - [x] GKE
- [x] ~Detailed Usage example and recipe~ Tracked separately in https://github.com/pingcap/tidb-operator/issues/1143

aylei:
Closing according to the todo list

PR #1048, opened 2019-10-24
Adopt aggregated apiserver

aylei:

#1002 

Add aggregated apiserver skeleton and toolchain.

More todos are listed in the original issue #1002 

### Check List 

Tests 

Manual test

- [x] Create
- [x] Update
- [x] List
- [x] Watch

Code changes

 - Has Go code change

 ```release-note
New component TiDB aggregated apiserver.
 ```

Here's an example about adding new resources in AA: https://github.com/aylei/tidb-operator/pull/1

To make review easier, I split this PR to multiple commits:

- Pin the versions of dependencies for Aggregated Apiserver, which actually can be ignored: https://github.com/pingcap/tidb-operator/pull/1048/commits/9103ca8f36e59c96f0ae5e3e00d2407a718d2012
- Add dataresource CRD, which introduce a new CRD called `DataResource` as the storage of our AA: https://github.com/pingcap/tidb-operator/pull/1048/commits/9103ca8f36e59c96f0ae5e3e00d2407a718d2012
- Generate code for dataresouce CRD, which can be ignored: https://github.com/pingcap/tidb-operator/pull/1048/commits/92fad18de2e4ce3d1806ce9f78bc898c5799dcf3
- **Important: Implement a kube-apiserver based `storage.Interface`**: https://github.com/pingcap/tidb-operator/pull/1048/commits/209a6e1ca04e23b3c9802c323653f97f054db58e
- Add aggregated apiserver skeleton and toolchain: https://github.com/pingcap/tidb-operator/pull/1048/commits/0fa2977740ba9993b9cd43405a2592ca0fa2a7d7

PTAL @weekface @cofyc @tennix @onlymellb @Yisaer @DanielZhangQD


docs/adding-resource-in-aa.md
```diff
@@ -0,0 +1,90 @@
+# Adding New Resources in TiDB Aggregated Apiserver
```
aylei: A brief introduction of the usage

pkg/apiserver/cmd/start.go
```diff
@@ -0,0 +1,309 @@
...
+	Codec            runtime.Codec
+}
+
+type PostStartHook struct {
+	Fn   genericapiserver.PostStartHookFunc
+	Name string
+}
+
+// StartApiServer starts an apiserver.
+func StartApiServer(
```
aylei: AA entrypoint

pkg/apiserver/cmd/start.go
```diff
@@ -0,0 +1,309 @@
...
+		o := genericoptions.DeprecatedInsecureServingOptions{}
+		return o.WithLoopback()
+	}()
+
+	return o
+}
+
+// Config create an apiserver config
+func (o ServerOptions) Config() (*apiserver.Config, error) {
+	if err := o.RecommendedOptions.SecureServing.MaybeDefaultWithSelfSignedCerts(
```
aylei: The CA issue is left for later resolution

aylei:
e2e is not working for this issue now, because the apiserver lib introduce a dependency `bitbucket.org/ww/goautoneg` that have to be downloaded thru hg, which is not installed in the CI env.

```
$ go mod why bitbucket.org/ww/goautoneg
# bitbucket.org/ww/goautoneg
github.com/pingcap/tidb-operator/pkg/apiserver/cmd
k8s.io/apiserver/pkg/endpoints/filters
k8s.io/apiserver/pkg/endpoints/handlers/responsewriters
k8s.io/apiserver/pkg/endpoints/handlers/negotiation
bitbucket.org/ww/goautoneg
```

I will fix it soon, tt does not block reviewing, though.

cofyc:
> e2e is not working for this issue now, because the apiserver lib introduce a dependency bitbucket.org/ww/goautoneg that have to be downloaded thru hg, which is not installed in the CI env.
I've added a command to install hg in CI script.

aylei:
@cofyc Thanks!

tools/tools.go
```diff
@@ -32,4 +32,5 @@ import (
 	_ "k8s.io/code-generator/cmd/openapi-gen"
 	_ "k8s.io/code-generator/cmd/register-gen"
 	_ "k8s.io/code-generator/cmd/set-gen"
+	_ "github.com/kubernetes-incubator/apiserver-builder-alpha/cmd/apiregister-gen"
```
cofyc: why not use `sigs.k8s.io/apiserver-builder-alpha/cmd/apiregister-gen`?

this process is moved to https://github.com/kubernetes-sigs/apiserver-builder-alpha/  now.

aylei: The pinned version v1.12 was still using the `kubernetes-incubator` import path, we can use `sigs.k8s.io....` after we bump the kubernetes dependencies to `v1.16`

cofyc:
because I renamed `pkg/apis/pingcap.com` to `pkg/apis/pingcap`, some conflicts must be resolved. regenerating code and updating imports should be enough

aylei:
@cofyc Thanks for noting!

pkg/pdapi/config.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/pingcap/log"
+	"github.com/pingcap/pd/pkg/typeutil"
+)
+
+// currently, import pd server.Config will introduce a great deal of dependencies and
+// the etcd dependency is conflicted with k8s.io/apiserver, as a workaround, we copy the
+// type definition from PD and get rid of the dependency of pd/server
+// TODO: use type definition from pd when pd separate the config to proper package
+
+// Copied from pd/server/config.go
```
tennix: Why copy this and is it for configuration management?

tennix: Oh, get it from the comment above.

hack/generate-internal-groups.sh
```diff
@@ -0,0 +1,141 @@
...
+    ALL_FQ_APIS+=("${EXT_APIS_PKG}/${GROUP_PATH}/${V}")
+    EXT_FQ_APIS+=("${EXT_APIS_PKG}/${GROUP_PATH}/${V}")
+  done
+done
+
+# apiregister-gen generate the REST boilerplate of resources and their internal version
+if [ "${GENS}" = "all" ] || grep -qw "registry" <<<"${GENS}"; then
+  echo "Generating api registries"
+  "${GOPATH}/bin/apiregister-gen" --input-dirs "$(codegen::join , "${APIS_DIR[@]}")" "$@"
+fi
```
cofyc: except this, do we have other changes?
when we can vendor code-generator root directory, I'd like to use scripts from code-generator.

pkg/apiserver/cmd/start.go
cofyc: self-sign certs and skip tls verify should be enough for internal use

pkg/apiserver/storage/store.go
```diff
@@ -0,0 +1,464 @@
...
+	"k8s.io/apiserver/pkg/storage"
+	"k8s.io/apiserver/pkg/storage/etcd"
+	"k8s.io/apiserver/pkg/storage/storagebackend/factory"
+	"k8s.io/client-go/rest"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// store implements a ConfigMap backed storage.Interface
+type store struct {
```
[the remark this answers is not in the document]
cofyc: add some unit tests to cover basic use cases?

hack/aa-codegen.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -e
```
cofyc: `-e` flag is redundant, it equals to `set -o errexit`.

hack/aa-codegen.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Copyright 2017 PingCAP, Inc.
```
cofyc: nit: 2019

cofyc:
The whole idea looks good to me. Because it's a standalone apiserver and not used by other components. I think we can add some basic unit tests and merge it soon.

pkg/apiserver/storage/store.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+func (s *store) GetToList(ctx context.Context, key string, resourceVersion string, p storage.SelectionPredicate, listObj runtime.Object) error {
+	listPtr, err := meta.GetItemsPtr(listObj)
+	if err != nil {
+		return err
+	}
+	v, err := conversion.EnforcePtr(listPtr)
+	if err != nil || v.Kind() != reflect.Slice {
+		panic("need ptr to slice")
```
tennix: return an error instead of just panic?

pkg/apiserver/storage/store.go
```diff
@@ -0,0 +1,464 @@
...
+	return s.versioner.UpdateList(listObj, 0, "")
+}
+
+func (s *store) List(ctx context.Context, key string, resourceVersion string, pred storage.SelectionPredicate, listObj runtime.Object) error {
+	listPtr, err := meta.GetItemsPtr(listObj)
+	if err != nil {
+		return err
+	}
+	li, err := conversion.EnforcePtr(listPtr)
+	if err != nil || li.Kind() != reflect.Slice {
```
tennix: ditto

pkg/apiserver/cmd/start.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	codec := builders.Codecs.LegacyCodec(versions...)
+	o := &ServerOptions{
+		RecommendedOptions: genericoptions.NewRecommendedOptions("", codec),
+		APIBuilders:        b,
+		RunDelegatedAuth:   true,
+		Codec:              codec,
+	}
+	o.RecommendedOptions.SecureServing.BindPort = 443
```
onlymellb: Why not use [SecureServingOptions](https://github.com/kubernetes/apiserver/blob/fd43ea72cae9a713aae01016b7a58da1bbaf547a/pkg/server/options/serving.go#L37)  instead of a fixed secure port?  just like  `InsecureServingOptions` above.

pkg/apiserver/cmd/start.go
```diff
@@ -0,0 +1,335 @@
...
+		o.RecommendedOptions.Audit.ApplyTo,
+		o.RecommendedOptions.Features.ApplyTo,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	loopbackKubeConfig, kubeInformerFactory, err := o.buildLoopback()
+	if err != nil {
+		klog.Warning("attempting to instantiate loopback client but failed...")
```
onlymellb: Add `err` message to the warning log.

pkg/apis/pingcap.com/v1alpha1/dataresource_types.go
```diff
@@ -0,0 +1,41 @@
...
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// +k8s:openapi-gen=false
+// DataResource is a generic custom resource for storing custom data
+type DataResource struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+
+	Data []byte `json:"data"`
```
DanielZhangQD: So for different CRs, they just store to the same structure []byte?
How to convert back to different CR structures from the Data?

hack/generate-internal-groups.sh
aylei: I change the client output package in case of collision with the generated code of CRD and change the package path from `a.b.c` to `a` so that the behavior is consistent with higher version of kubernetes.

I'd like to change this script to call the vendored script in the code-generator directory after we upgrade the k8s dependencies to `v1.16`

pkg/apiserver/storage/store.go
aylei: This is the behavior of the official implementation, according to the context that this method get called, this condition indicates a bug in code and better to be treated as panic.

tennix: OK, got it.

pkg/apiserver/cmd/start.go
aylei: good catch

pkg/apis/pingcap.com/v1alpha1/dataresource_types.go
aylei: The data decode all information about the resource, just same as etcd

aylei:
@cofyc @tennix @onlymellb @DanielZhangQD I've addressed the review comments and add the basic unit tests for `store.go`(https://github.com/pingcap/tidb-operator/pull/1048/commits/626f3f070c31d2367ad30bb8c80e26db26556bc3). Please take a look again and approve if there's no problem, thanks!

/run-e2e-in-kind

PR #1054, opened 2019-10-28
fix stability test, add `v1` to statefulset apiVersions

weekface:
### Check List 

Tests 

 - Stability test

Code changes

Side effects

Related changes

 ```release-note
NONE
 ```


pkg/webhook/statefulset/statefulset.go
```diff
@@ -68,22 +70,20 @@ func AdmitStatefulSets(ar v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
...
 
-	raw := ar.Request.OldObject.Raw
-	set := apps.StatefulSet{}
-	if _, _, err := deserializer.Decode(raw, nil, &set); err != nil {
-		err := fmt.Errorf("statefulset %s/%s, decode request failed, err: %v", namespace, name, err)
-		glog.Errorf(err.Error())
+	stsObjectMeta, stsPartition, err := getStsAttributes(ar.Request.OldObject.Raw, apiVersion)
+	if err != nil {
+		glog.Errorf("statefulset %s/%s, decode request failed, err: %v", namespace, name, err)
 		return util.ARFail(err)
```
onlymellb: I think this return error is not meaningful, should return a more detailed error, just like the error output above.

tests/actions.go
```diff
@@ -447,7 +447,7 @@ func (oa *operatorActions) DeployOperator(info *OperatorConfig) error {
 	}
 
 	// deploy statefulset webhook and configuration to hijack update statefulset opeartion
-	cmd = fmt.Sprintf("kubectl apply -f %s/webhook.yaml", oa.manifestPath(info.Tag))
+	cmd = fmt.Sprintf(`sed 's/apiVersions: \["v1beta1"\]/apiVersions: ["v1", "v1beta1"]/' %s/webhook.yaml | kubectl apply -f -`, oa.manifestPath(info.Tag))
```
onlymellb: `manifests/webhook.yaml` has been modified, so can you omit this operation?

weekface: no, `v1.0.1` or prior only has one `v1beta1`. We must change them too in the stability test case.

pkg/webhook/statefulset/statefulset.go
weekface: Done

PR #1096, opened 2019-11-01
Disable verbose logging and coverage by default and enable -race in CI

cofyc:

Github action treats `go test` logs as errors.

Note that even if we disabled verbose logging, logs will be printed when the test fails.

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

 ```release-note

 ```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ GOOS := $(if $(GOOS),$(GOOS),linux)
 GOARCH := $(if $(GOARCH),$(GOARCH),amd64)
 GOENV  := GO15VENDOREXPERIMENT="1" CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH)
 GO     := $(GOENV) go build
-GOTEST := CGO_ENABLED=0 go test -v -cover
 
 PACKAGE_LIST := go list ./... | grep -vE "pkg/client" | grep -vE "zz_generated"
 PACKAGE_DIRECTORIES := $(PACKAGE_LIST) | sed 's|github.com/pingcap/tidb-operator/||'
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -86,14 +85,27 @@ stability-test-push: stability-test-docker
...
 	$(GO) -ldflags '$(LDFLAGS)' -o tests/images/fault-trigger/bin/fault-trigger tests/cmd/fault-trigger/*.go
 
+# ARGS:
+#
+# GOFLAGS: Extra flags to pass to 'go' when building, e.g.
+# 		`-v` for verbose logging.
+# 		`-race` for race detector.
+# GO_COVER: Whether to run tests with code coverage. Set to 'y' to enable coverage collection.
+#
+ifeq ($(GO_COVER),y)
```
cofyc: I disable code coverage by default because it's not needed in local development. This will speed up local unit tests.

After this is accepted and merged, I'll update our master job to enable coverage.

Makefile
```diff
@@ -86,14 +85,27 @@ stability-test-push: stability-test-docker
...
+test:
+	@echo "Run unit tests"
+	@go test ./pkg/... && echo "\nUnit tests run successfully!"
+endif
 
 check-all: lint check-static check-shadow check-gosec staticcheck errcheck
 
 check-setup:
-	@which retool >/dev/null 2>&1 || go get github.com/twitchtv/retool
+	@which retool >/dev/null 2>&1 || GO111MODULE=off go get github.com/twitchtv/retool
```
cofyc: we should install retool in GOPATH mode, otherwise, it will change go.mod

PR #1107, opened 2019-11-05
set net.ipv4.tcp_keepalive_time and net.core.somaxconn for tidb and tikv in init container

DanielZhangQD:
fix https://github.com/pingcap/tidb-operator/issues/880 for GCP  and Alibaba
Set net.ipv4.tcp_keepalive_time and net.core.somaxconn for tidb in init container
Set net.core.somaxconn for tikv in init container
### Check List 

Tests 

 - Unit test

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Related changes

 ```release-note
Support configuring net.ipv4.tcp_keepalive_time and net.core.somaxconn for TiDB and configuring net.core.somaxconn for TiKV.
 ```

### DIFF
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -177,10 +177,6 @@ pd:
     # # when the kubelet is configured to allow unsafe sysctls
     # - name: net.core.somaxconn
     #   value: "32768"
-    # - name: net.ipv4.tcp_syncookies
-    #   value: "0"
-    # - name: net.ipv4.tcp_tw_recycle
-    #   value: "0"
 
   # Specify the priorityClassName for PD Pod.
   # refer to https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#how-to-use-priority-and-preemption
@@ -275,10 +271,6 @@ tikv:
     # # when the kubelet is configured to allow unsafe sysctls
     # - name: net.core.somaxconn
     #   value: "32768"
-    # - name: net.ipv4.tcp_syncookies
-    #   value: "0"
-    # - name: net.ipv4.tcp_tw_recycle
-    #   value: "0"
 
   # Specify the priorityClassName for TiKV Pod.
   # refer to https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#how-to-use-priority-and-preemption
@@ -362,10 +354,6 @@ tidb:
     # # when the kubelet is configured to allow unsafe sysctls
     # - name: net.core.somaxconn
     #   value: "32768"
-    # - name: net.ipv4.tcp_syncookies
-    #   value: "0"
-    # - name: net.ipv4.tcp_tw_recycle
-    #   value: "0"
 
     # # Load balancers usually have an idle timeout (eg. AWS NLB idle timeout is 350),
     # # the tcp_keepalive_time must be set to lower than LB idle timeout.
--- a/pkg/manager/member/tidb_member_manager.go
+++ b/pkg/manager/member/tidb_member_manager.go
@@ -278,7 +278,7 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster) *apps.StatefulSet {
 				privileged := true
 				initContainers = append(initContainers, corev1.Container{
 					Name:  "init",
-					Image: controller.GetSlowLogTailerImage(tc),
+					Image: controller.GetUtilImage(tc),
 					Command: []string{
 						"sh",
 						"-c",
--- a/pkg/manager/member/tidb_member_manager.go
+++ b/pkg/manager/member/tidb_member_manager.go
@@ -266,6 +266,36 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster) *apps.StatefulSet {
 		})
 	}
 
+	sysctls := "sysctl -w"
+	var initContainers []corev1.Container
+	if tc.Spec.TiDB.Annotations != nil {
+		init, ok := tc.Spec.TiDB.Annotations[label.AnnSysctlInit]
+		if ok && (init == label.AnnSysctlInitVal) {
+			if tc.Spec.TiDB.PodSecurityContext != nil && len(tc.Spec.TiDB.PodSecurityContext.Sysctls) > 0 {
+				for _, sysctl := range tc.Spec.TiDB.PodSecurityContext.Sysctls {
+					sysctls = sysctls + fmt.Sprintf(" %s=%s", sysctl.Name, sysctl.Value)
+				}
+				privileged := true
+				initContainers = append(initContainers, corev1.Container{
+					Name:  "init",
+					Image: controller.GetSlowLogTailerImage(tc),
+					Command: []string{
+						"sh",
+						"-c",
+						sysctls,
+					},
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: &privileged,
+					},
+				})
+			}
+		}
+	}
+	podSecurityContext := tc.Spec.TiDB.PodSecurityContext.DeepCopy()
+	if len(initContainers) > 0 {
+		podSecurityContext.Sysctls = []corev1.Sysctl{}
+	}
+
 	var containers []corev1.Container
 	if tc.Spec.TiDB.SeparateSlowLog {
 		// mount a shared volume and tail the slow log to STDOUT using a sidecar.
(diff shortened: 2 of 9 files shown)


DanielZhangQD:
After discussion with @tennix and @weekface, will update the solution as below:

- users still configure PodSecurityContext.sysctls in values.yaml
- if annotations for init container is set in values.yaml, tidb-operator will convert PodSecurityContext.sysctl to init container;
- otherwise, set PodSecurityContext with existing logic.

charts/tidb-cluster/templates/tidb-cluster.yaml
```diff
@@ -18,6 +18,13 @@ metadata:
 {{ toYaml .Values.extraLabels | indent 4 }}
     {{- end }}
 spec:
+  {{- if .Values.tcpKeepaliveTime }}
+  tcpKeepaliveTime: {{ .Values.tcpKeepaliveTime }}
+  tcpKeepaliveIntvl: 75
+  {{- end }}
+  {{- if .Values.soMaxConn }}
+  soMaxConn: {{ .Values.soMaxConn }}
```
aylei: Putting these parameters at cluster level is a little misleading, how about moving these to component? e.g. `.sepc.tidb.tcpKeepaliveTime`, `.spec.tikv.soMaxConn`

pkg/manager/member/tidb_member_manager.go
```diff
@@ -266,6 +266,31 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster) *apps.StatefulSet {
...
+		sysctls = sysctls + fmt.Sprintf(" net.core.somaxconn=%d", tc.Spec.SoMaxConn)
+	}
+
+	var initContainers []corev1.Container
+
+	if tc.Spec.TCPKeepaliveTime > 0 || tc.Spec.SoMaxConn > 0 {
+		privileged := true
+		initContainers = append(initContainers, corev1.Container{
+			Name:  "init",
+			Image: controller.GetSlowLogTailerImage(tc),
```
aylei: Use another image or promote the slow log tailer image to a general util image

pkg/manager/member/tidb_member_manager.go
```diff
@@ -266,6 +266,31 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster) *apps.StatefulSet {
 		})
 	}
 
+	sysctls := "sysctl -w"
+	if tc.Spec.TCPKeepaliveTime > 0 {
```
aylei:
```suggestion
	if tc.Spec.TCPKeepaliveTime > 0 && tc.Spec.TCPKeepaliveIntvl > 0 {
```

charts/tidb-cluster/templates/tidb-cluster.yaml
cofyc: I prefer a general solution. How about using [the sysctl struct from Kubernetes](https://github.com/kubernetes/api/blob/25240d7d6d90fc391c9f53db9d620a313264d5d5/core/v1/types.go#L3184)?
 
It's an alternative way when configuring via Kubelet is not avaiable, no need to use another structure.

cofyc: +1 config per component, like podSecurityContext which we are using now

charts/tidb-cluster/values.yaml
```diff
@@ -173,10 +173,6 @@ pd:
     # # when the kubelet is configured to allow unsafe sysctls
     # - name: net.core.somaxconn
     #   value: "32768"
-    # - name: net.ipv4.tcp_syncookies
-    #   value: "0"
```
DanielZhangQD: Remove them because they cannot be set in containers

DanielZhangQD:
@cofyc @aylei @tennix @weekface Please help review the new solution.

pkg/manager/member/tidb_member_manager.go
DanielZhangQD: Actually, the image does not matter, so I just reuse the image from the slow log tailer and we also have a default value for it, this does not involve any change to values.yaml and users do not need to know about it at all.

charts/tidb-cluster/templates/tidb-cluster.yaml
DanielZhangQD: Reuse the PodSecurityContext in each component.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -266,6 +266,36 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster) *apps.StatefulSet {
...
+		init, ok := tc.Spec.TiDB.Annotations[label.AnnSysctlInit]
+		if ok && (init == label.AnnSysctlInitVal) {
+			if tc.Spec.TiDB.PodSecurityContext != nil && len(tc.Spec.TiDB.PodSecurityContext.Sysctls) > 0 {
+				for _, sysctl := range tc.Spec.TiDB.PodSecurityContext.Sysctls {
+					sysctls = sysctls + fmt.Sprintf(" %s=%s", sysctl.Name, sysctl.Value)
+				}
+				privileged := true
+				initContainers = append(initContainers, corev1.Container{
+					Name:  "init",
+					Image: controller.GetSlowLogTailerImage(tc),
```
aylei: I mean we should use a new method name or change this method name to something like `GetBusyboxImage` or `GetUtilitymage`

pkg/manager/member/tikv_member_manager.go
```diff
@@ -315,6 +315,36 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster) (*apps.StatefulSet, e
...
+		init, ok := tc.Spec.TiKV.Annotations[label.AnnSysctlInit]
+		if ok && (init == label.AnnSysctlInitVal) {
+			if tc.Spec.TiKV.PodSecurityContext != nil && len(tc.Spec.TiKV.PodSecurityContext.Sysctls) > 0 {
+				for _, sysctl := range tc.Spec.TiKV.PodSecurityContext.Sysctls {
+					sysctls = sysctls + fmt.Sprintf(" %s=%s", sysctl.Name, sysctl.Value)
+				}
+				privileged := true
+				initContainers = append(initContainers, corev1.Container{
+					Name:  "init",
+					Image: controller.GetSlowLogTailerImage(tc),
```
tennix: I suggest using an annotation to configure the init container image, and also set a default image when the annotation is not available.

charts/tidb-cluster/values.yaml
tennix: This should still work by default if the kubelet has the unsafe sysctl enabled. When not enabled, users need to add an annotation to set this. So I don't think we should remove it.

DanielZhangQD: No, they are not for containers.

pkg/manager/member/tikv_member_manager.go
aylei: I prefer a dedicate field instead of annotation if we really want it to be configurable, for troubleshooting purpose, changing the image via `kubectl edit` is more straightforward.

pkg/manager/member/tidb_member_manager.go
DanielZhangQD: ok

pkg/manager/member/tikv_member_manager.go
DanielZhangQD: I think it's unnecessary to introduce user interface change for this image, it does not matter what it is and for TiDB version >= v2.1.8, slow log tailer must be enabled to check slow log.

DanielZhangQD:
@weekface @tennix @aylei comments addressed, PTAL again.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -315,6 +315,36 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster) (*apps.StatefulSet, e
...
+						sysctls,
+					},
+					SecurityContext: &corev1.SecurityContext{
+						Privileged: &privileged,
+					},
+				})
+			}
+		}
+	}
+	podSecurityContext := tc.Spec.TiKV.PodSecurityContext.DeepCopy()
```
tennix: Add a comment here: when using init container to set `sysctl`, do not set `sysctl` in pod security context.

DanielZhangQD: ok

DanielZhangQD:
@weekface @tennix @aylei comments addressed, PTAL again.

PR #1101, opened 2019-11-04
Add admission to delete PD Pod handler for webhook

Yisaer:
https://github.com/pingcap/tidb-operator/issues/992
Sub PR for https://github.com/pingcap/tidb-operator/pull/1040

### What problem does this PR solve? 

Currently Operator need to handle all the process during `Upgrade` / `Scale`.  In this request, Webhook ensure all the pd pod would be safely deleted while Operator only need to edit template or replicas of `StatefulSet`

For now webhook won't work during `Upgrading` and `Scaling` in stable verison according to https://github.com/pingcap/tidb-operator/pull/1040#issuecomment-548774197 's advice. 

Tests 
 - Unit test

Code changes
 - Has Go code change

Side effects

 - N/A

### Does this PR introduce a user-facing change?:
None

PR #1114, opened 2019-11-07
Fix transfer leader before delete member during pd scale in

Yisaer:
According to https://github.com/pingcap/tidb-operator/pull/1101/files#r343496778' comment, it should be transferring pd leader first then deleting pd member during pd scale-in if the target pod was pd leader.

Code changes
 - Has Go code change

### DIFF
--- a/pkg/webhook/pod/util.go
+++ b/pkg/webhook/pod/util.go
@@ -114,3 +114,11 @@ func addDeferDeletingToPDPod(podAC *PodAdmissionControl, pod *core.Pod) error {
 	_, err := podAC.kubeCli.CoreV1().Pods(pod.Namespace).Update(pod)
 	return err
 }
+
+func isPDLeader(pdClient pdapi.PDClient, pod *core.Pod) (bool, error) {
+	leader, err := pdClient.GetPDLeader()
+	if err != nil {
+		return false, err
+	}
+	return leader.Name == pod.Name, nil
+}
--- a/pkg/webhook/pod/pd_deleter.go
+++ b/pkg/webhook/pod/pd_deleter.go
@@ -49,7 +49,12 @@ func (pc *PodAdmissionControl) admitDeletePdPods(pod *corev1.Pod, ownerStatefulS
 
 	IsDeferDeleting := IsPodWithPDDeferDeletingAnnotations(pod)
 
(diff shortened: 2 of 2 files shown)


pkg/webhook/pod/pods.go
```diff
@@ -91,7 +96,9 @@ func (pc *PodAdmissionControl) AdmitPods(ar v1beta1.AdmissionReview) *v1beta1.Ad
 //// otherwise we will check it decided by component type.
 func (pc *PodAdmissionControl) admitDeletePods(name, namespace string) *v1beta1.AdmissionResponse {
 
-	pod, err := pc.podLister.Pods(namespace).Get(name)
+	// We would update pod annotations if they were deleted member by admission controller,
+	// so we shall find this pod from apiServer considering getting latest pod info.
+	pod, err := pc.kubeCli.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
```
Yisaer: We would update pod annotations if they were deleted member by admission controller, so we shall find this pod from apiServer considering getting latest pod info.

DanielZhangQD: What if it's changed after this get?

pkg/webhook/pod/util.go
```diff
@@ -69,7 +68,7 @@ func addDeferDeletingToPVC(podAC *PodAdmissionControl, tc *v1alpha1.TidbCluster,
 
 // check whether the former upgraded pd pods were healthy in PD cluster during PD upgrading.
 // If not,then return an error
-func checkFormerPDPodStatus(kubeCli kubernetes.Interface, pdClient pdapi.PDClient, tc *v1alpha1.TidbCluster, namespace string, ordinal int32, replicas int32) error {
+func checkFormerPDPodStatus(podLister corelisters.PodLister, pdClient pdapi.PDClient, tc *v1alpha1.TidbCluster, namespace string, ordinal int32, replicas int32) error {
```
DanielZhangQD: Not sure if it's reasonable, just feel a little wired to have a lister as a function argument.

pkg/webhook/pod/pods.go
Yisaer: In this case, we only need to get pod.name and pod.annotations ( defer deleting specificlly) from pod info. If `delfer deleting` is added, it wouldn't be remove again.

pkg/webhook/pod/util.go
Yisaer: It only used in pd_deleter, so I think that's ok.

PR #853, opened 2019-08-30
Support configuring resources requests and limits for all containers of tidb cluster

aylei:
close #852 
Signed-off-by: Aylei <rayingecho@gmail.com>

### Check List 

Tests 

 - E2E test

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has CI related scripts change
 - Has documents change

Side effects

 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

### Does this PR introduce a user-facing change?:
 
 ```release-note
Support configuring resources requests and limits for all containers of tidb cluster
 ```

PR #1119, opened 2019-11-07
Cherry pick of #853: configure resources for all containers

aylei:
Cherry pick of #853 on release-1.0.

#853: Support confiugring resources requests and limits for all


aylei:
@weekface @tennix PTAL

weekface:
should add `scheduledBackup.resources` in ` values.yaml`

aylei:
@weekface Nice catch
Addressed, PTAL.

PR #1124, opened 2019-11-08
Add cert Initializer and Refresh

Yisaer:
In Webhook Server and AA Server we need CA cert to communicated internal. 

This request adds 2 component as Initializer and Refresh as Job and CronJob to help us create cert and refresh it when it's expired automaticlly.

`Initializer` would create the component resources which defined in `cmd args` and the default values is `all`. As webhook server currently is  a `switch` feature ( which is disabled in default ), I also add a switch to enable or not `webhook` resources in Initializer.

`Refresh` would check all the resources created by `Initializer` once a day ( this could be configured in `values.yaml` ) and create the related job if the resources need to be refreshed.

Tests 

Code changes

 - Has Helm charts change
 - Has Go code change

Side effects

 - N/A

None


charts/tidb-operator/templates/scripts/_create_cert.sh.tpl
```diff
@@ -0,0 +1,159 @@
+set -e
```
gregwebs: Does this need to be a .tpl? It seems like it doesn't need to be templated but instead can take in any templating as arguments.

Yisaer: > Does this need to be a .tpl? It seems like it doesn't need to be templated but instead can take in any templating as arguments.

Thanks for pointing this.

cmd/initializer/main.go
```diff
@@ -0,0 +1,82 @@
...
+	podName          string
+	verifyPeriodDays int
+)
+
+func init() {
+	flag.BoolVar(&printVersion, "V", false, "Show version and quit")
+	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
+	flag.StringVar(&component, "component", "all", "The component to initializer resource for")
+	flag.IntVar(&verifyPeriodDays, "verifyPeriodDays", 365, "Number of days a certificate generated by -x509 is valid for.")
+	flag.BoolVar(&initializer.WebhookEnabled, "webhookEnabled", false, "switch to enable init webhook resource")
```
Yisaer: As webhook server need to be enabled to use, Intializer need to switch on  `webhookEnabled` to initialize its resources.

pkg/cert-refresh/cert_refresh.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+}
+
+// RefreshManager list all the secrets created by Initializer and check whether it need refresh ca cert
+func (rm *RefreshManager) Run(podName, namespace string, config *RefreshConfig) error {
+	list, err := rm.checkCertsNeedRefresh(namespace, config.RefreshIntervalDays)
+	if err != nil {
+		return err
+	}
+	err = rm.setConfigOwnerReferences(podName, namespace, config)
```
Yisaer: If `Refresh` find any cert need to be refreshed, it would create a job and set the job owner Reference as `Refresh` itself.

charts/tidb-operator/templates/scripts/_create_cert.sh
```diff
@@ -0,0 +1,159 @@
+set -e
```
Yisaer: Currently I choose to use script to generate ca cert, secrets and csr to avoid repeating cert and secret control as https://github.com/pingcap/tidb-operator/pull/782 is working on it either. I will generate all the above things in code after https://github.com/pingcap/tidb-operator/pull/782 is merged.

cmd/initializer/main.go
aylei: Webhook must be enabled when the component is `all` or `webhook`, so it seems like that this variable is redundant

cmd/initializer/main.go
```diff
@@ -0,0 +1,82 @@
...
+	namespace        string
+	component        string
+	podName          string
+	verifyPeriodDays int
+)
+
+func init() {
+	flag.BoolVar(&printVersion, "V", false, "Show version and quit")
+	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
+	flag.StringVar(&component, "component", "all", "The component to initializer resource for")
```
aylei: Better to be a string array, that is, able to specify multiple components but not all

Makefile
```diff
@@ -52,6 +52,12 @@ admission-controller:
 apiserver:
 	$(GO) -ldflags '$(LDFLAGS)' -o images/tidb-operator/bin/tidb-apiserver cmd/apiserver/main.go
 
+initializer:
+	$(GO) -ldflags '$(LDFLAGS)' -o images/tidb-operator/bin/tidb-initializer cmd/initializer/main.go
```
aylei: The `tidb-initializer` name is already taken by the initializer job in the `tidb-cluster` chart

images/tidb-operator/Dockerfile
```diff
@@ -1,8 +1,21 @@
...
+    apk del --purge deps && \
+    rm /var/cache/apk/*
+
 RUN apk add tzdata --no-cache
 ADD bin/tidb-scheduler /usr/local/bin/tidb-scheduler
 ADD bin/tidb-discovery /usr/local/bin/tidb-discovery
 ADD bin/tidb-admission-controller /usr/local/bin/tidb-admission-controller
 ADD bin/tidb-controller-manager /usr/local/bin/tidb-controller-manager
 ADD bin/tidb-apiserver /usr/local/bin/tidb-apiserver
+ADD bin/tidb-initializer /usr/local/bin/tidb-initializer
```
aylei: I recommend separate the initializer bin, openssl and kubectl to a dedicated image if possible

pkg/cert-refresh/cert_refresh.go
```diff
@@ -0,0 +1,215 @@
...
+	RefreshIntervalDays int
+}
+
+const (
+	configMapName  = "tidb-operator-initializer-config"
+	serviceAccount = "tidb-operator-initializer-sa"
+)
+
+var (
+	Days int
```
aylei: Where is this variable being assigned?

images/tidb-operator/Dockerfile
Yisaer: > I recommend separate the initializer bin, openssl and kubectl to a dedicated image if possible

Nice advice. Separating to different images needs to edit the scripts of e2e test in jenkins which is highly risky. I will seperate them after we transferred scripts into code repo.

pkg/initializer/webhook_initializer.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// update Webhook server to make sure the latest secret would be used.
+func (initializer *Initializer) updateWebhookServer(namespace string, secret *core.Secret) error {
+
+	server, err := initializer.kubeCli.ExtensionsV1beta1().Deployments(namespace).Get(AdmissionWebhookName, metav1.GetOptions{})
```
aylei: This hard-coded name an implicit dependency between the initializer binary and the helm chart.

IMHO, we should provide a command line argument to specify the deployment name, and coordinate the names in the helm chart.

charts/tidb-operator/templates/initializer-job.yaml
```diff
@@ -0,0 +1,51 @@
+{{ $a := split ":" .Values.operatorImage }}
+  {{ $length := len $a }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  {{ if eq $length 2 }}
+  name: tidb-operator-initializer-{{ $a._1 }}
+  {{ else }}
+  name: tidb-operator-initializer-{{ $a._2 }}
+  {{ end }}
```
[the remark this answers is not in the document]
Yisaer: > what is this for?

Adding tag version in `Initializer` Job name, if not , Operator would upgrade failed as Job already exists.

charts/tidb-operator/templates/cert-refresh-cronjob.yaml
```diff
@@ -0,0 +1,45 @@
...
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
+                - name: POD_NAME
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.name
+                - name: IMAGE
+                  value: {{ .Values.operatorImage }}
+          restartPolicy: OnFailure
```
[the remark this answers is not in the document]
Yisaer: fix

charts/tidb-operator/values.yaml
```diff
@@ -124,3 +124,11 @@ apiserver:
...
   #   effect: "NoSchedule"
 
+cert:
+  # initializer and cert-refresh would generate ca cert for verifyPeriodDays days
+  verifyPeriodDays: "365"
+  refresh:
+    # cert-refresh check cert whether expired by schedule
+    schedule: "0 1 * * *"
+    # cert-refresh would refresh ca certs if it would be expired after refreshIntervalDays at most
+    IntervalDays: "7"
```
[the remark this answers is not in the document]
Yisaer: fix

cmd/cert-refresh/main.go
```diff
@@ -0,0 +1,92 @@
...
+	refreshIntervalDays int
+	podName             string
+	image               string
+)
+
+func init() {
+	flag.BoolVar(&printVersion, "V", false, "Show version and quit")
+	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
+	flag.IntVar(&cert_refresh.Days, "verifyPeriodDays", 365, "Number of days a certificate generated by -x509 is valid for.")
+	flag.IntVar(&refreshIntervalDays, "intervalDays", 7, "Number of days cert should be refreshed before expired.")
```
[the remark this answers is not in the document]
Yisaer: fix

pkg/initializer/util.go
```diff
@@ -0,0 +1,39 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package initializer
+
+import (
+	"fmt"
+	glog "k8s.io/klog"
```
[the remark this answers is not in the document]
Yisaer: fix

charts/tidb-operator/templates/initializer-job.yaml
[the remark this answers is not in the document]
aylei: Sounds weird, what if the operator release is upgraded to change the configuration but the operatorImage is not changed?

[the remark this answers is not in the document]
aylei: I got it, this job reference the operatorImage in container spec. But a more serious problem is that this job may run multiple times accidentally.

pkg/initializer/webhook_initializer.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return err
+	}
+	if server.Spec.Template.Annotations == nil {
+		server.Spec.Template.Annotations = map[string]string{
+			"checksum/cert": certUtils.Checksum(secret.Data["cert.pem"]),
+		}
+	} else {
+		server.Spec.Template.Annotations["checksum/cert"] = certUtils.Checksum(secret.Data["cert.pem"])
```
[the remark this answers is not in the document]
Yisaer: fix

pkg/initializer/webhook_initializer.go
```diff
@@ -0,0 +1,156 @@
...
+
+	conf, err := initializer.kubeCli.AdmissionregistrationV1beta1().ValidatingWebhookConfigurations().Get(AdmissionWebhookName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	f := admissionregistration.Fail
+	for id, webhook := range conf.Webhooks {
+		webhook.ClientConfig.CABundle = secret.Data["cert.pem"]
+		webhook.FailurePolicy = &f
```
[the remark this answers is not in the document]
Yisaer: fix

charts/tidb-operator/templates/initializer-job.yaml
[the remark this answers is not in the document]
Yisaer: > I got it, this job reference the operatorImage in container spec. But a more serious problem is that this job may run multiple times accidentally.

yep, the job would be created and run each time Operator Upgraded, but it doesn't mean the resources needed to be created each times. Currently I found that the `manifests/create_cert.sh` would delete the older resources and create the new one and I reused it.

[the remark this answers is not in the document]
aylei: I don't get your point, would the recreation cause a disruption of the webhook?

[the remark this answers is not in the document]
aylei: At least, this job is not necessarily to run multiple times, seems like that we are working around helm.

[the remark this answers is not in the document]
Yisaer: > At least, this job is not necessarily to run multiple times, seems like that we are working around helm.

I think this would be revised as `Initializer` wouldn't recreate resources if the resources exists.

cmd/initializer/main.go
Yisaer: > Webhook must be enabled when the component is `all` or `webhook`, so it seems like that this variable is redundant

As webhook server is a `switch feature`, the webhook-server resources shouldn't be initialize when component is `all` and `webhookEnabled` is false which I think `webhookEnabled` is still necessary for now. It will be removed after Webhook enabled in default.

charts/tidb-operator/templates/initializer-job.yaml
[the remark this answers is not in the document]
Yisaer: fix by helm hook.

cmd/initializer/main.go
aylei: Sounds like the problem is in the design of the `component` field, if the client (in this case, helm chart) choose to disable webhook initialization, then it could have excluded webhook in the component list, which is the only purpose of the `webhook-enabled` argument.

cmd/initializer/main.go
Yisaer: fix

Makefile
Yisaer: fix

pkg/cert-refresh/cert_refresh.go
Yisaer: fix

pkg/initializer/webhook_initializer.go
Yisaer: fix

cmd/initializer/main.go
Yisaer: > Sounds like the problem is in the design of the `component` argument, if the client (in this case, helm chart) choose to disable webhook initialization, then it could have excluded webhook in the component list, which is the only purpose of the `webhook-enabled` argument.

In this case, I intend to use `component` argument to decide which resource should be initialize. If we want to initializer all (possibly less then 5 or 3) the component resources except `webhook`,  the componenet should be defined as `aaa,bbb,ccc,ddd`. `webhook-enabled` should be removed after `webhook` enabled in default.

images/tidb-operator/Dockerfile
```diff
@@ -1,8 +1,21 @@
 FROM alpine:3.5
 
+ARG KUBE_VERSION=1.12.10
```
tennix: Why package `kubectl` in tidb-operator Docker image? This increases the image size and binds specific `kubectl` to tidb-operator, besides it adds extra attack surface for tidb-operator. What if there is no code change but kubectl has a security bugfix, shall we update the docker image and tell user to upgrade tidb-operator image?

Yisaer: > Why package `kubectl` in tidb-operator Docker image? This increases the image size and binds specific `kubectl` to tidb-operator, besides it adds extra attack surface for tidb-operator. What if there is no code change but kubectl has a security bugfix, shall we update the docker image and tell user to upgrade tidb-operator image?

Right, I think installing openssl would be necessary here. I will remove `kubectl` and edit the scripts.

Yisaer: fix

images/tidb-operator/Dockerfile
Yisaer: remove openssl and kubectl now

pkg/initializer/util/csrUtil.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return approvedCSR, nil
+}
+
+func generateCSR(service, namespace string, ownerRefs []metav1.OwnerReference, requestPem []byte) *certificates.CertificateSigningRequest {
+	csr := &certificates.CertificateSigningRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            fmt.Sprintf("%s.%s", service, namespace),
```
aylei: why not use the `CSRNameForService` func?

aylei: or we can remove the func, which is better IMHO

pkg/initializer/util/csrUtil.go
```diff
@@ -0,0 +1,163 @@
...
+		return false, nil
+	})
+}
+
+func waitUntilCsrExist(csrName string, kubeCli kubernetes.Interface, timeout time.Duration) error {
+
+	return wait.PollImmediate(1*time.Second, timeout, func() (done bool, err error) {
+		_, err = kubeCli.CertificatesV1beta1().CertificateSigningRequests().Get(csrName, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
```
aylei: errors other than `notFound` should not be ignored

Yisaer: > errors other than `notFound` should not be ignored

In this case, I think there is no difference between `notFound` error and other error.

aylei: `notFound` is one of the expected situations, other unexpected errors should either be logged or returned, or you will end up with getting the timeout message and have no idea about what really happens.

Yisaer: fix

pkg/initializer/util/csrUtil.go
Yisaer: fix

weekface:
Use [cert-manager](https://docs.cert-manager.io/en/latest/index.html) to manage the certificates may be a better solution.

It supports [self signing](https://docs.cert-manager.io/en/latest/tasks/issuers/setup-selfsigned.html) Issuers.

It supports [renewal](https://docs.cert-manager.io/en/latest/reference/certificates.html#certificate-duration-and-renewal-window) certificates **automatically**.

It can [inject](https://docs.cert-manager.io/en/latest/reference/cainjector.html) a `Certificate` into the `caBundle` field of `ValidatingWebhookConfiguration` resource **annotated with**:
```
cert-manager.io/inject-apiserver-ca: "true" Injects the cluster CA.
cert-manager.io/inject-ca-from: <NAMESPACE>/<CERTIFICATE> Injects the CA from the specified certificate.
```

pkg/cert-refresh/cert_refresh.go
```diff
@@ -0,0 +1,196 @@
...
+	OwnerKind            string
+	OwnerName            string
+	OwnerUid             string
+	RefreshIntervalDays  int
+	WebhookAdmissionName string
+	Timeout              int
+}
+
+type RefreshManager struct {
+	kubecli kubernetes.Interface
```
AstroProfundis: `kubeCli` is used in other part of our codes

pkg/cert-refresh/cert_refresh.go
```diff
@@ -0,0 +1,196 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cert_refresh
+
+import (
+	"fmt"
+	"github.com/pingcap/tidb-operator/pkg/label"
+	certUtil "github.com/pingcap/tidb-operator/pkg/util"
```
AstroProfundis: I think we would have conflicts with #782

pkg/cert-refresh/cert_refresh.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+const (
+	configMapName  = "tidb-operator-initializer-config"
+	serviceAccount = "tidb-operator-initializer-sa"
+)
+
+func NewRefreshManager(kubecli kubernetes.Interface, config *RefreshConfig) *RefreshManager {
+	return &RefreshManager{
+		kubecli: kubecli,
```
AstroProfundis: ditto

PR #1129, opened 2019-11-12
upgrade Go to 1.13

tennix:
### What problem does this PR solve? 

This PR closes #1128 

### What is changed and how does it work?

The related build scripts are updated to Go 1.13 and also the CONTRIBUTION.md.

### Check List 

Tests 

 - Unit test

Code changes

 - Has CI related scripts change
 - Has documents change

Side effects

N/A

Related changes

 - Need to cherry-pick to the release-1.0 branch

### Does this PR introduce a user-facing change?:
 
 ```release-note
None
 ```

PR #1134, opened 2019-11-12
manually cherry-pick #1129 to upgrade Go compiler to 1.13

tennix:
Cherry-pick #1129 to release-1.0 branch to upgrade the Go compiler to 1.13.


cofyc:
this fails because in the master branch we're going to install CRDs in e2e test instead, I'll fix it

PR #1135, opened 2019-11-12
put e2e jenkinsfile from jenkins into repo

onlymellb:
Currently Jenkinsfile of e2e pr and master job are placed on jenkins master, there is no version control, and each time there is a change, you need to modify both places. So the purpose of this pr is to manage these scripts with git and then try to unify the scripts of e2e pr and master jobs.

### Check List 

Tests 

 - Manual test 

Code changes

Side effects

Related changes

 ```release-note

 ```


cofyc:
add jenkins file for master job too? maybe we can merge them together

onlymellb:
@cofyc This script can also be used by the master job so that the pr and master jobs can be unified.

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -0,0 +1,260 @@
...
+						export GOPATH=${WORKSPACE}/go
+						export PATH=${WORKSPACE}/go/bin:\$PATH
+						if ! hash hg 2>/dev/null; then
+							sudo yum install -y mercurial
+						fi
+						hg --version
+						make
+						rm -rf tests/images/test-apiserver/bin
+						make e2e-build
+						"""
```
cofyc: must run `make test GO_COVER=y` and report the results to `https://codecov.io` on master branch

note that I guess we should use the token in Jenkins credentials.

cofyc: or we can run it in github actions instead?

cofyc: cc @weekface @tennix what do you think?

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -0,0 +1,260 @@
...
+				}
+				stash excludes: "${PROJECT_DIR}/vendor/**", includes: "${PROJECT_DIR}/**", name: "tidb-operator"
+			}
+		}
+
+		node('k8s-kind') {
+				def WORKSPACE = pwd()
+				def E2E_IMAGE = "localhost:5000/pingcap/tidb-operator-e2e:${GITHASH.take(7)}"
+				def IMAGE_TAG = "localhost:5000/pingcap/tidb-operator:${GITHASH.take(7)}"
+				def APISERVER_IMAGE = "localhost:5000/pingcap/test-apiserver:${GITHASH.take(7)}"
```
aylei: looks like these lines have 2 indention

ci/pingcap_tidb_operator_build_kind.groovy
onlymellb: I think it's okay to add this logic to this script.

cofyc:
I'm going to merge this fix. We can add go coverage later.

PR #1140, opened 2019-11-13
fix webhook does not take effect

### LINKED ISSUE #1137: The webhook used by e2e does not take effect
## Bug Report

We use webhook in e2e to intercept pod delete events, only allow pods to be delete when it is not DDLOwner of tidb, not leader of pd.  But in fact, since the failurePolicy configured by this webhook is `ignore`, this causes the webhook check to have no effect in the e2e test. And we can see from the apiserver log that the certificate and domain name of this webhook also do not match. The error log is as follows:

```
E1113 05:55:45.792226       1 dispatcher.go:68] failed calling admission webhook "check-pod-before-delete.k8s.io": Post https://webhook-service.tidb-operator-e2e.svc:443/pods?timeout=30s: x509: certificate is valid for webhook-service.e2e.svc, not webhook-service.tidb-operator-e2e.svc
```

onlymellb:
fix #1137 
### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change

Side effects

Related changes

 ```release-note

 ```

### DIFF
--- a/tests/cmd/e2e/main.go
+++ b/tests/cmd/e2e/main.go
@@ -48,7 +48,7 @@ func main() {
 	ns := "e2e"
 
 	var err error
-	certCtx, err = apimachinery.SetupServerCert(ns, tests.WebhookServiceName)
+	certCtx, err = apimachinery.SetupServerCert("tidb-operator-e2e", tests.WebhookServiceName)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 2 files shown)


tests/cmd/e2e/main.go
```diff
@@ -48,7 +48,7 @@ func main() {
 	ns := "e2e"
 
 	var err error
-	certCtx, err = apimachinery.SetupServerCert(ns, tests.WebhookServiceName)
+	certCtx, err = apimachinery.SetupServerCert("tidb-operator-e2e", tests.WebhookServiceName)
```
shuijing198799: Why do you want to fix the ns used to generate the cert?

PR #1142, opened 2019-11-13
import cert-manager as depenedency

Yisaer:

Add `cert-manager` with latest released version (0.11.0) into `tidb-operator` chart dependecy and being default disabled. We use cert-manager to help us manage cert for `webhook`,`apiServer` and e2e tls. As cert-manager is not prod-ready yet, it is disabled in `values.yaml` in default.

Tests 

Code changes

 - Has Helm charts change

Side effects

 - N/A

None

### DIFF
--- a/charts/tidb-operator/charts/cert-manager/Chart.yaml
+++ b/charts/tidb-operator/charts/cert-manager/Chart.yaml
@@ -1,0 +1,16 @@
+appVersion: v0.11.0
+description: A Helm chart for cert-manager
+home: https://github.com/jetstack/cert-manager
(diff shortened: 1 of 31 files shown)


tennix:
I think it's unnecessary to copy cert-manager chart into our repo. We can use helm chart dependency or merely write document about how to deploy cert-manager.

Yisaer:
> I think it's unnecessary to copy cert-manager chart into our repo. We can use helm chart dependency or merely write document about how to deploy cert-manager.

ok, we could guide user in document how to install cert-manager.

Issue #1145, opened 2019-11-14
failed e2e test: pd is leader, can't be deleted namespace e2e name cluster5-pd-3


cofyc:
## Bug Report

https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/operator_ghpr_e2e_test_kind/detail/operator_ghpr_e2e_test_kind/930/pipeline

```
[2019-11-14T07:17:02.109Z] I1114 07:17:01.944146       1 actions.go:887] checking tidb cluster [e2e/cluster5] status

[2019-11-14T07:17:03.034Z] I1114 07:17:02.953017       1 pods.go:111] savely delete pod namespace e2e name cluster5-pd-4 leader name cluster5-pd-3

[2019-11-14T07:17:06.296Z] I1114 07:17:06.192993       1 actions.go:1206] index: 2,the schedulers: evict-leader-scheduler-5 = evict-leader-scheduler-5

[2019-11-14T07:17:06.296Z] E1114 07:17:06.193352       1 actions.go:1209] index: 1,the scheduler: evict-leader-scheduler-5 != evict-leader-scheduler-4

[2019-11-14T07:17:07.222Z] E1114 07:17:07.193978       1 actions.go:1209] index: 1,the scheduler: evict-leader-scheduler-5 != evict-leader-scheduler-4

[2019-11-14T07:17:08.583Z] E1114 07:17:08.193882       1 actions.go:1209] index: 1,the scheduler: evict-leader-scheduler-5 != evict-leader-scheduler-4

[2019-11-14T07:17:09.510Z] E1114 07:17:09.193856       1 actions.go:1209] index: 1,the scheduler: evict-leader-scheduler-5 != evict-leader-scheduler-4

[2019-11-14T07:17:10.435Z] E1114 07:17:10.193846       1 actions.go:1209] index: 1,the scheduler: evict-leader-scheduler-5 != evict-leader-scheduler-4

[2019-11-14T07:17:11.358Z] E1114 07:17:11.193850       1 actions.go:1209] index: 1,the scheduler: evict-leader-scheduler-5 != evict-leader-scheduler-4

[2019-11-14T07:17:12.282Z] E1114 07:17:12.193907       1 actions.go:1209] index: 1,the scheduler: evict-leader-scheduler-5 != evict-leader-scheduler-4
[...]

Issue #1112, opened 2019-11-07
Transfer PD leader before deleting member during PD scaling in


DanielZhangQD:
Currently, during scaling in PD, tidb-operator just deletes member directly, however, if the PD being scaled in is leader, it should transfer leader first.

DanielZhangQD:
Will be fixed by https://github.com/pingcap/tidb-operator/pull/1114

Yisaer:
This also shoud be fixed in Operator. https://github.com/pingcap/tidb-operator/pull/1114 only fix it in webhook server.

Yisaer:
fix in https://github.com/pingcap/tidb-operator/pull/1151

PR #1151, opened 2019-11-15
fix pd leader transferring during pd scale-in

Yisaer:
https://github.com/pingcap/tidb-operator/issues/1145
https://github.com/pingcap/tidb-operator/issues/1112
Currently we do not transfer pd leader during pd scale-in if the pod which was going to be deleted (and also was a pd leader). This request fix it by transfer pd leader to pd-0 during pd scale-in in this situation. 

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

 - N/A

Related changes

None

### DIFF
--- a/pkg/manager/member/pd_scaler.go
+++ b/pkg/manager/member/pd_scaler.go
@@ -116,7 +116,7 @@ func (psd *pdScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet,
 				return err
 			}
 			resetReplicas(newSet, oldSet)
-			return nil
+			return controller.RequeueErrorf("tc[%s/%s]'s pd pod[%s/%s] is transferring pd leader,can't scale-in now", ns, tcName, ns, memberName)
 		}
 	}
 
--- a/pkg/manager/member/pd_scaler.go
+++ b/pkg/manager/member/pd_scaler.go
@@ -116,6 +116,7 @@ func (psd *pdScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet,
 				return err
 			}
 		}
+		resetReplicas(newSet, oldSet)
 		return nil
 	}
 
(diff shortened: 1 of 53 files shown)


pkg/manager/member/pd_scaler.go
```diff
@@ -101,7 +101,26 @@ func (psd *pdScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet,
...
+	// If the pd pod was pd leader during scale-in, we would transfer pd leader to pd-0 directly
+	// If the pd statefulSet would be scale-in to zero and the pd-0 was going to be deleted,
+	// we would directly deleted the pd-0 without pd leader transferring
+	if ordinal > 0 {
+		leader, err := pdClient.GetPDLeader()
+		if err != nil {
+			return err
+		}
+		if leader.Name == memberName {
+			err = pdClient.TransferPDLeader(fmt.Sprintf("%s-pd-%d", tc.GetName(), 0))
```
cofyc: 1) this is used to transfer lead to the first pod (pod-0), but pod-0 maybe not working at the time of transferring 
2) does it take effect on the completion of this request? if not, the problem still exist

what do you think about this solution?

- delete current PD member (same as before)
- check if it is leader (if not return nil, if yes return error and let the controller to retry next time)

Yisaer: > 1. this is used to transfer lead to the first pod (pod-0), but pod-0 maybe not working at the time of transferring
> 2. does it take effect on the completion of this request? if not, the problem still exist
> 
> what do you think about this solution?
> 
> * delete current PD member (same as before)
> * check if it is leader (if not return nil, if yes return error and let the controller to retry next time)

Currently PD upgrading would transfer pd leader to the first and the last ordinal pd pod without considering whether both of them were available. I think it is necessary to check whether the target pd pod is working before transferring pd leader, maybe it should be fixed in another issue.

In my solution, if the target pd pod was the pd leader, it would be transferring pd leader to pd-0 requeue this request to wait for next round.

As for your solution, I don't understand how can the target pod ( which was also pd leader ) transferring pd leader If we only return nil.

cofyc: when the PD member is deleted and after all members receive this message, it will not be the leader anymore (in the background, the PD cluster will elect a new leader).

so, we can simply delete it and wait for it not be a leader (if it was a leader).

Yisaer: If we scale-in pd cluster from 5 to 3, it could be at most 2 times pd leader tranferring. First pd-4 to pd-3 , and Second pd-3 to pd -2, which I think would be high risky.

PR #1166, opened 2019-11-15
pdapi: update healthPrefix

cwen0:
use `pd/api/v1/health`  url instead of `pd/health` to check pd health 
```
$ curl pdip:2379/pd/health
404 page not found
$ curl pdip:2379/pd/api/v1/health
[
  {
    "name": "tidb-cluster-pd-0",
    "member_id": 17115972119073096751,
    "client_urls": [
      "http://tidb-cluster-pd-0.tidb-cluster-pd-peer.hibernate-regions-exp61-cat0-tidb-cluster.svc:2379"
    ],
    "health": true
  }
]
```
related pr: https://github.com/pingcap/pd/pull/1918

### DIFF
--- a/pkg/pdapi/pdapi.go
+++ b/pkg/pdapi/pdapi.go
@@ -246,6 +246,7 @@ func (pc *pdClient) GetHealth() (*HealthInfo, error) {
 	apiURL = fmt.Sprintf("%s/%s", pc.url, healthPrefix)
 	body, err = httputil.GetBodyOK(pc.httpClient, apiURL)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


gregwebs:
I think we should know the PD version when we make these API calls. There are different reasons why the API call could fail, and for the older version we are adding a useless API call

cwen0:
@gregwebs thanks! I will fix it!

cwen0:
I will close this pr, because pd have fix it in https://github.com/pingcap/pd/pull/1947

PR #1167, opened 2019-11-17
Add TidbCluster v1alpha2 type definition

aylei:
This PR adds the TidbCluster v1alpha2 type definition and the conversion between v1alpha1 <-> v1alpha2

core logic: https://github.com/pingcap/tidb-operator/pull/1167/commits/91294510ebe5481bfbc014c637d20a39a34c21bf


tennix:
I think we can extend the current v1alpha1 version for the CRD, it's compatible with the exsiting CRD definition. But use a different api group for AA api.

aylei:
> I think we can extend the current v1alpha1 version for the CRD, it's compatible with the exsiting CRD definition. But use a different api group for AA api.

Yes, AA will be introduced as needed, but not now for tidbcluster

PR #812, opened 2019-08-22
Keep the instance label of drainer same with the TiDB cluster in favor of monitoring

aylei:
close #811 

Signed-off-by: Aylei <rayingecho@gmail.com>

### Check List 

Tests 

 - Manual test (add detailed scripts or steps below)

Code changes

 - Has Helm charts change

### Does this PR introduce a user-facing change?:
 No.

PR #1170, opened 2019-11-18
Automated cherry pick of #812: Keep the instance label of drainer same with the TiDB cluster

aylei:
Cherry pick of #812 on release-1.0.

#812: Keep the instance label of drainer same with the TiDB cluster


aylei:
@tennix @weekface @cofyc PTAL

PR #1177, opened 2019-11-18
Limit the service accounts from pod request for webhook

Yisaer:
Currently pod delete webhook would affect each delete request for pd/tikv pod. This requests limit the service account from admission request to avoid affecting other client requests ( like kubectl )

Tests 

Code changes

 - Has Go code change

Side effects

 - N/A

 None

### DIFF
--- a/cmd/admission-controller/main.go
+++ b/cmd/admission-controller/main.go
@@ -37,9 +37,10 @@ import (
 )
 
 var (
-	printVersion bool
-	certFile     string
-	keyFile      string
+	printVersion         bool
+	certFile             string
+	keyFile              string
+	extraServiceAccounts string
 )
 
 func init() {
@@ -47,6 +48,7 @@ func init() {
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&certFile, "tlsCertFile", "/etc/webhook/certs/cert.pem", "File containing the x509 Certificate for HTTPS.")
 	flag.StringVar(&keyFile, "tlsKeyFile", "/etc/webhook/certs/key.pem", "File containing the x509 private key to --tlsCertFile.")
+	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "extra Service Accounts the Webhook should control")
 	features.DefaultFeatureGate.AddFlag(flag.CommandLine)
 	flag.Parse()
 }
@@ -72,6 +74,8 @@ func main() {
 		glog.Fatalf("failed to create Clientset: %v", err)
 	}
 
+	glog.Infof("extraServiceAccounts = %s", extraServiceAccounts)
+
 	var kubeCli kubernetes.Interface
 	kubeCli, err = kubernetes.NewForConfig(cfg)
 	if err != nil {
@@ -91,7 +95,7 @@ func main() {
 	informerFactory := informers.NewSharedInformerFactory(cli, controller.ResyncDuration)
 	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeCli, controller.ResyncDuration)
 
-	webhookServer := webhook.NewWebHookServer(kubeCli, cli, informerFactory, kubeInformerFactory, certFile, keyFile)
+	webhookServer := webhook.NewWebHookServer(kubeCli, cli, informerFactory, kubeInformerFactory, certFile, keyFile, extraServiceAccounts)
 	controllerCtx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
--- a/cmd/admission-controller/main.go
+++ b/cmd/admission-controller/main.go
@@ -48,7 +48,7 @@ func init() {
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&certFile, "tlsCertFile", "/etc/webhook/certs/cert.pem", "File containing the x509 Certificate for HTTPS.")
 	flag.StringVar(&keyFile, "tlsKeyFile", "/etc/webhook/certs/key.pem", "File containing the x509 private key to --tlsCertFile.")
-	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "extra Service Accounts the Webhook should control")
+	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "comma-separated, extra Service Accounts the Webhook should control. The full pattern for each common service account is system:serviceaccount:<namespace>:<serviceaccount-name>")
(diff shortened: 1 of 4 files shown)


tennix:
This feature may be used for graceful restart. So I suggest keeping it.

DanielZhangQD:
> This feature may be used for graceful restart. So I suggest keeping it for now.

@Yisaer Can we specify the service account or context to use the sts controller sa for API call or kubectl? If yes, the graceful restart can be done in this way.

Yisaer:
> the service account or context to

I don't get your point. Could you tell more detail ?

pkg/webhook/pod/pods.go
```diff
@@ -78,7 +82,10 @@ func (pc *PodAdmissionControl) AdmitPods(ar v1beta1.AdmissionReview) *v1beta1.Ad
 	name := ar.Request.Name
 	namespace := ar.Request.Namespace
 	operation := ar.Request.Operation
-	klog.Infof("receive admission to %s pod[%s/%s]", operation, namespace, name)
+	if stsControllerServiceAccounts != ar.Request.UserInfo.Username {
```
tennix: Make the skipping service account configurable

Yisaer: fix

cmd/admission-controller/main.go
```diff
@@ -37,16 +37,18 @@ import (
...
+	keyFile              string
+	extraServiceAccounts string
 )
 
 func init() {
 	flag.BoolVar(&printVersion, "V", false, "Show version and quit")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&certFile, "tlsCertFile", "/etc/webhook/certs/cert.pem", "File containing the x509 Certificate for HTTPS.")
 	flag.StringVar(&keyFile, "tlsKeyFile", "/etc/webhook/certs/key.pem", "File containing the x509 private key to --tlsCertFile.")
+	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "extra Service Accounts the Webhook should control")
```
tennix: It's comma separated service accounts. 
The pre-defined service account is `system:serviceaccount:kube-system:statefulset-controller`, how users find the service account full name?

pkg/webhook/pod/pods.go
```diff
@@ -78,7 +86,21 @@ func (pc *PodAdmissionControl) AdmitPods(ar v1beta1.AdmissionReview) *v1beta1.Ad
 	name := ar.Request.Name
 	namespace := ar.Request.Namespace
 	operation := ar.Request.Operation
-	klog.Infof("receive admission to %s pod[%s/%s]", operation, namespace, name)
+	serviceAccount := ar.Request.UserInfo.Username
+	klog.Infof("receive %s pod[%s/%s] by sa[%s]", operation, namespace, name, serviceAccount)
+
+	isUnknownServiceAccounts := true
```
tennix:
```suggestion
	skippableSA := true
```

pkg/webhook/pod/pods.go
```diff
@@ -78,7 +86,21 @@ func (pc *PodAdmissionControl) AdmitPods(ar v1beta1.AdmissionReview) *v1beta1.Ad
...
+	for _, sa := range pc.serviceAccounts {
```
DanielZhangQD: Suggest converting the SAs to a map[string]struct{}.

pkg/webhook/pod/pods.go
```diff
@@ -78,7 +86,21 @@ func (pc *PodAdmissionControl) AdmitPods(ar v1beta1.AdmissionReview) *v1beta1.Ad
...
+		if sa == serviceAccount {
+			isUnknownServiceAccounts = false
+			break
+		}
+	}
+
+	if isUnknownServiceAccounts {
+		klog.Infof("Request was not sent by Knowing Controlled ServiceAccounts,Admit to %s pod[%s/%s]", operation, namespace, name)
```
DanielZhangQD:
```suggestion
		klog.Infof("Request was not sent by known controlled ServiceAccounts, admit to %s pod [%s/%s]", operation, namespace, name)
```

cmd/admission-controller/main.go
Yisaer: > It's comma separated service accounts.
> The pre-defined service account is `system:serviceaccount:kube-system:statefulset-controller`, how users find the service account full name?

Currently the pattern for serviceaccounts for the common case would be `system:serviceaccount:<namespace>:<serviceaccount-name>`.
And I prefer not to pre-defined any service account pattern here.

pkg/webhook/pod/pods.go
Yisaer: The `pc. serviceAccounts` was a string list and I don't think it's necessary to fetch serviceAccounts from apiServer here. The comparison of string is enough.

cmd/admission-controller/main.go
DanielZhangQD: Please make it clear in the description, that this option is comma-separated SAs and how to get the full name.

pkg/webhook/pod/pods.go
Yisaer: fix

cmd/admission-controller/main.go
Yisaer: fix

pkg/webhook/pod/pods.go
```diff
@@ -51,25 +51,37 @@ type PodAdmissionControl struct {
...
+
+func NewPodAdmissionControl(kubeCli kubernetes.Interface, operatorCli versioned.Interface, PdControl pdapi.PDControlInterface, informerFactory informers.SharedInformerFactory, kubeInformerFactory kubeinformers.SharedInformerFactory, recorder record.EventRecorder, extraServiceAccounts []string) *PodAdmissionControl {
 
 	pvcInformer := kubeInformerFactory.Core().V1().PersistentVolumeClaims()
 	PVCControl := controller.NewRealPVCControl(kubeCli, recorder, pvcInformer.Lister())
 	tcLister := informerFactory.Pingcap().V1alpha1().TidbClusters().Lister()
 
 	podLister := kubeInformerFactory.Core().V1().Pods().Lister()
 	stsLister := kubeInformerFactory.Apps().V1().StatefulSets().Lister()
+	serviceaccounts := map[string]interface{}{}
```
weekface: Use [`sets.NewString`](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go#L30:6) instead?

Yisaer: fix.

Issue #1121, opened 2019-11-08
Migrating Helm chart template logic to server side


tennix:
## Feature Request

Currently, many resources are created and managed solely by the Helm chart. Embedding logic in the chart template makes it hard to add some features, such as the unit computation and conversion between PD/TiKV's and K8s's. It also makes it difficult to write automatic tests, validation logic and backward compatible code. Besides, users have to use both kubectl and helm to manage tidb clusters, this adds extra complexity.

Now we have introduced [aggregated apiserver](https://github.com/pingcap/tidb-operator/issues/1002). We should migrate Helm chart template logic to the controller-manager or aggregated apiserver.

Below are lists of resources we should migrate to server-side:

- [x] tidb-discovery service  #1322 
- [ ] tidb-initializer job #1262
- [x] tidb service #1190 #1242
- [x] tidb configmaps and start-script #1192 
- [x] pd configmaps and start-script #1263 https://github.com/pingcap/tidb-operator/pull/1330
- [x] tikv configmaps and start-script #1264 https://github.com/pingcap/tidb-operator/pull/1330
- [x] pump #1191 
- [ ] monitor #1261

ref: server dry-run https://github.com/pingcap/tidb-operator/issues/999

gregwebs:
This makes it harder for users to hack changes (Now they have to compile tidb-operator) or to even understand ahead of time what would happen. I am wondering if we can have a dry run feature to emit all the resources that tidb-operator would create. I guess this is similar to another feature request to have a plan feature like terraform, but that may be more focused on update?

tennix:
Since we will have the aggregated apiserver, before the CRs are written into k8s, we can emit a plan (server-side dry run). I guess this can solve your worries.
But yes, to change the spec or resources we don't expose requires changing Go code can re-compile tidb-operator. We lose some flexibility but gain more correctness. Just as dynamic programming languages vs. static programming languages.

gregwebs:
yes, server side dry-run (create and update) would help.

aylei:
@tennix @gregwebs I reference the dry-run issue in the description, however, one thing worries me is that the PD/TiKV/TiDB are still created with ordered dependencies in our controller. I remember that you have a proposal to create them simultaneously, how is it going then? @gregwebs

gregwebs:
> how is it going then? @gregwebs
We just need to finalize testing on the PR now.

cofyc:
should we remove dependencies on these annotations keys:

[...]

PR #1193, opened 2019-11-19
Extend tidbcluster v1alpha1 to support configuration and pump

### LINKED ISSUE #1159: Extend TidbCluster v1alpha to support the spec of pump, configuration and discovery service
## Proposal

As per #1121 , we want to extend TidbCluster Resource to replace  (or part of) the helm chart.

example:

```
apiVersion: pingcap.com/v1alpha2
kind: TidbCluster
metadata: {...}
spec:
 version: v3.0.5             # default: latest recommended stable version
 pd:
   baseImage: pingcap/pd     # default
   replicas: 3               # required
   resources: {...}          # default: {}
   config:
     log:
       level: info           # default
 tikv: {...}
 tidb: {...}
 pump: {...}
```

### LINKED ISSUE #1182: Support cluster-level setting for components
## Feature Request

Formerly, we rely on template tools like `helm` to coordinate fields and eliminate copy-paste redundancy. Now that we are going to migrate the logic of helm to server-side #1121 , `TidbCluster` should support cluster-level setting for components. Of course, each component could still specify the properties and override the cluster one.

## Proposal

Adding the following fields in `TidbClusterSpec`:

``` diff

type TidbClusterSpec struct {

+	// Time zone of TiDB cluster Pods
+	Timezone string `json:"timezone,omitempty"`
+	// ImagePullPolicy of TiDB cluster Pods
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
+	// Whether Hostnetwork is enabled for TiDB cluster Pods
+	HostNetwork *bool `json:"hostNetwork,omitempty"`
+	// Affinity of TiDB cluster Pods
+	Affinity *corev1.Affinity `json:"affinity,omitempty"`
+	// PriorityClassName of TiDB cluster Pods
+	PriorityClassName string `json:"priorityClassName,omitempty"`
+	// Base node selectors of TiDB cluster Pods, components may add or override selectors upon this respectively
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+	// Base annotations of TiDB cluster Pods, components may add or override selectors upon this respectively
+	Annotations map[string]string `json:"annotations,omitempty"`
+	// Base tolerations of TiDB cluster Pods, components may add more tolreations upon this respectively
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
}
```

aylei:
This  PR extends the TidbCluster v1alpha1 type to support the management of component configuration, service and pump.

Support cluster-level setting for components.

There is no logic change introduced.

Next step: controllers, as stated in #1121

close #1182 
close #1159 

 ```release-note
Support cluster-level setting for components 
 ```


weekface:
Is it compatible with `1.0.x`? How do these `1.0.x` users upgrade?

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -13,6 +13,166 @@
...
+	baseImage := a.ComponentSpec.BaseImage
+	// base image takes higher priority
+	if baseImage != "" {
+		version := a.ComponentSpec.Version
+		if version == "" {
+			version = a.ClusterSpec.Version
+		}
+		image = fmt.Sprintf("%s:%s", baseImage, version)
+	}
+	return image
```
DanielZhangQD: image requires a tag and baseImage cannot have a tag? Why define two images in componentSpec?

aylei:
> Is it compatible with `1.0.x`? How do these `1.0.x` users upgrade?

Yes, no actions required as stated in the design doc. If use specify new specs added in this PR, for example, `.spec.tidb.service`, the control-loop will try to adopt the orphaned objects created by helm first.

pkg/apis/pingcap/v1alpha1/tidbcluster.go
aylei: `version` is used as tag, effectively

For backward compatibility, we keep the `image` field.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -90,19 +91,64 @@ type TidbClusterList struct {
 // +k8s:openapi-gen=true
 // TidbClusterSpec describes the attributes that a user creates on a tidb cluster
 type TidbClusterSpec struct {
-	SchedulerName string   `json:"schedulerName,omitempty"`
-	PD            PDSpec   `json:"pd,omitempty"`
-	TiDB          TiDBSpec `json:"tidb,omitempty"`
-	TiKV          TiKVSpec `json:"tikv,omitempty"`
-	// +k8s:openapi-gen=false
-	TiKVPromGateway TiKVPromGatewaySpec `json:"tikvPromGateway,omitempty"`
```
aylei: This field is not used anymore, removing this is backward compatible as long as the validation code do not reject this field

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -13,6 +13,166 @@
...
+}
+
+func (a *componentAccessorImpl) Image() string {
+	image := a.ComponentSpec.Image
+	baseImage := a.ComponentSpec.BaseImage
+	// base image takes higher priority
+	if baseImage != "" {
+		version := a.ComponentSpec.Version
+		if version == "" {
+			version = a.ClusterSpec.Version
```
aylei: validation constraints will guarantee either .spec.version or .spec.<component>.version is not-empty if `baseImage` is present

pkg/util/util.go
```diff
@@ -26,7 +26,7 @@ import (
 
 // ResourceRequirement creates ResourceRequirements for MemberSpec
 // Optionally pass in a default value
-func ResourceRequirement(spec v1alpha1.ContainerSpec, defaultRequests ...corev1.ResourceRequirements) corev1.ResourceRequirements {
+func ResourceRequirement(spec v1alpha1.Resources, defaultRequests ...corev1.ResourceRequirements) corev1.ResourceRequirements {
```
tennix: Rename the variable name?

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -90,19 +91,64 @@ type TidbClusterList struct {
...
+	// SchedulerName of TiDB cluster Pods
+	SchedulerName string `json:"schedulerName,omitempty"`
+
+	// Persistent volume reclaim policy applied to the PVs that consumed by TiDB cluster
 	PVReclaimPolicy corev1.PersistentVolumeReclaimPolicy `json:"pvReclaimPolicy,omitempty"`
-	EnablePVReclaim bool                                 `json:"enablePVReclaim,omitempty"`
-	Timezone        string                               `json:"timezone,omitempty"`
+
+	// Whether enable PVC reclaim for orphan PVC left by statefulset scale-in
+	EnablePVReclaim bool `json:"enablePVReclaim,omitempty"`
```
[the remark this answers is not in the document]
aylei: I suggest `PVCReclaimPolicy` with enum value `Deferred` and `Immediate` if we are going to do the renaming

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -90,19 +91,64 @@ type TidbClusterList struct {
 // +k8s:openapi-gen=true
 // TidbClusterSpec describes the attributes that a user creates on a tidb cluster
 type TidbClusterSpec struct {
```
aylei: Modification of `.status` will be introduced along with the controller logic

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -202,15 +202,15 @@ type TiDBSpec struct {
 	ComponentSpec
 	// +k8s:openapi-gen=false
 	Resources
+	Replicas         int32            `json:"replicas"`
+	Service          *TiDBServiceSpec `json:"service,omitempty"`
+	StorageClassName string           `json:"storageClassName,omitempty"`
```
tennix: Where this is needed?

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -128,31 +128,42 @@ func (a *componentAccessorImpl) Annotations() map[string]string {
...
 func (tc *TidbCluster) BasePDSpec() ComponentAccessor {
 	return &componentAccessorImpl{&tc.Spec, &tc.Spec.PD.ComponentSpec}
 }
 
-func (tc *TidbCluster) BasePumpSpec() ComponentAccessor {
-	return &componentAccessorImpl{&tc.Spec, &tc.Spec.Pump.ComponentSpec}
+// BasePumpSpec returns two results:
+// 1. the base pump spec, if exists.
+// 2. whether the base pump spec exists.
+func (tc *TidbCluster) BasePumpSpec() (ComponentAccessor, bool) {
```
tennix: The `nil` already report the non-existence of pump spec. Is it duplicate to return a bool?

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -128,31 +128,42 @@ func (a *componentAccessorImpl) Annotations() map[string]string {
 }
 
 func (a *componentAccessorImpl) Tolerations() []corev1.Toleration {
-	tols := a.ClusterSpec.Tolerations
-	tols = append(tols, a.ComponentSpec.Tolerations...)
+	tols := a.ComponentSpec.Tolerations
```
tennix: So the component tolerations just override the cluster level tolerations? Better add a document for this method. For it's different from other fields like node selectors.

pkg/apis/pingcap/v1alpha1/tidbcluster.go
aylei: An extra returned result is designed for enforcing the programmer to check whether the pump spec is present.

Otherwise one has to remember check whether `ComponentAccssor` is nil, which is impossible in other `Base<Component>Spec` method

pkg/apis/pingcap/v1alpha1/types.go
aylei: Good catch

pkg/apis/pingcap/v1alpha1/tidbcluster.go
aylei: Yes, the field comment reveals this:

```
	// Tolerations of the component. Override the cluster-level tolerations if non-empty
	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
```

After we migrate from helm to API object, I plan to generate the api document based on the openapi schema, which has those comments information.

tennix: I mean users calling this function can know what this fuction does.

pkg/apis/pingcap/v1alpha1/tidbcluster.go
tennix: OK, got it.

aylei:
@tennix 
After I remove the `.spec.tidb.storageClassName`field, I find that this field has been existing from the very beginning and there is a unit test for it.

We can remove this field and adjust the UT because it is not used anywhere, but I suggest leave it as is because TiDB MAY use persistent volume according to the RFC https://github.com/pingcap/tidb/pull/13481, which introduce a built-in logging aggregation mechanism that requires local persistent (for a short time) log storage.

aylei:
@tennix 
After I remove the `.spec.tidb.storageClassName`field, I find that this field has been existing from the very beginning and there is a unit test for it.

We can remove this field and adjust the UT because it is not used anywhere, but I suggest leave it as is because TiDB MAY use persistent volume according to the RFC https://github.com/pingcap/tidb/pull/13481, which introduce a built-in logging aggregation mechanism that requires local persistent (for a short time) log storage.

aylei:
@tennix @weekface PTAL again

tennix:
> @tennix
> After I remove the `.spec.tidb.storageClassName`field, I find that this field has been existing from the very beginning and there is a unit test for it.
> 
> We can remove this field and adjust the UT because it is not used anywhere, but I suggest we leaving it as is because TiDB MAY use persistent volume according to the RFC [pingcap/tidb#13481](https://github.com/pingcap/tidb/pull/13481), which introduce a built-in logging aggregation mechanism that requires local persistent (for a short time) log storage.

OK, sounds good to me.

weekface:
the conflicts should be resolved

aylei:
@weekface conflicts addressed, PTAL

@tennix PTAL again

&
/run-e2e-in-kind

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -13,6 +13,177 @@
...
+
+func (a *componentAccessorImpl) PodSecurityContext() *corev1.PodSecurityContext {
+	return a.ComponentSpec.PodSecurityContext
+}
+
+func (a *componentAccessorImpl) ImagePullPolicy() corev1.PullPolicy {
+	pp := a.ComponentSpec.ImagePullPolicy
+	if pp == nil {
+		pp = &a.ClusterSpec.ImagePullPolicy
+	}
```
onlymellb: `ClusterSpec.ImagePullPolicy` must not be empty?

pkg/apis/pingcap/v1alpha1/tidbcluster.go
onlymellb: The tidb spec can also be empty, for example, I just want to deploy tikv for storage.

pkg/apis/pingcap/v1alpha1/tidbcluster.go
aylei: Yes, there will be a defaulting through mutating-webhook and validation to ensure clusterSpec.ImagePullPolicy and component.ImagePullPolicy is not both empty.

aylei: And it's not an issue if clusgerSpec.ImagePullPolicy is empty, we just write empty pull policy in PodSpec and use the default of kubernetes

PR #1195, opened 2019-11-20
add 3 ways for initialize webhook cert

Yisaer:
We need initialize cert for webhook server , and this pr provide 3 ways to initialize webhook cert.
1. Using helm hook to initialize webhook cert like [prometheus-operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator/templates/prometheus-operator/admission-webhooks/job-patch)
2. User self-provider.
3. Managing by cert-manager.

For the first option, helm hook would create a job to initialize secrets in the released namespace before `operator` installed.

For the second option, user could provide their own cert in the secret,and the `secret`  is part of operator chart template.

For the last option, the cert secret would be managed by [cert-manager](https://docs.cert-manager.io/en/latest/). And the `cert-manager` should >= 0.11.0 version and caInject enabled.

Currently the webhook cert initializer was disabled in `values.yaml` so there would be no user-facing change.

Tests 

Code changes

 - Has Go code change
 - Has Helm change

Side effects

 - N/A

None


charts/tidb-operator/templates/admission/certManager/certificate.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if and (eq .Values.admissionWebhook.enabled true) (eq .Values.admissionWebhook.certProvider "certManager") }}
+apiVersion: cert-manager.io/v1alpha2
+kind: Certificate
+metadata:
+  name: tidb-admission-webhook-crt
+  namespace: {{ $.Release.Namespace }}
```
aylei: This is redundant

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  template:
+    metadata:
+      name: {{ template "chart.name" . }}-admission-delete
+    spec:
+      restartPolicy: Never
+      serviceAccountName: tidb-admission-webhook-sa
+      containers:
+        - name: "{{.Release.Name}}-job-del"
+          image: bitnami/kubectl:1.15.3
```
aylei: we should always make image configurable, in case of private or firewalled env

charts/tidb-operator/templates/admission/jobPatch/pre-install-secret-job.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if and (eq .Values.admissionWebhook.enabled true) (eq .Values.admissionWebhook.certProvider "patch") }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name:  {{ template "chart.name" . }}-admission-create
+  namespace: {{ $.Release.Namespace }}
```
aylei: ditto

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if and (eq .Values.admissionWebhook.enabled true) (eq .Values.admissionWebhook.certProvider "patch") }}
```
aylei: `certProvider: patch` sounds weird, how can `patch` be a provider?

charts/tidb-operator/values.yaml
```diff
@@ -175,3 +175,21 @@ advancedStatefulset:
...
+    image:
+      repository: jettech/kube-webhook-certgen
+      tag: v1.2.0
+  ## `selfProvider` need user to provide their own ca/cert/key for webhook
+  selfProvider:
+    ca: ""
+    key: ""
+    cert: ""
+  ## `certManager` would create an self-signed Issuer and a Certificate for webhook cert managed by cert-manager
+  certManager:
```
aylei: what is this for?

Yisaer: Currently this is unused, only pre-defined for the future.

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
Yisaer: Using `autoGenerator` instead

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
Yisaer: fix

Yisaer: fix

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
```diff
@@ -0,0 +1,34 @@
...
+      containers:
+        - name: "{{.Release.Name}}-job-del"
+          image: {{ .Values.admissionWebhook.autoGenerator.kubectl.image.repository }}:{{ .Values.admissionWebhook.autoGenerator.kubectl.image.tag }}
+          imagePullPolicy: {{ .Values.imagePullPolicy | default "IfNotPresent" }}
+          command:
+            - "sh"
+            - "-c"
+            - |
+              set -e
+              kubectl delete secrets -n {{ .Release.Namespace }} tidb-admission-webhook-certs || true
```
aylei: could we retry?

Yisaer: > could we retry?

I think this depends on is there any necessary to ensure the secret would be deleted after deleting the chart.

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
```diff
@@ -0,0 +1,34 @@
...
+  labels:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component:  admission-cert
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": post-delete
```
tennix: With AA, we'll deprecate the Helm Chart management for TiDB cluster. But for tidb-operator, it's now more and more tightly coupled with Helm.
I'm wondering how much work we need to do in the future to make tidb-operator deployable with only `kubectl`.

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
tennix: Can we just add this in the document instead of doing it automatically?

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
Yisaer: Here using `helm hook` is better but not necessary as each resources could retry until their dependency ready.

charts/tidb-operator/templates/admission/jobPatch/post-del-secret-job.yaml
Yisaer: > Can we just add this in the document instead of doing it automatically? It's only a one time and deadly simple command.

sure

charts/tidb-operator/templates/admission/certManager/Issuer.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if and (eq .Values.admissionWebhook.enabled true) (eq .Values.admissionWebhook.certProvider "certManager") }}
```
aylei: filename should start in lower case

PR #1207, opened 2019-11-21
upload unit tests coverage report to codecov

onlymellb:
### What problem does this PR solve? 
* upload unit tests coverage report to codecov
* fix master job build failed. The solution comes from @cofyc 

### Check List 

Tests 

 - Unit test

Code changes

 - Has CI related scripts change

### Does this PR introduce a user-facing change?:
 
 ```release-note
None
 ```

PR #1208, opened 2019-11-21
fix e2e job on master branch

cofyc:

we should fetch heads refs too.

### Check List 

Tests 

 - Unit test
 - E2E test


onlymellb:
@cofyc I tested this solution and found that it still doesn't work, I fixed this problem in this pr #1207, I have tested it and can work

cofyc:
closing this in favor of https://github.com/pingcap/tidb-operator/pull/1207.

Issue #1139, opened 2019-11-13
webhook for Tikv being deleted.


Yisaer:
## Feature Request

Currently Operator needs to operate each componenet instance during upgrading and scaling. We want to put these dealing process into webhook when any instance being deleted and Operator would only need edit StatefulSet. 
https://github.com/pingcap/tidb-operator/issues/992

For now , we have succefully transfer the process of `PD` which is still in test.
https://github.com/pingcap/tidb-operator/pull/1101

Next, we'll start to transfer TiKV upgrading and scaling process into webhook.

PR #1218, opened 2019-11-22
Tikv Upgrading and Scale-in Controlled by Webhook

Yisaer:
https://github.com/pingcap/tidb-operator/issues/1139
This request let webhook control the `tikv` lifecycle during Upgrading and Scale-in by `delete pod` and `create pod` webhook

Also I trasnferred admission package we use from `v1beta1` version to `v1`

Tests 

 - Unit test

Code changes

 - Has Go code change


pkg/webhook/pod/pods.go
```diff
@@ -26,7 +29,7 @@ import (
 	"github.com/pingcap/tidb-operator/pkg/pdapi"
 	operatorUtils "github.com/pingcap/tidb-operator/pkg/util"
 	"github.com/pingcap/tidb-operator/pkg/webhook/util"
-	"k8s.io/api/admission/v1beta1"
+	admission "k8s.io/api/admission/v1beta1"
```
weekface: why not use the `v1`?

Yisaer: fix

pkg/webhook/pod/pods.go
```diff
@@ -192,8 +181,67 @@ func (pc *PodAdmissionControl) admitDeletePods(name, namespace string) *v1beta1.
...
+		klog.Errorf("failed to get owner statefulset,refuse to create pod[%s/%s]", namespace, name)
+		return util.ARFail(err)
+	}
+
+	ordinal, err := operatorUtils.GetOrdinalFromPodName(name)
+	if err != nil {
+		return util.ARFail(err)
+	}
+
+	if ownerStatefulSet.Status.CurrentReplicas < 1 && ownerStatefulSet.Status.UpdatedReplicas < 1 && ordinal == 0 {
```
DanielZhangQD: What's this case for?

pkg/webhook/pod/tikv_creater.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func (pc *PodAdmissionControl) admitCreateTiKVPod(pod *core.Pod, tc *v1alpha1.TidbCluster, pdClient pdapi.PDClient) *admission.AdmissionResponse {
+
+	name := pod.Name
+	namespace := pod.Namespace
+
+	stores, err := pdClient.GetStores()
+	if err != nil {
+		klog.Infof("failed to create pod[%s/%s],%v", namespace, name, err)
```
DanielZhangQD:
```suggestion
		klog.Infof("Failed to get stores during pod [%s/%s] creation, error: %v", namespace, name, err)
```

pkg/webhook/pod/tikv_creater.go
```diff
@@ -0,0 +1,52 @@
...
+	evictLeaderSchedulers, err := pdClient.GetEvictLeaderSchedulers()
+	if err != nil {
+		klog.Infof("failed to create pod[%s/%s],%v", namespace, name, err)
+		return util.ARFail(err)
+	}
+
+	// if the pod which is going to be created already have a store and was in evictLeaderSchedulers,
+	// we should end this evict leader
+	for _, store := range stores.Stores {
+		ip := strings.Split(store.Store.GetAddress(), ":")[0]
```
DanielZhangQD: Rename `ip`?

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -0,0 +1,208 @@
...
+	return util.ARSuccess()
+}
+
+func (pc *PodAdmissionControl) admitDeleteOfflineTiKVPod() *admission.AdmissionResponse {
+	return &admission.AdmissionResponse{
+		Allowed: false,
+	}
+}
+
+func (pc *PodAdmissionControl) admitDeleteDownTiKVPod(isInOrdinal bool, pod *core.Pod, ownerStatefulSet *apps.StatefulSet, tc *v1alpha1.TidbCluster, pdClient pdapi.PDClient) *admission.AdmissionResponse {
```
DanielZhangQD: The functions are the same for Down, NonStore, and Tombstone.

pkg/webhook/pod/tikv_util.go
```diff
@@ -0,0 +1,139 @@
...
+	klog.Infof("tikv upgrader: set pod %s/%s annotation %s to %s successfully",
+		namespace, name, EvictLeaderBeginTime, now)
+
+	return nil
+}
+
+func isTiKVReadyToUpgrade(upgradePod *core.Pod, store *pdapi.StoreInfo) bool {
+
+	if store.Status.LeaderCount == 0 {
+		klog.Infof("pod[%s/%s] is no region leader in store[%d]", upgradePod.Namespace, upgradePod.Name, store.Store.Id)
```
DanielZhangQD:
```suggestion
		klog.Infof("pod[%s/%s] has no region leader in store[%d]", upgradePod.Namespace, upgradePod.Name, store.Store.Id)
```

pkg/webhook/pod/tikv_deleter.go
Yisaer: > The functions are the same for Down, NonStore, and Tombstone.

Yep, separating into different functions is intended if they become different in future.

pkg/webhook/pod/pods.go
Yisaer: For initialization. When we receive request for creating tikv-0, there are 2 cases: 
  1. initialize tikv
  2.upgrading tikv

In second case, we shall end evict leader for tikv-1 and in first case, we shall let the request pass.

pkg/webhook/pod/tikv_creater.go
Yisaer: The `ip` here is temp variable. Only helping to get pod name from store info.

pkg/webhook/pod/tikv_creater.go
```diff
@@ -0,0 +1,65 @@
...
+
+	stores, err := pdClient.GetStores()
+	if err != nil {
+		klog.Infof("Failed to get stores during pod [%s/%s] creation, error: %v", namespace, name, err)
+		return util.ARFail(err)
+	}
+	evictLeaderSchedulers, err := pdClient.GetEvictLeaderSchedulers()
+	if err != nil {
+		klog.Infof("failed to create pod[%s/%s],%v", namespace, name, err)
+		return util.ARFail(err)
```
tennix: When TiKV is not bootstrapped, PD will respond an error when the client accesses `/pd/api/v1/schedulers` which is the endpoint we are visiting in `GetEvictLeaderSchedulers` method.
So will the cluster never get bootstrapped if we block tikv pod creation?

Yisaer: > When TiKV is not bootstrapped, PD will respond an error when the client accesses `/pd/api/v1/schedulers` which is the endpoint we are visiting in `GetEvictLeaderSchedulers` method.
> So will the cluster never get bootstrapped if we block tikv pod creation?

In `AdmitCreatePods` func in `pkg/webhook/pod/pods.go` would let the request pass directly when `ownerStatefulSet.Status.CurrentReplicas < 1 && ownerStatefulSet.Status.UpdatedReplicas < 1 && ordinal == 0` .  In this case, if we create tikv with 3 replicas , the request for creating tikv-0, tikv-1,tikv-2 would be sent at the same time. The tikv-0 would be admited to created while tikv-1 and tikv-2 would be refused to created as they couldn't visit the tikv store Info. When tikv-0 started successfully, tikv-1 and tikv-2 would be created successfully.

pkg/webhook/pod/tikv_creater.go
```diff
@@ -0,0 +1,52 @@
+package pod
+
+import (
+	"fmt"
```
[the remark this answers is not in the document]
Yisaer: fix

pkg/webhook/pod/tikv_creater.go
tennix: This is unreasonable, why tikv needs a seed pod? The way tikv works shouldn't rely on this. What do you think? @weekface

Yisaer: The main problem is that what should we do if we couldn't visit tikv store info (like http status error ) when we receive creating tikv pod request. If we refuses all the requests, then we couldn't create tikv during tikv bootstrapped. If we admit all the requests, then we couldn't end evict leader for former pod.

weekface: Now, `GetEvictLeaderSchedulers` will return an error when the TiKV cluster was not bootstrapped.

We should change the method `GetEvictLeaderSchedulers` to return an empty array(means there are no schedules) when TiKV was bootstrapping.

We needn't a seed pod.

Yisaer: > Now, `GetEvictLeaderSchedulers` will return an error when the TiKV cluster was not bootstrapped.
> 
> We should change the method `GetEvictLeaderSchedulers` to return an empty array(means there are no schedules) when TiKV was bootstrapping.
> 
> We needn't a seed pod.

I think this method is better. I would accept it and remove seed pod method.

Yisaer:
As @weekface 's advice, now tikv create hook would let the request pass if pd return `TiKV cluster not bootstrapped, please start TiKV first` instead of seed pod. @tennix

pkg/webhook/pod/util.go
```diff
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	FAIL_TO_FIND_TIDB_COMPONENT_OWNER_STATEFULSET = "failed to find owner statefulset for pod[%s/%s]"
+	FailToFindTidbComponentOwnerStatefulset = "failed to find owner statefulset for pod[%s/%s]"
```
DanielZhangQD: The error message has nothing to do with `TidbComponent`.
BTW, is it necessary to export here?

Yisaer: Yes, in case of deleting pod, we need to know whether the owner statefulset of tidb component ( not all the pods ) is whether existed. If not existed, the request to delete pod shall be admitted.

pkg/httputil/httputil.go
```diff
@@ -42,14 +42,14 @@ func GetBodyOK(httpClient *http.Client, apiURL string) ([]byte, error) {
...
 	defer DeferClose(res.Body)
-	if res.StatusCode >= 400 {
-		errMsg := fmt.Errorf("Error response %v URL %s", res.StatusCode, apiURL)
-		return nil, errMsg
-	}
 	body, err := ioutil.ReadAll(res.Body)
 	if err != nil {
 		return nil, err
 	}
+	if res.StatusCode >= 400 {
```
weekface: why move these lines here?

pkg/webhook/pod/tikv_creater.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+func (pc *PodAdmissionControl) admitCreateTiKVPod(pod *core.Pod, tc *v1alpha1.TidbCluster, pdClient pdapi.PDClient) *admission.AdmissionResponse {
+
+	name := pod.Name
+	namespace := pod.Namespace
+
+	stores, err := pdClient.GetStores()
+	if err != nil {
+		if strings.HasSuffix(err.Error(), tikvNotBootstrapped) {
```
weekface: getStore API also returns `tikvNotBootstrapped `?

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	switch storeInfo.Store.StateName {
+	case v1alpha1.TiKVStateTombstone:
+		return pc.admitDeleteTombStoneTiKVPod(isInOrdinal, pod, ownerStatefulSet, tc, pdClient)
+	case v1alpha1.TiKVStateOffline:
+		return pc.admitDeleteOfflineTiKVPod()
+	case v1alpha1.TiKVStateDown:
+		return pc.admitDeleteDownTiKVPod(isInOrdinal, pod, ownerStatefulSet, tc, pdClient)
+	case v1alpha1.TiKVStateUp:
```
weekface: add a default case to just print some logs?

pkg/webhook/pod/tikv_creater.go
```diff
@@ -0,0 +1,76 @@
...
+			return util.ARSuccess()
+		}
+		klog.Infof("failed to create pod[%s/%s],%v", namespace, name, err)
+		return util.ARFail(err)
+	}
+
+	// if the pod which is going to be created already have a store and was in evictLeaderSchedulers,
+	// we should end this evict leader
+	for _, store := range stores.Stores {
```
tennix: This too deeply nested block can be avoided by recording the schedulers in a map. And even better, when there is no evict leader schedulers, we don't have to loop over all the stores.

pkg/webhook/pod/tikv_creater.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+	"github.com/pingcap/tidb-operator/pkg/pdapi"
+	"github.com/pingcap/tidb-operator/pkg/webhook/util"
+	admission "k8s.io/api/admission/v1"
+	core "k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+const (
+	tikvNotBootstrapped = "\"TiKV cluster not bootstrapped, please start TiKV first\"\n"
```
tennix: Use backquote to avoid escape the quote.

pkg/httputil/httputil.go
Yisaer: > why move these lines here?

Cause I need to add the response body into error Info even when the status code of response is larger than 400.
Currently if the code > 400, the response body wouldn't be read.

pkg/webhook/pod/tikv_creater.go
Yisaer: > getStore API also returns `tikvNotBootstrapped `?

yes.

pkg/webhook/pod/pods.go
```diff
@@ -192,8 +181,52 @@ func (pc *PodAdmissionControl) admitDeletePods(name, namespace string) *v1beta1.
...
+
+	klog.Infof("receive admission to %s pod[%s/%s]", "create", namespace, name)
+
+	l := label.Label(pod.Labels)
+	if !(l.IsPD() || l.IsTiKV() || l.IsTiDB()) {
+		klog.Infof("pod[%s/%s] is not TiDB component,admit to create", namespace, name)
+		return util.ARSuccess()
+	}
+
+	tcName, exist := pod.Labels[label.InstanceLabelKey]
```
tennix: Why fail on this when there is no instance label? If a pod does not have the instance label, it may not be managed by tidb-operator.

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -0,0 +1,221 @@
...
+		if err != nil {
+			klog.Infof("tc[%s/%s]'s tikv pod[%s/%s] failed to delete,%v", namespace, tcName, namespace, name, err)
+			return util.ARFail(err)
+		}
+	}
+
+	return util.ARSuccess()
+}
+
+func (pc *PodAdmissionControl) admitDeleteTombStoneTiKVPod(isInOrdinal bool, pod *core.Pod, ownerStatefulSet *apps.StatefulSet, tc *v1alpha1.TidbCluster, pdClient pdapi.PDClient) *admission.AdmissionResponse {
```
aylei: This method is exactly same with `admitDeleteNonStoreTiKVPod`, could we eliminate the duplication by extract a general function and call in these functions?

Yisaer: fix

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/pingcap/tidb-operator/pkg/webhook/util"
+	admission "k8s.io/api/admission/v1"
+	apps "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
+	"k8s.io/klog"
+)
+
+const (
+	// EvictLeaderBeginTime is the key of evict Leader begin time
+	EvictLeaderBeginTime = "evictLeaderBeginTime"
```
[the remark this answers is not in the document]
Yisaer: fix

pkg/webhook/pod/tikv_creater.go
Yisaer: fix

pkg/webhook/pod/tikv_deleter.go
Yisaer: updated

pkg/webhook/pod/tikv_creater.go
Yisaer: updated

pkg/webhook/pod/pods.go
Yisaer: Updated. 
For now, the pod which is going to be created or deleted would be checked whether is managed by `tidb-operator` and is pd/tikv/tidb first. If not, the request would be admitted.

After that, if the pod don't have `InstanceLabelKey`, the request for creating would be refused while the request for deleting would be admitted.

pkg/webhook/pod/pods.go
```diff
@@ -192,8 +181,52 @@ func (pc *PodAdmissionControl) admitDeletePods(name, namespace string) *v1beta1.
...
+	if !exist {
+		klog.Errorf("pod[%s/%s] has no label: %s", namespace, name, label.InstanceLabelKey)
+		return util.ARFail(fmt.Errorf("pod[%s/%s] has no label: %s", namespace, name, label.InstanceLabelKey))
+	}
+
+	tc, err := pc.tcLister.TidbClusters(namespace).Get(tcName)
```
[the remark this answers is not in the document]
Yisaer: > It is possible for user to delete `TidbCluster` with `--cascade=false` and orphan the tikv, pd and tidb statefulsets, maybe we have to distinguish `notFound` and other errrors

If the user delete `TidbCluster` with `--cascade=false`, we can't and don't have to guarantee the avaiable or accessable for tidb both as operator or webhook.

So I think we don't have to consider this case.

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -0,0 +1,221 @@
...
+	// EvictLeaderTimeout is the timeout limit of evict leader
+	EvictLeaderTimeout = 3 * time.Minute
```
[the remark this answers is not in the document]
Yisaer: updated

pkg/webhook/pod/pods.go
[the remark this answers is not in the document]
aylei: In the case of operator, cluster still has basic functionality provided by statefulset

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -0,0 +1,221 @@
...
+	namespace := pod.Namespace
+	tcName := tc.Name
+
+	isStoreExist := false
+	storesInfo, err := pdClient.GetStores()
+	if err != nil {
+		return util.ARFail(err)
+	}
+	var storeInfo *pdapi.StoreInfo
+	for _, info := range storesInfo.Stores {
```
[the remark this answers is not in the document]
Yisaer: updated.

pkg/webhook/pod/pods.go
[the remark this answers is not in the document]
Yisaer: > In the case of operator, cluster still has basic functionality provided by statefulset

I think in the case of creating pod request, we would only deal with the situtaion we want and the other requests would be all admitted. So I think if we didn't find the `TidbCluster` here, I would let the request pass directly.

pkg/webhook/pod/tikv_util.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return nil
+}
+
+func endEvictLeader(storeInfo *pdapi.StoreInfo, pdClient pdapi.PDClient) error {
+	storeID := storeInfo.Store.Id
+	err := pdClient.EndEvictLeader(storeID)
+	if err != nil {
+		klog.Errorf("tikv upgrader: failed to end evict leader storeID: %d, %v", storeID, err)
+		return err
```
aylei: Is there any force strategy for scenarios like this? This pattern introduce a hard dependency from TidbCluster to operator (webhook), if the webhook suffers an network partition to PD (meanwhile, tikv could talk to PD), the Pod creation is blocked by this single point.
As a comparison, former controller is working in bypass and the cluster is still functional without operator.

Yisaer: I think `force create` is a nice and necessary choice for this case. Currently `deleting hook pod` is also have a `force` choice. I think it is better to leave it in anthoer issue.

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -0,0 +1,191 @@
...
+	return util.ARSuccess()
+}
+
+func (pc *PodAdmissionControl) admitDeleteOfflineTiKVPod() *admission.AdmissionResponse {
+	return &admission.AdmissionResponse{
+		Allowed: false,
+	}
+}
+
+func (pc *PodAdmissionControl) admitDeleteUpTiKVPod(isInOrdinal, isUpgrading bool, pod *core.Pod, ownerStatefulSet *apps.StatefulSet, tc *v1alpha1.TidbCluster, pdClient pdapi.PDClient, store *pdapi.StoreInfo, storesInfo *pdapi.StoresInfo) *admission.AdmissionResponse {
```
tennix: Could this method be simplified? It takes so many arguments which makes it hard to tell how it works.

Though there are some functions have many arguments in tidb-operator. They are entry point of the controller which is still pretty straightforward.

Yisaer: Sure, I suggest to create a new issue to simplify the param to make it more easy.

tennix: OK

PR #1229, opened 2019-11-25
bugfix: get time string from current time instead of pod name

### LINKED ISSUE #1227: scheduled-backup failed to start when clusterName is too long
## Bug Report

Scheduled backup [gets](https://github.com/pingcap/tidb-operator/blob/v1.0.4/charts/tidb-cluster/templates/scripts/_start_scheduled_backup.sh.tpl#L5) `timestamp` from the `POD_NAME`, this will be wrong if the `clusterName` is too long, for example, the `clusterName` is 32 words long, the `timestamp` var will be the word `backup`:

```
$ echo "ssssssssssssssssssssssssssssssss-scheduled-backup-157465068x4n5" | awk -F- '{print $(NF-1)}'
backup
```

weekface:

fixes #1227 

### Check List 

Tests 

Code changes

 - Has Helm charts change

Side effects

Related changes

 ```release-note
get time string from current time instead of pod name
 ```

### DIFF
--- a/charts/tidb-cluster/templates/scripts/_start_scheduled_backup.sh.tpl
+++ b/charts/tidb-cluster/templates/scripts/_start_scheduled_backup.sh.tpl
@@ -2,9 +2,7 @@ set -euo pipefail
 
 host=$(getent hosts {{ template "cluster.name" . }}-tidb | head | awk '{print $1}')
 
-timestamp=$(echo ${POD_NAME}|awk -F- '{print $(NF-1)}')
-## use UTC time zone to resolve timestamp, avoiding different parsing results due to different default time zones
-backupName=scheduled-backup-`date -u -d @${timestamp}  "+%Y%m%d-%H%M%S"`
+backupName=scheduled-backup-`date "+%Y%m%d-%H%M%S"`
 backupPath=/data/${backupName}
 
 echo "making dir ${backupPath}"


charts/tidb-cluster/templates/scripts/_start_scheduled_backup.sh.tpl
```diff
@@ -2,9 +2,7 @@ set -euo pipefail
 
 host=$(getent hosts {{ template "cluster.name" . }}-tidb | head | awk '{print $1}')
 
-timestamp=$(echo ${POD_NAME}|awk -F- '{print $(NF-1)}')
-## use UTC time zone to resolve timestamp, avoiding different parsing results due to different default time zones
-backupName=scheduled-backup-`date -u -d @${timestamp}  "+%Y%m%d-%H%M%S"`
+backupName=scheduled-backup-`date "+%Y%m%d-%H%M%S"`
```
weekface: Use [`TZ` env](https://github.com/pingcap/tidb-operator/blob/v1.0.4/charts/tidb-cluster/templates/scheduled-backup-cronjob.yaml#L66) timezone information of the pod instead of `UTC`.

onlymellb: I recommend continuing to use `UTC` timestamps `date  -u "+%Y%m%d-%H%M%S"`

weekface: It should be strange if the user set `TZ` already.

onlymellb: 👌

tennix:
Does this impact the restore from the scheduled backup?

weekface:
The generated name is the same as the old way: `"+%Y%m%d-%H%M%S"`, for example: `scheduled-backup-20190822-041004`. So it should not impact restore behavior.

PR #1247, opened 2019-11-27
run e2e build in container with kind

### LINKED ISSUE #1238: run CI builds in containers via Jenkins Kubernetes Plugin
## Feature Request

The current CI job is not scalable.

run CI builds in containers via Jenkins Kubernetes Plugin.

- GitHub actions
- Prow

however, it requires us to maintain another infra.

https://github.com/jenkinsci/kubernetes-plugin
example: https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/build_advanced_statefulset/detail/build_advanced_statefulset/20/pipeline/

cofyc:

fixes #1238 

- run e2e build in container (with kind)
- run two concurrent e2e jobs with Kubernetes v1.12.10 and v1.16.3
- add a script `hack/run-in-docker.sh` to run commands in isolated development environment (same in CI)
- refine pipeline (more small steps, easier to view CI process)

### Check List 

Tests 

 - Unit test
 - E2E test


charts/tidb-operator/templates/scheduler-rbac.yaml
```diff
@@ -121,6 +121,15 @@ rules:
 - apiGroups: [""]
   resources: ["persistentvolumeclaims"]
   verbs: ["get", "list", "update"]
+# followng permissions are required if CSINodeInfo/AttachVolumeLimit features are enabled
```
cofyc: if CSINodeInfo/AttachVolumeLimit  are enabled (defaults true in 1.16.3), we need this permission

this is a temporary fix, I think it's better to bind our service account to `system:kube-scheduler` cluster role, then we don't need to care about it at all 

https://github.com/pingcap/tidb-operator/issues/1252

tests/actions.go
```diff
@@ -476,6 +476,7 @@ func (oa *operatorActions) DeployOperator(info *OperatorConfig) error {
 	// deploy statefulset webhook and configuration to hijack update statefulset opeartion
 	cmd = fmt.Sprintf(`
 sed 's/apiVersions: \["v1beta1"\]/apiVersions: ["v1", "v1beta1"]/
+s#imagePullPolicy:.*#imagePullPolicy: IfNotPresent#g
```
cofyc: no local registry anymore, so we should use `IfNotPresent`

charts/tidb-operator/templates/scheduler-rbac.yaml
```diff
@@ -59,6 +58,15 @@ rules:
...
```
cofyc: if CSINodeInfo/AttachVolumeLimit  are enabled (defaults true in 1.16.3), we need this permission

this is a temporary fix, I think it's better to bind our service account to `system:kube-scheduler` cluster role, then we don't need to care about it at all 

https://github.com/pingcap/tidb-operator/issues/1252

charts/tidb-operator/templates/scheduler-rbac.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/component: scheduler
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+{{- if .Values.clusterScoped }}
```
cofyc: fix if-end statement

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -2,6 +2,92 @@
...
+      type: Directory
+  - name: cgroup
+    hostPath:
+      path: /sys/fs/cgroup
+      type: Directory
+  - name: docker-root
+    emptyDir: {}
+  - name: docker-graph
+    emptyDir: {}
+'''
```
onlymellb: I suggest moving this pod template to the Jenkins [system configuration](https://internal.pingcap.net/idc-jenkins/configure) so that all jobs on Jenkins can use this template just like before.

onlymellb: I suggest moving this pod template to the jenkins [system configuration](https://internal.pingcap.net/idc-jenkins/configure) so that all jobs on jenkins can use this template just like before.

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -2,6 +2,92 @@
...
+    # kind needs /lib/modules and cgroups from the host
+    volumeMounts:
+    - mountPath: /lib/modules
+      name: modules
+      readOnly: true
+    - mountPath: /sys/fs/cgroup
+      name: cgroup
+    # dind expects /var/lib/docker to be volume
+    - name: docker-root
+      mountPath: /var/lib/docker
```
aylei: Just to confirm: do all container runtimes serve on this socket?

ci/pingcap_tidb_operator_build_kind.groovy
cofyc: SRE team can copy this to system configuration for their own use. But I don't like to depend on global configurations. Different jobs have different requirements. Easier to update image and configurations (e.g. resources, volume configuration) and fewer dependencies (don't worry about affecting other jobs or being affected by changes in Jenkins configurations).

cofyc: And this makes it easier for people outside of our company to know how we run our e2e jobs.

hack/e2e.sh
```diff
@@ -56,40 +62,232 @@ while getopts "h?" opt; do
...
-    export NO_BUILD=y
-fi
+# https://github.com/kubernetes-sigs/kind/releases/tag/v0.6.0
+declare -A kind_node_images
+kind_node_images["v1.11.10"]="kindest/node:v1.11.10@sha256:44e1023d3a42281c69c255958e09264b5ac787c20a7b95caf2d23f8d8f3746f2"
+kind_node_images["v1.12.10"]="kindest/node:v1.12.10@sha256:e93e70143f22856bd652f03da880bfc70902b736750f0a68e5e66d70de236e40"
+kind_node_images["v1.13.12"]="kindest/node:v1.13.12@sha256:ad1dd06aca2b85601f882ba1df4fdc03d5a57b304652d0e81476580310ba6289"
+kind_node_images["v1.14.9"]="kindest/node:v1.14.9@sha256:00fb7d424076ed07c157eedaa3dd36bc478384c6d7635c5755746f151359320f"
+kind_node_images["v1.15.6"]="kindest/node:v1.15.6@sha256:1c8ceac6e6b48ea74cecae732e6ef108bc7864d8eca8d211d6efb58d6566c40a"
+kind_node_images["v1.16.3"]="kindest/node:v1.16.3@sha256:bced4bc71380b59873ea3917afe9fb35b00e174d22f50c7cab9188eac2b0fb88"
```
aylei: Shall we iterate all the patch versions for each minor version? e.g. the caller may specify `v1.16.2`

hack/run-in-docker.sh
```diff
@@ -0,0 +1,63 @@
...
+# required by kind
+docker_args+=(
+    -v /lib/modules:/lib/modules
+    -v /sys/fs/cgroup:/sys/fs/cgroup
+)
+
+docker run ${docker_args[@]} \
+    -v $ROOT:/go/src/github.com/pingcap/tidb-operator \
+    -w /go/src/github.com/pingcap/tidb-operator \
+    --entrypoint /usr/local/bin/runner.sh \
```
aylei: what's the relationship between `runner.sh` and our `e2e.sh`?

ci/pingcap_tidb_operator_build_kind.groovy
cofyc: this is docker directory not a socket, dockerd socket is created in the container

hack/e2e.sh
cofyc: this is a list of pre-built kind node images. If `v1.16.2` is specified, the script will fail with an "unsupported version" error.

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -2,6 +2,92 @@
...
+
+def build(SHELL_CODE) {
+	podTemplate(yaml: podYAML) {
+		node(POD_LABEL) {
+			container('main') {
+				def WORKSPACE = pwd()
+				dir("${WORKSPACE}/go/src/github.com/pingcap/tidb-operator") {
+					unstash 'tidb-operator'
+					stage("Debug Info") {
+						println "debug host: 172.16.5.5"
```
weekface: 172.16.5.5？

hack/run-in-docker.sh
cofyc: runner.sh is a start script. It first start DinD and then executes user-provided commands. It has no relationship with our `e2e.sh`.

ci/pingcap_tidb_operator_build_kind.groovy
cofyc: yes, this is master node IP of Kubernetes cluster we're running on. I hardcoded it here for debugging purposes.

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -15,6 +101,7 @@ def getChangeLogText() {
 }
 
 def call(BUILD_BRANCH, CREDENTIALS_ID, CODECOV_CREDENTIALS_ID) {
+	timeout(120) {
```
onlymellb: Must be completed in two minutes?

cofyc: 2 hours, unit is minutes

ci/pingcap_tidb_operator_build_kind.groovy
aylei: got it after reading `runner.sh`

ci/pingcap_tidb_operator_build_kind.groovy
onlymellb: Can we set the time unit explicitly? Like `timeout (time: 2, unit: 'HOURS')`?

ci/pingcap_tidb_operator_build_kind.groovy
onlymellb: Make sense.

hack/e2e.sh
```diff
@@ -56,40 +62,234 @@ while getopts "h?" opt; do
...
+        test -d $hostWorkerPath || mkdir $hostWorkerPath
+        cat <<EOF >> $tmpfile
+  extraMounts:
+  - containerPath: /mnt/disks/
+    hostPath: "$hostWorkerPath"
+    propagation: HostToContainer
+EOF
+    fi
+    # workers
+    for ((i=1; i <=3; i++)) {
```
onlymellb: Make the number of work nodes configurable?

hack/e2e.sh
```diff
@@ -56,40 +62,234 @@ while getopts "h?" opt; do
...
+        cat <<EOF >> $tmpfile
+- role: worker
+EOF
+        if [[ "$KIND_DATA_HOSTPATH" != "none" ]]; then
```
onlymellb: The script shebang is bash by default. Does the conditional judgment statement use double brackets uniformly?

ci/pingcap_tidb_operator_build_kind.groovy
cofyc: done

hack/e2e.sh
cofyc: done, `KUBE_WORKERS` is added

hack/e2e.sh
cofyc: I don't think it's necessary. In the future, we can use a static analysis tool [shellcheck](https://github.com/koalaman/shellcheck) which can help us in writing robust shell scripts.

onlymellb: I think there is no problem using single brackets or double brackets. The main question is whether we can use one of them in the script instead of mixing them.

cofyc: I'm fine with mixing them. `[[` is preferred over `[` in `bash` because it has some enhancements, however, `[` / `test` commands are often shorter in simple test cases.

cofyc: we can stick to `[[` when we have some verify scripts to force users not to use `[`, otherwise it's a burden to check it case by case

hack/run-in-docker.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
```
tennix:
```suggestion
#!/usr/bin/env bash
```

hack/run-in-docker.sh
```diff
@@ -0,0 +1,63 @@
...
+    -it --rm
+    --name tidb-operator-dev
+)
+
+# required by dind
+docker_args+=(
+    --privileged
+    -e DOCKER_IN_DOCKER_ENABLED=true
+    # Docker in Docker expects it to be a volume
+    --tmpfs /var/lib/docker
```
tennix: Will this consume lots of memory?

cofyc: it depends, can use docker local volume when necessary

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -47,104 +134,61 @@ def call(BUILD_BRANCH, CREDENTIALS_ID, CODECOV_CREDENTIALS_ID) {
...
-							make test
-						fi
-						make
-						make e2e-build
-						"""
+							sh """
+							export GOPATH=${WORKSPACE}/go
+							export PATH=${WORKSPACE}/go/bin:\$PATH
+							# hq is required to clone git repos from https://bitbucket.org 
+							if ! hash hg 2>/dev/null; then
```
tennix: Is `hg` needed? I don't have `hg` installed on my local machine and haven't got any issues yet.

cofyc: removed, `hg` is not needed anymore

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -47,104 +134,61 @@ def call(BUILD_BRANCH, CREDENTIALS_ID, CODECOV_CREDENTIALS_ID) {
...
+			build("IMAGE_TAG=${GITHASH} SKIP_BUILD=y GINKGO_NODES=8 KUBE_VERSION=v1.12.10 make e2e")
+        }
+        builds["E2E v1.16.3"] = {
+			build("IMAGE_TAG=${GITHASH} SKIP_BUILD=y GINKGO_NODES=8 KUBE_VERSION=v1.16.3 make e2e")
+        }
+        builds.failFast = false
+        parallel builds
+
+		// we requires ~/bin/config.cfg, filemgr-linux64 utilities on k8s-kind node
+		// TODO make it possible to run on any node
```
tennix: Maybe we can upload the binary and chart into internal file server and do the publish in a separate jenkins job which downloads the file from file server and publish to qiniu and other public registries.

ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -47,104 +134,56 @@ def call(BUILD_BRANCH, CREDENTIALS_ID, CODECOV_CREDENTIALS_ID) {
...
-							lockfile=/k8s-locks/\$cluster
-							if [ ! -e \$lockfile ]; then
-									touch \$lockfile
-									echo ${ghprbPullId} > \$lockfile
-									clusterName=\$cluster
-									echo "####### branch ${ghprbPullId}/${BUILD_BRANCH} get env:\$cluster #######"
-									echo "####### if you want to debug,please exec the command: export KUBECONFIG=`/root/go/bin/kind get kubeconfig-path --name=\$cluster ` in jenkins slave node ########"
-									break 2
+					stage("Build and Test") {
+						ansiColor('xterm') {
```
tennix: What shell is used by default in Jenkins? Shall we make it explicit?

cofyc: `sh`. I think it is not necessary because we don't write complex inline scripts in Jenkins pipeline anymore.

hack/run-in-docker.sh
cofyc: switched to use local volumes now, with local volume, we can reuse the cache to avoid pulling images every time (enabled if `SKIP_CLEANUP` is set)

cofyc:
fixed a race in e2e test
cc @aylei @onlymellb @tennix @weekface PTAL

tests/pkg/webhook/pods.go
```diff
@@ -37,6 +37,12 @@ func (wh *webhook) admitPods(ar v1beta1.AdmissionReview) *v1beta1.AdmissionRespo
 	reviewResponse := v1beta1.AdmissionResponse{}
 	reviewResponse.Allowed = false
 
+	if !wh.namespaces.Has(namespace) {
```
cofyc: when another spec is in deleting the namespace, this webhook will reject the DeleteCollection request, we can simply ignore requests in other namespaces because each test spec runs in an isolated namespace now

PR #1248, opened 2019-11-27
charts: backup: flexible credentials

gregwebs:
allow retrieving credentials from metadata on AWS and GCP

on GCP, stream the backup upload

The current backup chart does not allow the recommended approach to getting credentials on AWS and GCP

Don't require a k8s secret for credentials

### Check List 

 - Manually tested on GKE. A non-final version of this was tested on AWS.

Code changes

 - Has Helm charts change

Side effects

 - Should maintain backwards compatibility

Related changes

 - Need to cherry-pick to the release branch ?
 - docs: update the docs to indicate that k8s secrets are unnecessary.

 ```release-note
backup chart: allow retrieving credentials from metadata on AWS and GCP
 ```


tennix:
We have the backup controller which makes it much easier to do with backup and restore. Better to add this feature to the controller. /cc @onlymellb @DanielZhangQD

gregwebs:
yes, the backup controller has all these same issues. If we are going to have the chart in this repo, this PR should be accepted. Otherwise we should delete the chart from the repo.

gregwebs:
I suggest though that we may want to keep a chart for mydumper if we move the controller to the new br.

tennix:
We aren't going to remove mydumper backup from the backup controller even if we have integrated the new BR.

charts/tidb-backup/values.yaml
```diff
@@ -94,6 +106,7 @@ s3: {}
   # region: ""
   # bucket: ""
   # secretName is the name of the secret which stores s3 object store access key and secret key
+  # This is not necessary on AWS. Instead you should be able to get the credentials from the EKS service IAM role.
```
DanielZhangQD: Then how to use it in the job container?

charts/tidb-backup/values.yaml
```diff
@@ -72,6 +83,7 @@ tikvGCLifeTime: 720h
 # backup to or restore from gcp bucket, the backup path is in the form of <clusterName>-<name>
 gcp: {}
   # bucket: ""
+  # secretName is not necessary on GKE if you use the workload identity feature
```
DanielZhangQD: No need to inject any credentials to the job container?

gregwebs: that's correct. If you enable the workload identity feature you just set a service account.

charts/tidb-backup/values.yaml
gregwebs: The AWS SDK will find the credentials, it just needs to run with a service account. You have to be using EKS or properly running something like iam-authenticator in the k8s cluster.

DanielZhangQD: We just need to attach the service account, no need to install the SDK or call any command to run with it?

tennix: Does iam-authenticator need to be packaged in the Docker image?

gregwebs: yes, these changes are tested as working. Just set the service account.
On GCP the association with the k8s service account to the cloud service account is done with

```
kubectl -n operations annotate serviceaccount ${svcacct} iam.gke.io/gcp-service-account=${svcacct}@${PROJECT}.iam.gserviceaccount.com
```

On EKS I believe one just needs to attach a policy. If not using EKS you run an authenticator program for the entire cluster.  https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html

onlymellb:
> > I suggest though that we may want to keep a chart for mydumper if we move the controller to the new br.
> 
> The backup controller does not use this chart.
> It seems that no need to keep this chart after the backup controller introduced? @onlymellb

In order to be compatible with the old version of tidb-operator, I think it can't be deleted in a short time.

charts/tidb-backup/templates/scripts/_start_backup.sh.tpl
```diff
@@ -36,15 +47,28 @@ fi
...
+cat <<EOF > /tmp/rclone.conf
+[gcp]
+type = google cloud storage
+bucket_policy_only = true
+$creds
+EOF
+
+  cd "${backup_base_dir}"
+  tar -cf - "${backup_name}" | pigz -p 16 \
+  | rclone --config /tmp/rclone.conf rcat gcp:${bucket}/${backup_name}/${backup_name}.tgz
```
onlymellb: Remove extra spaces at the beginning.

charts/tidb-backup/values.yaml
```diff
@@ -13,12 +13,19 @@ mode: backup # backup | restore | scheduled-restore
...
+# By default this will match a normal tidb-operator deploy
+# host:
+
 # scheduledBackupName is the name of a scheduled backup directory,
 # used to restore the tidb cluster from scheduled backup.
 # scheduledBackupName: scheduled-backup-20190822-041004
 
 image:
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
tennix: Why change this to `Always`?

gregwebs: For infrequent long-running operations I think Always might be better. If the image is already there, which is unlikely except in the case of retry, then IfNotPresent will save doing some requests to the registry (presumably this would take a few seconds). But the problem is that a different image may be used depending on which node the pod is scheduled to. So the behavior of Always is more consistent. If we use an image tag of a specific non-changing version, then switching back to IfNotPresent is preferably to avoid contacting the registry in some cases.

tennix: Yes, the image content will not be updated with the specific tag. Changing to `Always` will fail if there is not internet connection which is most of the case for on-prem deployment. Users may simply load the image on the backup node.

charts/tidb-backup/templates/scripts/_start_backup.sh.tpl
```diff
@@ -36,15 +47,28 @@ fi
...
-uploader \
-  --cloud=gcp \
-  --bucket={{ .Values.gcp.bucket }} \
-  --backup-dir=${dirname}
+# Once we know there are no more credentials that will be logged we can run with -x
+set -x
+bucket={{ .Values.gcp.bucket }}
+creds=${GOOGLE_APPLICATION_CREDENTIALS:-""}
+if ! [[ -z $creds ]] ; then
+creds = "service_account_file = ${creds}"
```
tennix:
```suggestion
creds="service_account_file = ${creds}"
```

charts/tidb-backup/values.yaml
gregwebs: Okay, I didn't know about that use case

gregwebs:
This PR was also tested on AWS now.

PR #1250, opened 2019-11-28
Simplify the params for pod webhook

Yisaer:
Currently the params for pod webhook is complex. This request tries to simplify the params for current webhook deleting hook function.

Also, I add the `force` option for `tikv` creating pod hook.

Tests 

Code changes

 - Has Go code change


pkg/webhook/pod/pods.go
```diff
@@ -186,12 +201,20 @@ func (pc *PodAdmissionControl) admitDeletePods(name, namespace string) *admissio
...
+
 	if l.IsPD() {
 		pdClient := pc.pdControl.GetPDClient(pdapi.Namespace(namespace), tcName, tc.Spec.EnableTLSCluster)
-		return pc.admitDeletePdPods(pod, ownerStatefulSet, tc, pdClient)
+		payload.pdClient = pdClient
+		return pc.admitDeletePdPods(payload)
 	} else if l.IsTiKV() {
 		pdClient := pc.pdControl.GetPDClient(pdapi.Namespace(namespace), tcName, tc.Spec.EnableTLSCluster)
-		return pc.admitDeleteTiKVPods(pod, ownerStatefulSet, tc, pdClient)
+		payload.pdClient = pdClient
```
weekface: the `payload.pdClient` is same with the PD section. It is better to add it to `payload` definition. 

	payload := &admitPayload{
		pod:              pod,
		tc:               tc,
		ownerStatefulSet: ownerStatefulSet,
		pdClient: pc.pdControl.GetPDClient(pdapi.Namespace(namespace), tcName, tc.Spec.EnableTLSCluster),
	}

pkg/webhook/pod/pd_deleter.go
```diff
@@ -59,12 +62,12 @@ func (pc *PodAdmissionControl) admitDeletePdPods(pod *corev1.Pod, ownerStatefulS
 	// NotMember represents this pod is deleted from pd cluster or haven't register to pd cluster yet.
 	// We should ensure this pd pod wouldn't be pd member any more.
 	if !isMember {
-		return pc.admitDeleteNonPDMemberPod(IsDeferDeleting, isInOrdinal, pod, ownerStatefulSet, tc, pdClient, ordinal)
+		return pc.admitDeleteNonPDMemberPod(IsDeferDeleting, isInOrdinal, payload, ordinal)
```
DanielZhangQD: Actually, all the other arguments except for payload can be calculated within the function with values of payload, right?
It's the same for the following functions.

pkg/webhook/pod/pods.go
Yisaer: updated.

pkg/webhook/pod/pd_deleter.go
```diff
@@ -102,7 +87,7 @@ func (pc *PodAdmissionControl) admitDeleteNonPDMemberPod(IsDeferDeleting, isInOr
 		// it would be existed an pd-3 instance with its deferDeleting label Annotations PVC.
 		// And the pvc can be deleted during upgrading if we use create pod webhook in future.
 		if !isInOrdinal {
-			err := addDeferDeletingToPVC(v1alpha1.PDMemberType, pc, tc, ownerStatefulSet.Name, namespace, ordinal)
+			err := addDeferDeletingToPVC(v1alpha1.PDMemberType, pc, payload.tc, ownerStatefulSet.Name, namespace, ordinal)
```
DanielZhangQD: If pass `pc` as an argument, why not `pc. addDeferDeletingToPVC `?

DanielZhangQD: BTW, `v1alpha1.PDMemberType, ownerStatefulSet.Name, ordinal` are only used to construct the PVC name, is it better to construct the name and pass the name as an argument?

pkg/webhook/pod/pods.go
```diff
@@ -239,10 +259,34 @@ func (pc *PodAdmissionControl) AdmitCreatePods(ar admission.AdmissionReview) *ad
...
+
 	if l.IsTiKV() {
 		pdClient := pc.pdControl.GetPDClient(pdapi.Namespace(namespace), tcName, tc.Spec.EnableTLSCluster)
 		return pc.admitCreateTiKVPod(pod, tc, pdClient)
 	}
 
 	return util.ARSuccess()
 }
+
+func fetchInfoFromPayload(payload *admitPayload) (pod *core.Pod, ownerStatefulSet *apps.StatefulSet, ordinal int32, podName, namespace, tcName string, isInOrdinal, isUpgrading, IsDeferDeleting bool, pdClient pdapi.PDClient, err error) {
```
DanielZhangQD: Why so many return values? For the values that can be referred to with payload directly, I think no need to return them in another function.
@weekface

Yisaer: updated.

pkg/webhook/pod/pd_deleter.go

pkg/webhook/pod/pd_deleter.go
```diff
@@ -97,11 +99,11 @@ func (pc *PodAdmissionControl) admitDeleteNonPDMemberPod(payload *admitPayload)
 		return util.ARSuccess()
 
 	}
-	err = pdClient.DeleteMember(name)
+	err = payload.pdClient.DeleteMember(name)
 	if err != nil {
 		return util.ARFail(err)
 	}
-	err = addDeferDeletingToPDPod(pc, pod)
+	err = addDeferDeletingToPDPod(pc, payload.pod)
```
DanielZhangQD: Maybe `err = pc.addDeferDeletingToPDPod(payload.pod)`

pkg/webhook/pod/util.go
```diff
@@ -56,9 +56,8 @@ func IsStatefulSetUpgrading(set *v1.StatefulSet) bool {
 // pd pod who would be deleted by statefulset controller
 // we add annotations to this pvc and delete it when we scale out the pd replicas
 // for the new pd pod need new pvc
-func addDeferDeletingToPVC(memberType v1alpha1.MemberType, podAC *PodAdmissionControl, tc *v1alpha1.TidbCluster, setName, namespace string, ordinal int32) error {
-	pvcName := operatorUtils.OrdinalPVCName(memberType, setName, ordinal)
-	pvc, err := podAC.pvcControl.GetPVC(pvcName, namespace)
+func addDeferDeletingToPVC(pvcName, namespace string, pvcControl controller.PVCControlInterface, tc *v1alpha1.TidbCluster) error {
```
DanielZhangQD: Can the `namespace` be get from `tc.Namespace`?

pkg/webhook/pod/pods.go
```diff
@@ -272,18 +272,17 @@ func (pc *PodAdmissionControl) AdmitCreatePods(ar admission.AdmissionReview) *ad
 	return util.ARSuccess()
 }
 
-func fetchInfoFromPayload(payload *admitPayload) (pod *core.Pod, ownerStatefulSet *apps.StatefulSet, ordinal int32, podName, namespace, tcName string, isInOrdinal, isUpgrading, IsDeferDeleting bool, pdClient pdapi.PDClient, err error) {
+func fetchInfoFromPayload(payload *admitPayload) (ordinal int32, podName, namespace, tcName string, isInOrdinal, isUpgrading, IsDeferDeleting bool, err error) {
```
DanielZhangQD: Still so many return values, the `podName, namespace, tcName` can be retrieved easily and the rest values are calculated only in one line, is it better to just call the target function in the caller?

Yisaer: updated.

pkg/webhook/pod/pd_deleter.go

pkg/webhook/pod/util.go
Yisaer: updated.

Yisaer:
@weekface @DanielZhangQD PTAL

Issue #1152, opened 2019-11-15
CI: Drop retool and gometalinter


aylei:
`twitchtv/retool` is out of maintenance and `alecthomas/gometalinter` has been deprecated. 

Now that we have upgraded to go `v1.13` in which go module is enabled by default, it is better to drop retool and manage these tools ~in go module~ in scripts instead.

The deprecated tool gometalinter should be dropped too.

yeya24:
Can I take this? What about using  https://github.com/golangci/golangci-lint?

gregwebs:
Yes, we should use golangci-lint. @yeya24 I opened a PR with what I think is a good setup for this: https://github.com/pingcap/tidb-operator/pull/1260

But it needs more work to be finalized: you are welcome to work on top of that.

cofyc:
It's not a good idea to manage external tools via `go.mod`. This is because each tool is a project which has its own dependencies which may conflict with each other. The most reliable and reproducible way is to download binary releases or run released docker images if they have. If tools do not have binaries or docker images released, we can compile them as a standalone project in an isolated environment.

cofyc:
I guess most tools like golangci-lint have binary or docker image releases. We can download and run them directly (check sums for safety if necessary).

Some simple binary utilities which are expected to compile with the latest pkgs (e.g. golang.org/x tools). I think it's fine to vendor (in the root or a dedicated sub go module).

aylei:
Yes, the method you propose would be better

cofyc:
@yeya24 do you have time to look at this?

PR #1260, opened 2019-12-01
update linter setup and use golangci-lint

gregwebs:
This is a starter for #1152

This is a setup I use in another project and it is quite similar to the setup in pingcap/tidb.

This has the only 2 linter tools we should need, but additional tools used by tidb-operator can be added to this system.

Currently golangci-lint fails.

- [ ] remove linters from the top-level Makefile
- [ ] get golangci-lint to pass (its is possible to tell it to ignore issues for now)


cofyc:
I don't think we need to compile tools if they have binary releases: https://github.com/golangci/golangci-lint/releases/tag/v1.21.0. We can download it by using curl or wget, e.g. https://github.com/pingcap/tidb-operator/blob/d4aed2a335c1d9617ed5d1d4afc33fbf7fedc0b8/hack/lib.sh#L99-L109.

hack/check
```diff
@@ -0,0 +1,40 @@
...
+	if ! command -v gfind ; then
+		echo "please install GNU Find"
+		echo ""
+		echo "    brew install findutils"
+		exit 1
+	fi
+	find=gfind
+fi
+
+pushd "$(dirname "$0")/.." >/dev/null
```
onlymellb: Can you use this [style](https://github.com/pingcap/tidb-operator/blob/master/hack/run-e2e.sh#L7-L8) uniformly?

hack/check
```diff
@@ -0,0 +1,40 @@
...
+NO_LINT=${NO_LINT:-""}
+if [[ -z $NO_LINT ]] ; then
+  echo "linting"
+  ./tools/bin/revive -formatter friendly -config revive.toml ${PKGS}
+fi
+
+echo "checking"
+pushd pkg
+for d in * ; do
+  pushd "$d"
```
onlymellb: Can we use the absolute path instead of switch dir?

gregwebs:
Feel free to take this PR over. What is here is the easy part. More work is to fix all the linter issues.

hack/check
```diff
@@ -0,0 +1,40 @@
...
+
+if [ "$(uname)" == "Darwin" ]; then
+	if ! command -v gfind ; then
+		echo "please install GNU Find"
+		echo ""
+		echo "    brew install findutils"
+		exit 1
+	fi
+	find=gfind
+fi
```
onlymellb: I think we can omit this code and use `PKGS=$(find pkg -type f -name '*.go'|awk '{print "github.com/pingcap/tidb-operator/"$1}'|sort -u)` instead.

gregwebs:
I agree that using binary releases would work well. I find myself sometimes needing a tool or version of a tool without such a release. Building also comes with security and reproducibility benefits that are probably not a big concern for codegen and linting tools.

PR #1265, opened 2019-12-02
Fixed a bug that changed tikv-gc-time during full backup

mantuliu:
When I read the tidb-operator code, I found that when the file manager.go is doing a full backup, it will modify the important gc parameter of TiDB, tikv-gc-life-time. But it is very likely that the value of the production environment is greater than 3 hours, so changing the tikv-gc-life-time to 3h hastily destroys the original historical version storage time of the database.
The code compares the value of the original tikv-gc-life-time in TiDB with the default value in backup. If the old value in TiDB is small, the tikv-gc-life-time value is replaced and written Operation, otherwise it will not be modified.
### Check List 

Tests 

 Because the k8s environment of tidb-operator is more complicated to build, I will add relevant test instructions later, thank you!

Code changes

 - Has Go code change

Side effects

Related changes

 ```release-note
NONE
 ```


weekface:
@onlymellb PTAL

onlymellb:
@mantuliu  Thank you for your contribution. The current implementation does have the problem that you said, but I think it is possible to solve this problem by setting `tikv-gc-life-time` to long enough time during the backup, just like this [solution](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-backup/templates/scripts/_start_backup.sh.tpl#L19). When the backup is complete, reset `tikv-gc-life-time` to the original value. what do you think?

mantuliu:
@onlymellb  Thank you for your answer. I studied the backup script carefully, this solution can indeed solve the problem mentioned in this PR. However, I still suggest that the corresponding restrictions be implemented at the code level. I think that the script's protection of the code is very correct, but the code should not completely rely on the script's protection of it. If one day we discard this script and use a new way is likely to introduce this problem again.

onlymellb:
@mantuliu  Well, your method is indeed optimal theoretically. By the way, can you set `tikv-gc-life-time` from the default `3h` to a longer time(such as `72h`)?  because `3h` may not be enough when the tidb cluster has a large amount of data.

cmd/backup-manager/app/backup/manager.go
```diff
@@ -114,42 +114,65 @@ func (bm *BackupManager) performBackup(backup *v1alpha1.Backup, db *sql.DB) erro
...
 			Status:  corev1.ConditionTrue,
-			Reason:  "SetTikvGCLifeTimeFailed",
+			Reason:  "ParseOldTikvGCLifeTimeFailed",
 			Message: err.Error(),
 		})
 	}
-	glog.Infof("set cluster %s %s to %s success", bm, constants.TikvGCVariable, constants.TikvGCLifeTime)
-
-	backupFullPath, err := bm.dumpTidbClusterData()
+	tikvGCTimeDuration, err := time.ParseDuration(constants.TikvGCLifeTime)
```
tennix: What about defining the default gc lifetime as `time.Duration` so we don't have to parse this every time.

cmd/backup-manager/app/backup/manager.go
```diff
@@ -114,42 +114,65 @@ func (bm *BackupManager) performBackup(backup *v1alpha1.Backup, db *sql.DB) erro
...
 			Status:  corev1.ConditionTrue,
-			Reason:  "ResetTikvGCLifeTimeFailed",
+			Reason:  "DumpTidbClusterFailed",
 			Message: err.Error(),
 		})
 	}
-	glog.Infof("reset cluster %s %s to %s success", bm, constants.TikvGCVariable, oldTikvGCTime)
+	glog.Infof("dump cluster %s data to %s success", bm, backupFullPath)
 
+	if oldTikvGCTimeDuration < tikvGCTimeDuration {
```
tennix: Should we try to reset this with backoff? @onlymellb Since this is an important parameter, we should ensure this is reset as possible. We don't have to do it in this PR.

mantuliu: @tennix If this value is not set, many TiDBs cannot be backed up. For example, the value set in our production environment is: 10m. This value cannot back up TiDB data. Only the setting is large enough to back up data.

mantuliu:
@onlymellb I have changed to 72h, please help me to see, help me /run-e2e-tests ,Thank you!

cmd/backup-manager/app/backup/manager.go
tennix: What I'm trying to say is when the backup is done, we need to ensure the gc lifetime is reset to the original value. Currently, this seems to be done only once and print error logs when fails. We can retry with backoff if the value is still what we set during backup for example 72h.

mantuliu: This is indeed a hidden question, and I agree with you very much, this possibility is very high.

cmd/backup-manager/app/backup/manager.go
mantuliu: a good idea.I'll try to change it later.

cmd/backup-manager/app/backup/manager.go
onlymellb: I think this is necessary, we can use this [function](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L290-L301) to do exponential backoff

Issue #1264, opened 2019-12-02
sync tikv configmap and start script in controller


aylei:
#1121

Sync the tikv configmap and start-script in controller, an example about syncing these could be found here:

Note:

- Rolling-update of configmap should be controlled via a strategy field.
- An integration test case (e2e) should be added to verify this change will not rolling-update the tikv statefulset created before if there's no change in `tidbcluster` spec.
- Unlike pump, configuration of TiKV must have schema, because TiKV is written in rust, it's hard to import the schema of TiKV's configuration directly, maybe maintain a go copy is an acceptable idea, for example, hackathon team `plainboring` did so : https://github.com/plainboring/config_client/blob/master/pkg/tikv/config.go

Issue #1263, opened 2019-12-02
sync pd configmap and start script in controller


aylei:
#1121

Issue #1257, opened 2019-11-29
e2e: able to isolate tidb-operator for test specs


aylei:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, we deploy a tidb-operator singleton for all ginkgo specs (https://github.com/pingcap/tidb-operator/blob/f8f4418e7287d2127ccbdc4b46bbfab9776c2175/tests/e2e/e2e.go#L124). Different specs are not isolated and it is tedious to test tidb-operator in different configurations or setups.

**Describe the feature you'd like:**

~~A dedicated tidb-operator for each spec.~~

Discussed with @cofyc , a new tag `[Serial]` could support synchronized modification on tidb-operator. 

**Describe alternatives you've considered:**

Support customizing operator configuration and setup of the suite, and run multiple e2e tests in parallel with different operator configuration. This is less flexible, IMHO.

cofyc:
Some resources are global, e.g. CRDs. We can have more jobs to test different kinds of specs. 
The current e2e job runs parallel specs in parallel. We can have another job to run serial specs (e.g. upgrade tidb-operator) serially (e.g. --ginkgo.focus='\[Serial\]').

aylei:
> Some resources are global, e.g. CRDs. We can have more jobs to test different kinds of specs.
> The current e2e job runs parallel specs in parallel. We can have another job to run serial specs (e.g. upgrade tidb-operator) serially (e.g. --ginkgo.focus='[Serial]').

That'd be great

cofyc:
> Some resources are global, e.g. CRDs. We can have more jobs to test different kinds of specs.
> The current e2e job runs parallel specs in parallel. We can have another job to run serial specs (e.g. upgrade tidb-operator) serially (e.g. --ginkgo.focus='[Serial]').

we have a serial job right now, this issue can be closed.

PR #1269, opened 2019-12-02
sync pump cluster in new pump_member_manager

### LINKED ISSUE #1191: Sync Pump in controllers

aylei:

close #1191 

Note: 

- **The strategy to manage start script and configmap will also applied to TiKV/TiDB/PD if this review passed**
- Documentation will be tracked in separate issues:
  - https://github.com/pingcap/docs-cn/issues/2092
  - https://github.com/pingcap/docs/issues/1684
- Verify in stability test is non goal for now.

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Helm charts change
 - Has Go code change

Related changes

 ```release-note
Custom resource `TidbCluster` is extended to support managing Pump cluster. 
 ```

@weekface @tennix @cofyc @Yisaer @AstroProfundis PTAL


charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -28,7 +28,7 @@ rules:
   - events
   verbs: ["*"]
 - apiGroups: [""]
-  resources: ["endpoints"]
+  resources: ["endpoints","configmaps"]
```
aylei: To operate configmaps of pump

manifests/webhook.yaml
```diff
@@ -95,7 +95,7 @@ spec:
             - -v=4
             # TODO webhook is not bundled with tidb-operator yet and tested in
             # e2e only, make it configurable in helm chart
-            - -features=AdvancedStatefulSet=true
+            # - -features=AdvancedStatefulSet=true
```
aylei: Disable asts temporarily, a new `[Serial]` tag will address this #1257

pkg/manager/member/pump_member_manager.go
```diff
@@ -0,0 +1,475 @@
...
+		return nil
+	}
+	if err := pmm.syncHeadlessService(tc); err != nil {
+		return err
+	}
+	return pmm.syncStatefulSet(tc)
+}
+
+// syncStatefulSet syncs the pump statefulset
+// TODO: sync statefulset status of pump to tidbcluster
```
aylei: https://github.com/pingcap/tidb-operator/issues/1271

pkg/manager/member/pump_member_manager.go
```diff
@@ -0,0 +1,475 @@
...
+		inUseName := FindPumpConfig(tc.Name, set.Spec.Template.Spec.Volumes)
+		// find an in-use configmap, will update it in-place
+		if inUseName != "" {
+			newCm.Name = inUseName
+		}
+	}
+
+	oldCmTemp, err := pmm.cmLister.ConfigMaps(newCm.Namespace).Get(newCm.Name)
+	if errors.IsNotFound(err) {
+		// TODO: garbage collection for pump configmaps
```
aylei: https://github.com/pingcap/tidb-operator/issues/1272

pkg/manager/member/pump_member_manager.go
```diff
@@ -0,0 +1,475 @@
...
+	var envs []corev1.EnvVar
+	if tc.Spec.Pump.SetTimeZone != nil && *tc.Spec.Pump.SetTimeZone {
+		envs = append(envs, corev1.EnvVar{
+			Name:  "TZ",
+			Value: tc.Spec.Timezone,
+		})
+	}
+	if spec.HostNetwork() {
+		// For backward compatibility, set HOSTNAME to POD_NAME in hostNetwork mode
+		envs = append(envs, corev1.EnvVar{
```
aylei: This can avoid rolling-update when user upgrade from helm chart, but I'm not sure if there is any side-effect, @cofyc PTAL

pkg/manager/member/pump_member_manager.go
```diff
@@ -0,0 +1,475 @@
...
+		Labels:          pumpLabel,
+		OwnerReferences: []metav1.OwnerReference{controller.GetOwnerRef(tc)},
+	}
+	return objMeta, pumpLabel
+}
+
+func getPumpStartScript(tc *v1alpha1.TidbCluster) (string, error) {
+	buff := new(bytes.Buffer)
+	// Keep the logic same as helm chart, but pump has not supported tls yet (no cert mounted)
+	// TODO: support tls
```
aylei: TLS of pump is not actually supported in tidb-cluster helm chart, nor in controller. We may document this limitation first @AstroProfundis  PTAL

pkg/manager/member/pump_member_manager.go
cofyc: you don't do this because this is a new sts to deploy (no need to keep back-compatibility like pd/tikv sts) 

you can always use `POD_NAME` for both pod network and host network mode because sts member address is always `<pod-name>-<pod-ordinal>`.

pkg/manager/member/pump_member_manager.go
```diff
@@ -0,0 +1,475 @@
...
+)
+
+// pumpStartScriptTpl is the template string of pump start script
+// Note: changing this will cause a rolling-update of pump cluster
+var pumpStartScriptTpl = template.Must(template.New("pump-start-script").Parse(`set -euo pipefail
+
+/pump \
+-pd-urls={{ .Scheme }}://{{ .ClusterName }}-pd:2379 \
+-L={{ .LogLevel }} \
+-advertise-addr=` + "`" + `echo ${HOSTNAME}` + "`" + `.{{ .ClusterName }}-pump:8250 \
```
cofyc: this is strange, does this work? `-advertise-addr={{.PodName}}.{{ .ClusterName }}-pump:8250`
we don't need to pass HOSTNAME environment for sts pod, because the pod address is predictable

tests/actions.go
```diff
@@ -205,6 +206,7 @@ type OperatorActions interface {
 	CheckUpgradeCompleteOrDie(info *TidbClusterConfig)
 	CheckInitSQL(info *TidbClusterConfig) error
 	CheckInitSQLOrDie(info *TidbClusterConfig)
+	DeployAndCheckPump(tc *TidbClusterConfig) error
```
cofyc: where is its implementation?

aylei: in `backup.go`, this is an existing function but was not included in the interface definition formerly.

pkg/manager/member/pump_member_manager.go
aylei: Use the `HOSTNAME` environment deliberately to keep backward compatibility with the helm chart, otherwise the pump statefulset will get rolling-update if we set the new env var `POD_NAME`.

pkg/manager/member/pump_member_manager.go
aylei: See the comment below, we have to consider the helm chart.

aylei: Actually I prefer to use `POD_NAME`, which is more accurate and has no hidden tricks, but we have to document that adopt pump via `TidbCluster` CR will trigger an rolling-update, this a trade-off, either looks not good but acceptable for me.

/cc @weekface @tennix how do you think?

manifests/webhook.yaml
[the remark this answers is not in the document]
aylei: As commented before:

> Disable asts temporarily, a new [Serial] tag will address this #1257

Current tidb-operator in e2e is a shared one with asts enabled, the case that user migrate cluster deployed by helm to `TidbCluster` CR with asts disabled cannot be tested.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -232,9 +243,19 @@ type PumpSpec struct {
...
 	// +k8s:openapi-gen=false
 	// TODO: add schema
-	Config map[string]json.JsonObject `json:"config,omitempty"`
+	config.GenericConfig `json:",inline"`
+
+	// +k8s:openapi-gen=false
+	// For backward compatibility with helm chart
+	SetTimeZone *bool `json:"setTimeZone,omitempty"`
+	// +k8s:openapi-gen=false
+	// For backward compatibility with helm chart
```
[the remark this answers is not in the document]
aylei: Good catch, formerly I did not want to operate on the schmealess config object, but seems like that extract `logLevel` requires little effort, will do.

pkg/manager/member/pump_member_manager.go
tennix: What about adding a feature gate for adopting helm managed resources?

aylei: No gate is needed, if user leave the `.spec.pump` as nil, the helm resource will not be adopted. The real concern is that whether user has the ability to **adopt pump cluster and do not trigger rolling-update**. If not, abandon helm for existing cluster will require extra effort.

pkg/manager/member/pump_member_manager.go
```diff
@@ -440,21 +444,37 @@ func getPumpStartScript(tc *v1alpha1.TidbCluster) (string, error) {
...
 		LogLevel    string
-	}{scheme, tc.Name, logLevel})
+	}{scheme, tc.Name, getPumpLogLevel(tc)})
 	if err != nil {
 		return "", err
 	}
 	return buff.String(), nil
 }
 
+func getPumpLogLevel(tc *v1alpha1.TidbCluster) string {
```
tennix: Why not pass the pump config field directly? Passing a `TidbCluster` pointer around is both unnecessary and unsafe.

pkg/manager/member/pump_member_manager.go
weekface: adapt old pump clusters and do not trigger rolling-update when update tidb-operator is reasonable for the user.

weekface: adapt old pump clusters and do not trigger rolling-update when update tidb-operator is reasonable for the user.

pkg/manager/member/pump_member_manager.go
aylei: Just wanted to hide the details about where the `pumpLogLevel` located in this function, this is a private function so I did not think too much, do you think it is necessary to change the function signature?

aylei:
In order to not blocking the development of #1192 #1264 #1263 , I will separate the `configmapControlInterface` to another PR.

pkg/manager/member/pump_member_manager.go
tennix: OK, sgtm.

pkg/manager/member/pump_member_manager.go
tennix: So users can allow the adoption when they need to update pump clusters? If so, then it's reasonable and we can document about this.

aylei: OK, if both approaches make sense, I pretend to keep it as is to avoid additional code change (mainly in e2e, there is a case to verify the sts do not get rolling-update after adoption, which need to be re-written if we choose to change the podTemplate in this PR), further improvement on the `PodTemplate` of pump could be made through other enhancement issues.

Issue #1253, opened 2019-11-28
make backup and restore more universal


onlymellb:
## Feature Request

**Is your feature request related to a problem? Please describe:**

The original design of backup and restore only considered the tidb cluster managed in k8s, so there are some restrictions.

- Can't backup and restore the tidb cluster outside k8s
- Can't specify a remote backup path to restore tidb cluste(only can specify a backup CR to restore tidb cluster)
- Backup and restore related resources must be created under the namespace of the backup and the restore cluster separately. Actually, this restriction is not necessary

At the same time, for convenience of use, some security and flexibility are sacrificed

**Describe the feature you'd like:**

The purpose here is mainly to make backup and recovery more versatile and solve the problems mentioned above
- Support backup and restore any accessible tidb cluster, whether inside or outside k8s.
- Support specify a remote backup path to restore any accessible tidb cluster, whether inside or outside k8s.
- Backup and restore related resources can be created in any namespace.
- Restore operation does not read secrets(remote storage access credentials) across namespaces

gregwebs:
> Can't backup and restore the tidb cluster outside k8s
I do this with the existing backup helm chart just by setting a variable for the host to connect to.

> Restore operation does not read secrets(remote storage access credentials) across namespaces
If we have control of the namespace that the backup/restore job runs in that can help a lot with avoiding inappropriate access across namespaces.

tennix:
We should also allow using IAM role to skip configuring secret explicitly as discussed here https://github.com/pingcap/tidb-operator/pull/1248#discussion_r351586748.

*Note:* current rclone does not support this feature [yet](https://github.com/rclone/rclone/issues/3753).

Issue #768, opened 2019-08-14
implement backup、backupSchedule and restore operators


onlymellb:
We list a TODO checklist to implement backup、backupSchedule and restore operators.

Main Logic:
- [x] implement the backup operator
- [x] implement the backupSchedule operator
- [x] implement the restore operator
- [x] tidb-backup-manager use to backup、clean and restore
- [x] the entry point script for tidb-backup-manager

Improve：
- [ ] support for specifying external tidb cluster to restore
- [ ] omit the `storageType` field in backup and backupSchedule CR
- [ ] add a switch to control whether to compress backup data
- [ ] support backup to local filesystem
- [x] support backup to GCP and restore from it
- [x] support backup to AWS and restore from it
- [ ] support backup to Azure and restore from it
- [ ] archive and upload backup data using streaming. refer mholt/archiver#132
- [ ] add a backup subcommand to tkctl to inspect the backup related information of the tidb cluster
- [ ] create RBAC related resources for the backup、clean and restore job in operator
- [ ] clean up the backup data in the backup PV
- [ ] support for using custom mydump options to backup
- [ ] support for using custom PVC to backup

Misc:
- [ ] unit test
- [ ] usage document
- [ ] design document
- [ ] e2e test

PR #1276, opened 2019-12-03
make backup and restore more universal and secure

### LINKED ISSUE #1217: Use env for rclone
## Feature Request

Rclone support [environment variable](https://rclone.org/docs/#environment-variables) to config, and we already inject the env for the backup, it's unnecessary to generate a temporary [`rclone.conf`](https://github.com/pingcap/tidb-operator/blob/f12f92cb93a8a6671b5387b82e1c786a29260869/images/backup-manager/entrypoint.sh#L4-L28). Besides, it should be more secure to use environment variables.

onlymellb:
resolve #1217 and #1253 

### Check List 

Tests 

Code changes

 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
make backup and restore more universal and secure
 ```


manifests/backup/restore-gcs.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: pingcap.com/v1alpha1
+kind: Restore
+metadata:
+  name: demo2-restore
+  namespace: test2
+spec:
+  from:
```
tennix: For `restore`, this field should be `to`.

manifests/backup/restore-gcs.yaml
```diff
@@ -0,0 +1,19 @@
...
+    host: 10.0.0.2
+    port: 4000
+    user: root
+    secretName: backup-demo2-tidb-secret
+  gcs:
+    projectId: <your-project-id>
+    secretName: gcs-secret
+  storageType: gcs
```
tennix: Putting `storageType`, `storageClassName` and `storageSize` together is confusing.

Is the PV a temporary storage? Maybe we can rename either the storage backend or the temporary PV.

Also, I think the backup path is specific to each storage backend, it's better to put under the corresponding storage backend.

tennix: How about this?
```yaml
spec:
  from: # storage backend info for backup file
    type: gcs
    gcs:
      projectId: <your-project-id>
      secretName: gcs-secret
      path: gcs://test1-demo1/backup-2019-11-11T16:06:05Z.tgz
  to: # downstream tidb cluster info
    host: 10.0.0.2
    port: 4000
    user: root
    secretName: backup-demo2-tidb-secret
  tempStorage: # use temporarily to download the backup file
    storageClassName: local-storage
    size: 1Gi
```
/cc @DanielZhangQD @shuijing198799

manifests/backup/backup-s3.yaml
```diff
@@ -5,12 +5,16 @@ metadata:
   name: demo1-backup-s3
   namespace: test1
 spec:
+  from:
+    host: 10.0.0.1
+    port: 4000
+    user: root
+    secretName: backup-demo1-tidb-secret
```
weekface: use `passwordSecretName`?

pkg/backup/util/util.go
```diff
@@ -123,83 +134,122 @@ func GenerateGcsCertEnvVar(secret *corev1.Secret, gcs *v1alpha1.GcsStorageProvid
...
 
-		certEnv, reason, err = GenerateGcsCertEnvVar(secret, backup.Spec.Gcs)
+		certEnv, reason, err = GenerateGcsCertEnvVar(provider.Gcs)
 
 		if err != nil {
 			return certEnv, reason, err
 		}
 	default:
-		err := fmt.Errorf("backup %s/%s don't support storage type %s", ns, name, backup.Spec.StorageType)
+		err := fmt.Errorf("don't support storage type %s", storageType)
```
tennix:
```suggestion
		err := fmt.Errorf("unsupported storage type %s", storageType)
```

pkg/backup/util/util.go
```diff
@@ -123,83 +134,122 @@ func GenerateGcsCertEnvVar(secret *corev1.Secret, gcs *v1alpha1.GcsStorageProvid
...
+			return bucketName, "S3ConfigIsEmpty", fmt.Errorf("backup %s/%s s3 config is empty", ns, name)
+		}
+		bucketName = backup.Spec.S3.Bucket
+	case v1alpha1.BackupStorageTypeGcs:
+		if backup.Spec.Gcs == nil {
+			return bucketName, "GcsConfigIsEmpty", fmt.Errorf("backup %s/%s gcs config is empty", ns, name)
+		}
+		bucketName = backup.Spec.Gcs.Bucket
+	default:
+		return bucketName, "NotSupportStorageType", fmt.Errorf("backup %s/%s don't support storage type %s", ns, name, backup.Spec.StorageType)
```
tennix:
```suggestion
		return bucketName, "UnsupportedStorageType", fmt.Errorf("backup %s/%s doesn't support storage type %s", ns, name, backup.Spec.StorageType)
```

pkg/backup/util/util.go
```diff
@@ -123,83 +134,122 @@ func GenerateGcsCertEnvVar(secret *corev1.Secret, gcs *v1alpha1.GcsStorageProvid
...
+		err := fmt.Errorf("don't support storage type %s", storageType)
 		return certEnv, "NotSupportStorageType", err
 	}
 	return certEnv, reason, nil
 }
 
-// GetTidbUserAndPassword get the tidb user and password from specific secret
-func GetTidbUserAndPassword(ns, name, tidbSecretName string, secretLister corelisters.SecretLister) (user, password, reason string, err error) {
+// GenerateTidbPasswordEnv generate the password EnvVar
+func GenerateTidbPasswordEnv(ns, name, tidbSecretName string, secretLister corelisters.SecretLister) ([]corev1.EnvVar, string, error) {
```
tennix: I think we can follow k8s builtin controller's behavior to not checking the referenced resources existence.

pkg/backup/util/util.go
```diff
@@ -123,83 +134,122 @@ func GenerateGcsCertEnvVar(secret *corev1.Secret, gcs *v1alpha1.GcsStorageProvid
...
+		}
+		gcsSecretName := provider.Gcs.SecretName
 		secret, err := secretLister.Secrets(ns).Get(gcsSecretName)
 		if err != nil {
-			err := fmt.Errorf("backup %s/%s get gcs secret %s failed, err: %v", ns, name, gcsSecretName, err)
+			err := fmt.Errorf("get gcs secret %s/%s failed, err: %v", ns, gcsSecretName, err)
 			return certEnv, "GetGcsSecretFailed", err
 		}
 
 		keyStr, exist := CheckAllKeysExistInSecret(secret, constants.GcsCredentialsKey)
```
tennix: This is already checked in `GenerateGcsCertEnvVar`.

pkg/backup/util/util.go
```diff
@@ -123,83 +134,122 @@ func GenerateGcsCertEnvVar(secret *corev1.Secret, gcs *v1alpha1.GcsStorageProvid
...
+		}
+		s3SecretName := provider.S3.SecretName
 		secret, err := secretLister.Secrets(ns).Get(s3SecretName)
 		if err != nil {
-			err := fmt.Errorf("backup %s/%s get s3 secret %s failed, err: %v", ns, name, s3SecretName, err)
+			err := fmt.Errorf("get s3 secret %s/%s failed, err: %v", ns, s3SecretName, err)
 			return certEnv, "GetS3SecretFailed", err
 		}
 
 		keyStr, exist := CheckAllKeysExistInSecret(secret, constants.S3AccessKey, constants.S3SecretKey)
```
tennix: This is already checked in `GenerateS3CertEnvVar`.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -506,7 +506,7 @@ type S3StorageProvider struct {
 	Provider S3StorageProviderType `json:"provider"`
 	// Region in which the S3 compatible bucket is located.
 	Region string `json:"region,omitempty"`
-	// Bucket in which to store the Backup.
+	// Bucket in which to store the backup data.
```
tennix: We do not support specifying a prefix?

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -725,18 +735,19 @@ type RestoreCondition struct {
...
-	// SecretName is the name of the secret which stores
-	// tidb cluster's username and password.
-	TidbSecretName string `json:"tidbSecretName"`
-	// StorageClassName is the storage class for restore job's PV.
+	// To is the tidb cluster that needs to restore.
+	To TiDBAccessConfig `json:"to"`
+	// BackupPath is the location of the backup.
+	BackupPath string `json:"backupPath"`
+	// Type is the backup type for tidb cluster.
+	Type BackupType `json:"backupType,omitempty"`
```
tennix: I think we can allow users to omit this field if there is only one storage provider to improve UX.

manifests/backup/restore-gcs.yaml
onlymellb: I agree putting `storageType`, `storageClassName` and `storageSize` together is confusing. These three fields all start with the word storage, which makes people think they have any relationship, but there is no relationship between them. Therefore I intend to remove `storageType` field, here is the related issue #768.  I prefer the revised definition below.

```yaml
spec:
  to:
    host: 10.0.0.2
    port: 4000
    user: root
    secretName: backup-demo2-tidb-secret
  gcs:
    projectId: <your-project-id>
    secretName: gcs-secret
    # "<bucket-name>/<path-to-backup-file>"
    path: test1-demo1/backup-2019-11-11T16:06:05Z.tgz
  storageClassName: local-storage
  storageSize: 1Gi
```
What do you think? @DanielZhangQD @shuijing198799

weekface: use `from` and `to` is more clear for users.
I agree to remove the `storageType` attribute, `gcs:` indicate the type already.

```
spec:
  storageClassName: local-storage
  storageSize: 1Gi
  from:
    gcs:
      projectId: <your-project-id>
      secretName: gcs-secret
      path: gcs://test1-demo1/backup-2019-11-11T16:06:05Z.tgz
  to:
    host: 10.0.0.2
    port: 4000
    user: root
    secretName: backup-demo2-tidb-secret
```

pkg/backup/util/util.go
onlymellb: The purpose of the check is to make it fail quickly when there is a problem. At the same time, the failure information and reason will be written into the backup CR status, which is helpful for upper-layer application integrated with us to diagnose the problem quickly. k8s only discovers problems when it encounters problems in the execution layer, and then records the problem to the underlying pod object, which is not very straightforward for diagnosing the problem of the upper-layer application

pkg/apis/pingcap/v1alpha1/types.go
onlymellb: What do you mean by specifying a bucket prefix and adding a random string to construct a complete bucket name?

pkg/backup/util/util.go
onlymellb: `GenerateGcsCertEnvVar` does not check whether the secret exists and contains the specified key.

pkg/backup/util/util.go
onlymellb: ditto

pkg/apis/pingcap/v1alpha1/types.go
onlymellb: The `BackupType` field is unused currently. Do you mean `StorageType`?

manifests/backup/backup-s3.yaml
onlymellb: I don't think it makes sense to include the name of the key in the `secretName`. What if the secret stores more stuff?

manifests/backup/restore-gcs.yaml
onlymellb: I think that defining a remote storage configuration(here is `gcs`) in restore has already indicated that the backup data must come from here, and there is no need to use `from` to indicate this relationship. In addition, we should ensure that the `spec` is as simple as possible to reduce the cost of use for users.

pkg/apis/pingcap/v1alpha1/types.go
tennix: Users may want to store the backup file under a specific directory (prefix), though we have generated a prefix composed by cluster name and namespace.

manifests/backup/restore-gcs.yaml
tennix: There are storage class and size defined, users may think the backup is in the storage defined with `storageClassName` and `storageSize`. Adding `to` makes it less confuse and more straight forward and also simple enough.

pkg/apis/pingcap/v1alpha1/types.go
onlymellb: In this pr,  users must specify the bucket name, since backup is not the same as before. There is no cluster and namespace, so the bucket name will not be generated automatically.

tennix: The bucket name is different from prefix. I'm wondering if we should provide another parameter `prefix`, so users can store backup file in a bucket under a specific directory.

cmd/backup-manager/app/cmd/backup.go
```diff
@@ -42,13 +43,15 @@ func NewBackupCommand() *cobra.Command {
...
+	cmd.Flags().StringVarP(&bo.Namespace, "namespace", "n", "", "Backup CR's namespace")
+	cmd.Flags().StringVarP(&bo.Host, "host", "h", "", "Tidb cluster access address")
+	cmd.Flags().Int32VarP(&bo.Port, "port", "P", bkconstants.DefaultTidbPort, "Port number to use for connecting tidb cluster")
+	cmd.Flags().StringVarP(&bo.Bucket, "bucket", "k", "", "Bucket in which to store the backup data")
+	cmd.Flags().StringVarP(&bo.Password, bkconstants.TidbPasswordKey, "p", "", "Password to use when connecting to tidb cluster")
 	cmd.Flags().StringVarP(&bo.User, "user", "u", "", "User for login tidb cluster")
-	cmd.Flags().StringVarP(&bo.StorageType, "storageType", "S", "", "Backend storage type")
+	cmd.Flags().StringVarP(&bo.StorageType, "storageType", "s", "", "Backend storage type")
 	cmd.Flags().StringVarP(&bo.BackupName, "backupName", "b", "", "Backup CRD object name")
+	util.SetFlagsFromEnv(cmd.Flags(), bkconstants.BackupManagerEnvVarPrefix)
```
DanielZhangQD: Actually, backup-manager will get the backup from kube-apiserver and you only need to pass the name and namespace of the backup CR, then all the other info can be retrieved from that CR.

manifests/backup/restore-gcs.yaml
DanielZhangQD: I think no need to add `from/to`, users know they are doing `backup` or `restore`, so even if no `from/to` indication, they know that the tidb cluster are used for what.
For the `storageClassName: local-storage  storageSize: 1Gi`, I prefer to make it clear in the doc that they are used for temporary storage whether to put them under `tempStorage` or not.
```
spec:
  cluster or tidb:
    host: 10.0.0.2
    port: 4000
    user: root
    secretName: backup-demo2-tidb-secret
  gcs:
    projectId: <your-project-id>
    secretName: gcs-secret
    # "<bucket-name>/<path-to-backup-file>"
    path: test1-demo1/backup-2019-11-11T16:06:05Z.tgz
  storageClassName: local-storage
  storageSize: 1Gi
```

manifests/crd.yaml
```diff
@@ -1020,31 +1035,114 @@ spec:
...
-              description: Backup represents the backup object to be restored.
-              type: string
-            backupNamespace:
-              description: Namespace is the namespace of the backup.
+            backupPath:
+              description: BackupPath is the location of the backup.
               type: string
-            cluster:
-              description: Cluster represents the tidb cluster to be restored.
+            backupType:
```
DanielZhangQD: There is a `backupType` in `backupSpec`, why add one more here?

pkg/apis/pingcap/v1alpha1/types.go
DanielZhangQD: I have added one on https://github.com/pingcap/tidb-operator/pull/1280

manifests/backup/restore-gcs.yaml
tennix: What if users provide multiple cloud storage configurations if we omit the `storageType` field?

cmd/backup-manager/app/cmd/backup.go
onlymellb: Yes, backup-manager can assemble its own parameters through backup & restore CR. But here I hope that the logic of these parameter checks can be done at the operator level, instead of waiting until the backup & restore job is executed to find that some parameters are problematic. We think that the backup-manager startup parameters obtained from the operator must be legal. The backup-manager only needs to focus on processing backup and restore tasks and updating the status of backup CR.

manifests/crd.yaml
onlymellb: This file is generated automatically

onlymellb:
I have tested backup and restore locally and it works as expected. @tennix @weekface @DanielZhangQD @shuijing198799  PTAL again

Issue #1249, opened 2019-11-28
Integrate new BR for backup and restore with S3 storage


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Integrate new BR for backup and restore with S3 storage
TODOs:

- [x] backup https://github.com/pingcap/tidb-operator/pull/1280
- [x] clean https://github.com/pingcap/tidb-operator/pull/1280
- [x] backupSchedule https://github.com/pingcap/tidb-operator/pull/1280
- [x] restore https://github.com/pingcap/tidb-operator/pull/1342

**Describe the feature you'd like:**

Integrate new BR for backup and restore with S3 storage
**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

https://pingcap.com/docs-cn/dev/how-to/maintain/backup-and-restore/br/

PR #1280, opened 2019-12-04
support backup to S3 with br

DanielZhangQD:
backup to S3 with br for issue #1249
Support new BR for backup to S3
### Check List 

Tests 

Code changes

 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
Support backup to S3 with BR
 ```


gregwebs:
I see this shelling out to run br but I expected br to be ran as a job. I thought br could take a very long time to complete.

DanielZhangQD:
> I see this shelling out to run br but I expected br to be ran as a job. I thought br could take a very long time to complete.

This PR reuses the backup-manager introduced in the current backup feature with mydumper, it is run in a job and backup-manager will call br to backup data.

pkg/controller/backup/backup_controller.go
```diff
@@ -215,3 +218,36 @@ func (bkc *Controller) enqueueBackup(obj interface{}) {
...
 	bkc.queue.Add(key)
 }
+
+func validBackup(backup *v1alpha1.Backup) bool {
+	ns := backup.Namespace
+	name := backup.Name
+	if backup.Spec.BR == nil {
+		if backup.Spec.Cluster == "" {
+			glog.Infof("Missing Cluster config in spec of %s/%s", ns, name)
+			return false
```
gregwebs: in this kind of situation I usually prefer to return `error`. With this approach the caller can decide to log the error, but also to support an interactive validation with a user the error message can actually be returned to the user. A test function can easily inspect error messages.

The logging level here seems like it should be error instead of into.

cmd/backup-manager/app/backup/backup.go
```diff
@@ -123,6 +126,51 @@ func (bo *BackupOpts) backupDataToRemote(source, bucketURI string) error {
...
+	args, path, err := constructBROptions(backup)
+	if err != nil {
+		return "", "", err
+	}
+	fullArgs := []string{
+		"backup",
+		fmt.Sprintf("%s", backup.Spec.Type),
+	}
+	fullArgs = append(fullArgs, args...)
+	output, err := exec.Command("br", fullArgs...).CombinedOutput()
```
gregwebs: Did we consider using br as a library here? The main advantage is passing error codes in go memory without serialization.
An issue is handling logging. br may log inconsistently with the logging here. This is solvable by injecting the appropriate log formatter into BR.
The alternative of logging all the BR output here in a single message only after it completes has its own problems though.

pkg/controller/backup/backup_controller.go
DanielZhangQD: If we return `error` here, this CR will be resynced later, actually, no need to resync the invalid CR, the best practice may be that we do validation in AA or webhook to reject the invalid CR.

For the logging level, I will change to `Error`, thanks!

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: The code structure of br is not suitable to be used as a library, all the work is done in the `Run` function directly in the `cobra.Command`.

gregwebs: I don't think it would be hard to make a library interface.

DanielZhangQD: Yes, but I prefer we make it work first, then later we can make a library interface for BR code and then we can use it in the operator, is it OK?

gregwebs: Okay. Sometimes there are advantages to running a separate process.

cmd/backup-manager/app/backup/backup.go
```diff
@@ -123,6 +126,51 @@ func (bo *BackupOpts) backupDataToRemote(source, bucketURI string) error {
...
 }
 
+func (bo *BackupOpts) brBackupData(backup *v1alpha1.Backup) (string, string, error) {
+	args, path, err := constructBROptions(backup)
+	if err != nil {
+		return "", "", err
+	}
+	fullArgs := []string{
+		"backup",
+		fmt.Sprintf("%s", backup.Spec.Type),
```
shuijing198799: just backup.Spec.Type without fmt.Sprintf ?

cmd/backup-manager/app/backup/backup.go
```diff
@@ -211,3 +259,75 @@ func archiveBackupData(backupDir, destFile string) error {
 	}
 	return nil
 }
+
+// getBRCommitTs get backup position from the log output of BR
+// It really depends on the format in BR log,
+// currently, it's in the format of [BackupTS=412992336769581057]
+func getBRCommitTs(out string) (string, error) {
```
tennix: Please don't get such important info from logs, it's fragile that the log format is not standard and may change in the future.

DanielZhangQD:
Code updated due to BR cli options change in https://github.com/pingcap/br/pull/103/.
TODO:
1. Address https://github.com/pingcap/br/issues/95 and then update code to get commit TS and backup size from br backup meta.
2. Import github.com/pingcap/br/pkg/storage for remote operation.

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: fixed

cmd/backup-manager/app/backup/backup.go
```diff
@@ -14,200 +14,128 @@
 package backup
```
DanielZhangQD: This is the backup file for BR.

cmd/backup-manager/app/backup/manager.go
```diff
@@ -14,43 +14,38 @@
 package backup
```
DanielZhangQD: This is the backup manager for BR.

cmd/backup-manager/app/cmd/export.go
```diff
@@ -0,0 +1,79 @@
+// Copyright 2019 PingCAP, Inc.
```
DanielZhangQD: This is renamed from cmd/backup-manager/app/cmd/backup.go.

cmd/backup-manager/app/export/export.go
```diff
@@ -0,0 +1,203 @@
+// Copyright 2019 PingCAP, Inc.
```
DanielZhangQD: This is renamed from cmd/backup-manager/app/backup/backup.go.

cmd/backup-manager/app/export/manager.go
```diff
@@ -0,0 +1,239 @@
+// Copyright 2019 PingCAP, Inc.
```
DanielZhangQD: This is renamed from cmd/backup-manager/app/backup/manager.go.

DanielZhangQD:
@tennix @onlymellb @shuijing198799 @gregwebs PTAL the new commits to line up with the latest BR update.

cmd/backup-manager/app/export/manager.go
```diff
@@ -0,0 +1,239 @@
...
+			Type:    v1alpha1.BackupFailed,
+			Status:  corev1.ConditionTrue,
+			Reason:  "GetBackupCRFailed",
+			Message: err.Error(),
+		})
+	}
+
+	var db *sql.DB
+	err = wait.PollImmediate(constants.PollInterval, constants.CheckTimeout, func() (done bool, err error) {
+		db, err = util.OpenDB(bm.getDSN(constants.TidbMetaDB))
```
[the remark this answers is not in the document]
DanielZhangQD: This is not for BR. This file is renamed from the original backup file with mydumper.
cmd/backup-manager/app/backup/* are for backup with br.
For backup with mydumper, I just renamed the files and change the subcommand to `export`.
BTW, cmd/backup-manager is built for a binary that calls mydumper or br to do backup and then updates the backup CR for the size, commitTS, status, etc., it's not part of tidb-operator.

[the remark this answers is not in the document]
gregwebs: okay, if this is just a rename that is fine.

DanielZhangQD:
Update scheduled backup with BR, PTAL.

cmd/backup-manager/app/backup/backup.go
```diff
@@ -14,200 +14,128 @@
...
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+)
 
-func (bo *BackupOpts) getBackupFullPath() string {
-	return filepath.Join(constants.BackupRootPath, bo.getBackupRelativePath())
-}
+const (
+	// MetaFile represents file name
+	MetaFile = "backupmeta"
+)
```
onlymellb: All constants are defined [here](https://github.com/pingcap/tidb-operator/blob/master/cmd/backup-manager/app/constants/constants.go), can we add more detailed comments to indicate that this variable is for `br`?

cmd/backup-manager/app/backup/backup.go
```diff
@@ -14,200 +14,128 @@
...
-}
-
-func (bo *BackupOpts) setTikvGCLifeTime(db *sql.DB, gcTime string) error {
-	sql := fmt.Sprintf("update %s set variable_value = ? where variable_name = ?", constants.TidbMetaTable)
-	_, err := db.Exec(sql, gcTime, constants.TikvGCVariable)
+	var btype string
+	if backup.Spec.Type == "" {
+		btype = string(v1alpha1.BackupTypeFull)
+	} else {
+		btype = string(backup.Spec.Type)
```
onlymellb: BR needs `BackupType` as args?

onlymellb: BR support `full` and `table` subcommand, but `BackupType` only contain `full` and `incremental`. So I think we need to add a new BackupType `table`. In addition, we need to check the three BackupTypes here. For br, only` full` and `table` BackupType is valid.

cmd/backup-manager/app/backup/backup.go
```diff
@@ -14,200 +14,128 @@
...
+	ctx := context.Background()
+	exist, err := s.Exists(ctx, MetaFile)
 	if err != nil {
-		return bfPath, fmt.Errorf("cluster %s, execute mydumper command %v failed, output: %s, err: %v", bo, args, string(output), err)
+		return commitTs, err
 	}
-	return bfPath, nil
-}
+	if !exist {
+		return commitTs, fmt.Errorf("%s not exist", MetaFile)
```
onlymellb: Add `bo.String()` info to error log

cmd/backup-manager/app/backup/backup.go
```diff
@@ -14,200 +14,128 @@
...
-
-	glog.Infof("upload cluster %s backup data to %s successfully, now move it to permanent URL %s", bo, tmpDestBucket, destBucket)
-
-	// the backup was a success
-	// remove .tmp extension
-	output, err = exec.Command("rclone", constants.RcloneConfigArg, "moveto", tmpDestBucket, destBucket).CombinedOutput()
+	metaData, err := s.ReadAll(ctx, MetaFile)
 	if err != nil {
-		return fmt.Errorf("cluster %s, execute rclone moveto command failed, output: %s, err: %v", bo, string(output), err)
+		return commitTs, err
```
onlymellb: Add more detail info to error message.

cmd/backup-manager/app/backup/backup.go
```diff
@@ -14,200 +14,128 @@
...
-}
-
-func (bo *BackupOpts) cleanRemoteBackupData(bucket string) error {
-	destBucket := util.NormalizeBucketURI(bucket)
-	output, err := exec.Command("rclone", constants.RcloneConfigArg, "deletefile", destBucket).CombinedOutput()
+	backupMeta := &kvbackup.BackupMeta{}
+	err = proto.Unmarshal(metaData, backupMeta)
 	if err != nil {
-		return fmt.Errorf("cluster %s, execute rclone deletefile command failed, output: %s, err: %v", bo, string(output), err)
+		return commitTs, err
```
onlymellb: ditto

pkg/backup/backupschedule/backup_schedule_manager.go
```diff
@@ -199,20 +200,26 @@ func (bm *backupScheduleManager) createBackup(bs *v1alpha1.BackupSchedule, times
...
+		if backupSpec.StorageSize == "" {
+			if bs.Spec.StorageSize != "" {
+				backupSpec.StorageSize = bs.Spec.StorageSize
+			} else {
+				backupSpec.StorageSize = constants.DefaultStorageSize
+			}
+		}
+	} else {
+		if backupSpec.S3 != nil {
+			backupSpec.S3.Prefix = path.Join(backupSpec.S3.Prefix, timestamp.UTC().Format(constants.TimeFormat))
```
onlymellb: User can't set Prefix?

pkg/controller/backup/backup_controller.go
```diff
@@ -173,6 +174,9 @@ func (bkc *Controller) sync(key string) error {
 		return err
 	}
 
+	if !validBackup(backup) {
+		return nil
+	}
```
onlymellb: We can define a BackupInvalid [BackupConditionType](https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/types.go#L622), if check backup failed, update this condition. Then make a judgment [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/backup/backup_controller.go#L194), so that the invalid backup CR will not be enqueued.

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: updated

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: I have added type `db` and `table` for br [here](https://github.com/pingcap/tidb-operator/pull/1280/files#diff-98e6ad9bd9688c018746d6bf46e0e6d6R596).
I will add validation in the [validBackup](https://github.com/pingcap/tidb-operator/pull/1280/files#diff-5d667fa465378b0536cbc88f73233d91R223) function.

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: Here is to return error, the caller will log the error with backup CR info.

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: Here is to return error, the caller will log the error with backup CR info.

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: Here is to return error, the caller will log the error with backup CR info.

pkg/backup/backupschedule/backup_schedule_manager.go
DanielZhangQD: Backup files from BR do not have timestamp info, we have to distinguish different backups with paths, I will change the path to include cluster info.

pkg/controller/backup/backup_controller.go
DanielZhangQD: updated

DanielZhangQD:
@onlymellb comments addressed, PTAL.

pkg/backup/backup/backup_manager.go
```diff
@@ -85,7 +81,21 @@ func (bm *backupManager) syncBackupJob(backup *v1alpha1.Backup) error {
...
+		uErr := bm.statusUpdater.Update(backup, &v1alpha1.BackupCondition{
+			Type:    v1alpha1.BackupInvalid,
+			Status:  corev1.ConditionTrue,
+			Reason:  "InvalidSpec",
+			Message: err.Error(),
+		})
+		if uErr != nil {
+			glog.Warningf("Update condition %s failed for backup %s/%s", v1alpha1.BackupInvalid, ns, name)
+		}
+		return nil
```
onlymellb: We can return `controller.IsIgnoreError` instead. refer [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/backupschedule/backup_schedule_controller.go#L143)

pkg/backup/backup/backup_manager.go
```diff
@@ -85,7 +81,21 @@ func (bm *backupManager) syncBackupJob(backup *v1alpha1.Backup) error {
...
+	err := backuputil.ValidateBackup(backup)
+	if err != nil {
+		uErr := bm.statusUpdater.Update(backup, &v1alpha1.BackupCondition{
+			Type:    v1alpha1.BackupInvalid,
+			Status:  corev1.ConditionTrue,
+			Reason:  "InvalidSpec",
+			Message: err.Error(),
+		})
+		if uErr != nil {
+			glog.Warningf("Update condition %s failed for backup %s/%s", v1alpha1.BackupInvalid, ns, name)
```
onlymellb: If the update failed, the error log will be output in the update method, so I think it is not necessary to repeatedly output the error log.

pkg/backup/backup/backup_manager.go
DanielZhangQD: updated

pkg/backup/backup/backup_manager.go
DanielZhangQD: updated

PR #1287, opened 2019-12-05
[host network] keep backward compatibility in release-1.0

cofyc:
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Helm charts change
 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
bugfix: don't set DNSPolicy if hostNetwork is not enabled to keep backward compatibility
 ```

### DIFF
--- a/pkg/manager/member/pd_member_manager.go
+++ b/pkg/manager/member/pd_member_manager.go
@@ -510,10 +510,41 @@ func (pmm *pdMemberManager) getNewPDSetForTidbCluster(tc *v1alpha1.TidbCluster) 
 			Value: tc.Spec.Timezone,
 		},
 	}
+	podSpec := corev1.PodSpec{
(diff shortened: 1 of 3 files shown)


pkg/manager/member/pd_member_manager.go
```diff
@@ -522,6 +553,7 @@ func (pmm *pdMemberManager) getNewPDSetForTidbCluster(tc *v1alpha1.TidbCluster)
 				},
 			},
 		})
+		podSpec.Containers[0].Env = env
```
weekface: only add `env` when using `hostNetwork`?

cofyc: ah.. sorry, fixed

Issue #1212, opened 2019-11-21
Deploy the webhook server as an AA Server to simplify deployment


weekface:
If we use [generic-admission-server](https://github.com/openshift/generic-admission-server#deployment)(or our own AA Server) to develop our webhook server, and deploy our webhook server as an AA server, then the AA Server will act as a proxy to our webhook server.

 If we set `ValidatingWebhookConfiguration`'s [caBundle](https://github.com/kubernetes/api/blob/master/admissionregistration/v1/types.go#L529) to `nil`, the webhook controller(in kube-apiserver) will authenticate the requests to the AA Server:

![image](https://user-images.githubusercontent.com/9758/69320922-0dc9dd80-0c7d-11ea-9f28-4d720188c32b.png)

We can also tell [APIService](https://github.com/kubernetes/kube-aggregator/blob/39a1409e2e9e9ffe96d3c2b8122f25933940b4ef/pkg/apis/apiregistration/v1/types.go#L59) to disables TLS certificate verification when communicating with AA Server.

So, by default:
- develop our webhook server using [generic-admission-server](https://github.com/openshift/generic-admission-server#deployment)(or our own AA Server)
- deploy it as an AA Server and set its `InsecureSkipTLSVerify ` to `true`

If the user wants to open TLS certificate verification to AA Server, just use [cert-manager](https://github.com/jetstack/cert-manager) which will automatically provisioning and managing TLS certificates (for examples: APIService's `caBundle` and the AA Server's server.cert and server.key).

Ref:
https://kubernetes.io/blog/2018/01/extensible-admission-is-beta/
https://github.com/openshift/generic-admission-server

@tennix @Yisaer

Yisaer:
This is helpful and worth to try.

Yisaer:
Closed in https://github.com/pingcap/tidb-operator/pull/1288

PR #1288, opened 2019-12-05
Make webhook as apiservice server

Yisaer:
https://github.com/pingcap/tidb-operator/issues/1212

Making admission-webhook as an Apiservice.

As the latest kubernetes version for [generic-admission-server](https://github.com/openshift/generic-admission-server) supported was 1.14, I have to make admission version into `v1beta1` instead of `v1`

I would try to make [generic-admission-server](https://github.com/openshift/generic-admission-server) support kubernetes 1.16 version later and make admission version back to `v1`

I also transferred current sts webhook and pod webhook into apiservice server.

As apiserver in operator is now disabled in default and generic-admission-server provided a easy interface, apiserver webhook would work as an independent server for now.
In future, apiserver and apiserver webhook would mixed into one server.


weekface:
What about moving the webhook server code into our own AA Server: https://github.com/pingcap/tidb-operator/tree/master/cmd/apiserver?

Yisaer:
> What about moving the webhook server code into our own AA Server: https://github.com/pingcap/tidb-operator/tree/master/cmd/apiserver?
 
As generic-admission-server now provided a easy and clear interface for us, I would try to research how much work we need to do to move it into current apiserver.

Yisaer:
As `apiserver` in `operator` is now disabled in default and generic-admission-server provided a easy interface, `apiserver webhook` would work as an independent server for now.
In future, `apiserver` and `apiserver webhook` would mixed into one server.

Makefile
```diff
@@ -40,7 +40,7 @@ docker: build
 endif
 	docker build --tag "${DOCKER_REGISTRY}/pingcap/tidb-operator:${IMAGE_TAG}" images/tidb-operator
 
-build: controller-manager scheduler discovery admission-controller apiserver backup-manager
+build: controller-manager scheduler discovery aggragated-webhook apiserver backup-manager
```
weekface: I think this is still an admission webhooks, `aggragated-webhook ` is strange.

Yisaer: updated

manifests/webhook.yaml
```diff
@@ -1,109 +1,155 @@
...
     metadata:
       labels:
-        app: admission-controller
+        app: aggragated-webhook
     spec:
-      serviceAccountName: admission-controller-sa
+      serviceAccountName: aggragated-webhook-sa
       containers:
-        - name: admission-controller
+        - name: aggragated-webhook
```
weekface: should fix all the `aggragated ` words

manifests/patch-e2e.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
weekface: Is this script used for k8s less than `v1.13`?

Yisaer: > Is this script used for k8s less than `v1.13`?

Yes, in e2e test with 1.12 k8s version, we had to declare cabundle.

manifests/webhook.yaml
Yisaer: updated

pkg/webhook/pod_admission_hooks.go
```diff
@@ -0,0 +1,132 @@
...
+		Interface: event.New(kubeCli.CoreV1().RESTClient()).Events("")})
+	recorder := eventBroadcaster.NewRecorder(v1alpha1.Scheme, core.EventSource{Component: "tidbcluster"})
+
+	pc := pod.NewPodAdmissionControl(kubeCli, cli, pdControl, informerFactory, kubeInformerFactory, recorder, strings.Split(a.ExtraServiceAccounts, ","), a.EvictRegionLeaderTimeout)
+	a.podAC = pc
+	informerFactory.Start(stopCh)
+	kubeInformerFactory.Start(stopCh)
+
+	// Wait for all started informers' cache were synced.
+	for _, synced := range informerFactory.WaitForCacheSync(wait.NeverStop) {
```
weekface: use `kubeCli` to get object instead of informers.

manifests/webhook.yaml
```diff
@@ -116,9 +162,9 @@ webhooks:
     failurePolicy: Fail
     clientConfig:
       service:
-        name: admission-controller-svc
-        namespace: ${NAMESPACE}
-        path: "/statefulsets"
+        name: kubernetes
+        namespace: default
+        path: "/apis/admission.tidb.pingcap.com/v1alpha1/statefulsetadmissionreviews"
```
weekface: use a single path for all webhooks: `/apis/admission.tidb.pingcap.com/v1alpha1/admissionreviews`.

weekface:
Rest LGTM

pkg/webhook/pod_admission_hooks.go
Yisaer: issued in https://github.com/pingcap/tidb-operator/issues/1315

manifests/webhook.yaml
Yisaer: issued in https://github.com/pingcap/tidb-operator/issues/1316

manifests/patch-e2e.sh
cofyc: to make this clear, I think we should run this script only if the k8s version is less than 1.13. here is an example on how to compare k8s version (https://github.com/pingcap/tidb-operator/blob/eb11bce9107ff7f0f484e8fb8cfe5d6b8cb10daf/tests/e2e/tidbcluster/tidbcluster.go#L128-L132)

Yisaer: Thanks! I would enhance it next issue.

cofyc: create an issue or a TODO item for this

Yisaer: issued in https://github.com/pingcap/tidb-operator/issues/1317

PR #1290, opened 2019-12-05
pdapi: only read secret for PD client cert (on TLS enabled clusters) …

AstroProfundis:
…when needed

`pdapi.GetPDClient()` reads `Secret` for PD client cert every time it is called for TLS enabled clusters, but that's not necessary.

The `pdapi.GetPDClient()` already implements a simple cache for PD clients it handled, so only read `Secret` at the first time it is called for a TLS enabled cluster could improve the performance a little.

Read `Secret` only when the PD client is not in cache (for TLS enabled clusters)

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test

Code changes

 - Has Go code change

Side effects

- None

Related changes

 ```release-note

 ```

### DIFF
--- a/pkg/pdapi/pdapi.go
+++ b/pkg/pdapi/pdapi.go
@@ -83,7 +83,7 @@ func (pdc *defaultPDControl) GetPDClient(namespace Namespace, tcName string, tls
 
 			rootCAs, tlsCert, err := certutil.LoadCerts(secret.Data["cert"], secret.Data["key"])
 			if err != nil {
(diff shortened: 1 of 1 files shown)


Yisaer:
What about if we need to update the secret when the cert was going to be expired? Would the controller continue to use the expired cert ?

AstroProfundis:
No expiration check at all (yet), as what we do for other certs at present. That should be added (replace the PD client in the cache) when implementing cert renewing, along with other TiDB components' certs.

PR #1291, opened 2019-12-05
sync tidb configmap in controller

### LINKED ISSUE #1192: Sync TiDB Configuration and start-script in control-loop

aylei:

close #1192 

New controller logic to sync tidb configmap.

Generify the `CreateOrUpdate` logic and the template rendering code.

This PR introduce a new dependency [`controller-runtime`](https://github.com/kubernetes-sigs/controller-runtime) in favor of its generic client for kubernetes. 

### Check List 

Tests 

 - E2E test

 ```release-note
The configuration of tidb-server can be managed in the `TidbCluster` resource now.
 ```

check the e2e case to get an overview of this PR: https://github.com/pingcap/tidb-operator/pull/1291/files#diff-92579a43c91e473b9716eeeee2068004

@onlymellb @cofyc @tennix @weekface @Yisaer PTAL


pkg/controller/configmap_control.go
```diff
@@ -85,32 +114,48 @@ func (cc *realConfigMapControl) UpdateConfigMap(tc *v1alpha1.TidbCluster, cm *co
...
 
+// mergeConfigMap merge the desired configmap to current one
+func mergeConfigMap(current runtime.Object, desired runtime.Object) error {
+	cmC := current.(*corev1.ConfigMap)
+	cmD := desired.(*corev1.ConfigMap)
+	cmC.Data = cmD.Data
+	cmC.Labels = cmD.Labels
+	for k, v := range cmD.Annotations {
+		cmC.Annotations[k] = v
+	}
```
cofyc: why we must keep old annotations?

pkg/manager/member/tidb_member_manager.go
```diff
@@ -433,11 +537,14 @@ func getNewTiDBHeadlessServiceForTidbCluster(tc *v1alpha1.TidbCluster) *corev1.S
...
 
-func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster) *apps.StatefulSet {
+func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap) *apps.StatefulSet {
 	ns := tc.GetNamespace()
 	tcName := tc.GetName()
 	instanceName := tc.GetLabels()[label.InstanceLabelKey]
 	tidbConfigMap := controller.MemberConfigMapName(tc, v1alpha1.TiDBMemberType)
+	if cm != nil {
+		tidbConfigMap = cm.Name
+	}
```
cofyc: in which case, cm will be `nil`?

aylei: when `.spec.tidb.config` is nil

pkg/controller/configmap_control.go
aylei: It is not a must. I preserve the un-overridden annotations deliberately, which allows user and third-party components add their annotations to resources owned by TidbCluster.

pkg/controller/configmap_control.go
```diff
@@ -85,32 +114,48 @@ func (cc *realConfigMapControl) UpdateConfigMap(tc *v1alpha1.TidbCluster, cm *co
...
-		cc.recorder.Event(tc, corev1.EventTypeWarning, reason, msg)
+		msg := fmt.Sprintf("%s ConfigMap %s for %s/%s failed error: %s",
+			strings.ToLower(verb), cmName, kind, name, err)
+		cc.recorder.Event(owner, corev1.EventTypeWarning, reason, msg)
 	}
 }
 
+// mergeConfigMap merge the desired configmap to current one
+func mergeConfigMap(current runtime.Object, desired runtime.Object) error {
+	cmC := current.(*corev1.ConfigMap)
```
onlymellb: From the name of this function, it is to merge two `ConfigMap` resources, so why the function parameters use interface types instead of specific `ConfigMap`  types?

aylei: Good catch, it was intended to match the function signature of `Merge`, I will make it more clear

aylei:
I'm not quite satisfied with the generalization made in this PR and find that [`controller-runtime`](https://github.com/kubernetes-sigs/controller-runtime) has better implementation for common controller operations, refactoring to use `controller-runtime` now.

Marked as `DNM`, sorry for the reviewers @cofyc @onlymellb

pkg/controller/generic_control.go
```diff
@@ -0,0 +1,256 @@
...
+	result, err := w.GenericControlInterface.CreateOrUpdate(controller, cm, MergeConfigMap)
+	if err != nil {
+		return nil, err
+	}
+	return result.(*corev1.ConfigMap), nil
+}
+
+// GenericControlInterface manages generic object that managed by an arbitrary controller
+type GenericControlInterface interface {
+	CreateOrUpdate(controller, obj runtime.Object, mergeFn MergeFn) (runtime.Object, error)
```
aylei: `Apply` is ambiguous with the [`server-side apply`](https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply) of kubernetes, rename to `CreateOrUpdate`

cmd/controller-manager/main.go
```diff
@@ -114,6 +116,11 @@ func main() {
 	if err != nil {
 		glog.Fatalf("failed to get advanced-statefulset Clientset: %v", err)
 	}
+	// TODO: optimize the read of genericCli with the shared cache
+	genericCli, err := client.New(cfg, client.Options{Scheme: scheme.Scheme})
```
aylei: Using the client of `controller-runtime` without using the `informer` of `controller-runtime` has a drawback that all read operations will go to kube-apiserver, it is acceptable now because the only read operation is to get the latest version when perform a `CreateOrUpdate`.

pkg/controller/generic_control.go
```diff
@@ -0,0 +1,256 @@
...
+//     existing.spec.externalTrafficPolicy = desired.spec.externalTrafficPolicy
+// Instead of:
+//     existing.spec = desired.spec
+//
+// However, this could be tedious for large object if the caller want to control lots of the fields,
+// if there is no one else will mutate this object or cooperation is not needed, it is okay to do aggressive
+// override. Note that aggressive override usually causes unnecessary updates because the object will be mutated
+// after POST/PUT to api-server (e.g. Defaulting), an annotation based technique could be used to avoid such
+// updating: set a last-applied-config annotation and diff the annotation instead of the real spec.
+type MergeFn func(existing, desired runtime.Object) error
```
aylei: `server-side apply` is a better solution in the future.

ref: https://kubernetes.io/docs/reference/using-api/api-concepts/#server-side-apply

pkg/controller/configmap_control.go
aylei: I've re-write the "apply" part, PTAL again

pkg/controller/controller_utils.go
```diff
@@ -335,3 +338,17 @@ func (rt *RequestTracker) SetRequests(requests int) *RequestTracker {
 func (rt *RequestTracker) GetError() error {
 	return rt.err
 }
+
+// MergeConfigMap is a MergeFn for generic configmap.
+// parameter types are set to runtime.Object to match the signature of MergeFn
+func MergeConfigMap(existing, desired runtime.Object) error {
+	existingCm := existing.(*corev1.ConfigMap)
```
tennix: Should we check this type assertion?

aylei: I will inline this function to the `CreateOrUpdateConfigMap` interface for simplicity, a public method without type assertion checking is dangerous indeed.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -433,6 +436,57 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+		ginkgo.By("Test managing TiDB configmap in TidbCluster CRD")
+		tc.Spec.TiDB.Config = &v1alpha1.TiDBConfig{}
+		tc.Spec.TiDB.ConfigUpdateStrategy = v1alpha1.ConfigUpdateStrategyInPlace
+
+		_, err = cli.PingcapV1alpha1().TidbClusters(tc.Namespace).Update(tc)
+		framework.ExpectNoError(err, "Expected update tidbcluster")
+
+		tidbSetName := controller.TiDBMemberName(tc.Name)
+		oldTiDBSet, err := c.AppsV1().StatefulSets(tc.Namespace).Get(tidbSetName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "Expected get TiDB statefulset")
```
cofyc: should we get old TiDB statefulset before updating TiDB cluster?

aylei: Yes, this case has to verify the controllerrevision will not be changed after we manage configmap in controller (set `.spec.tidb.config` to non-nil)

cofyc: this is what I think:
- deploy tidb cluster
- wait tidb cluster is fully up
- get tidb statefulset
- update tidb cluster
- check if tidb cluster is not upgraded in a few minutes (e.g.  if the tidb cluster is not upgraded in 2 minutes, then we think tidb-operator will not upgrade tidb cluster because of our change)
- wait for our check is ok in 5 minutes
  - compare with old tidb set before updating tidb cluster

aylei: Yes, wait for a certain period is a safer choice. I would check the tidb-servers are not upgrade for 1 minutes after the tidb-statefulset get adopted by `tidbcluster`, because it may takes time for `tidb-controller-manager` to sync the statefulset, and after the statefulset is fully synced, 1 minute should be enough for `kube-controller-manager` to determine whether the statefulset should be rolling-udpated

aylei: oh, tidb-controller do not adopt statefulset, my fault

aylei:
@weekface @cofyc I've addressed the comments, PTAL again

aylei:
@weekface Could you please take a look again?

pkg/controller/configmap_control.go
```diff
@@ -27,16 +28,21 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ConfigMapControlInterface manages configmaps used by TiDB clusters
 type ConfigMapControlInterface interface {
-	CreateConfigMap(*v1alpha1.TidbCluster, *corev1.ConfigMap) error
-	UpdateConfigMap(*v1alpha1.TidbCluster, *corev1.ConfigMap) (*corev1.ConfigMap, error)
-	DeleteConfigMap(*v1alpha1.TidbCluster, *corev1.ConfigMap) error
+	// CreateConfigMap create the given ConfigMap
+	CreateConfigMap(runtime.Object, *corev1.ConfigMap) (*corev1.ConfigMap, error)
```
tennix: Should add a comment for what the `runtime.Object` is.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -355,6 +369,96 @@ func (tmm *tidbMemberManager) syncTiDBService(tc *v1alpha1.TidbCluster) error {
...
+	if config == nil {
+		return nil, nil
+	}
+
+	// override CA if tls enabled
+	if tc.Spec.EnableTLSCluster {
+		if config.Security == nil {
+			config.Security = &v1alpha1.Security{}
+		}
+		config.Security.ClusterSSLCA = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
```
tennix: It's better to define these 3 paths as constant.

aylei: Good catch!

Issue #1255, opened 2019-11-29
Feature Request: Add common wait-unit init-container 


Yisaer:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently we need a common wait-unit init container for `tidb-initializer` / ~~`tikv` / `tidb`~~ (`pump` and `drainer`) to wait until their dependency ready. We also want it to have a `timeout` configuration in order to avoid wating forever.

gregwebs:
This is already developed on this PR: https://github.com/pingcap/tidb-operator/pull/581
Also the discovery service is more developed on this PR: https://github.com/pingcap/tidb-operator/pull/935

Yisaer:
Hi @gregwebs , I checked the pr #581 and found the `wait-for-pd` might be more uncommon or complex for this feature request.

I think a `busybox` image with `nc` checking whether the target port was opened is enough for a common wait-until init container.

cofyc:
we need to wait for the `host:port` of TiDB service is ready (as @Yisaer said a simple TCP port check is enough) before trying to run commands in `tidb-initializer`, `pump` and `drainer` (the description has been updated) because the kubelet will wait more and more time to restart these pods to retry if they fail fast many times and we can't configure restart backoff time in k8s.

gregwebs:
Yes that sounds great for TiDB. Then my endpoint ready check for TiDB initializer would work properly.

cofyc:
I'm going to add support for tidb-initializer (https://github.com/pingcap/tidb-operator/pull/1296). This saves my time in local development.

PR #1296, opened 2019-12-06
Before proceeding to initialize TiDB, we wait for the TiDB host:port is open.

cofyc:

https://github.com/pingcap/tidb-operator/issues/1255

Before proceeding to initialize TiDB, we wait for the TiDB `host:port` is open. This avoids unnecessary pod restarts which may result in a very long restart delay (10s, 20s, .., 5 mins).
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy

### Check List 

Tests 

 - Unit test
 - E2E test

 ```release-note
tidb-initializer: wait for the TiDB host:port is open before processing to initialize TiDB. This speeds up TiDB initialization.
 ```


charts/tidb-cluster/templates/tidb-initializer-job.yaml
```diff
@@ -19,6 +19,31 @@ spec:
...
+      # restart delay (10s, 20s, .., 5 mins).
+      # https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy
+      - name: wait
+        image: {{ .Values.busybox.image }}
+        imagePullPolicy: {{ .Values.busybox.imagePullPolicy | default "IfNotPresent" }}
+        command:
+        - sh
+        - -c
+        - |
+          trap exit TERM
```
cofyc: able to terminate fast on kill container signal (TERMINAL) (xref: https://github.com/kubernetes/kubernetes/pull/64015#issue-188961717)

gregwebs: I always run something like [dumb-init](https://github.com/Yelp/dumb-init) in my pods so that these issues don't show up

PR #1304, opened 2019-12-09
e2e: run process outside of the cluster

### LINKED ISSUE #1274: run e2e process outside of the cluster
## Feature Request

- stability test will restart k8s nodes which may affect e2e pod

run e2e process outside of the cluster

- access k8s DNS/IPs
  - via proxy

cofyc:

fixes #1274 

- run e2e webhook in a delegated pod
  - check if the e2e webhook failed in upgrading process to verify our logic
- run blockwriter via a delegated pod
- use port forwarding mechanism to access pod ports (tidb, pd, prometheus, grafana)
  - portforward package is provided (see [docs](https://godoc.org/k8s.io/client-go/tools/portforward))
  - a few utilities are provided
    - proxiedpdclient
    - proxiedtidbclient

### Check List 

Tests 

 - Unit test
 - E2E test


Makefile
```diff
@@ -90,15 +90,16 @@ endif
 e2e-build: test-apiserver-build
 	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/ginkgo github.com/onsi/ginkgo/ginkgo
 	$(GO) test -c -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/e2e.test ./tests/e2e
+	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/webhook ./tests/cmd/webhook
 
-test-apiserver-dokcer-push: test-apiesrver-docker
+test-apiserver-dokcer-push: test-apiserver-docker
```
cofyc: typo only

Makefile
```diff
@@ -90,15 +90,16 @@ endif
 e2e-build: test-apiserver-build
 	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/ginkgo github.com/onsi/ginkgo/ginkgo
 	$(GO) test -c -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/e2e.test ./tests/e2e
+	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/webhook ./tests/cmd/webhook
```
cofyc: add webhook into e2e image (no need an extra image)

hack/e2e.sh
```diff
@@ -120,13 +120,12 @@ function e2e::image_build() {
     fi
     DOCKER_REGISTRY=$DOCKER_REGISTRY IMAGE_TAG=$IMAGE_TAG make docker
     DOCKER_REGISTRY=$DOCKER_REGISTRY IMAGE_TAG=$IMAGE_TAG make e2e-docker
-    DOCKER_REGISTRY=$DOCKER_REGISTRY IMAGE_TAG=$IMAGE_TAG make test-apiesrver-docker
+    DOCKER_REGISTRY=$DOCKER_REGISTRY IMAGE_TAG=$IMAGE_TAG make test-apiserver-docker
```
cofyc: typo only

hack/run-e2e.sh
```diff
@@ -112,41 +85,11 @@ e2e_args=(
...
-    # TODO find a better way to print logs starting from the last place
-    execType=1
-done
+docker run -it --rm \
+    --net=host \
+    -v $KUBECONFIG:/etc/kubernetes/admin.conf \
+    --env KUBECONFIG=/etc/kubernetes/admin.conf \
+    --env KUBECONTEXT=$KUBECONTEXT \
+    $E2E_IMAGE \
+    ${e2e_args[@]}
```
cofyc: run e2e process locally

pkg/controller/tidb_control.go
```diff
@@ -34,16 +34,16 @@ const (
 	timeout          = 5 * time.Second
 )
 
-type dbInfo struct {
+type DBInfo struct {
```
cofyc: export it, then I can implement `proxiedtidbclient.NewProxiedTiDBClient`.

pkg/pdapi/pdapi.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	timeout = 5 * time.Second
+	DefaultTimeout = 5 * time.Second
```
cofyc: export it, then I an use the same timeout

pkg/pdapi/pdapi.go
```diff
@@ -61,12 +61,31 @@ func NewDefaultPDControl(kubeCli kubernetes.Interface) PDControlInterface {
 	return &defaultPDControl{kubeCli: kubeCli, pdClients: map[string]PDClient{}}
 }
 
+// GetTLSConfig returns *tls.Config for given TiDB cluster.
+func GetTLSConfig(kubeCli kubernetes.Interface, namespace Namespace, tcName string) (*tls.Config, error) {
```
cofyc: export it then I can reuse this utility in proxiedpdclient package.

tests/actions.go
```diff
@@ -92,9 +96,14 @@ func NewOperatorActions(cli versioned.Interface,
...
 		// tcStsGetter:  helper.NewHijackClient(kubeCli, asCli).AppsV1(),
-		tidbControl:  controller.NewDefaultTiDBControl(),
 		pollInterval: pollInterval,
 		cfg:          cfg,
+		fw:           fw,
+	}
+	if fw != nil {
+		oa.tidbControl = proxiedtidbclient.NewProxiedTiDBClient(fw)
+	} else {
+		oa.tidbControl = controller.NewDefaultTiDBControl()
```
cofyc: keep backward compatibility if `fw` is not provided. we can remove this when we can run stability test cases outside of the Kubernetes cluster.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -69,11 +73,23 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+		fwCancel = cancel
+		oa = tests.NewOperatorActions(cli, c, asCli, tests.DefaultPollInterval, e2econfig.TestConfig, nil, fw)
 		cfg = e2econfig.TestConfig
 		ocfg = e2econfig.NewDefaultOperatorConfig(cfg)
 	})
 
+	ginkgo.AfterEach(func() {
+		if fwCancel != nil {
+			fwCancel()
+		}
```
cofyc: after each test, cancel the up-level context

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -69,11 +73,23 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
 		asCli, err = asclientset.NewForConfig(config)
 		framework.ExpectNoError(err, "failed to create clientset")
-		oa = tests.NewOperatorActions(cli, c, asCli, tests.DefaultPollInterval, e2econfig.TestConfig, nil)
+		clientRawConfig, err := e2econfig.LoadClientRawConfig()
+		framework.ExpectNoError(err, "failed to load raw config")
+		ctx, cancel := context.WithCancel(context.Background())
+		fw, err := portforward.NewPortForwarder(ctx, e2econfig.NewSimpleRESTClientGetter(clientRawConfig))
+		framework.ExpectNoError(err, "failed to create port forwarder")
+		fwCancel = cancel
+		oa = tests.NewOperatorActions(cli, c, asCli, tests.DefaultPollInterval, e2econfig.TestConfig, nil, fw)
```
cofyc: we use port-forwarding in e2e, then we can access pod ports outside of the Kubernetes cluster

charts/tidb-cluster/templates/scripts/_start_drainer.sh.tpl
```diff
@@ -27,7 +27,7 @@ done
 /drainer \
 -L={{ .Values.binlog.drainer.logLevel | default "info" }} \
 -pd-urls={{ template "cluster.scheme" . }}://{{ template "cluster.name" . }}-pd:2379 \
--addr=`echo ${HOSTNAME}`.{{ template "cluster.name" . }}-drainer:8249 \
+-addr=:8249 \
```
cofyc: able to port forward  to 8249 (k8s [hard coded host as localhost](https://github.com/kubernetes/kubernetes/blob/928817a26a84d9e3076d110ea30ba994912aa477/pkg/kubelet/dockershim/docker_streaming_others.go#L47))

charts/tidb-drainer/templates/scripts/_start_drainer.sh.tpl
```diff
@@ -27,7 +27,7 @@ done
 /drainer \
 -L={{ .Values.logLevel | default "info" }} \
 -pd-urls=http://{{ .Values.clusterName }}-pd:2379 \
--addr=`echo ${HOSTNAME}`.{{ include "drainer.name" . }}:8249 \
+-addr=:8249 \
```
cofyc: able to port forward  to 8249 (k8s [hard coded host as localhost](https://github.com/kubernetes/kubernetes/blob/928817a26a84d9e3076d110ea30ba994912aa477/pkg/kubelet/dockershim/docker_streaming_others.go#L47))

tests/e2e/e2e.go
```diff
@@ -136,7 +135,26 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
+	// pdClient, cancel, err := proxiedpdclient.NewProxiedPDClient(kubeCli, fw, "default", "quelling-wolverine", false)
+	// framework.ExpectNoError(err, "failed to create proxied pd client")
+	// for {
+	// cluster, err := pdClient.GetCluster()
+	// fmt.Printf("cluster: %v, err: %v\n", cluster, err)
+	// time.Sleep(3 * time.Second)
+	// }
+	// time.Sleep(time.Minute)
+	// cancel()
+	// os.Exit(0)
```
cofyc: ah, forgot to remove this

charts/tidb-cluster/templates/scripts/_start_drainer.sh.tpl
tennix: This might cause the existing drainer to be restarted. I'm not sure if we should allow this for major tidb-operator upgrade.

tests/e2e/util/proxiedtidbclient/proxiedtidbclient.go
```diff
@@ -0,0 +1,103 @@
...
+		Path:   "/settings",
+	}
+	req, err := http.NewRequest("GET", u.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	res, err := p.httpClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
```
tennix: Use `resp, err := p.httpClient.Get(u.String())` directly.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -160,26 +176,30 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+
+		ginkgo.By("Register webhook")
+		oa.RegisterWebHookAndServiceOrDie(ocfg.WebhookConfigName, ns, svc.Name, certCtx)
+
+		ginkgo.By(fmt.Sprintf("Deploying tidb cluster %s", cluster.ClusterVersion))
 		oa.DeployTidbClusterOrDie(&cluster)
 		oa.CheckTidbClusterStatusOrDie(&cluster)
 		oa.CheckDisasterToleranceOrDie(&cluster)
 
-		cluster.ScalePD(3)
```
cofyc: I removed this because scaling logic has already been tested in another test. Each test focus on its own purpose, then every test will be small and easier to test/debug.

tests/e2e/util/proxiedtidbclient/proxiedtidbclient.go
```diff
@@ -0,0 +1,103 @@
...
+func (p *proxiedTiDBClient) GetInfo(tc *v1alpha1.TidbCluster, ordinal int32) (*controller.DBInfo, error) {
+	panic("implement when necessary")
+}
+
+func (p *proxiedTiDBClient) GetSettings(tc *v1alpha1.TidbCluster, ordinal int32) (*config.Config, error) {
+	tcName := tc.GetName()
+	ns := tc.GetNamespace()
+	scheme := tc.Scheme()
+
+	if tc.Spec.EnableTLSCluster {
```
tennix: I think this should be the client side TLS not cluster level TLS.

charts/tidb-cluster/templates/scripts/_start_drainer.sh.tpl
cofyc: I'll revert this. Alternatively, we can exec into the pod to check the API.

tests/e2e/util/proxiedtidbclient/proxiedtidbclient.go
cofyc: you mean `tc.TiDB.EnableTLSClient`? this client is used to access tidb API port (10080), the switch is `tc.Spec.EnableTLSCluster` (I copied this from https://github.com/pingcap/tidb-operator/blob/b81862e568e8e1224282d7168631000b1151378d/pkg/controller/tidb_control.go#L138-L140)

tests/e2e/util/proxiedtidbclient/proxiedtidbclient.go
cofyc: good suggestion! done

tests/e2e/util/proxiedtidbclient/proxiedtidbclient.go
tennix: @AstroProfundis PTAL

tests/e2e/util/portforward/portforward.go
```diff
@@ -0,0 +1,163 @@
...
+	return forwardedPorts, cancel, nil
+}
+
+func (f *portForwarder) Forward(namespace, resourceName string, addresses []string, ports []string) (forwardedPorts []portforward.ForwardedPort, cancel context.CancelFunc, err error) {
+	builder := resource.NewBuilder(f).
+		WithScheme(scheme.Scheme, scheme.Scheme.PrioritizedVersionsAllGroups()...).
+		ContinueOnError().
+		NamespaceParam(namespace).DefaultNamespace()
+
+	getPodTimeout := time.Minute
```
aylei: defined as const?

tests/e2e/util/proxiedpdclient/proxiedpdclient.go
```diff
@@ -0,0 +1,78 @@
...
+	scheme := "http"
+	if tlsEnabled {
+		scheme = "https"
+		tlsConfig, err = pdapi.GetTLSConfig(kubeCli, pdapi.Namespace(namespace), tcName)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	ports := []string{fmt.Sprintf("0:%d", 2379)}
+	forwardedPorts, cancel, err := fw.Forward(namespace, fmt.Sprintf("svc/%s", controller.PDMemberName(tcName)), []string{"127.0.0.1"}, ports)
```
aylei: consider using `forwardOnePort` here?

tests/actions.go
```diff
@@ -2007,7 +2089,18 @@ func (oa *operatorActions) checkTiKVConfigUpdated(tc *v1alpha1.TidbCluster, clus
...
+	if oa.fw != nil {
+		localHost, localPort, cancel, err := portforward.ForwardOnePort(oa.fw, ns, fmt.Sprintf("svc/%s-prometheus", tcName), 9090)
+		if err != nil {
+			return err
+		}
+		defer cancel()
+		prometheusAddr = fmt.Sprintf("%s:%d", localHost, localPort)
+	} else {
+		prometheusAddr = fmt.Sprintf("%s-prometheus.%s:9090", tcName, ns)
+	}
```
aylei: This pattern appears multiple times, can we have a `oa.GetAddr()` (or separately, `oa.GetSvcAddr()` and `oa.GetPodAddr()`) method so that the detail of pod/svc connectivity could be handled in one place and port-forward could be used seamlessly in the new code without extra attention.

charts/tidb-cluster/templates/scripts/_start_drainer.sh.tpl
aylei: We can make this configurable and use the original one by default, in e2e, switch to listen on `0.0.0.0`

tests/actions.go
aylei: The implementation detail I mentioned above does not matter, the important thing is that we need a more friendly interface so that the author of test case don't have to care about whether the access to Pod need port-forward or not.

cofyc: I didn't add subroutines because these addresses are used in one place. Another reason is the `oa.fw != nil` check is temporary, after the stability tests are reimplemented in the same framework, this will be simplified, so I think current situation is acceptable. All users should use port-forwarding to access ports of the pod. (or `kubectl exec ...`)

tests/e2e/util/portforward/portforward.go
cofyc: done

tests/e2e/util/proxiedpdclient/proxiedpdclient.go
cofyc: good catch! done

charts/tidb-cluster/templates/scripts/_start_drainer.sh.tpl
cofyc: yes, it's a good solution. But I'm still going to use `exec` API because this is a general solution to access k8s resources too. K8S e2e framework provides [a lot of utilities] (https://github.com/kubernetes/kubernetes/blob/e76619931f805e3ba2ceece405b662891810c5e7/test/e2e/framework/exec_util.go#L50) around `exec` API. It's simple to use. No more dependencies and no need to change current charts.

Port forwarding is more useful in MySQL query checks because images normally do not contain MySQL client.

cofyc: implemented: https://github.com/pingcap/tidb-operator/pull/1304/files#diff-67dacabf080ca4dc8e95d2acf12ee36bR3161-R3171 
the alpine image has builtin wget (busybox)

cofyc:
@tennix @aylei PTAL

tests/actions.go
aylei: OK, LGTM

aylei:
LGTM, but I think we should wait @AstroProfundis to verify the tidb tls client config first

tests/e2e/util/proxiedtidbclient/proxiedtidbclient.go
AstroProfundis: `tc.TiDB.EnableTLSCluster` enables the server side TLS connection, so accessing TiDB's status API is controlled by this param.

`tc.TiDB.EnableTLSClient` enables the MySQL client <-> TiDB Server encryption, it uses another set of cert/key and only take affect to the SQL port (4000 by default).

So, I think it's correct to use `tc.TiDB.EnableTLSCluster` for the check.

hack/run-e2e.sh
```diff
@@ -103,50 +77,20 @@ e2e_args=(
...
-        echo "error: e2e failed, phase: $phase"
-        exit 1
-    fi
-    # if we failed on "$KUBECTL_BIN ${kubectl_args[@]:-} logs -f", try to print one line each time
-    # TODO find a better way to print logs starting from the last place
-    execType=1
-done
+docker run --rm \
+    --net=host \
+    -v $KUBECONFIG:/etc/kubernetes/admin.conf \
```
tennix: It's better to mount this as readonly.

cofyc: done in https://github.com/pingcap/tidb-operator/pull/1304/commits/6e094d012de80360d2754d8084ca0ecaeb99bc6d

Issue #1309, opened 2019-12-09
Add basic default values and validations for configuration schema


aylei:
follow up #1308

PR #1310, opened 2019-12-09
Add detailed schema for tidb cluster configurations

### LINKED ISSUE #1308: Generate detailed openapi schema for configurations

aylei:

close #1308 
This PR add a validation schema for TiDB configurations.

Prior to `v1.15`, kubernetes do not have [mature support for CRD schema](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema), in our scenario, kube-apiserver do not publish the schema of CRD and CRD do not allow specifying properties for fields like `default`, `maximum`. So the current work around is generate one and provide it via a go interface.

This PR adjust the struct of config in favor of schema generation, `default` value and `validation` of fields will be added in #1309 

UPDATE: discussed with @shuijing198799 , it is better to generate schema in client.

@weekface @LinuxGit @tennix @shuijing198799 PTAL


pkg/schema/types.go
```diff
@@ -0,0 +1,36 @@
...
+// +groupName=schema.pingcap.com
+// +versionName=v1alpha1
+package schema
+
+import (
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:resource
```
aylei: `controller-gen` do not support filtering resources in the same package, create a dummy type to generate the configuration schema.

pkg/apis/pingcap/v1alpha1/pd_config.go
```diff
@@ -261,10 +239,10 @@ type PDScheduleConfig struct {
 	MaxMergeRegionKeys *uint64 `toml:"max-merge-region-keys,omitempty,omitempty" json:"max-merge-region-keys,omitempty"`
 	// SplitMergeInterval is the minimum interval time to permit merge after split.
 	// +optional
-	SplitMergeInterval typeutil.Duration `toml:"split-merge-interval,omitempty,omitempty" json:"split-merge-interval,omitempty"`
+	SplitMergeInterval string `toml:"split-merge-interval,omitempty,omitempty" json:"split-merge-interval,omitempty"`
```
tennix: Two `omitempty` for toml?

aylei: Good cach

aylei:
@tennix @shuijing198799 @LinuxGit  Addressed, PTAL again

pkg/apis/pingcap/v1alpha1/pd_config.go
shuijing198799: catch ？ hahahahhahahahah

PR #1330, opened 2019-12-12
sync pd and tikv configmap in controller

### LINKED ISSUE #1263: sync pd configmap and start script in controller

### LINKED ISSUE #1264: sync tikv configmap and start script in controller

aylei:

close #1263 
close #1264 

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

 ```release-note
pd and tikv configs are now managed in TidbCluster resource.
 ```


pkg/apis/pingcap/v1alpha1/pd_config.go
```diff
@@ -250,7 +246,7 @@ type PDScheduleConfig struct {
 	// MaxStoreDownTime is the max duration after which
 	// a store will be considered to be down if it hasn't reported heartbeats.
 	// +optional
-	MaxStoreDownTime Duration `toml:"max-store-down-time,omitempty" json:"max-store-down-time,omitempty"`
+	MaxStoreDownTime string `toml:"max-store-down-time,omitempty" json:"max-store-down-time,omitempty"`
```
aylei: `BurntSushi/toml` do not respect `MarshalText`, use `string` instead

cofyc: `BurntSushi/toml` seems not maintained anymore...

we can use `https://github.com/pelletier/go-toml` and implement `toml.MarshalTOML()` (like MarshalJSON())

cofyc: https://play.golang.org/p/_SfokqsaCPo it works

aylei: Yes, that's exactly what I've thought.

Actually I've tried `go-toml`( and `naoina/toml`), it turns out that `go-toml` do not support `interface{}`. Also, PD and TiDB uses `BurntSushi/toml` so there is little verifying work, seems like that `BurntSushi/toml` is not an ideal but the only choice

Issue #1326, opened 2019-12-11
Scale TiDB to 0 did not clear failureMembers


aylei:
## Bug Report

#1325 
https://github.com/pingcap/tidb-operator/issues/1299

aylei:
closed via #1334

Issue #1325, opened 2019-12-11
Support scale TiDB to 0


aylei:
## Feature Request

See https://github.com/pingcap/tidb-operator/issues/1299 for original discussion.

Currently, scale TiDB to 0 will not clear the `failureMembers`, cause some tidb pods still stick in crashed state.

Also, the test, document also should be added about scaling TiDB to 0.

PR #1334, opened 2019-12-13
clear TiDB failureMembers when scale TiDB to 0 

mikechengwei:
#1325 
#1326 

When tidb member all crash，we can set replica to 0 which destory tidb sts.
### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

NONE


cofyc:
hi, Thanks for your contribution! Can you follow the instructions [here] to sign the CLA?

mikechengwei:
@cofyc  done

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -300,7 +300,11 @@ func (tc *TidbCluster) TiDBAllMembersReady() bool {
 }
 
 func (tc *TidbCluster) TiDBStsDesiredReplicas() int32 {
-	return tc.Spec.TiDB.Replicas + int32(len(tc.Status.TiDB.FailureMembers))
+	if tc.Spec.TiDB.Replicas == int32(0) {
+		return tc.Spec.TiDB.Replicas
+	} else {
+		return tc.Spec.TiDB.Replicas + int32(len(tc.Status.TiDB.FailureMembers))
+	}
```
aylei: I tend not to change this method to keep its semantic simple, how to handle `failureMembers` when replias is 0 is just another concern.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -208,7 +208,7 @@ func (tmm *tidbMemberManager) syncTiDBStatefulSetForTidbCluster(tc *v1alpha1.Tid
 	}
 
 	if tmm.autoFailover {
-		if tc.TiDBAllPodsStarted() && tc.TiDBAllMembersReady() && tc.Status.TiDB.FailureMembers != nil {
+		if (tc.TiDBAllPodsStarted() && tc.TiDBAllMembersReady() && tc.Status.TiDB.FailureMembers != nil) || *newTiDBSet.Spec.Replicas == int32(0) {
```
aylei: I think we should not perform `Recover` when `.failureMembers` is empty

pkg/manager/member/tidb_member_manager_test.go
```diff
@@ -224,6 +224,41 @@ func TestTiDBMemberManagerSyncUpdate(t *testing.T) {
 				g.Expect(set.Spec.Template.Spec.Containers).To(HaveLen(2))
 			},
 		},
+		{
+			name: "scale TiDB to 0 did not clear failureMembers",
```
aylei: It is better to use a behavior description as the case name, e.g. TiDB should clear failureMembers when scale to 0

pkg/manager/member/tidb_member_manager_test.go
```diff
@@ -224,6 +224,41 @@ func TestTiDBMemberManagerSyncUpdate(t *testing.T) {
...
+				}
+			},
+			errWhenUpdateStatefulSet: false,
+			err:                      false,
+			expectStatefulSetFn: func(g *GomegaWithT, set *apps.StatefulSet, err error) {
+				g.Expect(*set.Spec.Replicas).To(Equal(int32(0)))
+			},
+		},
+		{
+			name: "scale TiDB to 0 when stateful set is upgrading",
```
aylei: ditto

pkg/manager/member/tidb_upgrader.go
```diff
@@ -56,6 +56,10 @@ func (tdu *tidbUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Stateful
 		return nil
 	}
 
+	if *newSet.Spec.Replicas == int32(0) {
```
aylei: Need a comment to explain why we should return here, also, I think this predicate can be moved to the beginning of this method

aylei:
@mikechengwei Thanks for your contribution! This is a subtle change and I think we should add an e2e case to evaluate it, how do you think @mikechengwei ?

aylei:
@weekface Could you please check this change?

mikechengwei:
@aylei ok.I will add a e2e test and continue optimizing my code .

weekface:
It is better to do this behavior in the `validating` phase of the `TidbCluster` object. But we do not have a validating now.

Another option is to add this codes to the [`syncTidbClusterStatus`](https://github.com/pingcap/tidb-operator/blob/5ead47e9d230ed9bf0a6acd2d2a9cbde729e6c48/pkg/manager/member/tidb_member_manager.go#L775) method to solve the issue.

```
if tc.TiDB.Replicas == 0 {
    tc.Status.TiDB. FailureMembers = nil
}
```

mikechengwei:
@weekface I think clearing failmembers are required only when autofailover is enabled. So,add codes to 
```
	if tmm.autoFailover {
		if (tc.TiDBAllPodsStarted() && tc.TiDBAllMembersReady() || *newTiDBSet.Spec.Replicas == int32(0) ) && tc.Status.TiDB.FailureMembers != nil {
			tmm.tidbFailover.Recover(tc)
		} else if tc.TiDBAllPodsStarted() && !tc.TiDBAllMembersReady() {
			if err := tmm.tidbFailover.Failover(tc); err != nil {
				return err
			}
		}
	}
```
Is it better?

pkg/apis/pingcap/v1alpha1/tidbcluster.go
mikechengwei: If i don't change this method, I will need to change this code 	https://github.com/pingcap/tidb-operator/blob/5ead47e9d230ed9bf0a6acd2d2a9cbde729e6c48/pkg/manager/member/tidb_member_manager.go#L756 replicas and all methods using `TiKVStsDesiredReplicas` method .
I think change this method easier to understand. When scaling to 0 replicas,the  desired replicas is 0. Right?

pkg/manager/member/tidb_member_manager.go
mikechengwei: agree

pkg/apis/pingcap/v1alpha1/tidbcluster.go
aylei: Yes, but **that is exactly** the sematic I suppose, we do not magically ignore `failureMembers` in `tidbcluster` when `.spec.replica` is some specific value (e.g. 0), instead, we let the owner of `failureMembers`, aka `TiDBFailover` to manage the failureMembers.

mikechengwei: I get.

weekface:
> @weekface I think clearing failmembers are required only when autofailover is enabled. So,add codes to
> 
> ```
> 	if tmm.autoFailover {
> 		if ((tc.TiDBAllPodsStarted() && tc.TiDBAllMembersReady()) || *newTiDBSet.Spec.Replicas == int32(0)) && tc.Status.TiDB.FailureMembers != nil {
> 			tmm.tidbFailover.Recover(tc)
> 		} else if tc.TiDBAllPodsStarted() && !tc.TiDBAllMembersReady() {
> 			if err := tmm.tidbFailover.Failover(tc); err != nil {
> 				return err
> 			}
> 		}
> 	}
> ```
> 
> Is it better?

The judgment conditions are a bit too complicated. In fact, the thing to do here is simple. If the user sets `spec.tidb.replicas` to 0 and `autoFailover` is open, then we need to clear the `.status.tidb.failoverMember` property.

```
if tc.TiDB.Replicas == 0 && tmm.autoFailover {
    tc.Status.TiDB. FailureMembers = nil
}
```

However, this is just a suggestion. I am fine with your solution if the logic is correct.

mikechengwei:
@aylei @weekface  I have optimize code ,could you help me recheck code .

pkg/manager/member/tidb_member_manager.go
```diff
@@ -208,6 +208,10 @@ func (tmm *tidbMemberManager) syncTiDBStatefulSetForTidbCluster(tc *v1alpha1.Tid
 	}
 
 	if tmm.autoFailover {
+		if tc.Spec.TiDB.Replicas == int32(0) && tc.Status.TiDB.FailureMembers != nil {
+			tmm.tidbFailover.Recover(tc)
+			*newTiDBSet.Spec.Replicas = int32(0)
```
aylei: I think it is not necessary to mutate the replicas here, in the next round of reconciling, the desired replicas will calculated as 0 by nature.

pkg/manager/member/tidb_member_manager_test.go
```diff
@@ -1616,3 +1616,105 @@ func TestGetTiDBConfigMap(t *testing.T) {
 		})
 	}
 }
+
+func TestTiDBMemberManagerScaleToOneReplica(t *testing.T) {
```
aylei: `ToZero`?

pkg/manager/member/tidb_upgrader.go
```diff
@@ -35,6 +35,14 @@ func NewTiDBUpgrader(tidbControl controller.TiDBControlInterface, podLister core
 }
 
 func (tdu *tidbUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet, newSet *apps.StatefulSet) error {
+	/**
+	 * when scale replica to 0 , all nodes crash and tidb is in upgrade phase, this method will throw error about pod is upgrade.
+	 * so  directly return nil when scale replica to 0.
+	 */
+	if tc.Spec.TiDB.Replicas == int32(0) {
+		return nil
+	}
```
aylei: By convention, we use `//` for code comments

pkg/manager/member/tidb_member_manager.go
mikechengwei: done

pkg/manager/member/tidb_member_manager_test.go
mikechengwei: done

pkg/manager/member/tidb_upgrader.go
mikechengwei: done

PR #1343, opened 2019-12-16
Add Restarter Controller for tidbcluster controller

Yisaer:

Add Restarter Controller to let tidb component pod restart by adding annotations to it if the webhook feature is enabled. Currently the webhook feature is defaultly disabled.

#### TL;DR
The restarter controller would check whether there were pods needed to be restarting depends on their annotations and make sure each moment there would be 1 pod restarting at most. The webhook would ensure the whole process would be safe.

#### main points:
1. User restart the pod by adding annotation to it.
2. Restart controller would delete it during reconciliation period.
3. Webhook would ensure the whole pod deleting process would be safe.
4. Restart controller would ensure that there would be at most one pod restarting each moment.

Tests 

Code changes

 - Has Go code change


pkg/manager/member/restarter.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (gr *GeneralRestarter) pop(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType, sts *apps.StatefulSet) (*core.Pod, error) {
+	labelSelector := &meta.LabelSelector{
+		MatchLabels: map[string]string{
+			label.ComponentLabelKey: memberType.String(),
+			label.ManagedByLabelKey: "tidb-operator",
+			label.NameLabelKey:      "tidb-cluster",
+			label.InstanceLabelKey:  tc.Name,
+		},
+	}
```
aylei: use `label.New()` instead

pkg/manager/member/tikv_member_manager.go
```diff
@@ -209,6 +215,13 @@ func (tkmm *tikvMemberManager) syncStatefulSetForTidbCluster(tc *v1alpha1.TidbCl
 		return err
 	}
 
+	if tkmm.webhookEnabled {
+		err := tkmm.tikvRestarter.Sync(tc, v1alpha1.TiKVMemberType)
+		if err != nil {
+			return err
+		}
+	}
```
aylei: IIUC, restarting will have to wait `0.5 * ResyncPeroid` in average if there is no event triggers reconciling, and a Pod will not get restarted if there is anything wrong in the former reconciling procedures.

I think we should add a new controller and make it orthogonal with tdb-cluster controller

pkg/manager/member/restarter.go
```diff
@@ -0,0 +1,151 @@
...
+	return controller.RequeueErrorf("tc[%s/%s]'s pod[%s/%s] is restarted,requeue", namespace, tc.Name, namespace, podName)
+}
+
+// sync func would continue to try deleting the pod which is during restarting or pop one new pod with AnnPodDeferDeleting
+// to restart
+func (gr *GeneralRestarter) sync(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType, sts *apps.StatefulSet) error {
+	podName, existed := sts.Annotations[label.AnnPodRestarting]
+	if existed {
+		return gr.restart(tc, memberType, podName)
```
aylei: If the `podName` is not exist, `restart` will return a error and the restarting lock will never get cleaned

pkg/manager/member/restarter.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return err
+	}
+	return controller.RequeueErrorf("tc[%s/%s]'s pod[%s/%s] is restarted,requeue", namespace, tc.Name, namespace, podName)
+}
+
+// sync func would continue to try deleting the pod which is during restarting or pop one new pod with AnnPodDeferDeleting
+// to restart
+func (gr *GeneralRestarter) sync(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType, sts *apps.StatefulSet) error {
+	podName, existed := sts.Annotations[label.AnnPodRestarting]
```
aylei: I recommend not to queue restarting requests to avoid the extra complexibility, kubernetes do not queue `kubectl delete` neither.

aylei:
I think it deserves a design document if we are going to do something more than simply delete pods with specific annotation

pkg/manager/member/restarter.go
Yisaer: I think queue the restaring requests is necessary, if user want to restart 2 tikv and there were 3 tikv at all, it is resonable to make sure that pd/tikv won't be unavaiable during restarting.

pkg/manager/member/pd_member_manager.go
```diff
@@ -232,6 +239,12 @@ func (pmm *pdMemberManager) syncPDStatefulSetForTidbCluster(tc *v1alpha1.TidbClu
 		}
 	}
 
+	if pmm.webhookEnabled {
+		if err = pmm.pdRestarter.Sync(tc, v1alpha1.PDMemberType); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
Yisaer: `webhook` would ensure each deleting request would be safe and I tend to make `Restart Controller` as `alpha` as `webook`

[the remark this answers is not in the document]
aylei: This couples `webhook` and `restart controller`, anyway, `restart controller` will never do things worse than `kubectl delete` and `webhook` just make sure the deleting procedure is graceful

pkg/manager/member/restarter.go
aylei: Of course this make sense in the context of automation, but restart is an user action and they do this on theire own risk. e.g. `kubectl delete` more than one pod also lay down the service

Yisaer: Make sense. Without considering this, restarter would be a simple controller which only delete the pod with annotations.

Issue #1317, opened 2019-12-10
make patch-e2e.sh edit cabundle depends on kubernetes version


Yisaer:
In kubernetes under version 1.13, the webhook config cabundle can't be null.

Current patch-e2e.sh would always edit cabundle, this should be null when in e2e test with version >= 1.13

Yisaer:
fixed in #1345

Issue #1316, opened 2019-12-10
use a single path for all webhooks:


Yisaer:
Currently we use 2 different path for each admission-hooks.
We should use a single path for all webhooks as `/apis/admission.tidb.pingcap.com/v1alpha1/admissionreviews`

Yisaer:
fix in #1345

Issue #1315, opened 2019-12-10
use kubecli instead of lister in webhook


Yisaer:
In webhook case, we need to get the latest info in apiserver instead of local cache. wo should use kubecli instead of lister in webhook.

Yisaer:
# fix in #1345

PR #1345, opened 2019-12-16
revise apiservice webhook

Yisaer:
https://github.com/pingcap/tidb-operator/issues/1317
https://github.com/pingcap/tidb-operator/issues/1316
https://github.com/pingcap/tidb-operator/issues/1315

This request would fix the problems left in https://github.com/pingcap/tidb-operator/pull/1288

1.  Revising `manifests/patch-e2e.sh`. As cabundle can only be accepted as null by kube-apiserver with version >= 1.13.0, now `patch-e2e.sh` would only patch cabundle if kube-apiserver with version < 1.13.0 in e2e test

2. Using one single api path for all webhook admission path.

3. Removing all the `Lister` and using kubecli instead.


manifests/patch-e2e.sh
```diff
@@ -38,20 +32,32 @@ namespace=${namespace:-tidb-admin}
...
+esac
 
+if [ "$bundle" = true ] ; then
+
+CA_BUNDLE=$(kubectl get configmap -n kube-system extension-apiserver-authentication -o=jsonpath='{.data.client-ca-file}' | base64 | tr -d '\n')
 
 case ${os} in
   linux)
     sed -i "s/caBundle: .*$/caBundle: ${CA_BUNDLE}/g" $CURDIR/webhook.yaml
-    sed -i "s/namespace: \${NAMESPACE}/namespace: ${namespace}/g" $CURDIR/webhook.yaml
```
weekface: why move this line to 37?

weekface:
In webhook, there are many actions are deleting something, for example: `pdDeleter`, `tikvDeleter`, so for safe, we should not use any `xxxLister` cache in the webhook.

manifests/patch-e2e.sh
Yisaer: There are 2 `sed` actions in this script. The first `sed` would change the namespace for `webhook.yaml`. And the second `sed` would change the `cabundle: null` into `cabundle: <cabundle token>` depends on kube-apiserver version.

Yisaer:
Update, now all the lister is removed.

cmd/admission-webhook/main.go
```diff
@@ -51,10 +56,32 @@ func main() {
...
 	}
-	stsAh := &webhook.StatefulSetAdmissionHook{}
-	cmd.RunAdmissionServer(podAh, stsAh)
+	run(flagset, ah)
+}
+
+func run(flagset *flag.FlagSet, admissionHooks ...apiserver.AdmissionHook) {
+
+	if len(os.Getenv("GOMAXPROCS")) == 0 {
+		runtime.GOMAXPROCS(runtime.NumCPU())
```
aylei: why is this necessary?

pkg/webhook/admission_hooks.go
```diff
@@ -19,60 +19,68 @@ import (
...
+		return a.stsAC.AdmitStatefulSets(ar)
+	default:
+		return a.unknownAdmissionRequest(ar)
 	}
-	return a.podAC.AdmitPods(ar)
 }
 
 // any special initialization goes here
-func (a *PodAdmissionHook) Initialize(cfg *rest.Config, stopCh <-chan struct{}) error {
+func (a *AdmissionHook) Initialize(cfg *rest.Config, stopCh <-chan struct{}) error {
```
aylei: shall we add

```
if a.initialized {
  return nil
}
```
?

Yisaer: good advice, updated.

cmd/admission-webhook/main.go
```diff
@@ -15,29 +15,34 @@ package main
 
 import (
 	"flag"
+	"github.com/openshift/generic-admission-server/pkg/apiserver"
```
tennix: Could you configure your editor to use `goimports` to format code? This will ensure the `import` package order consistent.

cmd/admission-webhook/main.go
tennix: +1. If we need to configure GOMAXPROCS, we can just use uber's `automaxprocs` package.

Yisaer: This is introduced in the orgin framework by [cmd.RunAdmissionServer](https://github.com/openshift/generic-admission-server/blob/92cbaa8d51b1b9fa33a7eda76683b3482590e534/pkg/cmd/cmd.go#L28).
Because I want to import my extra flagset, I copied this function like [quack](https://github.com/pusher/quack/blob/5d7531613d6bf969b9018fde40a5cffb7d3efd7e/cmd/quack/main.go#L34).

cmd/admission-webhook/main.go
Yisaer: updated and add `goimport` into FileWatcher. Missing it was a mistake.

cmd/admission-webhook/main.go
tennix: I think Go runtime will use NumCPU if the GOMAXPROCS=0 automatically.

cmd/admission-webhook/main.go
```diff
@@ -51,10 +57,32 @@ func main() {
...
+	stopCh := genericapiserver.SetupSignalHandler()
+
+	cmd := server.NewCommandStartAdmissionServer(os.Stdout, os.Stderr, stopCh, admissionHooks...)
+
+	// Add admission hook flags
+	cmd.Flags().AddGoFlagSet(flagset)
+
+	// Flags for glog
+	cmd.PersistentFlags().AddGoFlagSet(flag.CommandLine)
+	// Fix glog printing "Error: logging before flag.Parse"
```
tennix: Does the klog still have this issue?

cmd/admission-webhook/main.go
Yisaer: updated, removing this logic.

cmd/admission-webhook/main.go
```diff
@@ -16,28 +16,34 @@ package main
...
 
 var (
+	flagset                  *flag.FlagSet
 	printVersion             bool
 	extraServiceAccounts     string
 	evictRegionLeaderTimeout time.Duration
 )
 
 func init() {
+	flagset = flag.NewFlagSet("tidb-admission-webhook", flag.ExitOnError)
```
tennix: Do we need to introduce the flagset? If so, I think we should use only one of flagset or the standard library flag.

cmd/admission-webhook/main.go
Yisaer: updated, this logic was useless after testing and removed it now.

cmd/admission-webhook/main.go
Yisaer: updated, this `flagset` is not necessary and removed now.

manifests/patch-e2e.sh
```diff
@@ -38,20 +32,32 @@ namespace=${namespace:-tidb-admin}
...
+    sed -i "" "s/namespace: \${NAMESPACE}/namespace: ${namespace}/g" $CURDIR/webhook.yaml
+    ;;
+    *)
+    echo "invalid os ${os}, only support Linux and Darwin" >&2
+    ;;
+esac
 
+if [ "$bundle" = true ] ; then
+
+CA_BUNDLE=$(kubectl get configmap -n kube-system extension-apiserver-authentication -o=jsonpath='{.data.client-ca-file}' | base64 | tr -d '\n')
```
tennix: Indent?

manifests/patch-e2e.sh
```diff
@@ -38,20 +32,32 @@ namespace=${namespace:-tidb-admin}
 CURDIR=$(cd $(dirname ${BASH_SOURCE[0]}); pwd )
 os=`uname -s| tr '[A-Z]' '[a-z]'`
 
-CA_BUNDLE=$(kubectl get configmap -n kube-system extension-apiserver-authentication -o=jsonpath='{.data.client-ca-file}' | base64 | tr -d '\n')
-echo $CA_BUNDLE
+case ${os} in
+  linux)
+    sed -i "s/namespace: \${NAMESPACE}/namespace: ${namespace}/g" $CURDIR/webhook.yaml
```
tennix: Will this modify the file directly? This file is checked into the git. So I think it's better to do the modification on the fly by using pipe.

tennix: Or can we do the patch after apply to k8s?

Yisaer: Currently, this would only modify the file in e2e test. I think this could be marked as an issue to find a more elegant way in a future request.

manifests/patch-e2e.sh
Yisaer: updated

manifests/patch-e2e.sh
Yisaer: issed in https://github.com/pingcap/tidb-operator/issues/1358

PR #1349, opened 2019-12-17
terraform: fix docker ulimit configuring for latest EKS AMI

### LINKED ISSUE #1302: Considering adjust `containerd` systemd service ulimit on EKS
## Question

Docker on EKS worker node may have separated `dockerd` and `containerd` systemd service, adjusting the Docker service to increase ulimit of docker systemd service is not enough (https://github.com/pingcap/tidb-operator/blob/master/deploy/modules/aws/tidb-cluster/templates/userdata.sh.tpl#L14-L19). Some users find the `containerd` service also needs to set ulimit (https://tidbcommunity.slack.com/archives/CHD0HA3LZ/p1575780218010000?thread_ts=1575775693.006100&cid=CHD0HA3LZ).

### LINKED ISSUE #1344: Terraform script broken on latest AWS AMI
Ref: #1302 

Not only `v1.14`, EKS `v1.12` is also influenced.

Also, we may considering lock the AMI in terraform script to protect us from EKS breaking changes.

aylei:

close #1344 
close #1302 

The latest AMI of EKS has a break change that it runs containerd as a separate systemd service, for which we have not adjusted the ulimit. Configuring `ulimit` for containers is a more general method, ref: https://docs.docker.com/engine/reference/commandline/dockerd/

Tested on AWS with the latest EKS AMI.

 ```release-note
Terraform: fix docker ulimit configuring for latest EKS AMI.
 ```


deploy/modules/aws/tidb-cluster/templates/userdata.sh.tpl
```diff
@@ -18,6 +18,8 @@ sed -i 's/LimitNPROC=infinity/LimitNPROC=1048576/' /etc/systemd/system/docker.se
 systemctl daemon-reload
 systemctl restart docker
 
+echo 'OPTIONS="--default-ulimit nofile=1024000:1024000"' >> /etc/sysconfig/docker
```
tennix: Will this be executed again after the node is rebooted?

aylei: No, why is this needed?

cofyc: I guess it provides CLI options for dockerd. Shouldn't it be put before `systemctl restart docker`?

aylei: Good catch, it is mistake when I split this change from my local branch

cofyc: btw, add some comments about why this line is needed will be better

cofyc: Is it possible to configure all our ulimits via `--default-ulimit` and remove these lines?

https://github.com/pingcap/tidb-operator/blob/b7f3ef5ca7e669044951ff835f5304233148e488/deploy/modules/aws/tidb-cluster/templates/userdata.sh.tpl#L15-L17

aylei:
fix a mistake in command order @cofyc @tennix @weekface PTAL again

deploy/modules/aws/tidb-cluster/templates/userdata.sh.tpl
tennix: What I'm trying to say is that we'd better to make this idempotent.

aylei: @cofyc I think `nofile` is enough, I will verify locally with the modification of docker.service removed.

aylei: @tennix Nice advice, I will check if this line has been added first.

aylei: @cofyc It turns out that we should still configure the systemd service unit according to this part of `/etc/sysconfig/docker`:

```
# The max number of open files for the daemon itself, and all
# running containers.  The default value of 1048576 mirrors the value
# used by the systemd service unit.
DAEMON_MAXFILES=1048576
```

aylei:
Comments addressed @weekface @DanielZhangQD @cofyc PTAL again

PR #1354, opened 2019-12-17
Add apiserver admission webhook template

Yisaer:
Add template for admission webhook in `tidb-operator` chart template, as we are using `apiservice` to deploy our `webhook`. 

In this request, the cabundle for `ValidationWebhookConfiguration` is null and the `insecureSkipTLSVerify` for `apiservice` are true in default. If the `kube-apiserver`'s version >= 1.13.0, the user could enable webhook without setting any cert configuration as a quick trial.

I removed the previous `cert-patch` method and exposed the configuration in `values.yaml` for users to configure their own `key` and `cert` for `tidb-admission-webhook` service.

I also limit the validation request if the kube-apiserver is later than 1.15.0 which would let the webhook only affect the `tidbcluster` pod .

Code changes

 - Has Helm charts change

## ref
https://github.com/pingcap/tidb-operator/issues/1212


Yisaer:
@aylei  I would add e2e test after this request and https://github.com/pingcap/tidb-operator/pull/1345 are both merged.

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
```diff
@@ -0,0 +1,140 @@
...
+      - apps.pingcap.com
+    resources:
+      - statefulsets
+    verbs:
+      - '*'
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: tidb-admission-webhook-crb
```
tennix: Better to use the same naming convention for existing rbac resources.

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
```diff
@@ -0,0 +1,140 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: tidb-admission-webhook-cr
+---
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: tidb:system:api-server-auth
```
tennix: ditto

charts/tidb-operator/values.yaml
```diff
@@ -177,20 +177,30 @@ advancedStatefulset:
...
-  autoGenerator:
-    image:
-      repository: jettech/kube-webhook-certgen
-      tag: v1.2.0
-  ## `selfProvider` need user to provide their own ca/cert/key for webhook
-  selfProvider:
-    ca: ""
-    key: ""
+  create: false
+  replicas: 1
```
tennix: Can we deploy multiple webhook pods?

charts/tidb-operator/templates/admission/admission-webhook-service.yaml
```diff
@@ -0,0 +1,20 @@
...
+  labels:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: admission-webhook
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  ports:
+    - name: https-webhook # optional
+      port: 443
```
tennix: Do we need to specify the targetPort?

charts/tidb-operator/values.yaml
Yisaer: Yes.

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
Yisaer: updated.

charts/tidb-operator/templates/admission/admission-webhook-service.yaml
Yisaer: updated, targetPort was added now.

charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
```diff
@@ -0,0 +1,96 @@
...
+metadata:
+  name: validation-delete-tidb-admission-webhook-cfg
+  labels:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: admission-webhook
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+webhooks:
+  - name: delete.podadmission.pingcap.net
```
tennix: Why use `pingcap.net`?

tennix:
Before adding e2e test, have you tested this manually?

charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
Yisaer: A mistake, updated as `delete.podadmission.tidb.pingcap.com`

Yisaer:
@tennix Yes, I test this on `kind` with `1.15` version and `1.12` version. Both of the results performed right.

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
```diff
@@ -0,0 +1,140 @@
...
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["pingcap.com"]
+    resources:  ["tidbclusters"]
+    verbs: ["*"]
+  - apiGroups:
+      - apps.pingcap.com
+    resources:
+      - statefulsets
+    verbs:
+      - '*'
```
aylei: why not keep the style consistent?

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
```diff
@@ -0,0 +1,140 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: admission-webhook
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+rules:
+  - apiGroups:
+      - "admission.tidb.pingcap.com"
+    resources:
+      - admissionreviews
+    verbs:
+      - create
```
aylei: ditto

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
```diff
@@ -0,0 +1,140 @@
+{{- if eq .Values.admissionWebhook.create true }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.admissionWebhook.serviceAccount }}
```
aylei: Better to have an `admissionWebhook.rbac.create` variable, otherwise the `serviceAccount` variable does not make sense (user cannot override the serviceAccount, after all), refer to `controller-manager.rbac` for detail.

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
Yisaer: updated.

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
Yisaer: updated.

Issue #1366, opened 2019-12-19
Optimize discovery advertisePeerUrl length limit 


mikechengwei:
## Bug Report

**What version of Kubernetes are you using?**
1.14.8

**What version of TiDB Operator are you using?**
1.0.5

**What did you do?**
Hi , in my scene, I modify pd advertisePeerUrl from `"${POD_NAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc"` to `"${POD_NAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc.datacloud.xxx.com"`,
It's helpful to me that k8s internal dns can sync company dns server.
Then I meet a error 
```
E1219 03:39:39.742899       1 mux.go:58] failed to discover: ssd-tidb-cluster-pd-2.ssd-tidb-cluster-pd-peer.default.svc.datacloud.17usoft.com:2380
, advertisePeerUrl format is wrong: ssd-tidb-cluster-pd-2.ssd-tidb-cluster-pd-peer.default.svc.datacloud.17usoft.com:2380
E1219 03:39:39.742916       1 mux.go:60] failed to writeError: advertisePeerUrl format is wrong: ssd-tidb-cluster-pd-2.ssd-tidb-cluster-pd-peer.default.svc.datacloud.17usoft.com:2380
```

The reason of this error is discovery nodes limit advertisePeerUrl length, I think we can optimize like the following:

```
if len(strArr) < 4 {
	return "", fmt.Errorf("advertisePeerUrl format is wrong: %s", advertisePeerUrl)
}
```

Source code is :
https://github.com/pingcap/tidb-operator/blob/35db292f812459722d39ee3b1ddc1b7a1994795a/pkg/discovery/discovery.go#L70

**What did you expect to see?**
optimize the discovery  mechanism advertisePeerUrl length  limit .

Yisaer:
@mikechengwei Hi, did you change the `pd advertisePeerUrl` by modifying the `chart/tidb-cluster/templates/scripts/_start.pd.sh.tpl` or how ?

mikechengwei:
yes,i modify `pd advertisePeerUrl`  config  in  chart/tidb-cluster/templates/scripts/_start.pd.sh.tpl and `--advertise-addr ` config in _start_tikv.sh.tpl file .

PR #1368, opened 2019-12-19
optimize discovery mechanism pd advertise peer url limit

mikechengwei:
#1366 

Optimize discovery mechanism advertisePeerUrl length limit

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

Side effects

None

Related changes

None

None


aylei:
The format of `advertise-peer-url` is determined by `_start_pd.sh.tpl`, so I think this PR would make more sense if the different form of advertised-peer-urls could be configured via public interface. Because `_start_pd_.sh.tpl` is supposed to be an implementation detail of tidb-operator and modification on this file is unexpected.

aylei:
BTW, the deprecating of the `tidb-cluster` helm chart is working in progress, it is better to apply the change to https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/template.go

mikechengwei:
@aylei different form of advertised-peer-urls is a good idea. But now , is  advertised-peer-url prefix format stable, like `${POD_NAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc`. Deploy a tikv cluster and external world can  connect to it, it is important advertised-peer-url can custom. Can we  add a common suffix? 
And discovery mechanism pd advertise peer limit url `.` character  must equal to 4 , can we optimize it to less than.

Issue #1353, opened 2019-12-17
scheduler:  instead of return an error, return empty nodes when no node can be scheduled to


weekface:
We should improve error messages of tidb-scheduler. Ref: #1346

Issue #1346, opened 2019-12-16
Improve error messages for "Failed filter with extender"


kolbe:
## Feature Request

**Is your feature request related to a problem? Please describe:**

My deployment fails with this message:
```
Failed filter with extender at URL http://127.0.0.1:10262/scheduler/filter, code 500
```

But neither the logs for kube-scheduler or tidb-scheduler seem to contain any additional information that actually helps me resolve the issue.

**Describe the feature you'd like:**

This error message should be improved to provide information that can be used to resolve the problem.

[kube-scheduler.log](https://github.com/pingcap/tidb-operator/files/3970421/kube-scheduler.log)
[tidb-scheduler.log](https://github.com/pingcap/tidb-operator/files/3970422/tidb-scheduler.log)

cofyc:
Closed via https://github.com/pingcap/tidb-operator/pull/1373. If #1373 does not address this issue, please re-open it.

PR #1373, opened 2019-12-19
scheduler: refine scheduler error message

weekface:

This PR fixes: #1353 , refines the `tidb-scheduler` error messages.

- return empty nodes when no node can be scheduled to, not an error (https://github.com/pingcap/tidb-operator/issues/1346)
- add more error message to K8s `Events`
- refine the error messages

Before this PR:

image

Now:

image
image

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
scheduler: refine scheduler error message
 ```


pkg/scheduler/predicates/ha.go
```diff
@@ -47,15 +46,13 @@ type ha struct {
...
-	recorder      record.EventRecorder
 }
 
 // NewHA returns a Predicate
-func NewHA(kubeCli kubernetes.Interface, cli versioned.Interface, recorder record.EventRecorder) Predicate {
+func NewHA(kubeCli kubernetes.Interface, cli versioned.Interface) Predicate {
 	h := &ha{
-		kubeCli:  kubeCli,
-		cli:      cli,
-		recorder: recorder,
```
weekface: will emit `Event` in `scheduler` package, not here

pkg/scheduler/predicates/ha.go
```diff
@@ -144,53 +141,54 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
...
 	min := -1
 	minNodeNames := make([]string, 0)
-	for nodeName, podNames := range nodeMap {
-		podsCount := len(podNames)
-		maxPodsPerNode := 0
-
-		if component == label.PDLabelVal {
+	maxPodsPerNode := 0
+
+	if component == label.PDLabelVal {
```
weekface: this `if ... else ...` block calculate the `maxPodsPerNode` var, it is only related to `replicas` var. so i move these codes out of the `for ... range nodeMap` block.

and also we want to use the `maxPodsPerNode` in error message.

@cofyc

pkg/scheduler/predicates/stable_scheduling.go
```diff
@@ -102,11 +99,9 @@ func (p *stableScheduling) Filter(instanceName string, pod *apiv1.Pod, nodes []a
 				return []apiv1.Node{node}, nil
 			}
 		}
-		msg := fmt.Sprintf("cannot run on its previous node %q", nodeName)
-		p.recorder.Event(pod, apiv1.EventTypeWarning, UnableToRunOnPreviousNodeReason, msg)
-	} else {
-		glog.V(2).Infof("no previous node exists for pod %q in TiDB cluster %s/%q", podName, ns, tcName)
+		return nodes, fmt.Errorf("cannot run on its previous node %q", nodeName)
```
weekface: return an `error` here to let `scheduler` package emit an `Event`. @cofyc

pkg/scheduler/scheduler.go
```diff
@@ -115,7 +119,10 @@ func (s *scheduler) Filter(args *schedulerapiv1.ExtenderArgs) (*schedulerapiv1.E
 		glog.Infof("entering predicate: %s, nodes: %v", predicate.Name(), predicates.GetNodeNames(kubeNodes))
 		kubeNodes, err = predicate.Filter(instanceName, pod, kubeNodes)
 		if err != nil {
-			return nil, err
+			s.recorder.Event(pod, apiv1.EventTypeWarning, predicate.Name(), err.Error())
```
weekface: emit an `Event` if the `error` is not `nil`.

pkg/scheduler/predicates/ha.go
```diff
@@ -95,7 +92,7 @@ func (h *ha) Filter(instanceName string, pod *apiv1.Pod, nodes []apiv1.Node) ([]
 	}
 
 	if len(nodes) == 0 {
-		return nil, fmt.Errorf("kube nodes is empty")
+		return nil, fmt.Errorf("no nodes available to schedule pods")
```
```suggestion
		return nil, fmt.Errorf("no nodes available to schedule pods %s/%s"， ns, podName)
```

pkg/scheduler/scheduler.go
```diff
@@ -115,7 +119,10 @@ func (s *scheduler) Filter(args *schedulerapiv1.ExtenderArgs) (*schedulerapiv1.E
...
+			return &schedulerapiv1.ExtenderFilterResult{
```
onlymellb: When an error occurs, we need to determine whether `kubeNodes` is empty. If it is empty, it should returns an error normally. If it is not empty, it should continue to schedule.

pkg/scheduler/predicates/stable_scheduling.go
onlymellb:
```suggestion
		return nodes, fmt.Errorf("cannot run %s/%s on its previous node %q", ns, podName, nodeName)
```

pkg/scheduler/scheduler.go
weekface: fixed @onlymellb PTAL

pkg/scheduler/predicates/stable_scheduling.go
```diff
@@ -102,11 +99,9 @@ func (p *stableScheduling) Filter(instanceName string, pod *apiv1.Pod, nodes []a
 				return []apiv1.Node{node}, nil
 			}
 		}
-		msg := fmt.Sprintf("cannot run on its previous node %q", nodeName)
-		p.recorder.Event(pod, apiv1.EventTypeWarning, UnableToRunOnPreviousNodeReason, msg)
-	} else {
-		glog.V(2).Infof("no previous node exists for pod %q in TiDB cluster %s/%q", podName, ns, tcName)
+		return nodes, fmt.Errorf("cannot run %s/%s on its previous node %q", ns, podName, nodeName)
 	}
+	glog.V(2).Infof("no previous node exists for pod %q in TiDB cluster %s/%q", podName, ns, tcName)
```
tennix: From the log message, there is no node for scheduling, but the following returns the `nodes` and the error is `nil`.

cofyc: this happens when we can't find the previous node of this TiDB member. e.g. the tidb pod is new, so we can run it on any node

tennix: Yes, I got it. But we should use warning level log and also emit this event to users.

pkg/scheduler/scheduler.go
```diff
@@ -114,10 +118,14 @@ func (s *scheduler) Filter(args *schedulerapiv1.ExtenderArgs) (*schedulerapiv1.E
...
 		glog.Infof("entering predicate: %s, nodes: %v", predicate.Name(), predicates.GetNodeNames(kubeNodes))
 		kubeNodes, err = predicate.Filter(instanceName, pod, kubeNodes)
+		glog.Infof("leaving predicate: %s, nodes: %v", predicate.Name(), predicates.GetNodeNames(kubeNodes))
 		if err != nil {
-			return nil, err
+			s.recorder.Event(pod, apiv1.EventTypeWarning, predicate.Name(), err.Error())
+			if len(kubeNodes) == 0 {
+				// do not return error to k8s: https://github.com/pingcap/tidb-operator/issues/1353
+				return nil, nil
+			}
```
cofyc: Does `nil` semantically equal to filter result with empty nodes? If it is, these three lines seem unnecessary because, at the end of this function, it will return filter results with an empty node list.

pkg/scheduler/predicates/stable_scheduling.go
cofyc: After the change in this PR, the error returned here will be propagated to users.  @weekface can you help change the log level to warn and return the log message as an error too?

pkg/scheduler/scheduler.go
weekface: > Does nil semantically equal to filter result with empty nodes? 

Yes.

> If it is, these three lines seem unnecessary because, at the end of this function, it will return filter results with an empty node list.

This `return` is in a `for range` loop. No need to step the next predicate when the `kubeNodes` is empty.

pkg/scheduler/predicates/stable_scheduling.go
weekface: comments addressed, @tennix @cofyc PTAL

pkg/scheduler/scheduler.go
weekface: What about to change `return nil, nil` to break?

```
if len(kubeNodes) == 0 {
	break
}
```

cofyc: I think it's good

weekface: fixed PTAL

PR #1377, opened 2019-12-19
Using admission server in e2e test instead of applying `webhook.yaml` 

Yisaer:

1. Using the admission webhook server in e2e instead of applying `manifests/webhook.yaml`.

2. fix some admission webhook template bugs and add update statefulset webhook configuration in admission webhook template.

3. Upgrading Operator from `disable webhook` to `enable webhook` wouldn't restart `tidb-controller-manager` until we add webhook switch in it.

4. moving statefulset webhook configuration into `admission webhook template` and add a switch for each hook.

### What is changed?

- go code
- chart template
- e2e test


tests/webhook.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	klog.Info(cmd)
+	res, err := exec.Command("/bin/sh", "-c", cmd).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to deploy operator: %v, %s", err, string(res))
+	}
+	klog.Infof("success to execute helm operator upgrade")
+
+	//// ensure pods unchanged when upgrading operator
```
Yisaer: Currently, Upgrade Operator from `disable webhook` to `enable webhook` would only create a new `deployment` admission webhook. After we add `webhook-enabled` switch in `tidb-controller-manager`, it would be necessary to check whether upgrading Operator would cause `tidbcluster` restarted as `tidb-controller-manager` would restart.

tests/webhook.go
```diff
@@ -0,0 +1,136 @@
...
+		"admissionWebhook.hooksEnabled.pds", isPodWebhookEnable,
+		"admissionWebhook.hooksEnabled.statefulSets", isStsWebhookEnabled)
+
+	setString := info.OperatorHelmSetString(nil)
+	setString = strings.Join([]string{setString, switchWebhook}, ",")
+
+	cmd := fmt.Sprintf(`helm upgrade %s %s --set-string %s `,
+		info.ReleaseName,
+		oa.operatorChartPath(info.Tag),
+		setString)
```
cofyc: add webhook settings in OperatorCfg and use `UpgradeOperator`?

charts/tidb-operator/values.yaml
```diff
@@ -182,13 +182,22 @@ admissionWebhook:
   serviceAccount: tidb-admission-webhook
   rbac:
     create: true
+  hooksEnabled:
+    ## statefulsets hook would check requests for updating tidbcluster's statefulsets
+    ## If enabled it, the statefulsets of tidbcluseter would update in partition by tidbcluster's annotation
+    statefulSets: false
+    ## pods hook would check requests for creating and deleting tidbcluster's pods
+    ## if enabled it, the pods of tidbcluster would safely created or deleted by webhook instead of controller
+    pods: false
```
cofyc: because we will always deploy webhook eventually because many features must be implemented in webhook (defaulting & validating, etc). Maybe we can have only one gate `hooksEnabled`?

cc @aylei

charts/tidb-operator/templates/admission/admission-webhook-deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if eq .Values.admissionWebhook.create true }}
+{{- if .Values.admissionWebhook.create }}
```
aylei: why we have to change the criteria

tests/webhook.go
Yisaer: > add webhook settings in OperatorCfg and use `UpgradeOperator`?

That's a bit of hard cause we need to set cabundle when kube-apiserver under 1.13 version which would be complex to embed into `UpgradeOperator`

charts/tidb-operator/values.yaml
aylei: I am ok with these fine-grained control, but I think the default value is better to be `true`. The UX I expect is that when I enable webhook by setting `admission.create=true`, I get all the admissions out-of-box, and could disable some of them based on my customized requirement.

tests/webhook.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			err = ioutil.WriteFile(cabundleFile.Name(), data, 0644)
+			if err != nil {
+				return err
+			}
+			dataByte, err := ioutil.ReadFile(cabundleFile.Name())
+			if err != nil {
+				return err
+			}
+			klog.Infof("%s", string(dataByte[:]))
```
aylei: The log should be more accurate and readable

tests/webhook.go
```diff
@@ -0,0 +1,136 @@
...
+			cmd = fmt.Sprintf("%s -f %s", cmd, cabundleFile.Name())
+		}
+	}
+
+	klog.Info(cmd)
```
aylei: ditto

tests/webhook.go
```diff
@@ -0,0 +1,136 @@
...
+	res, err := exec.Command("/bin/sh", "-c", cmd).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to deploy operator: %v, %s", err, string(res))
+	}
+	klog.Infof("success to execute helm operator upgrade")
```
aylei: add the identity info of the upgraded operator in log

tests/webhook.go
cofyc: is it possible to configure `caBundle` via `--set-string`? if it is, we can configure `caBundle` in OperatorCfg and update `OperatorConfig.OperatorHelmSetString` to use it if it exists, e.g.

- configure webhook configurations in OperatorConfig
- (if < 1.13) get caBundle from `kube-system.extension-apiserver-authentication`
- (if < 1.13) configure caBundle in OperatorConfig
- UpgradeOperator(OperatorConfig)

tests/webhook.go
aylei: why should the check be commented out?

tests/webhook.go
```diff
@@ -0,0 +1,136 @@
...
+	//
+	//	return false, nil
+	//}
+	//
+	//err = wait.Poll(oa.pollInterval, 5*time.Minute, waitFn)
+	//if err == wait.ErrWaitTimeout {
+	//	return nil
+	//}
+
+	klog.Infof("success to upgrade operator with webhook switch %v, pod hook %v, sts hook %v", isWebhookEnabled, isPodWebhookEnable, isStsWebhookEnabled)
```
aylei: ditto

tests/webhook.go
Yisaer: updated.

charts/tidb-operator/templates/admission/admission-webhook-deployment.yaml
Yisaer: Cause `{{- if eq .Values.admissionWebhook.create true }}` didn't work well when I configure the setting by `helm upgrade <name> <path> --set key=value` which would casue and comparing type error.

tests/webhook.go
Yisaer: updated

charts/tidb-operator/values.yaml
Yisaer: updated, the `podsWebhook` is `true` in default while the `stsWebhook` is `false` as `stsWebhook` is a special feature.

tests/webhook.go
Yisaer: I set cabundle by `--set-string` at first and encounter a multi-line problem as the cabundle content is multi-line. I used `-f` to avoid this problem.

tests/webhook.go
Yisaer: updated.

tests/webhook.go
Yisaer: For now, upgrading Operator from `disable admission` to `enable admission` won't restart the `tidb-controller-manager` so it is not necessary to check the `tidbcluster`. 

After we add admission-webhook switch in `tidb-controller-manager`, the checking would be necessary.

tests/webhook.go
Yisaer: This log is useless and removed yet.

tests/webhook.go
Yisaer: This log helps to print the cmd we would execute, so it didn't need any extra info except itself.

tests/webhook.go
cofyc: you can fix the multi-line problem by escaping the value of the `--set-string` flag.

cofyc: The current code is fragile because it does not escape the shell arguments. another way is pass command and its arguments to `exec.Command` (don't use `sh`, then no need to escape shell arguments).

Yisaer: updated, for now, the switch of each hook and admission server and the cabundle if managed by `OperatorConfig`.

charts/tidb-operator/templates/admission/admission-webhook-deployment.yaml
aylei: This is should be our helm BP

Issue #1337, opened 2019-12-15
add feature of specifying  clusterIP when  creating pd service used ClusterIp type.


mikechengwei:
## Feature Request

**Is your feature request related to a problem? Please describe:**
I directly  use tikv cluster directly exclude tidb nodes,i provide a lvs ip  for my users,the lvs ip is pd service ip.
 When the cluster can't recover forever deep on helm tool, i need to execute helm reinstall command.
So I must guarantee  cluster pd service ip immutable.

**Describe the feature you'd like:**
add feature of specifying  clusterIP,when pd service use ClusterIp type.
```
services:
  - name: pd
    type: ClusterIP
    ClusterIP: 10.xx.xx.xx
```

**Describe alternatives you've considered:**
1. pd manager create service ,supporting specifying clusterIP.
2. helm charts add comment option.

aylei:
Thanks for your reporting! Would you like to take this?

BTW, new field for `ServiceSpec` is better to be added here:

https://github.com/pingcap/tidb-operator/blob/8425f87b1b78ee7d45fe7d8b3422cbab24af79d3/pkg/apis/pingcap/v1alpha1/types.go#L342

mikechengwei:
Ok,i will do it.

mikechengwei:
@aylei hi,i have seen pd create service using below class : 
https://github.com/pingcap/tidb-operator/blob/8425f87b1b78ee7d45fe7d8b3422cbab24af79d3/pkg/apis/pingcap/v1alpha1/types.go#L123
It was deprecated . Is the reason why this be  deprecated because we change to use headlessService?But now we expect that the tidb operator is also a tikv operator.
Should i continue to modify based on `Services`,or changing to use  `ServiceSpec ` class?

tennix:
This slice only contains service type and it's global. We are going to replace it with component-specific struct which contains more field.

I'm curious about in which scenario does this happen.
> When the cluster can't recover forever deep on helm tool, i need to execute helm reinstall command.

Also how do you make tikv accessible to the external world if the tikv client is not deployed inside the k8s cluster.

mikechengwei:
[...]

PR #1382, opened 2019-12-20
replace corev1.Services type with  ServiceType and rm useless code

mikechengwei:
#1337 

modify all reference `corev1.Services` method and add clusterIP type.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Helm charts change
 - Has Go code change

Side effects
```
services:		
  - name: pd		
    type: ClusterIP
```
I think the `name` of `service` has no  effect. So we change `Services ` to `ServiceType `, we only need to guarantee the type of `services ` equals to pd's `service` config. So  set `ClusterIP` as  default type ,it will not affect operator upgrade.
 

Related changes

None

None


charts/tidb-cluster/values.yaml
```diff
@@ -31,12 +31,6 @@ pvReclaimPolicy: Retain
 # When enabling, scale in pd or tikv will trigger PV reclaim automatically
 enablePVReclaim: false
 
-# services is the service list to expose, default is ClusterIP
-# can be ClusterIP | NodePort | LoadBalancer
-services:
-  - name: pd
-    type: ClusterIP
-
```
aylei: We should keep the old fields to keep the backward compatiability

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -363,14 +362,6 @@ type TiDBServiceSpec struct {
...
 }
 
-// +k8s:openapi-gen=true
-// Deprecated
-// Service represent service type used in TidbCluster
-type Service struct {
-	Name string `json:"name,omitempty"`
-	Type string `json:"type,omitempty"`
-}
-
```
aylei: ditto

charts/tidb-cluster/values.yaml
mikechengwei: I think  the type config of `Services`  is working,the name is not working  because the name use hard code.
If we guarantee the type not changing, it will not trigger compatibility issues .

aylei: That is, if the an user change the `ClusterIP` here to `NodePort`, and we remove these fields in the new version of helm chart, then the user's release would be broken on upgrading.

aylei: General guideline is we could never delete public fields

mikechengwei: Okay,i know.

mikechengwei:
@aylei I have optimize code , and you can  review  it again.

charts/tidb-cluster/values.yaml
```diff
@@ -73,6 +73,12 @@ pd:
     [replication]
     location-labels = ["region", "zone", "rack", "host"]
 
+  # pd Service
+  # we can only specify clusterIP and loadBalancerIP now
+  service:
+    clusterIP: "None"
```
aylei: Looks like that this variable is not rendered in the template, actually, I recommend to leave the helm chart as-is because we are migrating from helm to CRD #1121

pkg/manager/member/pd_member_manager.go
```diff
@@ -463,6 +463,12 @@ func (pmm *pdMemberManager) getNewPDServiceForTidbCluster(tc *v1alpha1.TidbClust
 			Selector: pdLabel,
 		},
 	}
+	svcSpec := tc.Spec.PD.Service
+	if svcSpec != nil {
+		pdService.Spec.ClusterIP = svcSpec.ClusterIP
+		pdService.Spec.LoadBalancerIP = svcSpec.LoadBalancerIP
+	}
```
aylei: `.spec.pd.service.type` and `.spec.pd.service.annotations` should also be respected

mikechengwei: done

pkg/manager/member/pd_member_manager.go
```diff
@@ -463,6 +463,15 @@ func (pmm *pdMemberManager) getNewPDServiceForTidbCluster(tc *v1alpha1.TidbClust
 			Selector: pdLabel,
 		},
 	}
+	// if set pd service type ,overwrite global variable services
+	svcSpec := tc.Spec.PD.Service
+	if svcSpec != nil {
+		pdService.Spec.ClusterIP = svcSpec.ClusterIP
+		pdService.Spec.Type = svcSpec.Type
+		pdService.Spec.LoadBalancerIP = svcSpec.LoadBalancerIP
+		pdService.ObjectMeta.Annotations = svcSpec.Annotations
```
aylei: what if the fields in svcSpec are empty strings? We should check it first

mikechengwei: you re right, the type need to check, other values I think empty is not affect.

aylei:
@weekface @cofyc @Yisaer @DanielZhangQD PTAL

PR #1386, opened 2019-12-20
add ephemeral-storage resources and make QOS guaranteed

cofyc:

CI pod is evicted because of low ephemeral-storage on the node.

```
status:
  message: 'The node was low on resource: ephemeral-storage. Container main was using
    12012Ki, which exceeds its request of 0. Container jnlp was using 96Ki, which
    exceeds its request of 0. '
  phase: Failed
  reason: Evicted
  startTime: "2019-12-20T05:19:51Z"
```

- add ephemeral-storage resources
- make QOS guaranteed

### Check List 

Tests 

 - Unit test
 - E2E test


ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -26,6 +26,11 @@ spec:
       requests:
         memory: "8000Mi"
         cpu: 8000m
+        ephemeral-storage: "40Gi"
```
aylei: would it be too large to schedule?

PR #1389, opened 2019-12-20
Add webhook switch for tidb-controller-manager

Yisaer:
Add webhook switch for tidb-controller-manager. 
Currently, if tidb-operator's admission pod webhook is enabled, the `upgrading` and `scaling-in` for tikv and pd would be managed by webhook instead of controller.

What changed:

1. Add webhook switch and logic for tidb-controller-manager.
2. Add e2e-test for upgrading
3. Add a pre-delete-job to delete `delete-pod` webhook before `helm del --purge <operator> ` in order to avoid pending during `helm del` when pod admission enabled


ci/pingcap_tidb_operator_build_kind.groovy
```diff
@@ -180,14 +180,17 @@ def call(BUILD_BRANCH, CREDENTIALS_ID, CODECOV_CREDENTIALS_ID) {
...
 		builds["E2E v1.16.3"] = {
-			build("IMAGE_TAG=${GITHASH} SKIP_BUILD=y GINKGO_NODES=8 KUBE_VERSION=v1.16.3 DOCKER_IO_MIRROR=https://dockerhub.azk8s.cn REPORT_DIR=\$(pwd)/artifacts REPORT_PREFIX=v1.16.3_ ./hack/e2e.sh -- --ginkgo.skip='\\[Serial\\]'", artifacts)
+			build("IMAGE_TAG=${GITHASH} SKIP_BUILD=y GINKGO_NODES=8 KUBE_VERSION=v1.16.3 DOCKER_IO_MIRROR=https://dockerhub.azk8s.cn REPORT_DIR=\$(pwd)/artifacts REPORT_PREFIX=v1.16.3_ ./hack/e2e.sh -- --ginkgo.focus='\\[TiDBCluster\\]'", artifacts)
 		}
 		builds["E2E v1.12.10 Serial"] = {
 			build("IMAGE_TAG=${GITHASH} SKIP_BUILD=y KUBE_VERSION=v1.12.10 DOCKER_IO_MIRROR=https://dockerhub.azk8s.cn REPORT_DIR=\$(pwd)/artifacts REPORT_PREFIX=v1.12.10_serial_ ./hack/e2e.sh -- --ginkgo.focus='\\[Serial\\]' --install-operator=false", artifacts)
 		}
+		builds["E2E v1.12.10 PodAdmission"] = {
+        		build("IMAGE_TAG=${GITHASH} SKIP_BUILD=y KUBE_VERSION=v1.12.10 DOCKER_IO_MIRROR=https://dockerhub.azk8s.cn REPORT_DIR=\$(pwd)/artifacts REPORT_PREFIX=v1.12.10_serial_ ./hack/e2e.sh -- --ginkgo.focus='\\[PodAdmission\\]' --install-operator=false", artifacts)
+        	}
```
cofyc: no need to add a new test group, put it in serial group with tag `[Serial]` in test name
when tests in serial group take a long time to finish, we can split the serial group into multiple jobs

Yisaer: updated.

tests/webhook.go
```diff
@@ -46,3 +53,166 @@ func (oa *operatorActions) setCabundleFromApiServer(info *OperatorConfig) error
...
+		return err
+	}
+
+	tc, err := oa.cli.PingcapV1alpha1().TidbClusters(ns).Get(tcName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get tidbcluster: %s/%s, %v", ns, tcName, err)
+	}
+
+	f := func(stsName, namespace string, desiredReplicas int64) (*apps.StatefulSet, bool, error) {
+		sts, err := oa.kubeCli.AppsV1().StatefulSets(ns).Get(stsName, metav1.GetOptions{})
```
cofyc: use `oa.tcStsGetter`

tests/webhook.go
```diff
@@ -46,3 +53,166 @@ func (oa *operatorActions) setCabundleFromApiServer(info *OperatorConfig) error
...
+		return err
+	}
+	tikvDesiredReplicas, err := strconv.ParseInt(info.Resources["tikv.replicas"], 10, 32)
+	if err != nil {
+		return err
+	}
+	tidbDesiredReplicas, err := strconv.ParseInt(info.Resources["tidb.replicas"], 10, 32)
+	if err != nil {
+		return err
+	}
```
cofyc: use `tc.Spec.TiDB/TiKV/PD.Replicas` fields

tests/webhook.go
```diff
@@ -46,3 +53,166 @@ func (oa *operatorActions) setCabundleFromApiServer(info *OperatorConfig) error
...
+}
+
+// check target tikv store whether have desired replicas and were they all Up and no evictSchedulers
+func (oa *operatorActions) checkDesiredStoreHealthy(tc *v1alpha1.TidbCluster, info *TidbClusterConfig) error {
+	klog.Infof("start to check tc[%s/%s]'s upgraded tikv store info", tc.Namespace, tc.Name)
+	desiredTikvNumber, err := strconv.Atoi(info.Resources["tikv.replicas"])
+	if err != nil {
+		return err
+	}
```
cofyc: `tc.Spec.TiKV.Replicas` is desired replicas of tikv member

tests/webhook.go
```diff
@@ -46,3 +53,166 @@ func (oa *operatorActions) setCabundleFromApiServer(info *OperatorConfig) error
...
+			return false, nil
+		}
+		return true, nil
+	})
+}
+
+// check target pd member whether have desired pd members and had one pd leader
+func (oa *operatorActions) checkDesiredPDHealthy(tc *v1alpha1.TidbCluster, info *TidbClusterConfig) error {
+	klog.Infof("start to check tc[%s/%s]'s upgraded pd member info", tc.Namespace, tc.Name)
+	desiredPDNumbers, err := strconv.Atoi(info.Resources["pd.replicas"])
```
cofyc: ditto

tests/e2e/tidbcluster/serial.go
```diff
@@ -109,4 +112,60 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+			cluster.Resources["tidb.replicas"] = "2"
+
+			oa.DeployTidbClusterOrDie(&cluster)
+			oa.CheckTidbClusterStatusOrDie(&cluster)
+			upgradeVersions := cfg.GetUpgradeTidbVersionsOrDie()
+			ginkgo.By(fmt.Sprintf("Upgrading tidb cluster from %s to %s", cluster.ClusterVersion, upgradeVersions[0]))
+			cluster.UpgradeAll(upgradeVersions[0])
+			oa.UpgradeTidbClusterOrDie(&cluster)
+			oa.CheckUpgradeWithPodWebhookOrDie(&cluster)
+			oa.CheckTidbClusterStatusOrDie(&cluster)
```
cofyc: what's the difference between `CheckUpgradeWithPodWebhookOrDie` and `CheckTidbClusterStatusOrDie`? Is it possible to merge them?

tests/webhook.go
Yisaer: updated.

tests/e2e/tidbcluster/serial.go
Yisaer: It seems 'CheckUpgradeWithPodWebhookOrDie' do the part of things of `CheckTidbClusterStatusOrDie`,so it is removed now and `CheckTidbClusterStatusOrDie` is enough to check Upgrading.

pkg/manager/member/tikv_scaler.go
```diff
@@ -80,6 +80,11 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
 		return err
 	}
 
+	if controller.PodWebhookEnabled {
+		decreaseReplicas(newSet, oldSet)
```
cofyc: Can webhook control the process of one by one scaling in? If it can now, `decreaseReplicas` is not needed.

cofyc:
I guess the serial e2e job fails because we didn't wait for the tidb-operator resources (e.g. tidb-controller-manager deployment) is fully cleaned.
We can install tidb-operator into test namespace instead of shared namespace `pingcap`.

Yisaer:
@cofyc  The reason why e2e job fails is that when we execute `helm del --purge <operator>`, the tiller server delete the `admission-webhook` service first. When pod admission webhook is enabled and the `failure Policy` is Fail, since the service of webhook is deleted, the request for deleting pod was reject and the tiller server get pending. 
Now I add a pre-delete job hook to delete webhook-config when pod admission webhook is enabled.

pkg/manager/member/tikv_scaler.go
Yisaer: Currently, the webhook didn't control the process running step by step. We could transfer this scale-in logic in webhook in the future.

charts/tidb-operator/templates/admission/admission-webhook-rbac.yaml
```diff
@@ -39,6 +39,10 @@ rules:
   - apiGroups: ["apps.pingcap.com"]
     resources: ["statefulsets"]
     verbs: ["*"]
+  - apiGroups: ["admissionregistration.k8s.io"]
+    resources: ["validatingwebhookconfigurations"]
+    verbs: ["delete"]
```
cofyc: I'm not familiar with helm. It seems that we need extra permission which is not part of our admission webhook because of `pre-delete-job`, do we have an alternative solution for this helm problem?
cc @aylei @weekface @onlymellb

Yisaer: The pre-delete-job should have its own service-account which had `validatingwebhookconfigurations` permission and the admission webhook server won't have this permission.
To get rid of requesting `validatingwebhookconfigurations` permission, we should make helm delete `validatingwebhookconfigurations` first when we execute `helm del --purge`

cofyc: better we can clean by running `helm del --purge`. add a service-account for `pre-delete-job`?

Yisaer: updated

tests/e2e/tidbcluster/serial.go
```diff
@@ -109,4 +112,59 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+			cluster.Resources["tikv.replicas"] = "3"
+			cluster.Resources["tidb.replicas"] = "2"
+
+			oa.DeployTidbClusterOrDie(&cluster)
+			oa.CheckTidbClusterStatusOrDie(&cluster)
+			upgradeVersions := cfg.GetUpgradeTidbVersionsOrDie()
+			ginkgo.By(fmt.Sprintf("Upgrading tidb cluster from %s to %s", cluster.ClusterVersion, upgradeVersions[0]))
+			cluster.UpgradeAll(upgradeVersions[0])
+			oa.UpgradeTidbClusterOrDie(&cluster)
+			oa.CheckTidbClusterStatusOrDie(&cluster)
```
aylei: This method only checks the final state of tidbcluster is ok, we should take the graceful checking during upgrading into account.

charts/tidb-operator/templates/admission/pre-delete-job.yaml
```diff
@@ -0,0 +1,100 @@
...
+spec:
+  template:
+    metadata:
+      name: webhook-del
+    spec:
+      restartPolicy: Never
+      serviceAccountName: {{ .Release.Name }}
+      containers:
+        - name: "{{.Release.Name}}-job-del"
+          image: bitnami/kubectl:latest
```
aylei: This should be configurable, some environments cannot access docker.io

tests/e2e/tidbcluster/serial.go
Yisaer: Currently, I haven't found a graceful way to check tidbcluster state during upgrading as the previous method used in basic e2e is to use webhook.

aylei: I admit that there is no clear way to do so, maybe we could mark it as TODO and track it in a separate issue

charts/tidb-operator/templates/admission/pre-delete-job.yaml
Yisaer: updated.

tests/e2e/tidbcluster/serial.go
Yisaer: updated and issue in https://github.com/pingcap/tidb-operator/issues/1407

charts/tidb-operator/templates/admission/pre-delete-job.yaml
Yisaer: updated

Issue #1262, opened 2019-12-02
New custom resource TidbInitializer


aylei:
## Feature Request

#1121 

Some variables are only applied on cluster initialization and are actually immutable hereafter. e.g. `initSql` and `tidbPasswordName`.

One possible solution is to expand these fields in `TidbCluster` and ensure immutability in validation webhook or AA.

Another solution is encode those variables to a Job-like custom resource, `TidbInitializer`.

We may prefer `TidbInitializer` because a separate CR is pluggable, a `TidbCluster` could have multiple `TidbInitializer`, and users are free to mutate the state configured by former initializer by simply apply an new initializer.

PR #1391, opened 2019-12-20
add tidbinitializer crd

aylei:

#1262 
 
 ```release-note
Add new CRD TidbInitializer.
 ```
@Yisaer @tennix @sdojjy @DanielZhangQD @LinuxGit PTAL


pkg/apis/pingcap/v1alpha1/tidbinitializer_types.go
```diff
@@ -0,0 +1,90 @@
...
+	// Spec defines the desired state of TidbInitializer
+	Spec TidbInitializerSpec `json:"spec"`
+
+	// +k8s:openapi-gen=false
+	// Most recently observed status of the TidbInitializer
+	Status TidbInitializerStatus `json:"status"`
+}
+
+// +k8s:openapi-gen=true
+// TidbInitializer spec encode the desired state of tidb initializer Job
```
DanielZhangQD: Also need o add `permitHost`, `initSqlConfigMapName`.

Yisaer: +1,  `initSqlConfigMapRef` is used widely in `initSql` case.

aylei: Is it still necessary if we change the type of `initSql` to string?

aylei: I will add it to make the functionality consistent with helm

aylei:
@DanielZhangQD @Yisaer PTAL

pkg/apis/pingcap/v1alpha1/tidbinitializer_types.go
```diff
@@ -53,21 +54,27 @@ type TidbInitializer struct {
 // +k8s:openapi-gen=true
 // TidbInitializer spec encode the desired state of tidb initializer Job
 type TidbInitializerSpec struct {
+	Image string `json:"image"`
```
DanielZhangQD: The `Image` and `ImagePullPolicy` are to cover below configs in `values.yaml`, right?
```
# mysqlClient is used to set password for TiDB
# it must has Python MySQL client installed
mysqlClient:
  image: tnir/mysqlclient
  imagePullPolicy: IfNotPresent
```
We also have to cover the initializer.resources config:
```
  initializer:
    resources: {}
    # limits:
    #  cpu: 100m
    #  memory: 100Mi
    # requests:
    #  cpu: 100m
    #  memory: 100Mi
```

aylei: Good catch!

charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -49,15 +49,7 @@ rules:
...
-  resources:
-  - tidbclusters
-  - tidbclusters/finalizers
-  - backups
-  - backups/finalizers
-  - backupschedules
-  - backupschedules/finalizers
-  - restores
-  - restores/finalizers
+  resources: ["*"]
```
aylei: Grant all resources of our own group should be safe

PR #1396, opened 2019-12-23
e2e case of operating tidb-cluster without helm

aylei:
Signed-off-by: Aylei <rayingecho@gmail.com>

close #1378 

### Check List 

Tests 

 - Unit test
 - E2E test

### Does this PR introduce a user-facing change?:
 
 ```release-note
e2e case of operating tidb-cluster without helm
 ```

A fix of TiKVConfig schema is also included, please check https://github.com/pingcap/tidb-operator/pull/1396/commits/97a075a361b6e1ae83c6ad1128c5e92275ab9b90 for non-codegen change

PR #1403, opened 2019-12-24
add controller for tidb initializer

DanielZhangQD:
fix https://github.com/pingcap/tidb-operator/issues/1262
Add controller for TidbInitializer CR
### Check List 

Tests 

Code changes

 - Has Go code change
 - Has documents change

Side effects

Related changes

 ```release-note
Support initializing TiDB Cluster with TidbInitializer Custom Resource.
 ```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/tidbcluster.go
+++ b/pkg/apis/pingcap/v1alpha1/tidbcluster.go
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	// defaultHelperImage is default image of helper
-	defaultHelperImage = "busybox:1.26.2"
+	// DefaultHelperImage is default image of helper
+	DefaultHelperImage = "busybox:1.31.1"
 )
 
 // ComponentAccessor is the interface to access component details, which respects the cluster-level properties
@@ -176,7 +176,7 @@ func (tc *TidbCluster) HelperImage() string {
 		image = tc.Spec.TiDB.SlowLogTailer.Image
 	}
 	if image == "" {
-		image = defaultHelperImage
+		image = DefaultHelperImage
 	}
 	return image
 }
--- a/pkg/apis/pingcap/v1alpha1/tidbcluster.go
+++ b/pkg/apis/pingcap/v1alpha1/tidbcluster.go
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	// DefaultHelperImage is default image of helper
-	DefaultHelperImage = "busybox:1.31.1"
+	// defaultHelperImage is default image of helper
+	defaultHelperImage = "busybox:1.26.2"
 )
 
 // ComponentAccessor is the interface to access component details, which respects the cluster-level properties
@@ -176,7 +176,7 @@ func (tc *TidbCluster) HelperImage() string {
 		image = tc.Spec.TiDB.SlowLogTailer.Image
 	}
 	if image == "" {
-		image = DefaultHelperImage
+		image = defaultHelperImage
 	}
 	return image
 }
--- a/pkg/apis/pingcap/v1alpha1/tidbinitializer_types.go
+++ b/pkg/apis/pingcap/v1alpha1/tidbinitializer_types.go
@@ -71,13 +71,17 @@ type TidbInitializerSpec struct {
 
 	// InitSqlConfigMapName reference a configmap that provide init-sql, take high precedence than initSql if set
 	// +optional
-	InitSqlConfigMap *corev1.LocalObjectReference `json:"initSqlConfigMap,omitempty"`
+	InitSqlConfigMap *string `json:"initSqlConfigMap,omitempty"`
 
 	// +optional
-	PasswordSecret *corev1.LocalObjectReference `json:"passwordSecret,omitempty"`
+	PasswordSecret *string `json:"passwordSecret,omitempty"`
 
 	// +optional
 	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Time zone of TiDB initializer Pods
+	// +optional
+	Timezone string `json:"timezone,omitempty"`
 }
 
 // +k8s:openapi-gen=true
--- a/pkg/controller/configmap_control.go
+++ b/pkg/controller/configmap_control.go
@@ -18,6 +18,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -28,7 +29,6 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ConfigMapControlInterface manages configmaps used by TiDB clusters
@@ -39,28 +39,36 @@ type ConfigMapControlInterface interface {
 	UpdateConfigMap(controller runtime.Object, cm *corev1.ConfigMap) (*corev1.ConfigMap, error)
 	// DeleteConfigMap delete the given ConfigMap owned by the controller object
 	DeleteConfigMap(controller runtime.Object, cm *corev1.ConfigMap) error
+	// ConfigMapExist check if the given ConfigMap exist
+	ConfigMapExist(ns, name string) (bool, error)
 }
 
 type realConfigMapControl struct {
-	client   client.Client
 	kubeCli  kubernetes.Interface
-	cmLister corelisters.ConfigMapLister
 	recorder record.EventRecorder
 }
 
 // NewRealSecretControl creates a new SecretControlInterface
 func NewRealConfigMapControl(
 	kubeCli kubernetes.Interface,
-	cmLister corelisters.ConfigMapLister,
 	recorder record.EventRecorder,
 ) ConfigMapControlInterface {
 	return &realConfigMapControl{
 		kubeCli:  kubeCli,
-		cmLister: cmLister,
 		recorder: recorder,
 	}
 }
 
+func (cc *realConfigMapControl) ConfigMapExist(ns, name string) (bool, error) {
+	if _, err := cc.kubeCli.CoreV1().ConfigMaps(ns).Get(name, metav1.GetOptions{}); err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return true, err
+	} else {
+		return true, nil
+	}
+}
 func (cc *realConfigMapControl) CreateConfigMap(owner runtime.Object, cm *corev1.ConfigMap) (*corev1.ConfigMap, error) {
 	created, err := cc.kubeCli.CoreV1().ConfigMaps(cm.Namespace).Create(cm)
 	cc.recordConfigMapEvent("create", owner, cm, err)
@@ -81,7 +89,7 @@ func (cc *realConfigMapControl) UpdateConfigMap(owner runtime.Object, cm *corev1
 			return nil
 		}
 
-		if updated, err := cc.cmLister.ConfigMaps(cm.Namespace).Get(cmName); err != nil {
+		if updated, err := cc.kubeCli.CoreV1().ConfigMaps(cm.Namespace).Get(cmName, metav1.GetOptions{}); err != nil {
 			utilruntime.HandleError(fmt.Errorf("error getting updated ConfigMap %s/%s from lister: %v", ns, cmName, err))
 		} else {
 			cm = updated.DeepCopy()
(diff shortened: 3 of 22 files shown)


pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -20,8 +20,8 @@ import (
 )
 
 const (
-	// defaultHelperImage is default image of helper
-	defaultHelperImage = "busybox:1.26.2"
+	// DefaultHelperImage is default image of helper
+	DefaultHelperImage = "busybox:1.31.1"
```
aylei: Is this necessary? I think change this image will rolling-update some clusters

DanielZhangQD:
Manual test passed, UT and e2e test will be done in https://github.com/pingcap/tidb-operator/issues/1404.

pkg/apis/pingcap/v1alpha1/tidbinitializer_types.go
```diff
@@ -71,13 +71,17 @@ type TidbInitializerSpec struct {
 
 	// InitSqlConfigMapName reference a configmap that provide init-sql, take high precedence than initSql if set
 	// +optional
-	InitSqlConfigMap *corev1.LocalObjectReference `json:"initSqlConfigMap,omitempty"`
```
aylei: It would be more extensible to use a struct, for example, we could support reference configmap cross namespaces by adding `namespace` to the object, that is, change `LocalObjectReference` to `ObjectReference`

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+	ok, err := tm.cmControl.ConfigMapExist(ns, name)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		err = tm.syncTiDBInitConfigMap(ti)
+		if err != nil {
+			return err
+		}
+	}
```
aylei: Can we use `CreateOrUpdateConfigMap`?

pkg/controller/controller_utils.go
```diff
@@ -124,6 +127,20 @@ func GetOwnerRef(tc *v1alpha1.TidbCluster) metav1.OwnerReference {
...
+	blockOwnerDeletion := true
+	return metav1.OwnerReference{
+		APIVersion:         InitControllerKind.GroupVersion().String(),
+		Kind:               InitControllerKind.Kind,
+		Name:               ti.GetName(),
+		UID:                ti.GetUID(),
+		Controller:         &controller,
+		BlockOwnerDeletion: &blockOwnerDeletion,
+	}
+}
```
aylei: It is recommended to use `GenericControl.CreateOrUpdateXXX()` method, which will set owner reference automatically and make the reconcile process be idempotent.

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+
+	return job, nil
+}
+
+func getTiDBInitConfigMap(ti *v1alpha1.TidbInitializer) (*corev1.ConfigMap, error) {
+	var permitHost string
+	var initSQL, passwdSet bool
+
+	if ti.Spec.PermitHost == nil || *ti.Spec.PermitHost == "" {
+		permitHost = "%"
```
aylei:
```suggestion
		permitHost = `%`
```

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	return job, nil
+}
+
+func getTiDBInitConfigMap(ti *v1alpha1.TidbInitializer) (*corev1.ConfigMap, error) {
+	var permitHost string
+	var initSQL, passwdSet bool
+
+	if ti.Spec.PermitHost == nil || *ti.Spec.PermitHost == "" {
```
aylei: for `*string`, empty string is explicitly set by user so it should be respected, if empty string is illegal, we should reject it in validation webhook

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+		permitHost = "%"
+	} else {
+		permitHost = *ti.Spec.PermitHost
+	}
```
aylei: How about adding a `GetPermitHost()` method to `TidbInitializer` struct so that the default value could be hidden

pkg/label/label.go
```diff
@@ -119,6 +124,14 @@ func New() Label {
 	}
 }
 
+// NewInitializer initialize a new Label for Jobs of TiDB initializer
+func NewInitializer() Label {
+	return Label{
+		NameLabelKey:      "initializer",
```
onlymellb: Use const `InitJobLabelVal` instead. By the way, change the following backup related too.

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+		jobControl,
+		cmControl,
+		cli,
+	}
+}
+
+func (tm *tidbInitManager) Sync(ti *v1alpha1.TidbInitializer) error {
+	if ti.DeletionTimestamp != nil {
+		return nil
+	}
```
onlymellb: This logic should be placed [here](https://github.com/pingcap/tidb-operator/pull/1403/files#diff-5297cabf3cf775db17d396cac519f87aR154), make it fail fast.

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+
+	name := controller.TiDBInitializerMemberName(ti.Spec.Clusters.Name)
+	ns := ti.Namespace
+	ok, err := tm.cmControl.ConfigMapExist(ns, name)
```
onlymellb: Use `exist` instead of `ok`?

pkg/controller/configmap_control.go
```diff
@@ -81,7 +89,7 @@ func (cc *realConfigMapControl) UpdateConfigMap(owner runtime.Object, cm *corev1
 			return nil
 		}
 
-		if updated, err := cc.cmLister.ConfigMaps(cm.Namespace).Get(cmName); err != nil {
+		if updated, err := cc.kubeCli.CoreV1().ConfigMaps(cm.Namespace).Get(cmName, metav1.GetOptions{}); err != nil {
```
onlymellb: I think it is not necessary to operate kube-apiserver directly, refer [here](url).

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	if !errors.IsNotFound(err) {
+		return fmt.Errorf("TiDBInitializer %s/%s get job %s failed, err: %v", ns, ti.Name, name, err)
+	}
+
+	return tm.syncTiDBInitJob(ti)
+}
+
+func (tm *tidbInitManager) checkStatusUpdate(ti *v1alpha1.TidbInitializer, job *batchv1.Job) error {
```
onlymellb: Renamed to `UpdateTiJobStatus`?

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,381 @@
...
+	if job == nil || ti == nil {
+		return nil
+	}
+	if apiequality.Semantic.DeepEqual(ti.Status.JobStatus, job.Status) {
+		return nil
+	}
+	job.Status.DeepCopyInto(&ti.Status.JobStatus)
+	_, err := tm.cli.PingcapV1alpha1().TidbInitializers(ti.Namespace).Update(ti)
```
onlymellb: Add `RetryOnConflict`?

pkg/apis/pingcap/v1alpha1/tidbcluster.go
DanielZhangQD: ok, I will revert this.

pkg/apis/pingcap/v1alpha1/tidbinitializer_types.go
DanielZhangQD: A Pod cannot mount a configmap from other ns, from the definition of `ConfigMapVolumeSource`, only `LocalObjectReference` is included.

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: I will add new methods to `GenericControlInterface`.

pkg/controller/controller_utils.go
DanielZhangQD: ok.

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: updated.

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: updated.

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: updated

pkg/apis/pingcap/v1alpha1/tidbcluster.go
cofyc: not revert yet?

pkg/label/label.go
DanielZhangQD: updated

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: updated

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: updated

cofyc:
do we have a basic TiDBInitializer e2e case right now? if not, I prefer adding one with this feature PR instead of another PR. This help reviewers to review PR.

now, we can create tidb cluster without helm, it's very easy to start a tidb cluster. A basic e2e case can be:

- create a TiDBInitializer object
- create a TiDBCluster object
- wait and verify tidb cluster is ready in a certain time

This is an example e2e case which create and verify tidb cluster without helm:  https://github.com/pingcap/tidb-operator/pull/1396/files#diff-92579a43c91e473b9716eeeee2068004R653

pkg/controller/configmap_control.go
DanielZhangQD: Cache the configmaps will increase the memory usage a lot.

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: updated

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: done

pkg/apis/pingcap/v1alpha1/tidbcluster.go
DanielZhangQD: done

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: done

DanielZhangQD:
@onlymellb @aylei comments addressed, PTAL.
@cofyc I will add e2e soon.

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,428 @@
...
+		update = true
+	}
+	if ti.Status.Phase != phase {
+		ti.Status.Phase = phase
+		update = true
+	}
+	if update {
+		return tm.retryUpdateStatus(ti)
+	}
+	return nil
```
aylei: This could be simplified using `GuaranteedUpdate`, we don't have to check the equality in client-side actually, if the update do not mutate the object, the `resourceVersion` of object won't be changed, that is, the object will not be affected at all.

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,428 @@
...
+	}, cm)
+	if err != nil {
+		return err
+	}
+	if !exist {
+		err = tm.syncTiDBInitConfigMap(ti)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
DanielZhangQD: For a one-time deal, I agree that we can just create the object, but here is the logic in sync, it will be done at least periodically, it's not reasonable to create the object each time.

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,428 @@
...
+	status := ti.Status.DeepCopy()
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var updateErr error
+		_, updateErr = tm.cli.PingcapV1alpha1().TidbInitializers(ns).Update(ti)
+		if updateErr == nil {
+			klog.Infof("TidbInitializer: [%s/%s] status updated successfully", ns, name)
+			return nil
+		}
+		klog.Errorf("failed to update TidbInitializer: [%s/%s], error: %v", ns, name, updateErr)
```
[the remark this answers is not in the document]
DanielZhangQD: updated

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,428 @@
...
+	if ti.Status.Phase != phase {
+		ti.Status.Phase = phase
+		update = true
+	}
+	if update {
+		return tm.retryUpdateStatus(ti)
+	}
+	return nil
+}
+func (tm *tidbInitManager) retryUpdateStatus(ti *v1alpha1.TidbInitializer) error {
```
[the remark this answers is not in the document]
DanielZhangQD: updated

pkg/manager/member/tidb_init_manager.go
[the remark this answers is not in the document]
DanielZhangQD: Check before creation can reduce the API call to kube-apiserver, and I have moved the check block to separate sync functions.

pkg/manager/member/tidb_init_manager.go
```diff
@@ -0,0 +1,428 @@
...
+	if err == nil {
+		// job has been created
+		return tm.updateStatus(ti.DeepCopy(), job)
+	}
+
+	if !errors.IsNotFound(err) {
+		return fmt.Errorf("TiDBInitializer %s/%s get job %s failed, err: %v", ns, ti.Name, name, err)
+	}
+
+	return tm.syncTiDBInitJob(ti)
```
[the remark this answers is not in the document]
DanielZhangQD: Check before creation can reduce the API call to kube-apiserver, and I have moved the check block to separate sync functions.

pkg/manager/member/tidb_init_manager.go
DanielZhangQD: Check before creation can reduce the API call to kube-apiserver, and I have updated to use `GuaranteedUpdate`.

DanielZhangQD:
@aylei comments addressed, PTAL again.

PR #1409, opened 2019-12-25
refactor the tidbcluster type definition in favor of defaulting

### LINKED ISSUE #1399: refactor the type definition of tidbcluster to ease validation, defaulting and apply
## Feature Request

This is a prerequisite of #1332 . The problem is, the Go definition of `TidbCluster` is failed to distinct "empty" and "unset" of a field. For example:

```
EnablePVReclaim bool `json:"enablePVReclaim,omitempty"`
```

Whether the user explicitly **sets it to `false`** or **omits this field**, the admission-server will see `EnablePVReclaim=false`. The admission-server cannot distinguish whether this field is "unset" or default to "false" because of Go type defaults. Nor can it set defaults.

[`server-side apply`](https://github.com/kubernetes/enhancements/issues/555) the same, the kube-apiserver cannot distinguish whether `EnablePVReclaim=false` is explicitly by user, hence cannot apply this change to the existing resource, that is, user can never change `EnablePVReclaim` to `false` via `server-side apply`.

I propose to refactor the type definition of tidbcluster, that is, change most of the fields to pointer, before proceed to #1332 . The guideline about whether a field type should be pointer can be easily concluded from above:

- Generally, in semantic, pointer type is the superset of non-pointer type, so we should prefer pointer in most of the time.
- Pointer type is harder to manipulate, which adds extra coding burden, so **we optionally could avoid pointer type if we are sure that the "empty value" of the type will never be used**.
  - Example:
    ```
    Version string `json:version`
    ```
    We could use `string` instead of `*string` for the cluster version because empty string `""` is not allowed in our business logic.

aylei:
close #1399 
Some field in TidbCluster type definition is changed to pointer type, and related changes are made.

Core changes are in `types.go`: https://github.com/pingcap/tidb-operator/pull/1409/files?file-filters%5B%5D=.go#diff-98e6ad9bd9688c018746d6bf46e0e6d6
### Check List 

Tests 

 - Unit test
 - E2E test

There should be no user-facing changes.


pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -104,62 +104,91 @@ type TidbClusterList struct {
 type TidbClusterSpec struct {
 
 	// PD cluster spec
-	PD PDSpec `json:"pd,omitempty"`
+	PD PDSpec `json:"pd"`
```
aylei: Do not omit empty for mandatory fields, so that the it is more clear that this struct is wrongly empty if the validation is bypassed or not present.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -104,62 +104,91 @@ type TidbClusterList struct {
...
+
+	// Time zone of TiDB cluster Pods
+	// Optional: Defaults to UTC
+	// +optional
+	Timezone string `json:"timezone,omitempty"`
+
+	// Services list non-headless services type used in TidbCluster
+	// Deprecated
+	// +k8s:openapi-gen=false
+	Services []Service `json:"services,omitempty"`
```
aylei: For deprecated fields, skip the openapi gen, so that they could be hidden from user in the schema and documents.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -174,91 +203,150 @@ type TidbClusterStatus struct {
...
+	// The desired ready replicas
+	// +kubebuilder:validation:Minimum=1
 	Replicas int32 `json:"replicas"`
-	// +k8s:openapi-gen=false
-	Service          *ServiceSpec `json:"service,omitempty"`
-	StorageClassName string       `json:"storageClassName,omitempty"`
 
+	// Base image of the component, image tag is now allowed during validation
+	// +kubebuilder:default=pingcap/pd
+	// +kubebuilder:validation:Pattern=^[^:]+$
```
aylei: Do not allow `:` character, this is merely disallow image tag in `baseImage`, a complete image regex is too complex to write here.

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -21,178 +21,112 @@ import (
...
+		return nil
 	}
-	return &componentAccessorImpl{&tc.Spec, &tc.Spec.Pump.ComponentSpec}, true
+	image := tc.Spec.Pump.Image
+	baseImage := tc.Spec.Pump.BaseImage
+	// base image takes higher priority
+	if baseImage != "" {
+		version := tc.Spec.Pump.Version
+		if version == nil {
+			version = &tc.Spec.Version
```
onlymellb: `tc.Spec.Version` must not be nil?

aylei: Yes, it is a `string`

pkg/apis/pingcap/v1alpha1/tidbcluster_component.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (a *componentAccessorImpl) Annotations() map[string]string {
+	anno := map[string]string{}
+	for k, v := range a.ClusterSpec.Annotations {
+		anno[k] = v
+	}
+	for k, v := range a.ComponentSpec.Annotations {
+		anno[k] = v
+	}
+	return anno
```
onlymellb: This pattern appears more than once, extracted as a function to merge two `map`?

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -104,62 +104,93 @@ type TidbClusterList struct {
...
+	// ImagePullPolicy of TiDB cluster Pods
+	// +kubebuilder:default=IfNotPresent
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
 
-	// Enable TLS connection between TiDB server components
-	EnableTLSCluster bool `json:"enableTLSCluster,omitempty"`
+	// ConfigUpdateStrategy determines how the configuration change is applied to the cluster.
+	// UpdateStrategyInPlace will update the ConfigMap of configuration in-place and an extra rolling-update of the
+	// cluster component is needed to reload the configuration change.
+	// UpdateStrategyRollingUpdate will create a new ConfigMap with the new configuration and rolling-update the the
```
onlymellb:
```suggestion
	// UpdateStrategyRollingUpdate will create a new ConfigMap with the new configuration and rolling-update the
```

pkg/apis/pingcap/v1alpha1/types.go
onlymellb: We can't use this regex，the image reference may include the port number, such as `10.0.0.1:5000/pingcap/tikv`. We can refer to [here](https://github.com/docker/distribution/blob/master/reference/regexp.go#L68-L70) to generate the correct regular expression, I tested it works. but the regular expression is too long. This is the correct regular expression `^((?:(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9])(?:(?:\.(?:[a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9-]*[a-zA-Z0-9]))+)?(?::[0-9]+)?/)?[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:/[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?)$ `

aylei: Good catch, I'd remove it for now and do it programmatically in the validation webhook

pkg/apis/pingcap/v1alpha1/tidbcluster_component.go
aylei: The content of this file is moved from `tidbcluster.go`, there are many other places that use for range to merge maps that should be abstracted, I thought we could leave it as a future enhancement.

aylei:
@onlymellb I've addressed a merge conflict, PTAL again

PR #1411, opened 2019-12-25
Add default behavior in config field description

### LINKED ISSUE #986: TidbCluster: configuration management
## Feature Request

dependent on Helm, and hard to integrate, not compatible when supporting different tidb versions, no validation.

- Not dependent on Helm
- Get/Update configuration items of components(PD/TiDB/TiKV) through HTTP API
- Compatible with different TiDB versions，no more tasks when updating a tidb cluster
- Validate configuration items
- configuration item need to contain following fields
  - configuration item name
  - default value
  - current value
  - need to reboot or not
  - range of value
  - configuration item description

aylei:

close #986 
1. some configuration fields that are not expected to modified by user are removed
2. add default behavior of fields
3. mark the some PD configurations as immutable, because they can only be modified via pd-ctl after cluster creation, modify through the CR is supposed to cause an `ValidateUpdate` error in the later validation logic

Note that we don't use `// +kubebuilder:default=xxx` comment tag to set defaults of the configuration because the TiKV/PD/TiDB server will automatically perform the behavior when the fields are omitted and we want to keep the spec of TidbCluster simple.

### Check List 

Tests 

 - Unit test
 - E2E test

 ```release-note
Refine the configuration schema of PD/TiKV/TiDB.
 ```
@shuijing198799 @sdojjy @tennix @Yisaer @LinuxGit PTAL


pkg/apis/pingcap/v1alpha1/pd_config.go
```diff
@@ -33,26 +33,14 @@ import (
...
-	ClientUrls string `toml:"client-urls,omitempty" json:"client-urls,omitempty"`
-	// +optional
-	PeerUrls string `toml:"peer-urls,omitempty" json:"peer-urls,omitempty"`
-	// +optional
-	AdvertiseClientUrls string `toml:"advertise-client-urls,omitempty" json:"advertise-client-urls,omitempty"`
-	// +optional
-	AdvertisePeerUrls string `toml:"advertise-peer-urls,omitempty" json:"advertise-peer-urls,omitempty"`
 
 	// +optional
 	Name string `toml:"name,omitempty" json:"name,omitempty"`
```
LinuxGit: Name and DataDir should be removed?

pkg/apis/pingcap/v1alpha1/pd_config.go
```diff
@@ -33,26 +33,14 @@ import (
...
 	// +optional
 	DataDir string `toml:"data-dir,omitempty" json:"data-dir,omitempty"`
 	// +optional
 	ForceNewCluster *bool `json:"force-new-cluster,omitempty"`
+	// Optional: Defaults to true
```
LinuxGit: It seems that configuration items like `ForceNewCluster`, `EnableGRPCGateway ` are hidden.

PR #1414, opened 2019-12-25
avoid trigger rolling-update when template is actually not changed

### LINKED ISSUE #1340: last applied pod template of statefulset will trigger rolling-update unexpectedly
## Bug Report

https://github.com/pingcap/tidb-operator/pull/1287 fix a compatiability issue of tidb-operator. However, in Kubernetes, `dnsPolicy` is default to `ClusterFirst`, so setting this field explicitly to `ClusterFirst` is compatible with leaving it as empty. The change that triggers rolling update is happened in the `.spec.template.annotation[last-applied-config]` field, which marshals the un-defaulted Pod template.

We should move this annotation from `.spec.template.annotations` to `.annotations` to avoid more cases like #1287 .

Existing `.spec.template.annotation[last-applied-config]` will be retained to keep backward compatibility but no longer be read.

ref: https://github.com/pingcap/tidb-operator/issues/1335

aylei:

close #1340 

The problem is that we marshal the local **un-defaulted** pod template to the annotations of pod template, which will cause unexpected rolling-update even if the change is compatible.

### Check List 

Tests 

 - Unit test
 - Stability test

There should be no user-facing change.


pkg/manager/member/utils.go
```diff
@@ -32,6 +32,8 @@ import (
 const (
 	// LastAppliedConfigAnnotation is annotation key of last applied configuration
 	LastAppliedConfigAnnotation = "pingcap.com/last-applied-configuration"
+	// LastAppliedPodTemplateAnnotation is annotation key of last applied configuration
+	LastAppliedPodTemplateAnnotation = "pingcap.com/last-applied-pod-template"
```
cofyc: It is possible to use `LastAppliedConfigAnnotation` because StatefuSetSpec contains PodTemplateSpec?

pkg/manager/member/utils.go
```diff
@@ -148,15 +154,20 @@ func statefulSetEqual(new apps.StatefulSet, old apps.StatefulSet) bool {
...
+		lastAppliedConfig, ok = old.Spec.Template.Annotations[LastAppliedConfigAnnotation]
+	}
+	if ok {
 		err := json.Unmarshal([]byte(lastAppliedConfig), &oldConfig)
 		if err != nil {
 			glog.Errorf("unmarshal PodTemplate: [%s/%s]'s applied config failed,error: %v", old.GetNamespace(), old.GetName(), err)
 			return false
 		}
-		return apiequality.Semantic.DeepEqual(oldConfig, new.Spec)
+		return apiequality.Semantic.DeepEqual(oldConfig, new.Spec.Template.Spec)
```
cofyc: when a new field with the default value (e.g. `DNSPolicy`) is added in client-side, this function will return false, because the old configuration does not have this field. `tkmm.tikvUpgrader.Upgrade` will be called. 

however, the stateful set controller will not upgrade pods because after defaulting in apiserver, the pod template spec does not change.

does this have problems?

pkg/manager/member/utils.go
aylei: Make sense

pkg/manager/member/utils.go
aylei: Yes, that's right

pkg/manager/member/utils.go
aylei: Addressed in https://github.com/pingcap/tidb-operator/pull/1414/commits/498009cb1c95aa36c9ca3202d20612d02997f6dd

aylei:
@Yisaer @cofyc comments addressed, PTAL

pkg/manager/member/utils.go
```diff
@@ -83,20 +83,7 @@ func SetStatefulSetLastAppliedConfigAnnotation(set *apps.StatefulSet) error {
 
 // SetLastAppliedConfigAnnotation set last applied config info to Statefulset's annotation and the podTemplate's annotation
 func SetLastAppliedConfigAnnotation(set *apps.StatefulSet) error {
```
cofyc: remove `SetLastAppliedConfigAnnotation` entirely?

aylei: make sense, I bet there's no external usage of this method, though public

aylei: removed in https://github.com/pingcap/tidb-operator/pull/1414/commits/e2868a6a7e5843f57cd763867cb995c4edf78b76

PR #1415, opened 2019-12-25
fix error importer config

aylei:

Cannot start tikv-importer using the default config
```
thread 'main' panicked at 'invalid auto generated configuration file /etc/tikv-importer/tikv-importer.toml, err unknown field `metrics`, expected one of `log-level`, `log-file`, `log-rotation-timespan`, `storage`, `server`, `metric`, `rocksdb`, `security`, `import`', src/import/config.rs:134:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Reference: https://github.com/tikv/importer/blob/master/etc/tikv-importer.toml
A typo is fixed
### Check List 

Tests 

 - Unit test
 - E2E test

 ```release-note
Fix the default tikv-importer configuration
 ```

### DIFF
--- a/charts/tidb-cluster/values.yaml
+++ b/charts/tidb-cluster/values.yaml
@@ -768,7 +768,7 @@ importer:
   pushgatewayImagePullPolicy: IfNotPresent
   config: |
     log-level = "info"
-    [metrics]
+    [metric]
     job = "tikv-importer"
     interval = "15s"
     address = "localhost:9091"


aylei:
> LGTM
> 
> add an e2e case for tikv-importer?

It is helpful but we may not do this, tikv-importer is going to integrated into the tikv process and we've not included tikv-importer in #1121 neither

PR #1419, opened 2019-12-25
make controller logic independable with certain tidbcluster label

### LINKED ISSUE #1418: remove dependency on tidbcluster labels
## Feature Request

follow up: https://github.com/pingcap/tidb-operator/issues/1121#issuecomment-568899391

aylei:

close #1418 

### Check List 

Tests 

 - Unit test
 - E2E test

 ```release-note
The instance label key of resources owned by `tidbcluster` is default to the cluster name now. 
```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/tidbcluster.go
+++ b/pkg/apis/pingcap/v1alpha1/tidbcluster.go
@@ -409,7 +409,7 @@ func (tc *TidbCluster) GetInstanceName() string {
 	labels := tc.ObjectMeta.GetLabels()
 	// Keep backward compatibility for helm.
 	// This introduce a hidden danger that change this label will trigger rolling-update of most of the components
-	// TODO: disallow mutation of this label or adding this label with value other than the cluster name in ValidateUpdate()
+	// TODO(aylei): disallow mutation of this label or adding this label with value other than the cluster name in ValidateUpdate()
 	if inst, ok := labels[label.InstanceLabelKey]; ok {
 		return inst
 	}
--- a/pkg/apis/pingcap/v1alpha1/tidbcluster.go
+++ b/pkg/apis/pingcap/v1alpha1/tidbcluster.go
@@ -16,6 +16,7 @@ package v1alpha1
 import (
 	"fmt"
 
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: pkg/apis/pingcap/v1alpha1/tidbcluster.go [hunk 1]
```diff
@@ -403,3 +404,14 @@ func (tc *TidbCluster) Timezone() string {
 	}
 	return tz
 }
+
+func (tc *TidbCluster) GetInstanceName() string {
```
aylei: Alternatively, we could define a `GetLabels()` method and shadow the one of  `ObjectMeta` so that the controller logic don't have to be modified, but that way we have to copy a map every time this method get called.

pkg/manager/member/pd_member_manager_test.go
```diff
@@ -1088,7 +1088,7 @@ func TestGetPDConfigMap(t *testing.T) {
 					Labels: map[string]string{
 						"app.kubernetes.io/name":       "tidb-cluster",
 						"app.kubernetes.io/managed-by": "tidb-operator",
-						"app.kubernetes.io/instance":   "",
+						"app.kubernetes.io/instance":   "foo",
```
onlymellb: Use const vars in label package to replace these keys?

pkg/manager/member/pump_member_manager_test.go
```diff
@@ -676,7 +676,7 @@ func TestGetNewPumpConfigMap(t *testing.T) {
 					Labels: map[string]string{
 						"app.kubernetes.io/name":       "tidb-cluster",
 						"app.kubernetes.io/managed-by": "tidb-operator",
-						"app.kubernetes.io/instance":   "",
+						"app.kubernetes.io/instance":   "foo",
```
onlymellb: ditto

pkg/manager/member/pd_member_manager_test.go
aylei: I think both make sense, by not using the const var, the correctness of the `labelKey` itself is also tested, it just depends on the purpose of the UTs

PR #1444, opened 2019-12-30
fix go boilerplate file: use YEAR and remove dot

cofyc:

- use YEAR instead of a hard-coded value
- `.` is not needed

### Check List 

Tests 

 - Unit test
 - E2E test


hack/boilerplate.go.txt
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019. PingCAP, Inc.
+// Copyright YEAR PingCAP, Inc.
```
cofyc: - `.` is not needed, see https://github.com/pingcap/tidb/blob/master/cmd/importer/job.go#L1

cofyc: `YEAR` is a place holder, see example https://github.com/kubernetes/kubernetes/blob/master/hack/boilerplate/boilerplate.go.txt.

PR #1445, opened 2019-12-30
Add tidbmonitor e2e test

Yisaer:

1. Add common e2e monitor test for `TidbMonitor` and previous `tidb-cluster-monitor`

2. fix label bugs for TidbMonitor

3. fix prometheus config template for `TidbMonitor`


pkg/monitor/monitor/util.go
```diff
@@ -215,9 +215,11 @@ func getMonitorDeploymentSkeleton(sa *core.ServiceAccount, monitor *v1alpha1.Tid
 	monitorLabel := label.New().Instance(monitor.Name).Monitor().Labels()
 	replicas := int32(1)
 
+	labels := label.NewMonitor().Instance(monitor.Name).Monitor().Labels()
```
DanielZhangQD: No need to call `Labels()` in the end.
BTW, what does L215 for?

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -705,6 +705,23 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+		cluster.Resources["tidb.replicas"] = "1"
+		oa.DeployTidbClusterOrDie(&cluster)
+		oa.CheckTidbClusterStatusOrDie(&cluster)
+
+		tc, err := cli.PingcapV1alpha1().TidbClusters(cluster.Namespace).Get(cluster.ClusterName, metav1.GetOptions{})
+		framework.ExpectNoError(err, "Expected get tidbcluster")
+
+		tm := newTidbMonitor("monitor", tc.Namespace, tc, true, false)
+		oa.DeployTidbMonitorOrDie(tm)
+		oa.CheckTidbMonitorOrDie(tm)
```
cofyc: inline these two functions

Current `operatorActions` is too complicated. We should not add more methods to it.

cofyc: if you think the logic of `CheckTidbMonitor` will be used by other e2e tests, you can add a simple utility function for it.

tests/monitor.go
```diff
@@ -0,0 +1,90 @@
...
+		if !podutil.IsPodReady(pod) {
+			klog.Infof("tm[%s/%s]'s pod[%s/%s] is not ready", tm.Namespace, tm.Name, pod.Namespace, pod.Name)
+			return false, nil
+		}
+		if tm.Spec.Grafana != nil && len(pod.Spec.Containers) != 3 {
+			return false, fmt.Errorf("tm[%s/%s]'s pod didn't have 3 containers with grafana enabled", tm.Namespace, tm.Name)
+		}
+		if tm.Spec.Grafana == nil && len(pod.Spec.Containers) != 2 {
+			return false, fmt.Errorf("tm[%s/%s]'s pod didnt' have 2 containers with grafana disabled", tm.Namespace, tm.Name)
+		}
```
aylei: this is somehow related too much to the implementation rather than the user-facing interface, IMHO, the checks are better to be "Are rule reloading, Grafana and Prometheus functional?", "Are the dashboards properly initialized?", etc.

pkg/monitor/monitor/util.go
Yisaer: `monitorLabel` is for the label of the `tidbmonitor` deployment and `labels` here is for the label selector of `tidbmonitor` deployment

pkg/label/label.go
```diff
@@ -167,6 +167,14 @@ func NewBackupSchedule() Label {
 	}
 }
 
+// NewMonitor initialize a new label for monitor of tidb-monitor
+func NewMonitor() Label {
+	return Label{
+		NameLabelKey:      TiDBMonitorVal,
+		ManagedByLabelKey: "monitor-operator",
```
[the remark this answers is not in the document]
Yisaer: updated

tests/e2e/tidbcluster/tidbcluster.go
Yisaer: updated

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -756,3 +772,58 @@ func newTidbClusterConfig(cfg *tests.Config, ns, clusterName, password, tidbVers
 		ClusterVersion:         tidbVersion,
 	}
 }
+
+func newTidbMonitor(name, namespace string, tc *v1alpha1.TidbCluster, grafanaEnabled, persist bool) *v1alpha1.TidbMonitor {
```
cofyc: @aylei created a fixture pkg (https://github.com/pingcap/tidb-operator/tree/master/tests/pkg/fixture), move it to this pkg?

the old `newTidbClusterConfig` function in this file can be removed when we abandoned helm entirely.

tests/monitor.go
Yisaer: updated, now we would check both the pod status and Prometheus and Grafana function.

PR #1446, opened 2019-12-30
fix tidb-controller rbac bug

Yisaer:

Add `clusterrolebindings.rbac.authorization.k8s.io` RBAC authority for tidb-controller as tidb-controller would create `clusterrolebindings` or `rolebindings` for tidbmonitor.


charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -78,9 +78,9 @@ Ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#privilege-esc
 */}}
 - apiGroups: ["rbac.authorization.k8s.io"]
   resources: [clusterroles,roles]
-  verbs: ["escalate","create","get","update"]
+  verbs: ["escalate","create","get","update","bind"]
```
aylei: `bind` is used to create or update rolebinding to the referenced role (listed in `resourceNames`), we don't need it here

Yisaer: `bind` here is necessary here,  if I removd  `bind`, then the tidb-controller would sync failed and print error info like 
``` bash
tm[tidb-admin/monitor]'s rbac failed to sync,err: clusterrolebindings.rbac.authorization.k8s.io "monitor-monitor" is forbidden: user "system:serviceaccount:tidb-admin:tidb-controller-manager" (groups=["system:serviceaccounts" "system:serviceaccounts:tidb-admin" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
```

aylei: This is indicates that controller-manager is going to grant permissions it doesn't have, look like using `bind` without specifying `resourceNames` blindly let the controller-manager could grant everything, which is unsecure

Yisaer: Updated. 
Currently, the operator have `/metrics` authority and `bind` authority was removed.

PR #1452, opened 2019-12-31
set provider versions for all modules

### LINKED ISSUE #1172: terraform: all modules should lock provider versions too
## Feature Request

`hack/check–terraform.sh` will install dependencies for each module, and run validate and fmt in that module. Currently, it will install the latest providers for modules under `deploy/modules`.

gcp/tidb-operator has a problem with the latest `google`, I'm going to fix it in https://github.com/pingcap/tidb-operator/pull/1102.

```
Error: "ip_allocation_policy.0.use_ip_aliases": [REMOVED] This field is removed as of 3.0.0. If previously set to true, remove it from your config. If false, remove it.

  on main.tf line 1, in resource "google_container_cluster" "cluster":
   1: resource "google_container_cluster" "cluster" {


```

all modules should lock provider versions.

DanielZhangQD:
fix #1172 
set provider versions for all modules
### Check List 

Tests 

 - Manual test 

Code changes

 - Has terraform scripts change

 ```release-note
NONE.
 ```


DanielZhangQD:
release-1.0 does not check terraform scripts for internal modules and we have set the provider versions in the modules that the user will run `init`, so no need to cherry-pick this to release-1.0.

PR #1454, opened 2019-12-31
fix pd client tls and use table driven tests to cover tls case

### LINKED ISSUE #1450: TLS e2e test fails
## Bug Report

https://github.com/pingcap/tidb-operator/pull/1430#issuecomment-569205786.

I think it's better to have a separate test for TLS too.

cofyc:
fixes #1450 

- fix pd client tls bug
- use table driven tests to cover tls case (more extensible)

### Check List 

Tests 

 - Unit test
 - E2E test

### DIFF
--- a/tests/e2e/tidbcluster/tidbcluster.go
+++ b/tests/e2e/tidbcluster/tidbcluster.go
@@ -45,7 +45,6 @@ import (
 	"github.com/pingcap/tidb-operator/tests/e2e/util/portforward"
 	"github.com/pingcap/tidb-operator/tests/pkg/apimachinery"
 	"github.com/pingcap/tidb-operator/tests/pkg/blockwriter"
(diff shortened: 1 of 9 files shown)


tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -105,19 +104,41 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+					"tikv.resources.limits.storage": "1G",
+				},
+			},
+			{
+				Version: "v3.0.5",
+				Name:    "basic-v3-cluster-tls",
+				Values: map[string]string{
+					"enableTLSCluster": "true",
+				},
+			},
```
cofyc: we can add more to cover various configurations of cluster

PR #1458, opened 2019-12-31
fix license headers and verify boilerplate

### LINKED ISSUE #1022: hack/verify: verify boilerplate
## Feature Request

Make sure all our files have the PingCAP author info.

cofyc:

fixes #1022 

### Check List 

Tests 

 - Unit test
 - E2E test


hack/verify-boilerplate.sh
```diff
@@ -0,0 +1,60 @@
...
+		-o -path './vendor/*' \
+		-o -path './pkg/client/*' \
+        -o -path './*/.terraform/*' \
+        -o -path './deploy/*' \
+        -o -path '*/*.sh' \
+        -o -path '*/Makefile' \
+        -o -path '*/Dockerfile' \
+        -o -path '*/*.register.go' \
+    \)
+))
```
aylei: can we consider gitignore? it is okay for CI but broken in my local development env

cofyc: They use different syntaxes. Which pattern you want to ignore, I can add it.

aylei: `./.idea/*`

hack/boilerplate/boilerplate.py
```diff
@@ -0,0 +1,242 @@
...
+    return True
+
+
+def file_extension(filename):
+    return os.path.splitext(filename)[1].split(".")[-1].lower()
+
+
+skipped_dirs = ['Godeps', 'third_party', '_gopath', '_output', '.git', 'cluster/env.sh',
+                "vendor", "test/e2e/generated/bindata.go", "hack/boilerplate/test",
+                "staging/src/k8s.io/kubectl/pkg/generated/bindata.go"]
```
weekface: do we have this directory?

cofyc: these directories are hardcoded in this k8s script, we don't have these directories, it's ok to remove or keep them

hack/verify-boilerplate.sh
cofyc: fixed in f7306b8c67a5716b08eb7eff779d08ba671b2db6, all hidden files/directories are ignored

hack/boilerplate/boilerplate.py
cofyc: fixed in e095abbe072cfdd720c6c1ab560c352cd59e053c, hardcoded rules are ignored (except the script itself)

PR #1462, opened 2020-01-02
add license header for shell scripts

cofyc:

- add license header for shell scripts
- verify license header in the check phase

### Check List 

Tests 

 - Unit test
 - E2E test

### DIFF
--- a/hack/verify-boilerplate.sh
+++ b/hack/verify-boilerplate.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2019 PingCAP, Inc.
+# Copyright 2020 PingCAP, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


pkg/apis/pingcap/v1alpha1/tidbcluster_component.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019. PingCAP, Inc.
+// Copyright 2019 PingCAP, Inc.
```
cofyc: bugfix (in #1458, I didn't check license header in the check phase)

hack/verify-boilerplate.sh
```diff
@@ -39,12 +43,12 @@ files=($(find . -type f -not \( \
         -o -path './vendor/*' \
         -o -path './pkg/client/*' \
         -o -path './*/.terraform/*' \
+        -o -path './tests/images/*/*' \
```
cofyc: `./tests/images/*/*` matches the docker image build directories which we should ignore

Issue #1465, opened 2020-01-02
Emit k8s events when pd failover


weekface:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #1466, opened 2020-01-02
failover: emit events when pd failover

weekface:
fixes: #1465 

This PR tries to emit three  k8s `Events`:  `PDMemberUnhealthy`, `PDMemberMarkedAsFailure` and `PDMemberDeleted` to indicate the pd `failover` procedure when the use type `kubectl describe tc`:

image

But there is a problem that there are many not very useful `Events`:  `SuccessfulUpdate`, I will open another PR to remove them. @cofyc @aylei PTAL.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

Related changes

 ```release-note
Emit events when PD failover
 ```


pkg/controller/tidbcluster/tidb_cluster_controller.go
```diff
@@ -76,11 +76,11 @@ func NewController(
 	tikvFailoverPeriod time.Duration,
 	tidbFailoverPeriod time.Duration,
 ) *Controller {
-	eventBroadcaster := record.NewBroadcaster()
-	eventBroadcaster.StartLogging(glog.Infof)
+	eventBroadcaster := record.NewBroadcasterWithCorrelatorOptions(record.CorrelatorOptions{QPS: 10})
```
weekface: the default `QPS` is too low: `1/ 300`.

pkg/controller/tidbcluster/tidb_cluster_controller.go
```diff
@@ -76,11 +76,11 @@ func NewController(
...
+	eventBroadcaster.StartLogging(glog.V(4).Infof)
```
weekface: `glog.Info` is too noisy.

pkg/controller/tidbcluster/tidb_cluster_controller.go
```diff
@@ -76,11 +76,11 @@ func NewController(
...
 	eventBroadcaster.StartRecordingToSink(&eventv1.EventSinkImpl{
 		Interface: eventv1.New(kubeCli.CoreV1().RESTClient()).Events("")})
-	recorder := eventBroadcaster.NewRecorder(v1alpha1.Scheme, corev1.EventSource{Component: "tidbcluster"})
+	recorder := eventBroadcaster.NewRecorder(v1alpha1.Scheme, corev1.EventSource{Component: "tidb-controller-manager"})
```
weekface: like k8s does

pkg/controller/tidbcluster/tidb_cluster_controller.go
cofyc: 10 events per object per seconds seems too large, it might generate a lot of load on the API server or unable to do other CRUD requests because of exceeding the total throttling limit. Most burstable tokens (25) are consumed by [unnecessary events](https://github.com/pingcap/tidb-operator/issues/1471). We can increase the burst and QPS together but not too much at the first.

weekface: Yeah, 10 is too large.

I will decrease it to a proper value after removing [those events](https://github.com/pingcap/tidb-operator/issues/1471).

pkg/manager/member/pd_failover.go
```diff
@@ -73,9 +78,12 @@ func (pf *pdFailover) Failover(tc *v1alpha1.TidbCluster) error {
 	}
 
 	healthCount := 0
-	for _, pdMember := range tc.Status.PD.Members {
+	for podName, pdMember := range tc.Status.PD.Members {
 		if pdMember.Health {
 			healthCount++
+		} else {
+			pf.recorder.Eventf(tc, apiv1.EventTypeWarning, "PDMemberUnhealthy",
+				"%s(%s) is unhealthy", podName, pdMember.ID)
```
aylei: IMHO, this is a `status` instead of `event`, recording this will be too noisy (despite the flow control, the controller will emit an event in each round of control loop if there is an unhealthy PD member).

weekface: The is already a `.status.pd[].health` attribute.

Emit it as an event is good for `kubectl describe`

aylei: Yes, seems like it is already in the describe result?
```
    Members:
      Yutengqiu - Demo - Pd - 0:
        Client URL:            http://yutengqiu-demo-pd-0.yutengqiu-demo-pd-peer.yutengqiu.svc:2379
        Health:                true
        Id:                    12697782363740270066
        Last Transition Time:  2020-01-06T01:48:58Z
        Name:                  yutengqiu-demo-pd-0
      Yutengqiu - Demo - Pd - 3:
        Client URL:            http://yutengqiu-demo-pd-3.yutengqiu-demo-pd-peer.yutengqiu.svc:2379
        Health:                true
        Id:                    10833084519111696661
        Last Transition Time:  2020-01-06T05:37:58Z
        Name:                  yutengqiu-demo-pd-3
      Yutengqiu - Demo - Pd - 4:
        Client URL:            http://yutengqiu-demo-pd-4.yutengqiu-demo-pd-peer.yutengqiu.svc:2379
        Health:                true
        Id:                    10563190389194377650
        Last Transition Time:  2020-01-06T05:44:25Z
        Name:                  yutengqiu-demo-pd-4
      Yutengqiu - Demo - Pd - 6:
        Client URL:            http://yutengqiu-demo-pd-6.yutengqiu-demo-pd-peer.yutengqiu.svc:2379
        Health:                true
        Id:                    6735927804110166558
        Last Transition Time:  2020-01-06T05:32:10Z
        Name:                  yutengqiu-demo-pd-6
```

weekface: emit an event is more user-friendly, it is a progress of failover.

weekface: or change `PDMemberUnhealthy ` to a more proper name?

aylei: the progress of failover makes sense, it is okay for me to emit events based on the the status of PD when we cannot actually capture the the "PD turning from healthy to unhealthy" event.

The naming issue is just trivial, I think current name is ok

pkg/controller/tidbcluster/tidb_cluster_controller.go
cofyc: what's the current rate of events?

weekface: `rate of events`, does it mean the default value of `QPS` and `Burstable`?

pkg/controller/tidbcluster/tidb_cluster_controller.go
cofyc: events are often the most important messages that end users care, I'd prefer to have it on in logs by default, how about `v(2)`?
If the logging of the event is too noisy, we should reduce the rate of events we reported to the API server. 
Another reason is the event can be dropped, in that case, we can find the full events from the logs.

weekface: reasonable

pkg/controller/tidbcluster/tidb_cluster_controller.go
cofyc: create an issue about setting the rate to proper values?

pkg/manager/member/pd_failover.go
cofyc: is it possible to report the event on state change?

pkg/controller/tidbcluster/tidb_cluster_controller.go
weekface: https://github.com/pingcap/tidb-operator/issues/1494

an issue opened

pkg/manager/member/pd_failover.go
weekface: good suggestion, we can do this in the [`syncTidbClusterStatus `](https://github.com/pingcap/tidb-operator/blob/d314b251735693a000088d7011f9a5b2932f579f/pkg/manager/member/pd_member_manager.go#L312) method, not failover method, an issue opened: https://github.com/pingcap/tidb-operator/issues/1495

PR #1425, opened 2019-12-26
Add TidbMonitor Controller

Yisaer:
### What problem does this PR solve? 

resolve: https://github.com/pingcap/tidb-operator/issues/1261 controller part.

PR #1474, opened 2020-01-03
Make Monitor Prometheus config template as Go struct

Yisaer:
https://github.com/pingcap/tidb-operator/pull/1425#discussion_r361419097
Currently, the template for `TiDB Monitor`'s Prometheus Config is raw defined. In this request, we defined the Prometheus config template by Go struct and marshal it into yaml.

```release-note
NONE
```


Yisaer:
As https://github.com/pingcap/tidb-operator/pull/1445 add e2e test for `TidbMonitor`, we could wait until https://github.com/pingcap/tidb-operator/pull/1445 are merged.

pkg/monitor/monitor/template.go
```diff
@@ -14,131 +14,67 @@
...
-    - source_labels: [__meta_kubernetes_pod_name]
-      action: replace
-      target_label: instance
-    - source_labels: [__meta_kubernetes_pod_label_app_kubernetes_io_instance]
-      action: replace
-      target_label: cluster
-  {{- end }}
-rule_files:
-  - '/prometheus-rules/rules/*.rules.yml'`))
+type PrometheusConfig struct {
```
qiffang: I think it is better to use prometheus Config struct.
These structs have been defined in prometheus Config.

pkg/monitor/monitor/template.go
```diff
@@ -162,15 +98,184 @@ type MonitorConfigModel struct {
...
+				},
+			},
+		},
+		TlsConfig: TlsConfig{
+			InsecureSkipVerify: true,
+		},
+		Schema: "http",
+		RelabelConfigs: []RelabelConfig{
+			{
+				SourceLabels: "[__meta_kubernetes_pod_label_app_kubernetes_io_instance]",
```
aylei: these sourceLabels appears multiple times, we could define constants for them

pkg/monitor/monitor/template.go
```diff
@@ -162,15 +98,184 @@ type MonitorConfigModel struct {
...
+				Action:       "keep",
+				Regex:        model.ReleaseTargetRegex,
+			},
+			{
+				SourceLabels: "[__meta_kubernetes_pod_annotation_prometheus_io_scrape]",
+				Action:       "keep",
```
aylei: I believe there is a enum in prometheus for this

pkg/monitor/monitor/template.go
```diff
@@ -162,15 +98,184 @@ type MonitorConfigModel struct {
...
+	if len(model.AlertmanagerURL) > 0 {
+		addAlertManagerUrl(pc, model)
+	}
+	bs, err := yaml.Marshal(pc)
 	if err != nil {
 		return "", err
 	}
-	return buff.String(), nil
+	// remove character "'"
+	return strings.ReplaceAll(string(bs), "'", ""), nil
```
aylei: why is this necessary?

A suggestion for good comment is to comment the intent of the code, not the things it does

pkg/monitor/monitor/template_test.go
```diff
@@ -0,0 +1,37 @@
...
+			"ns1",
+			"ns2",
+		},
+		EnableTLSCluster: false,
+	}
+	content, err := RenderPrometheusConfig(model)
+	g.Expect(err).NotTo(HaveOccurred())
+	encodedContent := base64.StdEncoding.EncodeToString([]byte(content))
+	g.Expect(encodedContent).Should(Equal(encodedConfig))
+}
```
aylei: This is case is obscure, it is hard to reason why when it fails, I'd suggest just to expect the marshaled text of the config

pkg/monitor/monitor/template.go
aylei: +1 if possible

Yisaer: updated.

pkg/monitor/monitor/template.go
Yisaer: updated.

pkg/monitor/monitor/template.go
Yisaer: removed now.

pkg/monitor/monitor/template_test.go
Yisaer: updated.

pkg/monitor/monitor/template.go
Yisaer: updated.

pkg/monitor/monitor/template.go
```diff
@@ -154,23 +48,239 @@ var dashBoardConfig = `{
...
+		},
+	}
 }
 
-func renderTemplateFunc(tpl *template.Template, model interface{}) (string, error) {
-	buff := new(bytes.Buffer)
-	err := tpl.Execute(buff, model)
+func addTlsConfig(pc *config.Config, cmodel *MonitorConfigModel) {
+
+	r, _ := config.NewRegexp(".*\\-tikv\\-\\d*$")
```
aylei: Do not ignore error, for compiling regexp from string literal, I recommend creating a package level variable in init block, so it can fail fast on start if the regex is wrong

pkg/monitor/monitor/template.go
```diff
@@ -154,23 +48,239 @@ var dashBoardConfig = `{
...
+	for id, sconfig := range pc.ScrapeConfigs {
+		if sconfig.JobName == "tidb-cluster" {
+			sconfig.HTTPClientConfig.TLSConfig = config.TLSConfig{
+				CAFile:   "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+				CertFile: "/var/lib/pd-client-tls/cert",
+				KeyFile:  "/var/lib/pd-client-tls/key",
```
aylei: There are constants for these paths

Yisaer: updated.

pkg/monitor/monitor/template.go
Yisaer: updated.

aylei: I didn't see how it was updated

pkg/monitor/monitor/template.go
```diff
@@ -154,23 +52,255 @@ var dashBoardConfig = `{
...
+					},
+				},
+			},
+		},
+	}
+}
+
+func addTlsConfig(pc *config.Config, cmodel *MonitorConfigModel) {
+
+	r, _ := config.NewRegexp(".*\\-tikv\\-\\d*$")
```
aylei: Looks like that this line is not changed

Yisaer: updated.

Issue #1440, opened 2019-12-30
api: apis/pingcap/v1alpha1 compatibility unit tests


cofyc:
## Feature Request

**Describe the feature you'd like:**

Add unit tests to make sure any changes in `apis/pingcap/v1alpha1 ` does not break API compatibility.

**Describe alternatives you've considered:**

- integration tests, we can add in later PRs
- e2e tests, we can add in later PRs

**Teachability, Documentation, Adoption, Migration Strategy:**

aylei:
which version should be considered? For v1.0.x it is a little tricky because it cannot work without helm actually

cofyc:
verify changes in the same version group (v1alpha1) should be backward compatible, the basic goal is 

- we should generate the same stateful sets for the old tidbcluster objects created in tidb-operator 1.0.

aylei:
That'd be great

PR #1489, opened 2020-01-06
Retain last applied pod template to keep backward-compatibility

### LINKED ISSUE #1475: Update statefulset should retain the last-applied-configuration
## Bug Report

#1414 will trigger rolling-update of legacy cluster because the annotations of the old cluster are totally overridden:

Before:
```
        annotations:
          prometheus.io/path: /metrics
          prometheus.io/port: "2379"
          prometheus.io/scrape: "true"
```

After upgrading operator:
```
        annotations:
          pingcap.com/last-applied-configuration: '{"volumes"....'
          prometheus.io/path: /metrics
          prometheus.io/port: "2379"
          prometheus.io/scrape: "true"
```
Though we do not set this annotation anymore, we should keep this during update to keep the backward compatibility.

aylei:

close #1475 
Now the `last-applied-config` annotation of statefulset's PodTemplate will be retained during update to keep the backward-compatibility.

In addition, we need #1440 to help us prevent such issues.
### Check List 

Tests 

 - Unit test
 - E2E test

```release-note
Retain last-applied-config of StatefulSet's `PodTemplate` during updates to keep backward compatibility.
```

### DIFF
--- a/pkg/manager/member/utils.go
+++ b/pkg/manager/member/utils.go
@@ -27,6 +27,7 @@ import (
 	apps "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	glog "k8s.io/klog"
 )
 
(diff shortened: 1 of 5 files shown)


aylei:
@weekface @cofyc @DanielZhangQD @Yisaer PTAL

pkg/manager/member/utils.go
```diff
@@ -262,3 +263,36 @@ func MapContainers(podSpec *corev1.PodSpec) map[string]corev1.Container {
...
+		// Retain the deprecated last applied pod template annotation for backward compatibility
+		var podConfig string
+		var hasPodConfig bool
+		if oldSet.Spec.Template.Annotations != nil {
+			podConfig, hasPodConfig = oldSet.Spec.Template.Annotations[LastAppliedConfigAnnotation]
+		}
+		set.Spec.Template = newSet.Spec.Template
+		if hasPodConfig {
+			set.Spec.Template.Annotations[LastAppliedConfigAnnotation] = podConfig
+		}
```
cofyc: `SetStatefulSetLastAppliedConfigAnnotation` is used to set last-applied-config annotation(s), move the `Spec.Template.Annotations[LastAppliedConfigAnnotation]` backward compatility logic into it?

aylei: It will change the interface of this method (which needs the old annotation)

PR #1491, opened 2020-01-06
add some small, single-purpose scripts

### LINKED ISSUE #1490: unable to verify crd groups
## Bug Report

Before verifying, we must copy the original file(s) into a temporary place. However, crd-groups.sh runs in `hack/codegen.sh` and we didn't back up it before it (`crd-groups.sh verify` runs after `hack/codegen.sh`), so the check is always ok.

It will be less error-prone to maintain small, single-purpose scripts.

cofyc:

fixes #1490 

- split `hack/crd-groups.sh` into `hack/{verify,update}-openapi-spec.sh, hack/{verify-update}-crd-groups.sh`
- add `hack/verify-all.sh` to run all verify scripts
- add `hack/update-all.sh` to run all update scripts
- rename `hack/codegen.sh` to `hack/update-codegen.sh`
- rename `hack/test-codegen.sh` to `hack/update-test-codegen.sh`

### Check List 

Tests 

 - Unit test
 - E2E test

```release-note
NONE
```


hack/update-all.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT=$(unset CDPATH && cd $(dirname "${BASH_SOURCE[0]}")/.. && pwd)
+cd $ROOT
+
+hack/update-codegen.sh
+hack/update-test-codegen.sh
```
cofyc: @aylei 
hi, I renamed the `hack/test-codegen.sh` to `hack/update-test-codegen.sh`  to follow the update script patterh `hack/update-xxx.<ext>`.  I think it's better to name the same type of scripts consistently.

aylei: Thanks!

Issue #1482, opened 2020-01-03
TidbMonitor: should not specify log level when prometheus.logLevel is empty


aylei:
## Bug Report

Currently, when `prometheus.logLevel` is left unset, the monitoring Pod will be unable to started:

```
Error parsing commandline arguments: unrecognized log level ""
usage: prometheus [<flags>]

The Prometheus monitoring server
```

Looks like we should not set the logLevel flag when it is actually empty https://github.com/pingcap/tidb-operator/blob/master/pkg/monitor/monitor/util.go#L374

aylei:
closed via #1500

Issue #1481, opened 2020-01-03
clearer message when TidbMonitor failed syncing because of the absence of referenced cluster


aylei:
## Bug Report

When I try to troubleshoot why my monitoring Pod did not get created, I could only find this clue in the log:

```
E0103 15:58:39.584097       1 tidb_monitor_controller.go:118] TidbMonitor: wuyelei/basic, sync failed, err: tidbcluster.pingcap.com "basic" not found, requeuing
```

There are several problems I suppose:

1. tidbcluster `basic` exists, why the controller cannot find it? It turns out there is no proper defaults for the namespace of the cluster, as mentioned in #1480 
2. Users need more clearer message to indicates why the Pod not get created, a status of TidbMonitor or an event could be much more friendly for users, who may not be willing to dig into a great deal of logs
3. The sync loop did not get requeue actually, the log is misleading

Yisaer:
fixed in https://github.com/pingcap/tidb-operator/pull/1500

Issue #1480, opened 2020-01-03
no default for `.spec.clusterRef.namespace` of TiDB monitor


aylei:
There is no default for `.spec.clusterRef.namespace` of TiDB monitor, which could be default to the local namespace (where the `TidbMonitor` located) to reduce the configuration complexity.

e.g.

I want to deploy `tidbmonitor` and `tidbcluster` in one yaml:

```
apiVersion: pingcap.com/v1alpha1
kind: TidbCluster
metadata:
  name: basic
spec:
  version: v3.0.7
  imagePullPolicy: IfNotPresent
  pvReclaimPolicy: Delete
  schedulerName: "tidb-scheduler"
  timezone: "Asia/Shanghai"
  pd:
    baseImage: pingcap/pd
    replicas: 3
    storageClassName: shared-nvme-disks
    requests:
      storage: "10Gi"
    config: {}
  tikv:
    baseImage: pingcap/tikv
    replicas: 3
    storageClassName: shared-nvme-disks
    requests:
      storage: "50Gi"
    config: {}
  tidb:
    baseImage: pingcap/tidb
    replicas: 2
    service:
      type: NodePort
    config: {}
---
apiVersion: pingcap.com/v1alpha1
kind: TidbMonitor
metadata:
  name: basic
spec:
  clusters:
  - name: basic
  persistent: false
  prometheus:
    baseImage: prom/prometheus
    version: v2.11.1
  grafana:
    baseImage: grafana/grafana
    version: 6.0.1
  initializer:
    baseImage: pingcap/tidb-monitor-initializer
    version: v3.0.5
  reloader:
    baseImage: pingcap/tidb-monitor-reloader
    version: v1.0.1
  imagePullPolicy: IfNotPresent
```

The namespace of these 2 resources are not determined until I actually apply them with

```
> kubectl apply -n my-ns -f ./
```

But this way I will not get the monitoring Pod get created because the `namespace` is empty -- therefore, the cluster referenced by the tidbmonitor could not be found. **I have to modify the yaml to set the namespace to "my-ns"** first, and actually I have to rollback this change after apply because I don't want the YAML definition to be coupled with specific namespace.

aylei:
@Yisaer PTAL

Yisaer:
fixed in https://github.com/pingcap/tidb-operator/pull/1500

PR #1500, opened 2020-01-07
Revise TidbMonitor controller 

Yisaer:
1. fix the bug for the log level of the Prometheus Spec in TidbMonitor which would cause TidbMonitor Failed to start. https://github.com/pingcap/tidb-operator/issues/1482

2. Add default namespace for the`ClusterRef` if it's empty in TidbMonitor Spec. https://github.com/pingcap/tidb-operator/issues/1480

3. Add events for TidbMonitor Controller for a more user-friendly experience. https://github.com/pingcap/tidb-operator/issues/1481

```release-note
- Add the feature of adding default ClusterRef namespace for TidbMonitor as same as it deployed
- Fix the bug when TidbMonitor's Pod can't be created when Spec.PrometheusSpec.logLevel is missing
```


Yisaer:
As https://github.com/pingcap/tidb-operator/pull/1445 add e2e test for `TidbMonitor`, we could wait until https://github.com/pingcap/tidb-operator/pull/1445 are merged.

aylei:
please add a release note

Yisaer:
@aylei @cofyc  PTAL

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -22,46 +22,73 @@ import (
...
+		message := fmt.Sprintf("Sync TidbMonitor[%s/%s] failed, it's going to be deleted", monitor.Namespace, monitor.Name)
+		mm.recorder.Event(monitor, corev1.EventTypeNormal, FailedSync, message)
 		return nil
 	}
 
+	setDefaultNamespaceForClusterRef(monitor)
+
 	// Sync Service
 	if err := mm.syncTidbMonitorService(monitor); err != nil {
+		message := fmt.Sprintf("Sync TidbMonitor[%s/%s] Service failed,err:%v", monitor.Namespace, monitor.Name, err)
```
aylei:
```suggestion
		message := fmt.Sprintf("Sync TidbMonitor[%s/%s] Service failed, err: %v", monitor.Namespace, monitor.Name, err)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -22,46 +22,73 @@ import (
...
+		mm.recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, message)
 		return err
 	}
+	mm.recorder.Event(monitor, corev1.EventTypeNormal, SuccessSync, fmt.Sprintf("Sync TidbMonitor[%s/%s] Service Success", monitor.Namespace, monitor.Name))
```
aylei: I suggest not recording successful sync to reduce verbosity, also, note that there is a global event rate-limit of tidb-controller-manager

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -22,46 +22,73 @@ import (
...
 	klog.V(4).Infof("tm[%s/%s]'s service synced", monitor.Namespace, monitor.Name)
 	// Sync PVC
 	if monitor.Spec.Persistent {
 		if err := mm.syncTidbMonitorPVC(monitor); err != nil {
+			message := fmt.Sprintf("Sync TidbMonitor[%s/%s] PVC failed,err:%v", monitor.Namespace, monitor.Name, err)
+			mm.recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, message)
 			return err
 		}
 	}
+	mm.recorder.Event(monitor, corev1.EventTypeNormal, SuccessSync, fmt.Sprintf("Sync TidbMonitor[%s/%s] PVC Success", monitor.Namespace, monitor.Name))
```
aylei: ditto

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -22,46 +22,73 @@ import (
...
 	klog.V(4).Infof("tm[%s/%s]'s pvc synced", monitor.Namespace, monitor.Name)
+
 	// Sync Deployment
-	return mm.syncTidbMonitorDeployment(monitor)
+	if err := mm.syncTidbMonitorDeployment(monitor); err != nil {
+		message := fmt.Sprintf("Sync TidbMonitor[%s/%s] Deployment failed,err:%v", monitor.Namespace, monitor.Name, err)
+		mm.recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, message)
+		return err
+	}
+	mm.recorder.Event(monitor, corev1.EventTypeNormal, SuccessSync, fmt.Sprintf("Sync TidbMonitor[%s/%s] Deployment Success", monitor.Name, monitor.Name))
```
aylei: ditto

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -182,3 +209,13 @@ func (mm *MonitorManager) syncTidbMonitorRbac(monitor *v1alpha1.TidbMonitor) (*c
 
 	return sa, nil
 }
+
+// If the namespace in ClusterRef is empty, we would set the TidbMonitor's namespace in the default
+func setDefaultNamespaceForClusterRef(tm *v1alpha1.TidbMonitor) {
```
aylei: Please do not mutate spec in controller, spec is user intent and by definition, is owned by user

Yisaer: updated.

pkg/monitor/monitor/monitor_manager.go
Yisaer: updated, all successful sync wouldn't be recorded.

pkg/monitor/monitor/util.go
```diff
@@ -729,3 +734,10 @@ func getMonitorPVC(monitor *v1alpha1.TidbMonitor) *core.PersistentVolumeClaim {
 		},
 	}
 }
+
+// If the namespace in ClusterRef is empty, we would set the TidbMonitor's namespace in the default
+func setDefaultNamespaceForClusterRef(model *MonitorConfigModel, tm *v1alpha1.TidbMonitor) {
```
cofyc: I think we don't need to create a new function for this

Yisaer: updated.

Yisaer:
> By the way, can we add some basic tests for functions in `pkg/monitor/monitor`? No tests yet right now.

Sure, and issued in https://github.com/pingcap/tidb-operator/issues/1554.

aylei:
BTW, it is recommended to separate PR to logic complete unit, so that each feature/fix could be independently rollbacked or cherry-picked, also, the release note will be easier to write.

pkg/monitor/monitor/monitor_manager.go
cofyc: the good practice is to record a successful event once the operation is done successfully and skip if nothing needs to do because the actual state matches the desired state
in this case, the events of objects will be like:

- <optional warning events about failures>
- one successful event

cofyc: this pattern can be applied to the most scenarios that the desired state is often only changed by the reconciliation controller
we can improve it in the future

PR #1501, opened 2020-01-07
fix some broken urls

weekface:
### Check List 

Tests 

 - No code

Code changes

Side effects

Related changes

```release-note
NONE
```


Yisaer:
`docs-cn` uses [verify-link.sh`](https://github.com/pingcap/docs-cn/blob/master/scripts/verify-links.sh) to check whether there are any links unaccessible. Maybe we could add it into our CI checking process?

weekface:
I think it is not necessary to check automatically now, veryfy-link.sh use `markdown-link-check`, values.yaml is YAML format however.

PR #1504, opened 2020-01-08
Remove unnecessary informer caches

### LINKED ISSUE #1327: significant increase in network and memory footprint after upgrading to 1.1.x from 1.0.x
## Question

After upgrade tidb-operator from `v1.0.3` to the latest master (https://github.com/pingcap/tidb-operator/commit/a3f4b8dfb1bbda1aacecb53cce1852f8e3a61534), there is a significant increase in network and memory footprint:

image
image

cluster scale:
- Node: 60+
- TiDB clusters: 70+

### LINKED ISSUE #1406: just cache the k8s native resources that need concerned
## Feature Request

Currently in the tidb-operator, a lot of k8s native resources are watched to the local cache, such as configmap, secret, pod, etc. When there are many such resources in the k8s cluster, this will cause the tidb-operator to consume too much memory. But in fact, the operator only pays attention to a small part of these resources. 

We can add SharedInformerOption to filter resources managed by tidb-operator like follow:
```go
kubeoptions := []kubeinformers.SharedInformerOption{
	kubeinformers.WithNamespace(ns),
	kubeinformers.WithTweakListOptions(
	func(listOptions *metav1.ListOptions) {
		listOptions.LabelSelector = fmt.Sprintf("%s=%s", label.ManagedByLabelKey, label.TiDBOperator)
		}),
	}
kubeInformerFactory = kubeinformers.NewSharedInformerFactoryWithOptions(kubeCli, controller.ResyncDuration, kubeoptions...)
```

https://github.com/pingcap/tidb-operator/blob/51b682016b580790d6c19a733494de91c5979bf5/cmd/controller-manager/main.go#L135-L148

**Breaking backward compatibility**
At present, most k8s native resources related to the tidb-operator have this label `app.kubernetes.io/managed-by: tidb-operator`. If this function is enabled, the `tidb-secret`(used by tidb-initializer-job) and `ceph-secret`(used by Backup)  or other resources created manually before need to be labeled `app.kubernetes.io/managed-by: tidb-operator`.

If we don't want to breaking backward compatibility, for these resources created outside of the tidb-operator,  we can obtain from the apiserver instead of the local cache.

aylei:

close #1327 
close #1406 

- Remove informer caches for configmap and secret

These will significantly reduce the memory footprint of tidb-controller-manager.

### Check List 

Tests 

 - Unit test
 - E2E test

```release-note
Remove unnecessary informer caches to reduce the memory footprint of tidb-controller-manager
```


onlymellb:
I think the configmaps can still be obtained by using the cache because we can control all the configmaps that are managed by tidb-operator to have a corresponding filter label. @aylei

aylei:
> by

Yes, we did not cache configmap and secrets for two reasons:

- changing of configmaps or secrets do not trigger reconcile
- their size (and numbers, because of configmap rollouts) is more likely to be the bottleneck

cofyc:
If in regular non-op (nothing needs to do because the actual state matches the desired state) reconciliation loops, we don't need to get `secrets/configMaps` from the API server. I think we don't need to list/watch them.

cmd/controller-manager/main.go
```diff
@@ -135,20 +137,23 @@ func main() {
...
-		options := []informers.SharedInformerOption{
-			informers.WithNamespace(ns),
-		}
-		informerFactory = informers.NewSharedInformerFactoryWithOptions(cli, controller.ResyncDuration, options...)
-
-		kubeoptions := []kubeinformers.SharedInformerOption{
-			kubeinformers.WithNamespace(ns),
-		}
-		kubeInformerFactory = kubeinformers.NewSharedInformerFactoryWithOptions(kubeCli, controller.ResyncDuration, kubeoptions...)
+	labelSelector := fmt.Sprintf("%s=%s", label.ManagedByLabelKey, label.TiDBOperator)
```
cofyc: why this is added? I guess we need to list/watch all our CRD objects with or without this label.

aylei: good catch

Yisaer:
@aylei  Is this PR still WIP and Does it need to be merged now?

Issue #1512, opened 2020-01-08
Follow istio's service port requirements


PaulGrandperrin:
Hi,
I'm trying to configure an HTTP reverse proxy to expose the Grafana service.
This reverse proxy uses istio which requires to set a name so that it can know about the protocol used on top of TCP:
https://istio.io/docs/setup/additional-setup/requirements/

Without this, the service is unusable as-is from istio and requires us to fork the service definition and add `name: http` in it.

It would be very useful for all Istio users to have the `name` property correctly set in all exposed services.

I already opened a similar issue for the elastic operator if you want some context:
https://github.com/elastic/cloud-on-k8s/issues/2069

**What version of TiDB Operator are you using?**
checked on v1.0.6 and v1.1.0-alpha.4

aylei:
Good catch.

BTW, is there any standard or de facto standard of service naming ? The name should be configurable if not, otherwise we just follow the standards.

Yisaer:
AFAIK, only Istio need this config requirements. I agree with configurable port name even without considering the istio requirements.

PaulGrandperrin:
I agree with @Yisaer but that would be nice to have a default name that works out of the box with Istio.
For instance, right now, the grafana service defines:
`name: grafana`
which could be changed to
`name: http-grafana`
to follow Istio guidelines/requirements

Anyway, thanks for your work, your operator is awesome :-)

aylei:
Agreed, @Yisaer would you like to take this?

Yisaer:
Sure @aylei

Yisaer:
merged in #1521  @PaulGrandperrin

PR #1521, opened 2020-01-09
Make service port-name configurable in monitor

Yisaer:
https://github.com/pingcap/tidb-operator/issues/1512
Make service port-name configurable in the monitor (Both TidbMonitor and monitor config in `values.yaml`).

### Check List 

Tests 

 - Manual test


charts/tidb-cluster/values.yaml
```diff
@@ -473,6 +473,7 @@ monitor:
     imagePullPolicy: IfNotPresent
     service:
       type: NodePort
+      portName: reloader
```
aylei: Shall we make the default on follow the requirements of istio?

pkg/monitor/monitor/util.go
```diff
@@ -681,6 +681,13 @@ func getMonitorService(monitor *v1alpha1.TidbMonitor) []*core.Service {
 			},
 		},
 	}
+	if monitor.Spec.Prometheus.Service.PortName != nil {
+		prometheusService.Spec.Ports[0].Name = *monitor.Spec.Prometheus.Service.PortName
```
aylei: Please assign the port name to the specific port object instead of to an index

pkg/monitor/monitor/util.go
```diff
@@ -681,6 +681,13 @@ func getMonitorService(monitor *v1alpha1.TidbMonitor) []*core.Service {
...
+	}
+	if monitor.Spec.Reloader.Service.PortName != nil {
+		reloaderService.Spec.Ports[0].Name = *monitor.Spec.Reloader.Service.PortName
```
aylei: ditto

pkg/monitor/monitor/util.go
```diff
@@ -707,6 +714,9 @@ func getMonitorService(monitor *v1alpha1.TidbMonitor) []*core.Service {
 				},
 			},
 		}
+		if monitor.Spec.Grafana.Service.PortName != nil {
+			grafanaService.Spec.Ports[0].Name = *monitor.Spec.Grafana.Service.PortName
```
aylei: ditto

charts/tidb-cluster/values.yaml
Yisaer: update.

pkg/monitor/monitor/util.go
Yisaer: update.

pkg/monitor/monitor/util.go
Yisaer: update

pkg/monitor/monitor/util.go
Yisaer: update

Yisaer:
> what about the PD and TiDB services?

I think we could create a new issue as `good first issue` for community contributors to finish other service port name config work. @aylei

pkg/monitor/monitor/util.go
```diff
@@ -709,6 +728,15 @@ func getMonitorService(monitor *v1alpha1.TidbMonitor) []*core.Service {
...
 		}
+		if monitor.Spec.Grafana.Service.PortName != nil {
+			for id, port := range grafanaService.Spec.Ports {
+				if port.Name == "http-grafana" {
+					port.Name = *monitor.Spec.Grafana.Service.PortName
+					grafanaService.Spec.Ports[id] = port
+					break
+				}
+			}
+		}
```
aylei: This is redundant, why not use the desired port name at the first place?

```
grafanaPortName := "http-grafana"
if monitor.Spec.Grafana.Service.PortName != nil {
  grafanaPortName := *monitor.Spec.Grafana.Service.PortName
}
...
Name:      grafanaPortName,
...
```

pkg/monitor/monitor/util.go
```diff
@@ -638,6 +638,21 @@ func getMonitorService(monitor *v1alpha1.TidbMonitor) []*core.Service {
 	var services []*core.Service
 	monitorLabel := label.New().Instance(monitor.Name).Monitor()
 	labels := label.NewMonitor().Instance(monitor.Name).Monitor()
+
+	reloaderPortName := "tcp-reloader"
+	prometheusPortName := "http-prometheus"
+	grafanaPortName := "http-grafana"
+
+	if monitor.Spec.Reloader.Service.PortName != nil {
```
cofyc: what if `PortName` is non-nil but empty, should we use the default name in that case?
cc @aylei do you recommend to use `*string` or `string`?

Yisaer: > what if `PortName` is non-nil but empty, should we use the default name in that case?
> cc @aylei do you recommend to use `*string` or `string`?

I think operator don't need to consider this case, if monitor service synced failed because of setting wrong by users, users should be responsible for it.

aylei: It is about our purpose. For kubernetes, port name is optional and could be empty, so the current code is okay. If we think empty is not a valid value, then we could consider empty as `unset` and use `string`, otherwise, just keep `*string`

aylei: I recommend add an access method for each Spec to get the port name to encapsulate the details of port name calculation.

Yisaer: updated.

pkg/apis/pingcap/v1alpha1/tidbmonitor.go
```diff
@@ -0,0 +1,29 @@
...
+func (tm *TidbMonitor) GrafanaPortName() *string {
+	if tm.Spec.Grafana != nil {
+		return tm.Spec.Grafana.Service.PortName
+	}
+	return nil
+}
+
+func (tm *TidbMonitor) ReloaderPortName() *string {
+	return tm.Spec.Reloader.Service.PortName
+}
```
cofyc: move all default values here, and in controller logic we can simply use `TidbMonitor. (PrometheusPortName| GrafanaPortName| ReloaderPortName)`? like https://github.com/pingcap/tidb-operator/blob/d314b251735693a000088d7011f9a5b2932f579f/pkg/manager/member/pump_member_manager.go#L373-L379

PR #1562, opened 2020-01-15
Check bash version in e2e test

### LINKED ISSUE #1560: e2e test doesn't work with default version of bash on macOS
## Feature Request

When run `./hack/e2e.sh` with the default bash on macOS (version 3.2), there's a error message:
```bash
./hack/e2e.sh: line 136: declare: -A: invalid option
```

It is because bash v3.2 doesn't support `declare  -A`. 

Either the developer should upgrade their bash or some shell script which compatible with bash v3.2 should be used here.

Use a method which also works in bash v3.2 or describe this in doc.

longfangsong:
Add bash version requirements note in doc.
fix #1560

Check bash version in e2e test script.
It add bash version requirements note in doc.

`e2e.sh` and `CONTRIBUTIN.md`.

In `e2e.sh` , we check the `$BASH_VERSION` and make sure it >= 4 before really execute anything which requires bash newer than version 4.

```release-note
NONE
```

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -131,6 +131,9 @@ Now you can run the following command to run all e2e test.
 $ ./hack/e2e.sh
 ```
 
+**NOTE**: We don't support bash version < 4 for now. For those who using a not supported version of bash, especially macOS (which default bash version is 3.2)
+users, please use containerized environment or upgrade your bash manually.
+
 It's possible to limit specs to run, for example:
 
 ```
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -132,6 +132,15 @@ echo "DOCKER_IO_MIRROR: $DOCKER_IO_MIRROR"
 echo "GCR_IO_MIRROR: $GCR_IO_MIRROR"
 echo "QUAY_IO_MIRROR: $QUAY_IO_MIRROR"
 
+# check bash version
+BASH_MAJOR_VERSION=$(echo "$BASH_VERSION" | grep -o "^\d\{1,3\}")
+# we need bash version >= 4
+if [ $BASH_MAJOR_VERSION -lt 4 ]
+then
+  echo "error: e2e.sh could not work with bash version earlier than 4 for now, please upgrade your bash"
+  exit 1
+fi
+
 # https://github.com/kubernetes-sigs/kind/releases/tag/v0.6.1
 declare -A kind_node_images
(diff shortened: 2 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -131,6 +131,9 @@ Now you can run the following command to run all e2e test.
 $ ./hack/e2e.sh
 ```
 
+**NOTE**: We don't support bash version < 4 for now. For those who using a not supported version of bash, especially macOS (which default bash version is 3.2)
+users, please use containerized environment or upgrade your bash manually.
```
cofyc:
```suggestion
users, please run `hack/run-in-container.sh` to start a containerized environment or install bash 4+ manually.
```

docs/CONTRIBUTING.md
```diff
@@ -131,6 +131,9 @@ Now you can run the following command to run all e2e test.
 $ ./hack/e2e.sh
 ```
 
+**NOTE**: We don't support bash version < 4 for now. For those who using a not supported version of bash, especially macOS (which default bash version is 3.2)
```
cofyc:
```suggestion
**NOTE**: We don't support bash version < 4 for now. For those who are using a not supported version of bash, especially macOS (which default bash version is 3.2)
```

hack/e2e.sh
```diff
@@ -132,6 +132,15 @@ echo "DOCKER_IO_MIRROR: $DOCKER_IO_MIRROR"
 echo "GCR_IO_MIRROR: $GCR_IO_MIRROR"
 echo "QUAY_IO_MIRROR: $QUAY_IO_MIRROR"
 
+# check bash version
+BASH_MAJOR_VERSION=$(echo "$BASH_VERSION" | grep -o "^\d\{1,3\}")
```
cofyc:
```suggestion
BASH_MAJOR_VERSION=$(echo "$BASH_VERSION" | cut -d '.' -f 1)
```

cofyc: because it's pre-check of the script, can you move this check after the `source "${ROOT}/hack/lib.sh"`? 
no need to run any logic if the bash is not the version after 4.

longfangsong: @cofyc even `./hack/e2e.sh --help`?

cofyc: it makes sense to allow `./hack/e2e.sh -h`, you can update #1565 

by the way, normally you don't need to create another PR to submit new changes.  you can push the branch again to update the PR (add `-f` flag to force it if necessary or pull again to merge remote change)

PR #1565, opened 2020-01-15
Check bash version for hack/e2e.sh

### LINKED ISSUE #1560: e2e test doesn't work with default version of bash on macOS

longfangsong:
Add bash version requirements note in doc.
fix #1560

Check bash version in e2e test script.
It add bash version requirements note in doc.

`e2e.sh` and `CONTRIBUTIN.md`.

In `e2e.sh` , we check the `$BASH_VERSION` and make sure it >= 4 before really execute anything.

```release-note
NONE
```

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -132,6 +132,15 @@ echo "DOCKER_IO_MIRROR: $DOCKER_IO_MIRROR"
 echo "GCR_IO_MIRROR: $GCR_IO_MIRROR"
 echo "QUAY_IO_MIRROR: $QUAY_IO_MIRROR"
 
+# check bash version
+BASH_MAJOR_VERSION=$(echo "$BASH_VERSION" | grep -o "^\d\{1,3\}")
+# we need bash version >= 4
+if [ $BASH_MAJOR_VERSION -lt 4 ]
+then
+  echo "error: e2e.sh could not work with bash version earlier than 4 for now, please upgrade your bash"
+  exit 1
+fi
+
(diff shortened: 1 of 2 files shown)


hack/e2e.sh
```diff
@@ -26,6 +26,16 @@ cd $ROOT
 
 source "${ROOT}/hack/lib.sh"
 
+# check bash version
+BASH_MAJOR_VERSION=$(echo "$BASH_VERSION" | cut -d '.' -f 1)
+# we need bash version >= 4
```
Yisaer: Could we merge these 2 notes into on note?

longfangsong: The first one explain the usage of the following "block" and the second one explain the usage of the following "statement".

But it is not that obvious so I think it is ok to merge.

cofyc:
https://github.com/pingcap/tidb-operator/pull/1562#discussion_r367204433

cofyc:
Thanks for the fix! This PR can guide people in local development.
If you think it's better to allow `./hack/e2e.sh -h` to run or want to improve comments, etc., feel free to submit another PR.

Issue #1567, opened 2020-01-16
PD Pod cannot be deleted


DanielZhangQD:
## Bug Report

**What version of Kubernetes are you using?**

v1.12.8
**What version of TiDB Operator are you using?**

tidb-operator:bfa1983847315df80b4c3f5d88f340de16018aa2
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

local-storage
**What's the status of the TiDB cluster pods?**

Running
**What did you do?**

Scale in PD
**What did you expect to see?**
PD is scaled in successfully
**What did you see instead?**
PD is not scaled in

DanielZhangQD:
Here is the log from kube-controller-manager:
```
E0116 02:12:38.461930       1 stateful_set.go:400] Error syncing StatefulSet tiflash-dt3/tiflash-dt-cluster3-pd, requeuing: admission webhook "delete.podadmission.tidb.pingcap.com" denied the request: persistentvolumeclaims "tikv-tiflash-dt-cluster3-pd-3" not found
I0116 02:12:38.462072       1 event.go:221] Event(v1.ObjectReference{Kind:"StatefulSet", Namespace:"tiflash-dt3", Name:"tiflash-dt-cluster3-pd", UID:"8022e2d1-3786-11ea-8799-eeeeeeeeeeee", APIVersion:"apps/v1", ResourceVersion:"392308714", FieldPath:""}): type: 'Warning' reason: 'FailedDelete' delete Pod tiflash-dt-cluster3-pd-3 in StatefulSet tiflash-dt-cluster3-pd failed error: admission webhook "delete.podadmission.tidb.pingcap.com" denied the request: persistentvolumeclaims "tikv-tiflash-dt-cluster3-pd-3" not found
```

Here is the log from admission controller:
```
I0116 02:02:38.387886       1 pods.go:88] receive DELETE pod[tiflash-dt3/tiflash-dt-cluster3-pd-3] by sa[system:serviceaccount:kube-system:statefulset-controller]
I0116 02:02:38.387916       1 pods.go:114] receive admission to delete pod[tiflash-dt3/tiflash-dt-cluster3-pd-3]
[...]

PR #1568, opened 2020-01-16
Fix pd deleting admission webhook bug

Yisaer:
https://github.com/pingcap/tidb-operator/issues/1567
Fix PD deleting admission webhook bug and add unit tests to cover this case.
I also found there were some bugs in the previous unit test and I fixed it in this request.

```release-note
* Fix the bug for admission webhook which cause pd pod deleting error
* Permit the deleting pod request for pd and tikv when PVC is not found in admission webhook
```


pkg/webhook/pod/pd_deleter.go
```diff
@@ -121,13 +122,16 @@ func (pc *PodAdmissionControl) admitDeleteNonPDMemberPod(payload *admitPayload)
 		// it would be existed an pd-3 instance with its deferDeleting label Annotations PVC.
 		// And the pvc can be deleted during upgrading if we use create pod webhook in future.
 		if !isInOrdinal {
-			pvcName := operatorUtils.OrdinalPVCName(v1alpha1.TiKVMemberType, payload.ownerStatefulSet.Name, ordinal)
+			pvcName := operatorUtils.OrdinalPVCName(v1alpha1.PDMemberType, payload.ownerStatefulSet.Name, ordinal)
 			pvc, err := pc.kubeCli.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, meta.GetOptions{})
```
DanielZhangQD: Need to check `notFound` error here.

cofyc:
can we add some test cases to cover these scenarios? it helps to verify the change is good to merge.

pkg/webhook/pod/pd_deleter.go

Yisaer:
> can we add some test cases to cover these scenarios? it helps to verify the change is good to merge.

Sure, I would add more e2e cases to cover it.

cofyc:
no, unit tests in this PR please

cofyc:
it's best practice to add tests to avoid the bug we fixed in each PR. In this pkg, you already wrote some test cases already, we can extend it to cover new scenarios. client-go fake clientset can do most basic logic testing.

Yisaer:
@cofyc  Add unit test for `PVCNoutFound` case.

tests/e2e/tidbcluster/serial.go
```diff
@@ -351,6 +352,18 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
 			oa.CheckTidbClusterStatusOrDie(&cluster)
 			oa.CleanTidbClusterOrDie(&cluster)
 		})
+
+		ginkgo.It("[PodAdmissionWebhook] delete pvc before deleting pod directly with admission webhook", func() {
```
aylei: what is the purpose of this case?

Yisaer: https://github.com/pingcap/tidb-operator/issues/1567
In this issue, we found that the admission webhook sever denied the deleting request because of failure to find its `pvc` during scale-in.
We think the request should be permitted during scale-in when `pvc` is not found.

Yisaer: I also will add this case into e2e test to cover it.

cofyc: revert this unfinished change in this PR? we can submit a separate PR to address https://github.com/pingcap/tidb-operator/issues/1569 if necessary.

PR #1572, opened 2020-01-16
Refine local kind setup

aylei:

Makes local kind script compliant with kind 0.7.0, support specifying helm and kind version through environments

### Check List 

Tests 

Build a kind cluster locally and it works fine

```release-note
NONE
```


hack/kind-cluster-build.sh
```diff
@@ -22,20 +22,32 @@ source $ROOT/hack/lib.sh
...
+want to mutate the defaulting one:
+
+    export KUBECONFIG=${HOME}/.kube/kind-config-<cluster>
 
 Options:
        -h,--help               prints the usage message
        -n,--name               name of the Kubernetes cluster,default value: kind
        -c,--nodeNum            the count of the cluster nodes,default value: 6
        -k,--k8sVersion         version of the Kubernetes cluster,default value: v1.12.8
        -v,--volumeNum          the volumes number of each kubernetes node,default value: 9
```
cofyc: should we reduce the default num of the cluster too? 1 control plane and 3 workers are enough for most cases.

hack/kind-cluster-build.sh
```diff
@@ -232,19 +245,19 @@ $HELM_BIN init --service-account=tiller --wait
 # FIXME: remove this
 docker pull gcr.io/google-containers/kube-scheduler:${k8sVersion}
 docker tag gcr.io/google-containers/kube-scheduler:${k8sVersion} mirantis/hypokube:final
-kind load docker-image --name=${clusterName} mirantis/hypokube:final
+$KIND_BIN load docker-image --name=${clusterName} mirantis/hypokube:final
```
cofyc: I guess this is not necessary right now. I'm fine to remove this in this PR or another PR.

aylei: I'd like to remove it later because it requires additional test

hack/kind-cluster-build.sh
aylei: good point, changed

PR #1573, opened 2020-01-16
add basic yaml deployment example

aylei:
Signed-off-by: Aylei <rayingecho@gmail.com>

### What problem does this PR solve? 
This is blocked by https://github.com/pingcap/tidb-operator/issues/1478, we could merge it after #1478 solved
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has Go code change
 - Has CI related scripts change
 - Has Terraform scripts change

Side effects

 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

### Does this PR introduce a user-facing change?:

```release-note
Add a basic deployment example of TiDB cluster
```

PR #1581, opened 2020-01-18
Able to omit storageClassName to use the default storage class of Kubernetes

### LINKED ISSUE #1478: Able to omit `storageClassName`
## Feature Request

It could be much easier for user to get started if the storageClassName could be omitted in which case the default storageClass of the cluster will be used as the default. However, currently `tidb-operator` will enforce the storageClassName being set.

`tidb-operator` should be able to set the `defaultStorageClassName` as "omitted" (`nil`), so that the `TidbCluster` with `storageClassName` omitted could use the cluster defaults. 

ref: #1211

cofyc:
Signed-off-by: Yecheng Fu <fuyecheng@pingcap.com>

fixes #1478 

I've manually tested with the following YAML:

```
kind: TidbCluster
metadata:
  name: basic
spec:
  version: v3.0.8
  timezone: UTC
  pvReclaimPolicy: Delete
  pd:
    baseImage: pingcap/pd
    replicas: 3
    requests:
      storage: "1Gi"
    config: {}
  tikv:
    baseImage: pingcap/tikv
    replicas: 3
    requests:
      storage: "1Gi"
    config: {}
  tidb:
    baseImage: pingcap/tidb
    replicas: 2
    service:
      type: ClusterIP
    config: {}
```

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
ACTION REQUIRED: `--default-storage-class-name` and `--default-backup-storage-class-name `are abandoned, storage class defaults to Kubernetes default storage class right now. If you have set default storage class different than Kubernetes default storage class, please set them explicitly in your tidb cluster helm or YAML files.
```


pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -222,9 +222,8 @@ type PDSpec struct {
 	// +optional
 	Service *ServiceSpec `json:"service,omitempty"`
 
-	// The storageClassName of the persistent volume for PD data storage, empty string means not explicitly set
-	// and use the cluster default set by admission controller.
-	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator
+	// The storageClassName of the persistent volume for PD data storage.
+	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator or Kubernetes default storage classs.
```
cofyc: We shouldn't set default value for `storageClassName` in defaulting phase. 

`nil` storageClassName in PVC is a valid value that indicates the default storage class of the cluster to be used.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -261,9 +260,8 @@ type TiKVSpec struct {
 	// +optional
 	MaxFailoverCount *int32 `json:"maxFailoverCount,omitempty"`
 
-	// The storageClassName of the persistent volume for TiKV data storage, empty string means not explicitly set
-	// and use the cluster default set by admission controller.
-	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator
+	// The storageClassName of the persistent volume for TiKV data storage.
+	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator or Kubernetes default storage classs.
```
cofyc: ditto

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -343,9 +341,8 @@ type PumpSpec struct {
 	// +optional
 	BaseImage string `json:"baseImage"`
 
-	// The storageClassName of the persistent volume for Pump data storage, empty string means not explicitly set
-	// and use the cluster default set by admission controller.
-	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator
+	// The storageClassName of the persistent volume for Pump data storage.
+	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator or Kubernetes default storage classs.
```
cofyc: ditto

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -744,8 +741,10 @@ type BackupSpec struct {
 	Type BackupType `json:"backupType,omitempty"`
 	// StorageProvider configures where and how backups should be stored.
 	StorageProvider `json:",inline"`
-	// StorageClassName is the storage class for backup job's PV.
-	StorageClassName string `json:"storageClassName,omitempty"`
+	// The storageClassName of the persistent volume for Backup data storage.
+	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator or Kubernetes default storage classs.
+	// +optional
+	StorageClassName *string `json:"storageClassName,omitempty"`
```
cofyc: apply the same change from https://github.com/pingcap/tidb-operator/pull/1409 to backup/restore spec

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -866,8 +865,10 @@ type BackupScheduleSpec struct {
 	MaxReservedTime *string `json:"maxReservedTime,omitempty"`
 	// BackupTemplate is the specification of the backup structure to get scheduled.
 	BackupTemplate BackupSpec `json:"backupTemplate"`
-	// StorageClassName is the storage class for backup job's PV.
-	StorageClassName string `json:"storageClassName,omitempty"`
+	// The storageClassName of the persistent volume for Backup data storage.
+	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator or Kubernetes default storage classs.
+	// +optional
```
cofyc: ditto

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -866,8 +865,10 @@ type BackupScheduleSpec struct {
...
+	StorageClassName *string `json:"storageClassName,omitempty"`
```
cofyc: ditto

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -944,8 +945,10 @@ type RestoreSpec struct {
 	Type BackupType `json:"backupType,omitempty"`
 	// StorageProvider configures where and how backups should be stored.
 	StorageProvider `json:",inline"`
-	// StorageClassName is the storage class for backup job's PV.
-	StorageClassName string `json:"storageClassName"`
+	// The storageClassName of the persistent volume for Restore data storage.
+	// Optionals: Defaults to the default-storage-class-name set in the tidb-operator or Kubernetes default storage classs.
+	// +optional
+	StorageClassName *string `json:"storageClassName,omitempty"`
```
cofyc: ditto

pkg/manager/member/pd_member_manager.go
```diff
@@ -588,7 +588,7 @@ func getNewPDSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap) (
 	podAnnotations := CombineAnnotations(controller.AnnProm(2379), basePDSpec.Annotations())
 	stsAnnotations := getStsAnnotations(tc, label.PDLabelVal)
 	storageClassName := tc.Spec.PD.StorageClassName
-	if storageClassName == nil {
+	if storageClassName == nil && len(controller.DefaultStorageClassName) > 0 {
```
cofyc: if `tc.Spec.PD.StorageClassName` is not set, try to use the default storage class of tidb-operator. If the default value of tidb-operator is not set, try to use the default storage class of Kubernetes cluster.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -409,7 +409,7 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 	capacity := controller.TiKVCapacity(tc.Spec.TiKV.Limits)
 	headlessSvcName := controller.TiKVPeerMemberName(tcName)
 	storageClassName := tc.Spec.TiKV.StorageClassName
-	if storageClassName == nil {
+	if storageClassName == nil && len(controller.DefaultStorageClassName) > 0 {
```
cofyc: ditto

charts/tidb-operator/values.yaml
```diff
@@ -15,11 +15,28 @@ timezone: UTC
...
-defaultStorageClassName: local-storage
+#
+# Specify storage class for pd/tikv componnents. 
+#
+# By default, "standard" is used. If you want to use default storage class in
+# the Kubernetes cluster, set its value to "".
+#
+# https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/
+# 
+# defaultStorageClassName: standard
```
cofyc: - use user-supplied class if provided
- use tidb-operator default class if not-empty (defaults to "standard" for backward compatibility)
- use Kubernetes cluster default class

charts/tidb-operator/values.yaml
```diff
@@ -15,11 +15,28 @@ timezone: UTC
...
+
+#
+# Specify storage class for backup and restore. 
+#
+# By default, "standard" is used. If you want to use default storage class in
+# the Kubernetes cluster, set its value to "".
+#
+# https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/
+# 
+# defaultBackupStorageClassName: standard
```
cofyc: - use user-supplied class if provided
- use tidb-operator default class if not-empty (defaults to "standard" for backward compatibility)
- use Kubernetes cluster default class

charts/tidb-operator/values.yaml
cofyc: if we want to use the default storage class of Kubernetes cluster **by default**, we need to empty the default values of `--default-storage-class-name` and `--default-backup-storage-class-name`.

cc @aylei what do you think?

Yisaer:
I'm not sure is it reasonable to use kubernetes cluster default storageClass instead of throwing the error to mention the users to set the proper storageClass when both `Operator` and `TidbCluster` storageClassName are not setted. @cofyc @aylei

cofyc:
the same YAML can be applied to different clusters that may have different default storage implementation. it helps the user to try [our example YAML](https://github.com/pingcap/tidb-operator/pull/1573) in any Kubernetes provider without modification.

If users have not special requirements, they don't need to care about what storage implementation the default storage class is. It's a system administrator's duty to maintain it. Unlike the network, the cluster can have more than one storage providers.

cofyc:
actually, we have a hard-coded default storage class `standard`...
if we want to apply YAML to a cluster in which the default class is not "standard", we need to remove this hard-coded value, see [discussion](https://github.com/pingcap/tidb-operator/pull/1581/files#r368210912).

charts/tidb-operator/values.yaml
aylei: Yes, I think so

pkg/backup/util/util.go
```diff
@@ -294,7 +294,7 @@ func ValidateBackup(backup *v1alpha1.Backup) error {
 		if backup.Spec.From.SecretName == "" {
 			return fmt.Errorf("missing tidbSecretName config in spec of %s/%s", ns, name)
 		}
-		if backup.Spec.StorageClassName == "" {
+		if backup.Spec.StorageClassName == nil || *backup.Spec.StorageClassName == "" {
```
aylei: why this is still mandatory?

charts/tidb-operator/values.yaml
aylei: Looks like that this change is not backward compatible.

Thinks a user upgrade operator with defaultStorageClassName empty, the storageClass of current PVC and PVCTemplate will be synced to the cluster default, which breaks reconciliation because storageClassName is immutable IIRC

aylei:
Having a default storage class name in controller is problematic, even if we avoid mutating the storage-class when update resources, there is still a problem when we delete the StatefulSet with orphan dependents and the controller eventually re-create the StatefulSet with a different storageClass.

IMHO:

- To be backward compatible, we should at least do not mutate storageClass when update, deleting with orphan dependents is rare and could be documented as an known issue if it happens with storageClass name change at the same time;
- For better sematic, the defaultStorageClassName should works like the default storageClass name like kubernetes, i.e., set the storageClass in mutating webhook, of course, the operator one should be prior to the kubernetes one

aylei:
And I think this PR deserves a release note, maybe ACTION REQUIRED

cofyc:
It's better to abandon --default-storage-class-name and --default-backup-storage-class-name flags... what do you think?

The desired PVC spec should not depend on something which is not in CRDs and can be changed.

cofyc:
In tidb-operator 1.0, I guess most people who use tidb-cluster helm chart should have set storage class names in their value files. 
We can add an `ACTION REQUIRED` release note to ask people to set storage class names in CRD objects explicitly if they didn't and the default storage class of the Kubernetes cluster is not the storage class they're using.

If we really want to allow people to configure a operator-level default value for storage class names, we should update the user object in the defaulting phase before storing it in the API server.

cofyc:
cc @weekface what do you think?

aylei:
I agree, users (or their organization) should make sure the stored TidbCluster CR has storageClass full-filled, either through kubernetes defaults, operator webhook defaults or other mechanisms.

And I agree we should deprecate the default-storage-class-name flag of tidb-controller-manager. 

It is an orthogonal problem for tidb-operator chart, if we want to keep offering the operator level defaults in v1.1.0, we need a webhook based implementation.

weekface:
> cc @weekface what do you think?

Agree.

People who use `tidb-cluster` helm chart have set storage class names in their value files.

cofyc:
> I agree, users (or their organization) should make sure the stored TidbCluster CR has storageClass full-filled, either through Kubernetes defaults, operator webhook defaults or other mechanisms.

After the change I suggested, TidbCluster CR can have no storage class filled if they want to use Kubernetes default storage class and this will not trigger the upgrade of tidb cluster in tidb-operator because the desired volume claim templates in statefulset do not have storage class filled too.

If the system administrator changed the default storage class in the cluster, the storage class of the PVC created by statefulset controller or our backup/restore controller will not be changed until the PVC is deleted by the user. The newly created PVC will use the new default storage class. I think it's reasonable and safe.

cofyc:
ah, I was mistaken about this.

cofyc:
I've commented on the wrong page (the comment have been deleted)

cofyc:
@aylei @weekface  PR is updated, PTAL again.

PR #1601, opened 2020-01-22
automatically fix files not end with end-of-line

### LINKED ISSUE #1488: automated update script to add EOF if it does not exist
## Feature Request

now, we have [a script](https://github.com/pingcap/tidb-operator/pull/1412) to verify our files must end with EOF, but we don't have a script to add it automatically. 

EOF can be added automatically, we can have an automated update script to do this.

This helps people to fix the issue soon.

longfangsong:
fix #1488 
Now we can use `fix-EOF.sh` instead of `check-EOF.sh` to check and fix files not end with end-of-line.
### Check List 

Code changes

```release-note
NONE
```

### DIFF
--- a/hack/fix-EOF.sh
+++ b/hack/fix-EOF.sh
@@ -30,23 +30,16 @@ FILELIST=($(find . -type f -not \( -path './vendor/*' \
     -o -path './tests/images/*/bin/*' \
     -o -path '*.png' \
     -o -path './tkctl' \
+    -o -path './.idea/*' \
+    -o -path './.DS_Store' \
+    -o -path './*/.DS_Store' \
     \)))
-
-NUM=0
-declare FAILED_FILE
 
 for f in ${FILELIST[@]}; do
     c=$(tail -c 1 "$f" | wc -l)
     if [ "$c" -eq 0 ]; then
-        FAILED_FILE+=($f)
-        NUM=$((NUM + 1))
+        echo "find file $f do not end with newline, fixing it"
+        # echo will add the end-of-line automatically
+        echo "$(cat $f)" > $f
     fi
 done
-
-if [ $NUM -ne 0 ]; then
-    echo "error: following files do not end with newline, please fix them"
-    printf '%s\n' "${FAILED_FILE[@]}"
-    exit 1
-else
-    echo "all files pass checking."
-fi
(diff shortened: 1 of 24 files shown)


longfangsong:
PS. Shall we make a git hook and run this automatically before commit?

hack/fix-EOF.sh
```diff
@@ -30,23 +30,16 @@ FILELIST=($(find . -type f -not \( -path './vendor/*' \
     -o -path './tests/images/*/bin/*' \
     -o -path '*.png' \
     -o -path './tkctl' \
+    -o -path './.idea/*' \
+    -o -path './.DS_Store' \
+    -o -path './*/.DS_Store' \
```
longfangsong: Maybe it would be better to get these ignored paths via .gitignore?

Yisaer:
I suggest not to remove `check-EOF.sh` because we still need it to guarantee each request should be ended with `EOF`.
What's your opinion? @cofyc

cofyc:
don't remove hack/check-EOF.sh which runs in our verify phase in CI to make sure every PR is valid to merge.

please 1) rename `hack/fix-EOF.sh` to `hack/update-EOF.sh` and 2) add `hack/update-EOF.sh` into `hack/update-all.sh`.

you can also view our current verify/update scripts to see how they work. verify scripts are mandatory, update scripts are optional but useful in developer daily development.

hack/fix-EOF.sh
```diff
@@ -30,23 +30,16 @@ FILELIST=($(find . -type f -not \( -path './vendor/*' \
...
-declare FAILED_FILE
-
 for f in ${FILELIST[@]}; do
     c=$(tail -c 1 "$f" | wc -l)
     if [ "$c" -eq 0 ]; then
-        FAILED_FILE+=($f)
-        NUM=$((NUM + 1))
+        echo "find file $f do not end with newline, fixing it"
+        # echo will add the end-of-line automatically
+        echo "$(cat $f)" > $f
```
cofyc: we can add newline directly by `printf "\n" >> $f`

hack/fix-EOF.sh
cofyc: or we can use `git ls-files` to list files in git repo only

cofyc:
> PS. Shall we make a git hook and run this automatically before commit?

not necessary, because we run verify (check) scripts in CI to check every PR. if developers forgot to fix the EOF, the CI will fail. btw, you can update the `hack/check-EOF.sh` to tell the developer to use update script to fix EOF, e.g. https://github.com/pingcap/tidb-operator/blob/11f5b949a62a268cd08047157d1680628fb1705f/hack/verify-codegen.sh#L43-L46

later, we can add a note in our contributor doc to suggest the developer to run `hack/verify-all.sh` locally before submitting a PR. of course, this is optional because they can wait for our CI to run it.

PR #1604, opened 2020-01-29
make the drainer name configurable

gregwebs:
This is used for the statefulset/pod names.

The release name is already unique, so I would actually suggest just using that
without the cluster name. However, that is a backwards incompatible
change that I hesitate to make.

Consider this more of an issue than a pull request. I haven't tested this yet.

Code changes

- chart

Related changes

```release-note
Allow drainerName to be configured in the tidb-drainer chart.
```


gregwebs:
Right now this is backwards compatible so I don't know what you are suggesting.

charts/tidb-drainer/values.yaml
```diff
@@ -5,6 +5,10 @@
 # timezone is the default system timzone
 timezone: UTC
 
+# Change the name of the statefulset and pod
+# The default is clusterName-ReleaseName-drainer
```
aylei: Can we warn user that changing name on existing will cause data loss?

cofyc: cc @gregwebs add a notice then we merge this?

gregwebs: okay, warning added

Issue #1582, opened 2020-01-19
Check store owner before tikv manager failover


Yisaer:
## Bug Report

**What did you do?**

Trying to `PodChaos` for `TidbCluster` with `tiflash`. The `PodChaos` would delete `tiflash` pod randomly and periodically. As `tiflash` store become `Down`, the `tikv memeber manager` in `Operator` found this problem and trigger tikv failover to create new tikv pods which it shouldn't currently. We should check the store owner during failover and only let tikv can failover now.

**What did you expect to see?**
`tiflash` become `down` which won't affect tikv failover.

**What did you see instead?**
tikv failover due to `tiflash` store state.

PR #1621, opened 2020-02-04
Check store owner for tikv manager

Yisaer:
Fixes https://github.com/pingcap/tidb-operator/issues/1582

Currently, `tikv-manager` would sync all the store info in the `tikv status`, which is not suitable if we add `tiflash` in the future. In this request, the `tikv-manager` would check the store owner by comparing the address of each store. 

### Check List 

Tests 

Code changes

 - Has Go code change

```release-note
NONE
```

### DIFF
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -590,9 +590,8 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 	addressPrefix := fmt.Sprintf("%s-tikv-", tc.Name)
 	addressSuffix := fmt.Sprintf("%s-tikv-peer.%s.svc:20160", tc.Name, tc.Namespace)
 	for _, store := range storesInfo.Stores {
-		address := store.Store.Address
 		// make sure that the store belongs to the tikv
-		if !(strings.HasPrefix(address, addressPrefix) && strings.HasSuffix(address, addressSuffix)) {
+		if store.Store != nil && !(strings.HasPrefix(store.Store.Address, addressPrefix) && strings.HasSuffix(store.Store.Address, addressSuffix)) {
 			continue
 		}
 		status := tkmm.getTiKVStore(store)
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -588,12 +588,13 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 		return err
 	}
 
-	pattern, err := regexp.Compile(fmt.Sprintf("%s-tikv-\\d+\\.%s-tikv-peer\\.%s\\.svc\\:\\d+", tc.Name, tc.Name, tc.Namespace))
+	pattern, err := regexp.Compile(fmt.Sprintf(`%s-tikv-\d+\.%s-tikv-peer\.%s\.svc\:\d+`, tc.Name, tc.Name, tc.Namespace))
 	if err != nil {
 		return err
 	}
 	for _, store := range storesInfo.Stores {
-		// make sure that the store belongs to the tikv
+		// In theory, the external tikv can join the cluster, and the operator would only manage the internal tikv.
+		// So we check the store owner to make sure it.
 		if store.Store != nil && !pattern.Match([]byte(store.Store.Address)) {
 			continue
 		}
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -16,6 +16,7 @@ package member
 import (
 	"fmt"
 	"reflect"
+	"regexp"
 	"strings"
 
 	"github.com/pingcap/kvproto/pkg/metapb"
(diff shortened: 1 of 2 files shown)


pkg/manager/member/tikv_member_manager.go
```diff
@@ -587,7 +587,13 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 		return err
 	}
 
+	addressPrefix := fmt.Sprintf("%s-tikv-", tc.Name)
+	addressSuffix := fmt.Sprintf("%s-tikv-peer.%s.svc:20160", tc.Name, tc.Namespace)
 	for _, store := range storesInfo.Stores {
+		// make sure that the store belongs to the tikv
+		if store.Store != nil && !(strings.HasPrefix(store.Store.Address, addressPrefix) && strings.HasSuffix(store.Store.Address, addressSuffix)) {
+			continue
```
cofyc: - we should not hard code `20160` port here. although we didn't support it yet, the port number is configurable in theory, `net.SplitHostPort` can help to extract the host from the address
- I suggest using the regex to do the checking, e.g. check if the host matches the pattern `<cluster>-tikv-\d+-tikv-peer.<namespace>.svc`,

cofyc: by the way, do tiflash and tikv in the same cluster have other properties we can use to distinguish them?

Yisaer: > by the way, do tiflash and tikv in the same cluster have other properties we can use to distinguish them?

I think the address might be the best property we can use to distinguish them. Here is the store info example for the tidbcluster.
```json
{
      "store": {
        "id": 6,
        "address": "tidb-tikv-0.tidb-tikv-peer.tidb-admin.svc:20160",
        "labels": [
          {
            "key": "host",
            "value": "docker-desktop"
          }
        ],
        "version": "3.0.8",
        "state_name": "Up"
      },
      "status": {
        "capacity": "465.6GiB",
        "available": "187.5GiB",
        "leader_count": 7,
        "leader_weight": 1,
        "leader_score": 7,
        "leader_size": 7,
        "region_count": 21,
        "region_weight": 1,
        "region_score": 21,
        "region_size": 21,
        "start_ts": "2020-02-04T05:43:43Z",
        "last_heartbeat_ts": "2020-02-04T05:51:13.2362298Z",
        "uptime": "7m30.2362298s"
      }
    }

```

Yisaer: > * we should not hard code `20160` port here. although we didn't support it yet, the port number is configurable in theory, `net.SplitHostPort` can help to extract the host from the address
> * I suggest using the regex to do the checking, e.g. check if the host matches the pattern `<cluster>-tikv-\d+\.<cluster>-tikv-peer\.<namespace>\.svc`,

updated.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -587,7 +588,15 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 		return err
 	}
 
+	pattern, err := regexp.Compile(fmt.Sprintf("%s-tikv-\\d+\\.%s-tikv-peer\\.%s\\.svc\\:\\d+", tc.Name, tc.Name, tc.Namespace))
+	if err != nil {
+		return err
+	}
 	for _, store := range storesInfo.Stores {
+		// make sure that the store belongs to the tikv
```
cofyc: in theory, the external tikv can join the cluster like tiflash, and I guess we should ignore them too, we should explain the reason in detail here, e.g. external tiflash/tikv can join the cluster but they are not under operator management.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -587,7 +588,15 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 		return err
 	}
 
+	pattern, err := regexp.Compile(fmt.Sprintf("%s-tikv-\\d+\\.%s-tikv-peer\\.%s\\.svc\\:\\d+", tc.Name, tc.Name, tc.Namespace))
```
cofyc: raw literal is preferred here, `%s-tikv-\d+\.%s-tikv-peer\.%s\.svc\:\d+`

pkg/manager/member/tikv_member_manager.go
JaySon-Huang: For tiflash, we have "engine=tiflash" in store.labels.
But not guarantee for other situation such as "external tikv"

```
    {
      "store": {
        "id": 44,
        "address": "172.16.5.85:9512",
        "labels": [
          {
            "key": "tiflash_http_port",
            "value": "4512"
          },
          {
            "key": "engine",
            "value": "tiflash"
          }
        ],
        "version": "3.1.0-beta",
        "peer_address": "172.16.5.85:9012",
        "state_name": "Up"
      },
      "status": {
        "capacity": "1.063TiB",
        "available": "1.063TiB",
        "used_size": "42MiB",
        "leader_weight": 1,
        "region_count": 1,
        "region_weight": 1,
        "region_score": 1,
        "region_size": 1,
        "start_ts": "2020-02-27T13:15:29+08:00",
        "last_heartbeat_ts": "2020-02-27T13:25:59.825627093+08:00",
        "uptime": "10m30.825627093s"
      }
    }
```

Yisaer: Cool, this would be pretty helpful.

Issue #1651, opened 2020-02-07
[Feature] AutoScaling For TidbCluster


Yisaer:
## Feature Request

**Describe the feature you'd like:**
Operator is going to support `Auto-scaling` feature for `TiKV` and `TiDB` in `TidbCluster`. 

`Auto-scaling` would help the Operator users to auto-scale out/in for the `TiKV` and `TiDB` for the `TiDBCluster` by the metrics / value / resources which the users could provide. This issue is to discuss and track the whole process of the `Auto-scaling` design and realization.

**Describe alternatives you've considered:**

First, I think there are some prerequisites to the `Auto-Scaling` for `TidbClusters`:

- The Customized Ability for the Auto-scaling Algorithm.  **(must to have)**  `TidbCluster` is sensitive to the scaling as the distributed databases, we should have the ability to control the whole auto-scaling process.
- The Metrics configurations should be extendable. **(must to have)**  Currently, the `Auto-scaling` need the `TidbCluster` metrics info to decide the recommended numbers of TiKV and TiDB.  In the future, the `Platform` information (Kubernetes) or the external global metrics/values are also necessary.
- The Interval duration Control of the Auto-scaling should be Cluster Level. **(nice to have)** `Operator` could manager several Clusters in one Kubernetes. We should provide the cluster level control ability and Interval duration is also important to avoid performance jitter.

Yisaer:
We welcome everyone to help to realize the `Auto-scaling` feature in Operator by discuss/suggest/code review/pull request/ etc.

Yisaer:
Currently, the auto-scaling is under alpha feature. 
It only provides the ability below:
1. basic auto-scaling Availability
2. basic guarantee to avoid jitter during auto-scaling.

To achieve Production ready:
1. control the scaling step for auto-scaling
2. record the consecutive count timestamp in auto-scaling

Yisaer:
Here describes How the auto-scaling algorithm work by average CPU load.

#1722

Yisaer:
After #1731 merged, we would have auto-scaling ability based by cpu load feature under alpha stage.
Currently, there are still plenty of jobs to do:

1.  Add Syncing TidbClusterAutoCluster Status
2. Add proper events and logs
3. Design for noise reduction for auto-scaling process
4. Support specific pd label for the auto-scaling out tikv instances.

Yisaer:
[...]

PR #1672, opened 2020-02-10
Add Controller for TidbClusterAutoScale

Yisaer:
#1651 
Add Controller for TidbClusterAutoScaler and make it controlled by feature gate.

```release-note
NONE
```


pkg/controller/autoscaler/tidbcluster_autoscaler_control.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type defaultAutoScalerControl struct {
+	recoder      record.EventRecorder
+	typedControl controller.TypedControlInterface
+	//	autoScalerManager
+}
+
+func (tac *defaultAutoScalerControl) ResconcileAutoScaler(ta *v1alpha1.TidbClusterAutoScaler) error {
+	var errs []error
```
DanielZhangQD: An array is necessary here?

Yisaer: Not necessary and not important.

charts/tidb-operator/values.yaml
```diff
@@ -34,6 +34,7 @@ imagePullPolicy: IfNotPresent
 features: []
 # - AdvancedStatefulSet=false
 # - StableScheduling=true
+# - AutoScaling=true
```
cofyc: use default value `false`

pkg/features/features.go
```diff
@@ -41,6 +41,9 @@ const (
 
 	// AdvancedStatefulSet controls whether to use AdvancedStatefulSet to manage pods
 	AdvancedStatefulSet string = "AdvancedStatefulSet"
+
+	// AutoScaling controls whether to use TidbClusterAutoScaler to auto scale-in/out pods
+	AutoScaling string = "AutoScaling"
```
cofyc: add `AutoScaling: false` here

https://github.com/pingcap/tidb-operator/blob/0624cfe4694533ea9dca682811a4e4b7856cfb06/pkg/features/features.go#L30-L33

Yisaer: updated.

charts/tidb-operator/values.yaml
Yisaer: updated.

PR #1692, opened 2020-02-13
deploy: close connection after set privilege

shuijing198799:
close connection after set privilege

### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change


tennix:
@shuijing198799 This should also be cherry-picked to release-1.0 branch. And for master and release-1.1, the helm chart logic already migrated to the controller, so we should also fix this in the controller.

PR #1703, opened 2020-02-14
Add consecutive count check for Auto-scaling

Yisaer:
ref #1651 

**Noteice**
As `Auto-scaling` is under alpha feature, the first target is to achieve basic auto-scaling Availability.
The extra control to avoid jitter for production-ready is not considered in this request. These could be recorded in this issue first. 
https://github.com/pingcap/tidb-operator/issues/1651

Add 3 new configuration for Auto-scaling API and default them them Controller.
1. MetricsTimeDuration which is used in the Prometheus Query
 2. ScaleOutThreshold 
 3. ScaleInThreshold

ScaleOutThreshold/ScaleInThreshold is used to check the consecutive count in auto-scaling control to avoid jitter.

I also add the unit test for the current sync logic for tidb and tikv and the newly added function in `util.go`

In this request, each auto-scaling reconciliation would record the consecutive count in the target tidbclusterAutoScaler annotations marked as *ScaleOutCount* / *ScaleInCount*

For example:
In one tikv auto-scaling reconciliation, 
if recommended replicas > current replicas, then:
> tikvScaleOutCount = tikvScaleOutCount + 1; 
    tikvScaleInCount = 0;

if recommended replicas < current replicas, then:
> tikvScaleOutCount = 0; 
   tikvScaleInCount= tikvScaleInCount +1;

if recommended replicas == current replicas, then:
> tikvScaleOutCount = 0; 
   tikvScaleInCount= 0;

When the *ScaleOutCount/ *ScaleInCount is bigger than the defined property in autoscaler api, the auto-scaling would be performed and the record would be empty and recount.

```release-note
NONE
```


pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -112,6 +112,26 @@ type BasicAutoScalerSpec struct {
...
+	// MetricsTimeWindowSeconds describe the time window seconds for the metrics
+	// to be queried in the Prometheus.
+	// If not set, the default value would be 180.
+	// +optional
+	MetricsTimeWindowSeconds *int32 `json:"metricsTimeWindowSeconds,omitempty"`
+
+	// ScaleOutThreshold describe the consecutive threshold for the auto-scaling,
+	// if the consecutive counts of the scale-out result in auto-scaling reach this number,
+	// the auto-scaling would be performed.
+	// If not set, the default value is 1 which means it would perform with no threshold.
```
onlymellb: I think the default value should be conservative, so it might be better to set the default value to 3?

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -112,6 +112,26 @@ type BasicAutoScalerSpec struct {
...
+	// +optional
+	ScaleOutThreshold *int32 `json:"scaleOutThreshold,omitempty"`
+
+	// ScaleInThreshold describe the consecutive threshold for the auto-scaling,
+	// if the consecutive counts of the scale-int result in auto-scaling reach this number,
+	// the auto-scaling would be performed.
+	// If not set, the default value is 1 which means it would perform with no threshold.
+	// +optional
+	ScaleInThreshold *int32 `json:"scaleInThreshold,omitempty"`
```
onlymellb: ditto

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
Yisaer: It's ok to me.

pkg/label/label.go
```diff
@@ -102,6 +102,15 @@ const (
 	// AnnTiKVLastAutoScalingTimestamp is annotation key of tidbclusterto which ordinal is created by tikv auto-scaling
 	AnnTiKVLastAutoScalingTimestamp = "tikv.tidb.pingcap.com/last-autoscaling-timestamp"
 
+	// AnnTiDBConsecutiveScaleOutCount describes the least consecutive count to scale-out for tidb
```
weekface: least or largest?

Yisaer: least, the recorded consecutive count should be bigger than `AnnTiDBConsecutiveScaleOutCount` to perform auto-scaling out

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -112,6 +112,26 @@ type BasicAutoScalerSpec struct {
 	// If not set, the default metric will be set to 80% average CPU utilization.
 	// +optional
 	Metrics []v2beta2.MetricSpec `json:"metrics,omitempty"`
+
+	// MetricsTimeWindowSeconds describe the time window seconds for the metrics
+	// to be queried in the Prometheus.
+	// If not set, the default value would be 180.
+	// +optional
+	MetricsTimeWindowSeconds *int32 `json:"metricsTimeWindowSeconds,omitempty"`
```
cofyc: I prefer using the time duration here, e.g. metav1.Duration.

> Syntactically, a range duration is appended in square brackets ([]) at the end of a vector selector to specify how far back in time values should be fetched for each resulting range vector element. ...

https://prometheus.io/docs/prometheus/latest/querying/basics/#range-vector-selectors

Yisaer: updated.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -112,6 +112,24 @@ type BasicAutoScalerSpec struct {
...
+
+	// ScaleOutThreshold describe the consecutive threshold for the auto-scaling,
+	// if the consecutive counts of the scale-out result in auto-scaling reach this number,
+	// the auto-scaling would be performed.
+	// If not set, the default value is 3.
+	// +optional
+	ScaleOutThreshold *int32 `json:"scaleOutThreshold,omitempty"`
+
+	// ScaleInThreshold describe the consecutive threshold for the auto-scaling,
+	// if the consecutive counts of the scale-int result in auto-scaling reach this number,
```
DanielZhangQD:
```suggestion
	// if the consecutive counts of the scale-in result in auto-scaling reach this number,
```

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -112,6 +112,24 @@ type BasicAutoScalerSpec struct {
...
+	// the auto-scaling would be performed.
+	// If not set, the default value is 3.
```
DanielZhangQD: We should be more conservative for scaling in, default to a bigger value, e.g. 5?

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,151 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+			tac.Spec.TiKV.ScaleOutThreshold = pointer.Int32Ptr(3)
+		}
+		if tac.Spec.TiKV.MetricsTimeDuration == nil {
+			tac.Spec.TiKV.MetricsTimeDuration = pointer.StringPtr("3m")
+		}
+		if tac.Spec.TiKV.ScaleOutIntervalSeconds == nil {
+			tac.Spec.TiKV.ScaleOutIntervalSeconds = pointer.Int32Ptr(300)
+		}
+		if tac.Spec.TiKV.ScaleInIntervalSeconds == nil {
+			tac.Spec.TiKV.ScaleInIntervalSeconds = pointer.Int32Ptr(300)
```
DanielZhangQD: larger value for scaling in?

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,151 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+
+	targetScaleOutAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleOutSuffix)
+	targetScaleInAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleInSuffix)
+
+	var scaleOutCount int
+	var scaleInCount int
+	scaleOutCount, scaleInCount = 0, 0
+	var err error
+
+	if v, existed := tc.Annotations[targetScaleOutAnn]; existed {
```
DanielZhangQD: If store the count in annotation, e.g. the threshold is 3, after it increases to 2, operator is down, then after 10 minutes, operator is up and then the result is scaling (even if the resource utilization is below the resource usage threshold during the 10 minutes that the operaotor is down), then the scaling will be done.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,151 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+			tac.Spec.TiDB.ScaleOutIntervalSeconds = pointer.Int32Ptr(300)
+		}
+		if tac.Spec.TiDB.ScaleInIntervalSeconds == nil {
+			tac.Spec.TiDB.ScaleInIntervalSeconds = pointer.Int32Ptr(300)
+		}
+	}
+}
+
+// updateConsecutiveCount would update the tc annotation depended by the given replicas in each reconciling
+func updateConsecutiveCount(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler,
```
DanielZhangQD: Why update the annotation of tc instead of tac?

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,151 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+	default:
+		return false, fmt.Errorf(invalidMemberTypeErrorMsg, tac.Namespace, tac.Name, memberType)
+	}
+	return true, nil
+}
+
+func emptyConsecutiveCount(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType) {
+	targetScaleOutAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleOutSuffix)
+	targetScaleInAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleInSuffix)
+	tc.Annotations[targetScaleOutAnn] = "0"
```
DanielZhangQD: Check nil of annotation?

pkg/label/label.go
```diff
@@ -102,6 +102,15 @@ const (
...
+	AnnTiDBConsecutiveScaleOutCount = "tidb.tidb.pingcap.com/consecutive-scale-out-count"
+	// AnnTiDBConsecutiveScaleInCount describes the least consecutive count to scale-in for tidb
+	AnnTiDBConsecutiveScaleInCount = "tidb.tidb.pingcap.com/consecutive-scale-in-count"
```
DanielZhangQD: Can we store these fields in the status of tac?
What's the benefit with annotation? And why do we put them in tc insead of tac?

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -112,6 +112,26 @@ type BasicAutoScalerSpec struct {
...
+	// MetricsTimeWindowSeconds describe the time window seconds for the metrics
+	// to be queried in the Prometheus.
+	// If not set, the default value would be 180.
+	// +optional
+	MetricsTimeWindowSeconds *int32 `json:"metricsTimeWindowSeconds,omitempty"`
+
+	// ScaleOutThreshold describe the consecutive threshold for the auto-scaling,
+	// if the consecutive counts of the scale-out result in auto-scaling reach this number,
+	// the auto-scaling would be performed.
+	// If not set, the default value is 3.
```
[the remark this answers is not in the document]
Yisaer: Nor necessary, users could modify the threshold to be suitable for their Prometheus.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
Yisaer: There is no necessary that the default value of scale-in threshold should be bigger than scale-out's. 3 is big enough for both of them.

pkg/autoscaler/autoscaler/util.go
Yisaer: ditto.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -24,38 +24,59 @@ import (
...
 	}
-	tc.Spec.Annotations[label.AnnTiDBLastAutoScalingTimestamp] = time.Now().String()
-	tc.Spec.TiDB.Replicas = targetReplicas
+	updateTcTiDBAnnIfScale(tc)
+	tc.Spec.TiDB.Replicas = recommendedReplicas
 	return nil
 }
+
+func updateTcTiDBAnnIfScale(tc *v1alpha1.TidbCluster) {
+	tc.Annotations[label.AnnTiDBLastAutoScalingTimestamp] = time.Now().String()
```
DanielZhangQD: Suggest not updating tc, HPA records the time in status:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#horizontalpodautoscalerstatus-v1-autoscaling

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -24,38 +24,59 @@ import (
...
+	emptyConsecutiveCount(tc, v1alpha1.TiDBMemberType)
```
DanielZhangQD: Should we empty the count after scaling? What if the resource utilization continues to be scaling in or out, in this case, we should just scale after xxInterval, right?

pkg/label/label.go
Yisaer: If we stored the state in the tac annotation or status, there would be lots of caring about the tc status in tac controller logic. For example, changing the `tidbcluster Ref` or deleting the `TidbCluster`. More over, if I delete an existed tac, and recreated it soon, the auto-scaling status wouldn't be affected if we store is in Tc Annotations.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -24,38 +24,59 @@ import (
...
 	if err != nil {
 		return err
 	}
 	if !ableToScale {
 		return nil
 	}
-	tc.Spec.Annotations[label.AnnTiDBLastAutoScalingTimestamp] = time.Now().String()
-	tc.Spec.TiDB.Replicas = targetReplicas
+	updateTcTiDBAnnIfScale(tc)
+	tc.Spec.TiDB.Replicas = recommendedReplicas
```
DanielZhangQD: We'd better also have step control, we do not want to scale out many Pods in one time.

Yisaer: The step Control was controlled in `TidbCluster controller` or admission Webhooks.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: The reason why I choose store the auto-scaling status in TC anotations is answered before.
And I think there is no conflict between storing the timestamp in TC Annotations or TAC status. The Source of meta info In TAC Status could be TC Annotations.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: We should empty the count after scaling. If the recommended replicas continue to be changed after scaling, This is just a next round of auto-scaling.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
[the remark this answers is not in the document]
DanielZhangQD: For example, Prom scrapes the metrics in 1-minute interval, the data in Prom will be 0,1,2,..., If user configures the window 5m, scaleout threshold 3, auto scaler controller sync interval 30s, then at 04:20 (check metrics of 0-4), 04:50 (check metrics of 0-4), 05:20 (check metrics of 1-5), then it will scale, is this expected?

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
DanielZhangQD: It depends, check the third comment of https://github.com/pingcap/tidb-operator/pull/1703/files?file-filters%5B%5D=.go#r379976699, is it big enough?

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -24,38 +24,59 @@ import (
 
 func (am *autoScalerManager) syncTiDB(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler, client promClient.Client) error {
 	if tac.Spec.TiDB == nil {
+		emptyConsecutiveCount(tc, v1alpha1.TiDBMemberType)
```
DanielZhangQD: This is just in case? If no config for TiDB, the count will not be populated, right?

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -24,38 +24,59 @@ import (
...
 		return nil
 	}
 	sts, err := am.stsLister.StatefulSets(tc.Namespace).Get(operatorUtils.GetStatefulSetName(tc, v1alpha1.TiDBMemberType))
 	if err != nil {
 		return err
 	}
 	if !checkAutoScalingPrerequisites(tc, sts, v1alpha1.TiDBMemberType) {
+		emptyConsecutiveCount(tc, v1alpha1.TiDBMemberType)
```
DanielZhangQD: Why empty here? Even if the prerequisites are not met, the resource utilization result is correct.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
DanielZhangQD: Or should remote the anno in this case?

pkg/autoscaler/autoscaler/tidb_autoscaler.go
DanielZhangQD: You mean the Pods will be created one by one with sts controller? What I mean is that we may just increase the replicas with 2 or some reasonable number instead of 10 during one scale.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -24,38 +24,72 @@ import (
...
+	targetReplicas := calculateRecommendedReplicas(tac, v1alpha1.TiKVMemberType, client)
 	targetReplicas = limitTargetReplicas(targetReplicas, tac, v1alpha1.TiKVMemberType)
 	if targetReplicas == tc.Spec.TiKV.Replicas {
+		emptyConsecutiveCount(tc, v1alpha1.TiKVMemberType)
+		return nil
+	}
+	return syncTiKVAfterCalculated(tc, tac, currentReplicas, targetReplicas, tc.Spec.TiKV.Replicas-currentReplicas)
+}
+
+// syncTiDBAfterCalculated would check the Consecutive count to avoid jitter, and it would also check the interval
```
DanielZhangQD:
```suggestion
// syncTiKVAfterCalculated would check the Consecutive count to avoid jitter, and it would also check the interval
```

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -24,38 +24,59 @@ import (
...
+
+// syncTiDBAfterCalculated would check the Consecutive count to avoid jitter, and it would also check the interval
+// duration between each auto-scaling. If either of them is not meet, the auto-scaling would be rejected.
+// If the auto-scaling is permitted, the timestamp would be recorded and the Consecutive count would be zeroed.
+func syncTiDBAfterCalculated(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler, currentReplicas, recommendedReplicas int32) error {
+	if err := updateConsecutiveCount(tc, tac, v1alpha1.TiDBMemberType, currentReplicas, recommendedReplicas); err != nil {
+		return err
+	}
+
+	ableToScale, err := checkConsecutiveCount(tc, tac, v1alpha1.TiDBMemberType, currentReplicas, recommendedReplicas)
```
DanielZhangQD: `currentReplicas ` is got from `tc`.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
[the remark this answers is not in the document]
Yisaer: First thing first, we can't configure the Prometheus scrape configure in operator, right? And the factors you listed (time window / threshold / resync duration ) are all configured by users, not calculated by operators. 

So when the auto-scaling performed actually is when the auot-scaling performed expected.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -24,38 +24,72 @@ import (
...
 	if err != nil {
 		return err
 	}
 	if !ableToScale {
 		return nil
 	}
-	tc.Spec.Annotations[label.AnnTiKVLastAutoScalingTimestamp] = time.Now().String()
-	tc.Spec.TiKV.Replicas = targetReplicas
+	updateTcTiKVAnnIfScale(tc)
+	tc.Spec.TiKV.Replicas = recommendedReplicas + failureStoreCount
```
DanielZhangQD: failover function will add failure stores again.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
Yisaer: updated.

pkg/autoscaler/autoscaler/util.go
Yisaer: updated.

pkg/label/label.go
Yisaer: updated.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: The spec of AutoScaler config for Tidb could be emptied from an existing status.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -24,38 +24,72 @@ import (
...
 	return nil
 }
+
+func getStateUpReplicas(tc *v1alpha1.TidbCluster) int32 {
```
DanielZhangQD: What if the status is not updated in time?
It may not be enough to just calculate all the metrics and get the upstores from status, suggest to refer to the calculation in HPA.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: If the prerequisites are not met, the recorded count should be emptied or it can't b called as `consecutive threshold`

pkg/autoscaler/autoscaler/tikv_autoscaler.go
Yisaer: The further plan is to change the way for fetching store info by pdClient. Currently, using tc status instead.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
DanielZhangQD: `consecutive` means resource usage crosses the threshold continuously, why we empty the count for resource usage just because of upgrading or scaling?

pkg/autoscaler/autoscaler/util.go
Yisaer: This is a good catch, I suggest to record is as an issue in order to guarantee the production-ready stability.  For an alpha feature, the first thing we need to ensure is the available for auto-scaler.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -112,6 +112,24 @@ type BasicAutoScalerSpec struct {
 	// If not set, the default metric will be set to 80% average CPU utilization.
 	// +optional
 	Metrics []v2beta2.MetricSpec `json:"metrics,omitempty"`
+
+	// MetricsTimeDuration describe the Time duration to be queried in the Prometheus
```
DanielZhangQD: What's the format of this parameter? Any default value?

pkg/autoscaler/autoscaler/util.go
Yisaer: issued in https://github.com/pingcap/tidb-operator/issues/1651

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: issued in https://github.com/pingcap/tidb-operator/issues/1651

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,186 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+		if tac.Spec.TiKV.ScaleInThreshold == nil {
+			tac.Spec.TiKV.ScaleInThreshold = pointer.Int32Ptr(5)
+		}
+		if tac.Spec.TiKV.ScaleOutThreshold == nil {
+			tac.Spec.TiKV.ScaleOutThreshold = pointer.Int32Ptr(3)
+		}
+		if tac.Spec.TiKV.MetricsTimeDuration == nil {
+			tac.Spec.TiKV.MetricsTimeDuration = pointer.StringPtr("3m")
+		}
+		if tac.Spec.TiKV.ScaleOutIntervalSeconds == nil {
```
DanielZhangQD: Usually scaling in interval is bigger than that of scaling out.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
Yisaer: The format should be time.Duration validated. The check would be realized in calculation request.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
DanielZhangQD: ok

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: The future plan for `currentReplicas` is to get it from pdapi, so I expose it as function param.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -24,38 +24,59 @@ import (
...
-	//	// revive:disable:empty-block
-	//}
+	currentReplicas := tc.Spec.TiDB.Replicas
+	targetReplicas := calculateRecommendedReplicas(tac, v1alpha1.TiDBMemberType, client)
 	targetReplicas = limitTargetReplicas(targetReplicas, tac, v1alpha1.TiDBMemberType)
 	if targetReplicas == tc.Spec.TiDB.Replicas {
+		emptyConsecutiveCount(tc, v1alpha1.TiDBMemberType)
+		return nil
+	}
+	return syncTiDBAfterCalculated(tc, tac, currentReplicas, targetReplicas)
```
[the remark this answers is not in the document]
Yisaer: same as https://github.com/pingcap/tidb-operator/pull/1703#discussion_r380045512

pkg/autoscaler/autoscaler/util.go
Yisaer: updated.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
Yisaer: The recommended replicas is calculated by `currentReplicas` which is excluded the failure replicas. That's why we should add the failure replicas into recommended replicas.

Yisaer: updated.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: issued in https://github.com/pingcap/tidb-operator/issues/1651

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,186 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+		return false, fmt.Errorf(invalidMemberTypeErrorMsg, tac.Namespace, tac.Name, memberType)
+	}
+	return true, nil
+}
+
+func emptyAutoScalingCountAnn(tac *v1alpha1.TidbClusterAutoScaler, memberType v1alpha1.MemberType) {
+	targetScaleOutAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleOutSuffix)
+	targetScaleInAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleInSuffix)
+	if tac.Annotations == nil {
+		tac.Annotations = map[string]string{}
```
DanielZhangQD: `return` if nil?

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,186 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+		return
+	}
+	// If the name and namespace was set in annotations, compare it and update the annoataion
+	if len(name) > 0 && len(namespace) > 0 {
+		if name != tac.Spec.Cluster.Name || namespace != tac.Spec.Cluster.Namespace {
+			resetAutoScalingAnn(tac)
+		}
+		return
 	}
+	resetAutoScalingAnn(tac)
```
DanielZhangQD: Can this block be done like below?
```
if name== && namespace== {
  return
}
reset
```

pkg/label/label.go
```diff
@@ -102,6 +102,19 @@ const (
 	// AnnTiKVLastAutoScalingTimestamp is annotation key of tidbclusterto which ordinal is created by tikv auto-scaling
 	AnnTiKVLastAutoScalingTimestamp = "tikv.tidb.pingcap.com/last-autoscaling-timestamp"
 
+	// AnnTiDBConsecutiveScaleOutCount describes the least consecutive count to scale-out for tidb
+	AnnTiDBConsecutiveScaleOutCount = "tidb.tidb.pingcap.com/consecutive-scale-out-count"
+	// AnnTiDBConsecutiveScaleInCount describes the least consecutive count to scale-in for tidb
```
DanielZhangQD: The variables in this file still necessary?

pkg/autoscaler/autoscaler/util.go
Yisaer: I think not necessary,  update annotation value as "0" is also ok.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -126,13 +136,186 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+}
+
+func checkConsecutiveCount(tac *v1alpha1.TidbClusterAutoScaler,
+	memberType v1alpha1.MemberType, currentReplicas int32, recommendedReplicas int32) (bool, error) {
+	if currentReplicas == recommendedReplicas {
+		return false, nil
+	}
+	targetScaleOutAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleOutSuffix)
+	targetScaleInAnn := fmt.Sprintf("%s.%s", memberType.String(), annScaleInSuffix)
+	currentScaleOutCount, err := strconv.ParseInt(tac.Annotations[targetScaleOutAnn], 10, 32)
```
DanielZhangQD: Check nil for annotation?

pkg/autoscaler/autoscaler/util.go
Yisaer: Good Idea, updated.

pkg/label/label.go
Yisaer: Yes.

pkg/autoscaler/autoscaler/util.go
Yisaer: updated.

Yisaer:
Thanks all. Really huge code to review.

PR #1705, opened 2020-02-14
tidb-backup: restore respects resources, imagePullPolicy, nodeSelector

mightyguava:
# What problem does this PR solve? 

* Makes the templates for the restore job respect `resources` and `imagePullPolicy` from the `values.yaml`, making it mirror the backup job better.
* Add `nodeSelector` to both backup and restore.
* ~Fixes: the restore job expects a PVC, but doesn't attempt to create the PVC in
restore mode.~ Removed from PR

### DIFF
--- a/charts/tidb-backup/templates/backup-job.yaml
+++ b/charts/tidb-backup/templates/backup-job.yaml
@@ -33,6 +33,8 @@ spec:
       {{- if .Values.serviceAccount }}
       serviceAccount: {{ .Values.serviceAccount }}
       {{- end }}
+      nodeSelector:
+{{ toYaml .Values.nodeSelector | indent 8 }}
       containers:
       - name: backup
         image: {{ .Values.image.backup }}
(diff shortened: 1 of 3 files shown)


mightyguava:
I had to remove b3e0d04aeaacf51fff356e665ea75c13f61064ee from the PR due to failing tests https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/operator_ghpr_e2e_test_kind/detail/operator_ghpr_e2e_test_kind/2812/tests/.

It looks like for ad-hoc backup and restore (at least in the failing test), the helm chart expects the same pvc from the backup to be kept around and used for the subsequent restore. This only works if the source and target clusters are in the same namespace. For my use case, I'm trying to restore to a cluster running on a different Kubernetes cluster in a different region, via S3. I'll split this issue into a separate PR when I think more about how to modify the chart...

PR #1708, opened 2020-02-15
tidb-backup: add restoreUsingExistingVolume option

mightyguava:
It looks like for ad-hoc backup and restore (at least in the failing test), the helm chart expects the same pvc from the backup to be kept around and used for the subsequent restore. This only works if the source and target clusters are in the same namespace.

I would like to restore to a cluster without any local backups, running on a different Kubernetes cluster in a different region, via S3.

This is split from https://github.com/pingcap/tidb-operator/pull/1705

This adds a `restoreUsingExistingVolume` defaulting to `true`. If set to `false`, then a new volume in the `restore` mode to hold data loaded from s3, gcp, ceph, etc.

This option has the same effect as simply setting the mode to `scheduled-restore`, but I didn't really understand what `scheduled-restore` meant.

### DIFF
--- a/charts/tidb-backup/templates/backup-pvc.yaml
+++ b/charts/tidb-backup/templates/backup-pvc.yaml
@@ -1,8 +1,12 @@
-{{- if (or (eq .Values.mode "backup") (eq .Values.mode "scheduled-restore")) }}
+{{- if (or (eq .Values.mode "backup") (eq .Values.mode "scheduled-restore") (and (eq .Values.mode "restore") (not .Values.restoreUsingExistingVolume))) }}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
+  {{- if eq .Values.mode "restore" }}
+  name: restore-{{ tpl .Values.name . }}
+  {{- else }}
   name: {{ tpl .Values.name . }}
+  {{- end }}
   labels:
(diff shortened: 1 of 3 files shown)


gregwebs:
hmm, I don't think this should be necessary. But either way it is better to use the tidb-lightning chart. tidb-lightning with the tidb backend is preferred over the tidb-backup restore.

mightyguava:
> hmm, I don't think this should be necessary.

what is the alternative?

> But either way it is better to use the tidb-lightning chart.

@shenli recommended using DM for sub-TB and lightning for TB scale. DM has no chart so I was using the tidb-backup loader. What are the tradeoffs? I think we'll mostly be loading < 1TB of data.

tennix:
@mightyguava DM is for streaming migration, it continues running and synchronizes data from upstream. Lightning is for one-time importing job.

aylei:
I'd recommend tidb-lightning too, it is just like `loader` when using TiDB as the backend. But still this PR is an necessary complement for the `tidb-backup` chart, thanks!

onlymellb:
@mightyguava `scheduled-restore` means restore tidb cluster from a scheduled backup snapshot. I think we can remove this type and use `restore` instead. @weekface  PTAL

Issue #1709, opened 2020-02-16
Support user provided certificate for TiDB SQL server/client


tennix:
## Feature Request

Currently, the TLS certificate support in 1.1 can only use certificates generated automatically by K8s. The certificate path is hard-coded and auto-generated. Users may want to provide their own certificates for TiDB SQL server/client, especially if a company has its own CA.

Issue #1670, opened 2020-02-10
Certificate-based authentication for TiDB


tennix:
## Feature Request

TiDB support [certificate-based authentication](https://pingcap.com/docs/dev/reference/security/cert-based-authentication/). This enhances the security for TiDB compared to password-based authentication. We should support this in tidb-operator for better security.

tirsen:
FYI: This includes configuring the CA chain set and configuring the cert/key that tidb presents to the mysql clients. Btw, this may be (and likely is) different from the ca/cert/key it uses for internal TLS.

tennix:
Yes, I think once we support user-provided certificate to tidb SQL layer, we can have this feature.

tennix:
@weekface Is this already supported?

weekface:
Yes, https://github.com/pingcap/tidb-operator/pull/1867 added certificate support for TiDB.

PR #1714, opened 2020-02-17
Support user-defined tidb server/client certificate

weekface:
fixes: https://github.com/pingcap/tidb-operator/issues/1709
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Support user-defined tidb server/client certificate
```


charts/tidb-cluster/templates/_helpers.tpl
```diff
@@ -91,16 +91,20 @@ config-file: |-
...
     {{- end -}}
     {{- if .Values.enableTLSCluster }}
   cluster-ssl-ca = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
   cluster-ssl-cert = "/var/lib/tidb-tls/cert"
   cluster-ssl-key = "/var/lib/tidb-tls/key"
     {{- end -}}
-    {{- if .Values.tidb.enableTLSClient }}
+    {{- if .Values.tidb.tlsClient.enabled }}
+    {{- if .Values.tidb.tlsClient.userGenerated.secretName }}
+  ssl-ca = "/var/lib/tidb-server-tls/ca"
```
weekface: use user-defined ca file.

charts/tidb-cluster/templates/tidb-cluster.yaml
```diff
@@ -92,18 +92,9 @@ spec:
...
-    - {{ . }}
-    {{- end }}
-  {{- end }}
-  {{- if .Values.tidb.extraSANDomainList }}
-    extraSANDomainList:
-    {{- range .Values.tidb.extraSANDomainList }}
-    - {{ . }}
-    {{- end }}
+  {{- if .Values.tidb.tlsClient }}
+    tlsClient:
```
weekface: use the same struct with `values.yaml`

pkg/manager/member/tidb_member_manager.go
```diff
@@ -108,12 +108,13 @@ func (tmm *tidbMemberManager) Sync(tc *v1alpha1.TidbCluster) error {
 		return err
 	}
 
-	// Sync Tidb StatefulSet
-	if err := tmm.syncTiDBStatefulSetForTidbCluster(tc); err != nil {
+	// Sync TiDB Service before syncing TiDB StatefulSet
+	if err := tmm.syncTiDBService(tc); err != nil {
```
weekface: we need service ip to fill the SANs attribute of the certificate.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -434,7 +439,11 @@ func getTiDBConfigMap(tc *v1alpha1.TidbCluster) (*corev1.ConfigMap, error) {
 		if config.Security == nil {
 			config.Security = &v1alpha1.Security{}
 		}
-		config.Security.SSLCA = pointer.StringPtr(serviceAccountCAPath)
+		if tc.Spec.TiDB.IsUserGeneratedCertificate() {
+			config.Security.SSLCA = pointer.StringPtr(path.Join(serverCertPath, "ca"))
```
weekface: use user-defined ca file

charts/tidb-cluster/values.yaml
```diff
@@ -428,18 +428,25 @@ tidb:
...
+      #  - 2.2.2.2
+      # # extra SAN Domain list
+      # extraSANDomainList:
+      #  - example1.com
+      #  - example2.com
+    # User-generated certificate
+    userGenerated: {}
+      # # secretName is the name of the secret which stores user-defined tidb server certificate, key and ca, create it with:
+      # # kubectl create secret generic <secret-name> --namespace=<namespace> --from-file=cert=<tidb server certificate file path> --from-file=key=<tidb server key file path> --from-file=ca=<ca file path>
+      # secretName: "demo-tidb-server-secret"
```
cofyc: I think we can simplify this a bit, like [ingress tls config](https://kubernetes.io/docs/concepts/services-networking/ingress/):

```
  tls:
  - hosts:
    - sslexample.foo.com
    secretName: testsecret-tls
```

the difference is we can specify how to issue a certificate automatically at the same place. 

```
tlsClient:
    enabled: false
    #  secretName is the name of the secret that stores user-defined tidb server certificate, key and ca...
    # if not specified but tls client is enabled, certificated signed by k8s is created automatically
    # secretName: "<tidb-tls-secret-name>"
    autoGenerated:
       extraSANIPList: ...
       extraSANDomainList: ...
```

weekface: ingress `hosts` the listen address of the http server.  A little different from ours. 

in addition, if user-defined certificate is used, they can't set `extraSAN...` here, they should set other SANs when creating certificate.

cofyc: I don't mean we need to add the `hosts` field. I just copied it from the ingress page I linked. I mean we can move `secretName` field to the up level, like `secretName` in ingress tls config. The below example is what I suggested.

weekface: make sense

cofyc: `secretName` can have high precedence over `autoGenerated`, like `userGenerated` over `autoGenerated`. but because if a pre-created certificate is provided, I guess no extra configurations can be add in `userGenerated`, so we can replace `userGenerated` with `secretName`.

weekface:
I have tested it manually, it works as expected. Yes, I think I will add test e2e test in another PR.

charts/tidb-cluster/values.yaml
weekface: user may use their custom Cluster TLS which used between pd, tikv and tidb peers. so there may be extra configrations?

weekface: may other `userDefined` attributes will be added in the future that we don't know yet

cofyc: this `tlsClient` field is under `tidb` field and used to configure tidb only. if we configure tls secret for other components, we can add `secretName` in the configuration of other components.

weekface: I mean the TLS between TiDB components: https://pingcap.com/docs/stable/how-to/secure/enable-tls-between-components/

But i am fine with only `secretName`.

charts/tidb-cluster/values.yaml
```diff
@@ -428,18 +428,25 @@ tidb:
...
-  #   - 1.1.1.1
-  #   - 2.2.2.2
-  # # extra SAN Domain List when you set tidb.enableTLSClient to true
-  # extraSANDomainList:
-  #   - example1.com
-  #   - example2.com
+  tlsClient:
+    # When enabled, TiDB will accept TLS encrypted connections from MySQL client
+    enabled: false
+    # Auto-generated certificate
```
tennix: This is generated in k8s, you can add a document link here.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -288,13 +289,17 @@ func (tmm *tidbMemberManager) syncTiDBServerCerts(tc *v1alpha1.TidbCluster) erro
...
+	if tc.Spec.TiDB.TLSClient.AutoGenerated != nil {
+		hostList = append(hostList, tc.Spec.TiDB.TLSClient.AutoGenerated.ExtraSANDomainList...)
+	}
 
 	ipList := []string{
 		"127.0.0.1", "::1",
 		svc.Spec.ClusterIP,
 	}
-	ipList = append(ipList, tc.Spec.TiDB.ExtraSANIPList...)
+	if tc.Spec.TiDB.TLSClient.AutoGenerated != nil {
```
tennix: Could you merge this `if` condition with the previous, when the cert is auto generated, both the `ipList` and `hostList` need to be appended.

tennix:
BTW, could you help to verify if this issue https://github.com/pingcap/tidb-operator/issues/1670 is also resolved by your PR?

weekface:
> BTW, could you help to verify if this issue #1670 is also resolved by your PR?

Of course, I will do it.

weekface:
> > BTW, could you help to verify if this issue #1670 is also resolved by your PR?
> 
> Of course, I will do it.

I have tested this tutorial manually and it works.

PR #1722, opened 2020-02-18
Add auto-scaling calculation based by CPU load

Yisaer:
#1651 
Adding auto-scaling calculation algorithm based by CPU load. 

This request adds the realization of auto-scaling calculation algorithm based by CPU load. 
Currently, the auto-scaling is designed to support the cpu AverageUtilization in the following api spec:
```yaml
  tikv:
    MetricsTimeDuration: "5m"
    metrics:
    - type: "Resource"
      resource:
        name: "cpu"
        target:
          type: "Utilization"
          averageUtilization: 80
```
**notice**
To use the auto-scaling by CPU load, the CPU requests in resources must be defined.

I would take this spec as an example to explain how the calculation work.
The users create it to claim that the Operator could auto-scale tikv by CPU average load and the threshold value of average Utilization is 80%.

The Operator would query the CPU metrics from Prometheus (currently) to calculate the sum of CPU seconds for each tikv instances in the past 5mins every resync duration seconds.
If the sum of CPU seconds for total tikv instances in the past 5mins is 500 sec as `cpuSecsTotal`  The cpu requests in resources is defined as 1 (1000mi) and current replicas of tikv is 3.
```bash
cpuRequestsRatio = 1000 / 1000 =1
durationSecs = 5mins = 300s
utilizationRatio = 80%
expectedCpuSecsTotal = cpuRequestsRatio * durationSeconds * currentReplicas * utilizationRatio = 1 * 300 * 3 * 0.8 = 720
recommendedReplicas = ceil(cpuSecsTotal * currentReplicas / expectedCpuSecsTotal) = 3
```

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/autoscaler/autoscaler/calculate/calculate.go
+++ b/pkg/autoscaler/autoscaler/calculate/calculate.go
@@ -57,7 +57,7 @@ func queryMetricsFromPrometheus(tac *v1alpha1.TidbClusterAutoScaler,
 		return err
 	}
 	if resp.Status != statusSuccess {
-		return fmt.Errorf("tac[%s/%s]' query error, response stataus:%v", tac.Namespace, tac.Name, resp.Status)
+		return fmt.Errorf("tac[%s/%s] query error, response status: %v", tac.Namespace, tac.Name, resp.Status)
 	}
 	return nil
 }
--- a/pkg/autoscaler/autoscaler/calculate/calculate.go
+++ b/pkg/autoscaler/autoscaler/calculate/calculate.go
@@ -29,7 +29,7 @@ const (
 	TikvSumCpuMetricsPattern     = `sum(tikv_thread_cpu_seconds_total{cluster="%s"}) by (instance)`
 	TidbSumCpuMetricsPattern     = `sum(process_cpu_seconds_total{cluster="%s",job="tidb"}) by (instance)`
 	InvalidTacMetricConfigureMsg = "tac[%s/%s] metric configuration invalid"
-	CpuSumMetricsErrorMsg        = "tac[%s/%s] cpu sum metrics error,can't calculate past %s cpu metrics,might caused by prometheus restart with no persistence"
+	CpuSumMetricsErrorMsg        = "tac[%s/%s] cpu sum metrics error, can't calculate the past %s cpu metrics, may caused by prometheus restart while data persistence not enabled"
 	queryPath                    = "/api/v1/query"
 
 	float64EqualityThreshold = 1e-9
--- a/pkg/autoscaler/autoscaler/calculate/calculate.go
+++ b/pkg/autoscaler/autoscaler/calculate/calculate.go
@@ -1,0 +1,121 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package calculate
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"net/http"
+	"strconv"
+
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+	promClient "github.com/prometheus/client_golang/api"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	TikvSumCpuMetricsPattern     = `sum(tikv_thread_cpu_seconds_total{cluster="%s"}) by (instance)`
+	TidbSumCpuMetricsPattern     = `sum(process_cpu_seconds_total{cluster="%s",job="tidb"}) by (instance)`
+	InvalidTacMetricConfigureMsg = "tac[%s/%s] metric configure invalid"
+	CpuSumMetricsErrorMsg        = "tac[%s/%s] cpu sum metrics error,can't calculate past %s cpu metrics,might casued by promthesu restart with no persistance"
+	queryPath                    = "/api/v1/query"
+)
+
+func queryMetricsFromPrometheus(tac *v1alpha1.TidbClusterAutoScaler,
+	query string, client promClient.Client, timestamp int64, resp *Response) error {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s%s", *tac.Spec.MetricsUrl, queryPath), nil)
+	if err != nil {
+		return err
+	}
+	q := req.URL.Query()
+	q.Add("query", query)
+	q.Add("time", fmt.Sprintf("%d", timestamp))
+	req.URL.RawQuery = q.Encode()
+	r, body, err := client.Do(req.Context(), req)
+	if err != nil {
+		return err
+	}
+	if r.StatusCode != http.StatusOK {
+		return fmt.Errorf("tac[%s/%s]' query error,status code:%d", tac.Namespace, tac.Name, r.StatusCode)
+	}
+	err = json.Unmarshal(body, resp)
+	if err != nil {
+		return err
+	}
+	if resp.Status != statusSuccess {
+		return fmt.Errorf("tac[%s/%s]' query error, response stataus:%v", tac.Namespace, tac.Name, resp.Status)
+	}
+	return nil
+}
+
+// MetricType describe the current Supported Metric Type to calculate the recommended Replicas
+type MetricType string
+
+const (
+	MetricTypeCPU MetricType = "cpu"
+	//metricTypeQPS MetricType = "qps"
+)
+
+func sumByInstanceFromResponse(instances []string, resp *Response) (float64, error) {
+	s := sets.String{}
+	for _, instance := range instances {
+		s.Insert(instance)
+	}
+	sum := 0.0
+	for _, r := range resp.Data.Result {
+		if s.Has(r.Metric.Instance) {
+			v, err := strconv.ParseFloat(r.Value[1].(string), 64)
+			if err != nil {
+				return 0.0, err
+			}
+			sum = sum + v
+		}
+	}
+	return sum, nil
+}
+
+// calculate func calculate the recommended replicas by given usageRadio and currentReplicas
+func calculate(currentValue float64, targetValue float64, currentReplicas int32) int32 {
+	usageRadio := currentValue / targetValue
+	return int32(math.Ceil(usageRadio * float64(currentReplicas)))
+}
+
+const (
+	statusSuccess = "success"
+)
+
+type Response struct {
+	Status string `json:"status"`
+	Data   Data   `json:"data"`
+}
+
+type Data struct {
+	ResultType string   `json:"resultType"`
+	Result     []Result `json:"result"`
+}
+
+type Result struct {
+	Metric Metric        `json:"metric"`
+	Value  []interface{} `json:"value"`
+}
+
+type Metric struct {
+	Cluster             string `json:"cluster,omitempty"`
+	Instance            string `json:"instance"`
+	Job                 string `json:"job,omitempty"`
+	KubernetesNamespace string `json:"kubernetes_namespace,omitempty"`
+	KubernetesNode      string `json:"kubernetes_node,omitempty"`
+	KubernetesPodIp     string `json:"kubernetes_pod_ip,omitempty"`
+}
--- a/pkg/autoscaler/autoscaler/calculate/calculate.go
+++ b/pkg/autoscaler/autoscaler/calculate/calculate.go
@@ -81,13 +81,13 @@ func sumByInstanceFromResponse(instances []string, resp *Response) (float64, err
 	return sum, nil
 }
 
-// calculate func calculate the recommended replicas by given usageRadio and currentReplicas
+// calculate func calculate the recommended replicas by given usageRatio and currentReplicas
 func calculate(currentValue float64, targetValue float64, currentReplicas int32) (int32, error) {
 	if almostEqual(targetValue, 0.0) {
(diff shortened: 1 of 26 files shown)


pkg/autoscaler/autoscaler/util.go
```diff
@@ -265,7 +264,7 @@ func checkConsecutiveCount(tac *v1alpha1.TidbClusterAutoScaler,
 			}
 		} else {
 			// scale-in, no-scaling would be return nil at first
-			if int32(currentScaleInCount) < *tac.Spec.TiDB.ScaleInThreshold {
+			if int32(currentScaleInCount) < *tac.Spec.TiKV.ScaleInThreshold {
```
Yisaer: This is a bug I fixed and I also revise the unit test.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -36,7 +39,11 @@ func (am *autoScalerManager) syncTiDB(tc *v1alpha1.TidbCluster, tac *v1alpha1.Ti
 		return nil
 	}
 	currentReplicas := tc.Spec.TiDB.Replicas
-	targetReplicas := calculateRecommendedReplicas(tac, v1alpha1.TiDBMemberType, client)
+	instances := filterTidbInstances(tc)
+	targetReplicas, err := calculateTidbMetrics(tac, sts, client, instances)
```
Yisaer: This is the entrance function for TiDB to calculate auto-scaling by each metric.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -35,8 +38,12 @@ func (am *autoScalerManager) syncTiKV(tc *v1alpha1.TidbCluster, tac *v1alpha1.Ti
 		emptyAutoScalingCountAnn(tac, v1alpha1.TiKVMemberType)
 		return nil
 	}
-	currentReplicas := getStateUpReplicas(tc)
-	targetReplicas := calculateRecommendedReplicas(tac, v1alpha1.TiKVMemberType, client)
+	instances := filterTiKVInstances(tc)
+	currentReplicas := int32(len(instances))
+	targetReplicas, err := calculateTikvMetrics(tac, sts, client, instances)
```
Yisaer: This is the entrance function for TiKV to calculate auto-scaling by each metric.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -80,3 +87,25 @@ func updateTcTiDBAnnIfScale(tac *v1alpha1.TidbClusterAutoScaler) {
...
+	case calculate.MetricTypeCPU:
+		return calculate.CalculateCpuMetrics(tac, sts, client, instances, metric, calculate.TidbSumCpuMetricsPattern, *tac.Spec.TiDB.MetricsTimeDuration, v1alpha1.TiDBMemberType)
+	default:
+		return 0, fmt.Errorf(calculate.InvalidTacMetricConfigureMsg, tac.Namespace, tac.Name)
+	}
+}
+
+func filterTidbInstances(tc *v1alpha1.TidbCluster) []string {
+	var instances []string
+	for i := 0; int32(i) < tc.Spec.TiDB.Replicas; i++ {
```
DanielZhangQD: Should we get the replicas from sts considering the failover case and check if Pod is Running & Ready?

Yisaer: I think currently we can leave it be, whether changing the instances filter logic depends on the future performance.

pkg/autoscaler/autoscaler/calculate/calculate.go
```diff
@@ -0,0 +1,95 @@
...
+
+func queryMetricsFromPrometheus(tac *v1alpha1.TidbClusterAutoScaler,
+	query string, client promClient.Client, timestamp int64, resp *Response) error {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s%s", *tac.Spec.MetricsUrl, queryPath), nil)
+	if err != nil {
+		return err
+	}
+	q := req.URL.Query()
+	q.Add("query", query)
+	q.Add("time", fmt.Sprintf("%d", timestamp))
```
DanielZhangQD: With this query, the end time is `time` and what's the start time? Do we need to specify a range?

pkg/autoscaler/autoscaler/calculate/cpu.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+	promClient "github.com/prometheus/client_golang/api"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv2beta2 "k8s.io/api/autoscaling/v2beta2"
+	corev1 "k8s.io/api/core/v1"
+)
+
+//TODO: create issue to explain how auto-scaling algorithm based by cpu metrics work
+func CalculateCpuMetrics(tac *v1alpha1.TidbClusterAutoScaler, sts *appsv1.StatefulSet,
+	client promClient.Client, instances []string, metric autoscalingv2beta2.MetricSpec,
```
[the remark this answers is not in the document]
Yisaer: Yes, but for `tac` contains all the `matric` and `timeWindow` for both tidb and tikv, it would be simpler and more convenient to pass these variables from calling.

pkg/autoscaler/autoscaler/calculate/calculate.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// sumByInstanceFromResponse sum the value in Response of each instance from Prometheus
+func sumByInstanceFromResponse(instances []string, resp *Response) (float64, error) {
+	s := sets.String{}
+	for _, instance := range instances {
+		s.Insert(instance)
+	}
+	sum := 0.0
+	for _, r := range resp.Data.Result {
```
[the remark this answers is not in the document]
Yisaer: the `Data` and `Result` here are not pointer, and I updated it to check the length of `Result`

pkg/autoscaler/autoscaler/calculate/util.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, m := range metrics {
+		if m.Type == autoscalingv2beta2.ResourceMetricSourceType && m.Resource != nil {
+			return m
+		}
+	}
+	return metrics[0]
+}
+
+// MetricType describe the current Supported Metric Type to calculate the recommended Replicas
+type MetricType string
```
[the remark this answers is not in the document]
Yisaer: updated.

pkg/autoscaler/autoscaler/calculate/calculate.go
Yisaer: No, the metrics is to query the total counts from Prometheus initialization to the given timestamp. To query the D-value between the start time and end time is to query it twice with different timestamp.

pkg/autoscaler/autoscaler/calculate/calculate.go
```diff
@@ -0,0 +1,98 @@
...
+	query string, client promClient.Client, timestamp int64, resp *Response) error {
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s%s", *tac.Spec.MetricsUrl, queryPath), nil)
+	if err != nil {
+		return err
+	}
+	q := req.URL.Query()
+	q.Add("query", query)
+	q.Add("time", fmt.Sprintf("%d", timestamp))
+	req.URL.RawQuery = q.Encode()
+	r, body, err := client.Do(req.Context(), req)
```
DanielZhangQD: Do we need to specify a timeout context?
Refer to https://godoc.org/github.com/prometheus/client_golang/api/prometheus/v1#example-API--Query

Yisaer: Currently unnecessary, It could be record it first as future plan.

Yisaer: ref https://github.com/pingcap/tidb-operator/issues/1651#issuecomment-586873466

pkg/autoscaler/autoscaler/calculate/calculate.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	if resp.Status != statusSuccess {
+		return fmt.Errorf("tac[%s/%s] query error, response status: %v", tac.Namespace, tac.Name, resp.Status)
+	}
+	return nil
+}
+
+// sumByInstanceFromResponse sum the value in Response of each instance from Prometheus
+func sumByInstanceFromResponse(instances []string, resp *Response) (float64, error) {
+	s := sets.String{}
```
aylei: `sumByInstance` is misleading, which means a sum for each instance

pkg/autoscaler/autoscaler/calculate/cpu.go
```diff
@@ -0,0 +1,85 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv2beta2 "k8s.io/api/autoscaling/v2beta2"
+	corev1 "k8s.io/api/core/v1"
+)
+
+//TODO: create issue to explain how auto-scaling algorithm based on cpu metrics work
+func CalculateCpuMetrics(tac *v1alpha1.TidbClusterAutoScaler, sts *appsv1.StatefulSet,
+	client promClient.Client, instances []string, metric autoscalingv2beta2.MetricSpec,
+	queryPattern, timeWindow string, memberType v1alpha1.MemberType) (int32, error) {
+	if metric.Resource == nil || metric.Resource.Target.AverageUtilization == nil {
```
aylei: This function takes too many parameters, can we refine this?

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -80,3 +87,25 @@ func updateTcTiDBAnnIfScale(tac *v1alpha1.TidbClusterAutoScaler) {
...
+	metric := calculate.FilterMetrics(tac.Spec.TiDB.Metrics)
+	mType, err := calculate.GenMetricType(tac, metric)
+	if err != nil {
+		return 0, err
+	}
+	switch mType {
+	case calculate.MetricTypeCPU:
+		return calculate.CalculateCpuMetrics(tac, sts, client, instances, metric, calculate.TidbSumCpuMetricsPattern, *tac.Spec.TiDB.MetricsTimeDuration, v1alpha1.TiDBMemberType)
+	default:
+		return 0, fmt.Errorf(calculate.InvalidTacMetricConfigureMsg, tac.Namespace, tac.Name)
```
aylei: I recommend using `-1`, scaling to 0 is a useful feature and 0 should not be used to denote an error

pkg/autoscaler/autoscaler/tidb_autoscaler.go
aylei: we should take failover into account

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -79,17 +86,32 @@ func syncTiKVAfterCalculated(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbCluster
...
 func updateTcTiKVAnnIfScale(tac *v1alpha1.TidbClusterAutoScaler) {
 	tac.Annotations[label.AnnTiKVLastAutoScalingTimestamp] = time.Now().String()
 	emptyAutoScalingCountAnn(tac, v1alpha1.TiKVMemberType)
 }
+
+func calculateTikvMetrics(tac *v1alpha1.TidbClusterAutoScaler, sts *appsv1.StatefulSet, client promClient.Client, instances []string) (int32, error) {
+	metric := calculate.FilterMetrics(tac.Spec.TiKV.Metrics)
+	mType, err := calculate.GenMetricType(tac, metric)
+	if err != nil {
+		return 0, err
```
aylei: ditto

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -79,17 +86,32 @@ func syncTiKVAfterCalculated(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbCluster
...
+	}
+	switch mType {
+	case calculate.MetricTypeCPU:
+		return calculate.CalculateCpuMetrics(tac, sts, client, instances, metric, calculate.TikvSumCpuMetricsPattern, *tac.Spec.TiKV.MetricsTimeDuration, v1alpha1.TiKVMemberType)
+	default:
+		return 0, fmt.Errorf(calculate.InvalidTacMetricConfigureMsg, tac.Namespace, tac.Name)
```
aylei: ditto

pkg/autoscaler/autoscaler/calculate/calculate.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, instance := range instances {
```
DanielZhangQD: Why convert `instances` to set in this function but not other functions?
BTW, `instances` should be the slice of pod name, is it possible to be duplicated in the slice?

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: updated.

pkg/autoscaler/autoscaler/calculate/cpu.go
Yisaer: updated.

pkg/autoscaler/autoscaler/calculate/cpu.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return 0, err
+	}
+	sum2, err := sumByInstanceFromResponse(instances, r)
+	if err != nil {
+		return 0, err
+	}
+	if sum1-sum2 < 0 {
+		return 0, fmt.Errorf(CpuSumMetricsErrorMsg, tac.Namespace, tac.Name, timeWindow)
+	}
```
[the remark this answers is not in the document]
Yisaer: updated.

PR #1723, opened 2020-02-18
fix tidb-lightning errors

mightyguava:
fixes for https://github.com/pingcap/tidb-operator/commit/9a3b1e27603dd28deac8dae43e34d21e1a678629

### DIFF
--- a/charts/tidb-lightning/templates/rclone-conf.yaml
+++ b/charts/tidb-lightning/templates/rclone-conf.yaml
@@ -1,10 +1,9 @@
+{{- if .Values.dataSource.remote.rcloneConfig }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: rclone-{{ include "tidb-lightning.name" . }}
-type: Opaque
 data:
   config-file: |-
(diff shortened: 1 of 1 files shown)


charts/tidb-lightning/templates/rclone-conf.yaml
```diff
@@ -1,10 +1,9 @@
+{{- if .Values.dataSource.remote.rcloneConfig }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: rclone-{{ include "tidb-lightning.name" . }}
-type: Opaque
```
mightyguava: `type` is only allowed in `Secret`s

charts/tidb-lightning/templates/rclone-conf.yaml
```diff
@@ -1,10 +1,9 @@
+{{- if .Values.dataSource.remote.rcloneConfig }}
```
mightyguava: This configmap shouldn't be created at all if `rcloneConfig` is unset

DanielZhangQD:
@mightyguava Thanks for fixing the errors.

PR #1732, opened 2020-02-19
Remove consecutive check

Yisaer:
Currently, we use consecutive check in auto-scaling for noise reduction. During my test, I found that the obj `tidbclusterautoscaler` would be synced frequently and rapidly during consecutive check.
Because `tidbclusterautoscaler` would store the consecutive check count, this need update the `tidbclusterautoscaler` and cause the reconciliation again and again.

IMO, noise reduction is necessary for the production-ready auto-scaling feature, so I tend to remove consecutive check for now and discuss the noise reduction design in the future.

issued in https://github.com/pingcap/tidb-operator/issues/1651#issuecomment-586873466

Related changes

```release-note
NONE
```


pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -56,22 +53,11 @@ func (am *autoScalerManager) syncTiDB(tc *v1alpha1.TidbCluster, tac *v1alpha1.Ti
 // duration between each auto-scaling. If either of them is not meet, the auto-scaling would be rejected.
 // If the auto-scaling is permitted, the timestamp would be recorded and the Consecutive count would be zeroed.
 func syncTiDBAfterCalculated(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler, currentReplicas, recommendedReplicas int32) error {
-	if err := updateConsecutiveCount(tac, v1alpha1.TiDBMemberType, currentReplicas, recommendedReplicas); err != nil {
```
DanielZhangQD: Can we just store the count in memory?
During container restart, just re-count them, no need to be persistent now.

Yisaer: This is remain to be discussed.

PR #1747, opened 2020-02-23
Make TidbMonitor intergrated in AutoScaler

Yisaer:
Make TidbClusterAutoScaler query prometheus from TidbMonitor 

Related changes

```release-note
NONE
```


pkg/autoscaler/autoscaler/util.go
```diff
@@ -198,3 +205,17 @@ func checkAndUpdateTacAnn(tac *v1alpha1.TidbClusterAutoScaler) {
...
+		return "", fmt.Errorf("tac[%s/%s]' metrics url or monitor should be defined explicitly", tac.Namespace, tac.Name)
+	}
+	conn := ""
+	if tac.Spec.Monitor != nil {
+		conn = fmt.Sprintf("http://%s-prometheus.%s.svc:9090", tac.Spec.Monitor.Name, tac.Spec.Monitor.Namespace)
+	}
+	if tac.Spec.MetricsUrl != nil {
+		conn = *tac.Spec.MetricsUrl
+	}
+	return conn, nil
```
DanielZhangQD: Suggestion:
```
if tac.Spec.MetricsUrl != nil {
	return *tac.Spec.MetricsUrl
}
return fmt.Sprintf("http://%s-prometheus.%s.svc:9090", tac.Spec.Monitor.Name, tac.Spec.Monitor.Namespace)
```

pkg/autoscaler/autoscaler/util.go
```diff
@@ -198,3 +205,17 @@ func checkAndUpdateTacAnn(tac *v1alpha1.TidbClusterAutoScaler) {
 	// If not satisfied, reset tac Ann
 	resetAutoScalingAnn(tac)
 }
+
+func genMetricsEndpoint(tac *v1alpha1.TidbClusterAutoScaler) (string, error) {
+	if tac.Spec.MetricsUrl == nil && tac.Spec.Monitor == nil {
+		return "", fmt.Errorf("tac[%s/%s]' metrics url or monitor should be defined explicitly", tac.Namespace, tac.Name)
```
DanielZhangQD:
```suggestion
		return "", fmt.Errorf("tac[%s/%s] metrics url or monitor should be defined explicitly", tac.Namespace, tac.Name)
```

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -73,7 +73,15 @@ func updateTcTiDBAnnIfScale(tac *v1alpha1.TidbClusterAutoScaler) {
...
-func calculateTidbMetrics(tac *v1alpha1.TidbClusterAutoScaler, sts *appsv1.StatefulSet, client promClient.Client, instances []string) (int32, error) {
+func calculateTidbMetrics(tac *v1alpha1.TidbClusterAutoScaler, sts *appsv1.StatefulSet, instances []string) (int32, error) {
+	ep, err := genMetricsEndpoint(tac)
+	if err != nil {
+		return -1, err
+	}
+	client, err := promClient.NewClient(promClient.Config{Address: ep})
+	if err != nil {
+		return -1, err
+	}
```
DanielZhangQD: Why not get ep and initialize the client in manager.go as we do now?

pkg/autoscaler/autoscaler/util.go
Yisaer: resolved.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
Yisaer: Querying Metrics from Prometheus would be the one of the data source, so I think it would be more proper to initialize Prometheus client in `calculateTidbMetrics` func instead of  `syncAutoScaling` func.

Yisaer:
It seems weird of the GitHub action CI, do you have any idea? @cofyc

cofyc:
it seems the ref `81546a0cb4d0b5a138281a9976818fce10f23341` in github actions does not exist anymore. you can modify the last commit with `git commit --amend` and force push again.

cofyc:
according to error logs:

```
git checkout --progress --force 81546a0cb4d0b5a138281a9976818fce10f23341
##[error]fatal: reference is not a tree: 81546a0cb4d0b5a138281a9976818fce10f23341
##[warning]Git checkout failed on shallow repository, this might because of git fetch with depth '1' doesn't include the checkout commit '81546a0cb4d0b5a138281a9976818fce10f23341'.
Removed matchers: 'checkout-git'
##[error]Git checkout failed with exit code: 128
##[error]Exit code 1 returned from process: file name '/home/runner/runners/2.165.2/bin/Runner.PluginHost', arguments 'action "GitHub.Runner.Plugins.Repository.v1_0.CheckoutTask, Runner.Plugins"'.
```

maybe we should not use `--depth 1` ...

pkg/autoscaler/autoscaler/tidb_autoscaler.go
DanielZhangQD: But for one tac, you will initialize the prom client twice for tidb and tikv, i think it's not necessary.

PR #1801, opened 2020-02-27
Manage hot region label for the tikv created by auto-scaler

Yisaer:

For the tikv created by the auto-scaler, we should label the special hot region label for its store.

In this request, it manage life-cycle for the hot region label and auto-scaling slots annotation for the tikv pod created by the auto-scaler.

1. Record the ordinal for the tikv pod created by auto-scaler in the target tidbcluster annotations
2. Removed the ordinal records in annotation when the pod is omitted by scale-in.
3. Add hot region label for the tikv pod according to the recorded ordinal in annotations.

Related changes

```release-note
Add hot region schedule feature for TiKV in Auto-Scaling
```


pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -83,8 +82,25 @@ func filterTiKVInstances(tc *v1alpha1.TidbCluster) []string {
...
+			existed := operatorUtils.GetAutoScalingOutSlots(tc, v1alpha1.TiKVMemberType)
+			v, err := operatorUtils.GenJsonFromSets(newlyScaleOutOrdinalSets.Union(existed))
+			if err != nil {
+				return err
+			}
+			tc.Annotations[label.AnnTiKVAutoScalingOutOrdinals] = fmt.Sprintf("%v", v)
+		}
+	}
+	tc.Spec.TiKV.Replicas = recommendedReplicas
+	return nil
```
Yisaer: When the auto-scaling result is scale-out, we would record the ordinal of the tikv pod which is going to be created by auto-scaler  in tidbcluster annotations in order to add hot region labels during tikv scaling process in tidbcluster controller.

pkg/manager/member/tikv_scaler.go
```diff
@@ -190,6 +193,55 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
+			delete(tc.Annotations, label.AnnTiKVAutoScalingOutOrdinals)
+			return nil
+		}
+		v, err := util.GenJsonFromSets(currentScalingSlots)
+		if err != nil {
+			return err
+		}
+		tc.Annotations[label.AnnTiKVAutoScalingOutOrdinals] = fmt.Sprintf("%v", v)
+		return nil
+	}
```
Yisaer: Reclaimed the ordinal slots in the tidbclusters annotations which belongs to the pod already been eliminated by scale-in.

pkg/manager/member/tikv_scaler.go
```diff
@@ -190,6 +193,55 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
 	return fmt.Errorf("TiKV %s/%s not found in cluster", ns, podName)
 }
 
+// SyncAutoScalerAnn would reclaim the auto-scaling out slots if the target pod is no longer existed
+// For the existed auto-scaling slots, we would add special hot region label the its store by pdapi.
```
DanielZhangQD:
```suggestion
// For the auto-scaling slots, we would add the special hot region label to the store with pdapi.
```

pkg/manager/member/tikv_scaler.go
```diff
@@ -190,6 +193,55 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
+		}
+		v, err := util.GenJsonFromSets(currentScalingSlots)
+		if err != nil {
+			return err
+		}
+		tc.Annotations[label.AnnTiKVAutoScalingOutOrdinals] = fmt.Sprintf("%v", v)
+		return nil
+	}
+
+	// For the existed auto-scaling slots, we would add special hot region label the its store by pdapi.
```
DanielZhangQD: ditto

pkg/manager/member/tikv_scaler.go
```diff
@@ -190,6 +193,55 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
+	pdClient := tsd.pdControl.GetPDClient(pdapi.Namespace(tc.Namespace), tc.Name, *tc.Spec.EnableTLSCluster)
+	for k := range currentScalingSlots {
+		podName := util.GetPodName(tc, v1alpha1.TiKVMemberType, k)
+		for _, store := range tc.Status.TiKV.Stores {
```
DanielZhangQD: `TiKV.Stores` is a map.

DanielZhangQD: BTW, would it be better to get the stores with pdclient directly instead of checking the status of tc?

pkg/manager/member/tikv_scaler.go
```diff
@@ -190,6 +193,55 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
+				id, err := strconv.ParseUint(store.ID, 10, 64)
+				if err != nil {
+					return err
+				}
+				ok, err := pdClient.SetStoreLabels(id, hostRegionLabel)
+				if err != nil {
+					return err
+				}
+				if !ok {
+					return fmt.Errorf("tc[%s/%s]'s pod[%s] failed to add label hot special region", tc.Namespace, tc.Name, podName)
```
DanielZhangQD:
```suggestion
					return fmt.Errorf("tc[%s/%s]'s pod[%s] failed to add special hot region label", tc.Namespace, tc.Name, podName)
```

pkg/manager/member/tikv_scaler.go
```diff
@@ -190,6 +193,55 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
+	pdClient := tsd.pdControl.GetPDClient(pdapi.Namespace(tc.Namespace), tc.Name, *tc.Spec.EnableTLSCluster)
+	for k := range currentScalingSlots {
+		podName := util.GetPodName(tc, v1alpha1.TiKVMemberType, k)
+		for _, store := range tc.Status.TiKV.Stores {
+			if store.PodName == podName {
+				id, err := strconv.ParseUint(store.ID, 10, 64)
+				if err != nil {
+					return err
+				}
+				ok, err := pdClient.SetStoreLabels(id, hostRegionLabel)
```
DanielZhangQD: Is it necessary to check if the label is already set?

Yisaer: Not necessary.

pkg/manager/member/tikv_scaler.go
Yisaer: There is no need to get stores by pdclient here, the tikv status will be synced each reconciliation from pdapi.

Yisaer: > `TiKV.Stores` is a map.

I don't understand is there any problem to go through the stores here.

DanielZhangQD: I thought the key is the podName, but it turns to be the store ID.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -83,8 +82,25 @@ func filterTiKVInstances(tc *v1alpha1.TidbCluster) []string {
...
+		if newlyScaleOutOrdinalSets.Len() > 0 {
+			if tc.Annotations == nil {
+				tc.Annotations = map[string]string{}
+			}
+			existed := operatorUtils.GetAutoScalingOutSlots(tc, v1alpha1.TiKVMemberType)
+			v, err := operatorUtils.GenJsonFromSets(newlyScaleOutOrdinalSets.Union(existed))
+			if err != nil {
+				return err
+			}
+			tc.Annotations[label.AnnTiKVAutoScalingOutOrdinals] = fmt.Sprintf("%v", v)
```
cofyc: `v` is string type, so `fmt.Sprintf` is not necessary.

pkg/manager/member/tikv_scaler.go
```diff
@@ -190,6 +193,55 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
+		currentScalingSlots = currentScalingSlots.Delete(reclaimedSlots.List()...)
+		if currentScalingSlots.Len() < 1 {
+			delete(tc.Annotations, label.AnnTiKVAutoScalingOutOrdinals)
+			return nil
+		}
+		v, err := util.GenJsonFromSets(currentScalingSlots)
+		if err != nil {
+			return err
+		}
+		tc.Annotations[label.AnnTiKVAutoScalingOutOrdinals] = fmt.Sprintf("%v", v)
```
cofyc: ditto

pkg/autoscaler/autoscaler/tikv_autoscaler.go
Yisaer: updated.

pkg/manager/member/scaler.go
```diff
@@ -35,6 +35,12 @@ const (
 	skipReasonScalerAnnDeferDeletingIsEmpty = "scaler: pvc annotations defer deleting is empty"
 )
 
+var (
+	hostRegionLabel = map[string]string{
+		"specialUse": "hotRegion",
```
cofyc: describe the purpose of this special label? why we need to add this special label on newly created TiKV instances. If PD has public docs about this label, we can link too.

pkg/util/util.go
```diff
@@ -126,3 +126,40 @@ func IsStatefulSetScaling(set *appsv1.StatefulSet) bool {
...
+	var slice []int32
+	err := json.Unmarshal([]byte(v), &slice)
+	if err != nil {
+		return s
+	}
+	s.Insert(slice...)
+	return s
+}
+
+func GenJsonFromSets(sets sets.Int32) (string, error) {
```
cofyc: can you copy this `encode` method here and make it public?
https://github.com/pingcap/tidb-operator/blob/4cbffa7619e3fbc1acea45c5b92990c427634ab3/pkg/manager/member/utils.go#L100-L106

this is a more generic `JSON marshal to string` method, then we don't need to add many different but similar methods. in your case, you can pass `sets.List()` to it.

pkg/manager/member/scaler.go
Yisaer: Nice catch, I will add `TODO` to mark there should be some docs about this label after I finish the entire auto-scaling document.

pkg/util/util.go
Yisaer: updated.

pkg/manager/member/scaler.go
cofyc: great! comments describing the purpose or reasons behind will make it easier to understand the code.

PR #1811, opened 2020-02-27
[DNM] New Cluster-TLS API

weekface:
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


tennix:
Is this API compatible with cert-manager?

weekface:
We can add cert-manager support by add a new key:

```
tlsCluster:
  enabled: true
  secretName: "secretName"
  autoGenerated: {}
  certManager: {}
```

tennix:
> We can add cert-manager support by add a new key:
> 
> ```
> tlsCluster:
>   enabled: true
>   secretName: "secretName"
>   autoGenerated: {}
>   certManager: {}
> ```

OK, sounds good to me.

charts/tidb-cluster/values.yaml
```diff
@@ -55,10 +55,35 @@ discovery:
...
+  # #   2. Then create the secret with the following command:
+  # #   kubectl create secret generic <secret-name> --namespace=<namespace>
+  # #       --from-file=pd-cert=<pd server certificate file path>
+  # #       --from-file=pd-key=<pd server key file path>
+  # #       --from-file=tikv-cert=<tikv server certificate file path>
+  # #       --from-file=tikv-key=<tikv server key file path>
+  # #       --from-file=tidb-cert=<tidb server certificate file path>
+  # #       --from-file=tidb-key=<tidb server key file path>
+  # #       --from-file=ca=<ca file path>
+  # secretName: "demo-tidb-components-secret"
```
cofyc: - one secret for one component
- for serects, I suggest using dedicated tls secret generation method in Kubernetes, `kubectl create secret tls <secret>`

the reasons are:
- automation tool like cert-manager manages one secret in each Certificate CRD 
- we should following k8s tls secret format, then third party generated secrets can be used directly

weekface: We are considering using one certificate for all tidb components. If we can, just need one `secret`.

weekface: Or TiDB core Components(include pd tikv and tidb) use one single cert, and TiDB Binlog(include pump and drainer) use another cert. It’s the same for future TiFlash and other components...

weekface: @cofyc The `kubectl create secret tls` only have `--cert` and `--key`, but without `--ca` flag, so we can't use it to create the user-defined certificate secret?

cofyc: yes, we still need to use the `generic` type in this case.

weekface:
comments addressed PTAL @cofyc @AstroProfundis @tennix

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -299,11 +299,7 @@ func (tc *TidbCluster) GetClusterID() string {
 }
 
 func (tc *TidbCluster) IsTLSClusterEnabled() bool {
-	enableTLCluster := tc.Spec.EnableTLSCluster
-	if enableTLCluster == nil {
-		return defaultEnableTLSCluster
-	}
-	return *enableTLCluster
+	return tc.Spec.TLSCluster != nil && tc.Spec.TLSCluster.Enabled
```
Yisaer: It seems we are no longer using the `tc.Spec.EnableTLSCluster`, do we need to remove this property or set it as deprecated?

weekface: removed

weekface:
Open a new PR instead: https://github.com/pingcap/tidb-operator/pull/1870

PR #1816, opened 2020-02-28
Enable defaulting for tidbcluster

Yisaer:

Enable tidbcluster webhook defaulting both in idc configuration and operat chart values.

Related changes

```release-note
Enable TidbCluster defaulting mutation by default which is recommended when admission webhook is used
```

### DIFF
--- a/charts/tidb-operator/values.yaml
+++ b/charts/tidb-operator/values.yaml
@@ -195,7 +195,7 @@ admissionWebhook:
     ## validating hook validates the correctness of the resources under pingcap.com group
     validating: false
     ## defaulting hook set default values for the the resources under pingcap.com group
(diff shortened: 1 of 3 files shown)


cofyc:
release note `Enable TidbCluster Mutation web` is misleading. I guess you suggest users to enable defaulting when admission webhook is used. how about this?

`Enable TidbCluster defaulting mutation by default which is recommended when admission webhook is used`

Yisaer:
> release note `Enable TidbCluster Mutation web` is misleading. I guess you suggest users to enable defaulting when admission webhook is used. how about this?
> 
> `Enable TidbCluster defaulting mutation by default which is recommended when admission webhook is used`

Nice advice.

Issue #1538, opened 2020-01-10
Make Service port name configurable for tidb and pd service


Yisaer:
## Feature Request

**Is your feature request related to a problem? Please describe:**

We need to make service port name configurable for tidb and pd service.  Currently, https://github.com/pingcap/tidb-operator/pull/1521 would make monitor service port name configurable. 

**Describe the feature you'd like:**

The default service port name for tidb is `mysql-tidb`. For the default pd service port name, when `enableTLS` is not enabled, it is `http-pd`. If enabled, the default service port name is `https-pd`.

mikechengwei:
I have optimize pd service config before, so this issue can be assigned to me. @Yisaer

Yisaer:
@mikechengwei Sure, feel free to go ahead.

Yisaer:
@mikechengwei Hi, is this issue still WIP? Do I need reassign this issue?

mikechengwei:
Sorry.I will finish it this week.ok?

Yisaer:
@mikechengwei  Thanks, take your time.

cofyc:
has it been resolved now?

mikechengwei:
I think it have been finished.

cofyc:
Thanks.
Closed via https://github.com/pingcap/tidb-operator/pull/1823.

cofyc:
@mikechengwei next time, you can use [these keywords](https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#about-issue-references] in your PR to close issue automatically ;-)

PR #1823, opened 2020-02-28
feature:make Service port name configurable for tidb and pd service

mikechengwei:
#1538 
Make Service port name configurable for tidb and pd service
### Check List

Tests 

 - Unit test
Code changes

 - Has Go code change

Side effects

 None

Related changes

None


mikechengwei:
@Yisaer help me review this pr.

charts/tidb-cluster/values.yaml
```diff
@@ -77,6 +77,7 @@ pd:
   # we can only specify clusterIP and loadBalancerIP now
   service:
     clusterIP: "None"
+    portName: "None"
```
cofyc: why it defaults to `None`? I guess "None" (string) port name will be used. shouldn't it be omitted by default?

charts/tidb-cluster/values.yaml
```diff
@@ -396,6 +397,7 @@ tidb:
   service:
     type: NodePort
     exposeStatus: true
+    portName: "None"
```
cofyc: ditto

charts/tidb-cluster/values.yaml
mikechengwei: For backward compatibility, I want to set nil in code like  clusterIP field.
 I can set original portName.

cofyc: you mean `None` in yaml is parsed as `nil` in go?

mikechengwei: It's my question,I  have optimized code again. And I comment  pd `clusterIP` field, because `clusterIP` set` None` is also an error.

charts/tidb-cluster/values.yaml
```diff
@@ -396,6 +397,7 @@ tidb:
   service:
     type: NodePort
     exposeStatus: true
+    portName: "mysql-client"
```
Yisaer:
```suggestion
    portName: "mysql"
```

charts/tidb-cluster/values.yaml
```diff
@@ -76,7 +76,8 @@ pd:
   # pd Service
   # we can only specify clusterIP and loadBalancerIP now
   service:
-    clusterIP: "None"
+#    clusterIP: "ClusterIP"
+    portName: "client"
```
Yisaer: Why note the clusterIP? I suggest the portName of pd would be like `http-pd`

charts/tidb-cluster/values.yaml
mikechengwei: For compatibility, should we use the original portName？

charts/tidb-cluster/values.yaml
```diff
@@ -77,6 +77,7 @@ pd:
   # we can only specify clusterIP and loadBalancerIP now
   service:
     clusterIP: "None"
+    portName: "client"
```
cofyc: what if we don't set `portName` by default in helm value. will the original name in the code be used? if it is, I think we can omit it in the values.

mikechengwei: It will use original name in the code. But do we need to remind users of this feature.

cofyc: we can give an example in the value files, like this:

```
service:
    clusterIP: "None"
    # portName: "<the default name will be used>"
```

what do you think?

mikechengwei: I agree, i will follow this.

cofyc: the absent field in yaml will be parsed as `nil` in go struct. this is backward compatible but we also provide a way for users to configure this field.

cofyc: cc @Yisaer what do you think?

mikechengwei: yes, you re right.

charts/tidb-cluster/values.yaml
```diff
@@ -77,6 +77,7 @@ pd:
   # we can only specify clusterIP and loadBalancerIP now
   service:
     clusterIP: "None"
+    # portName: "<the default name will be used>"
```
cofyc: can you replace `<the default name will be used>` with the actual port name used for pd? I guess it's `client` here?

charts/tidb-cluster/values.yaml
```diff
@@ -396,6 +397,7 @@ tidb:
   service:
     type: NodePort
     exposeStatus: true
+    # portName: "<the default name will be used>"
```
cofyc: can you replace `<the default name will be used>` with the actual default name used for tidb?

charts/tidb-cluster/values.yaml
Yisaer:
```suggestion
      portName: "http-pd"
```

charts/tidb-cluster/values.yaml
Yisaer:
```suggestion
      portName: mysql
```

charts/tidb-cluster/values.yaml
cofyc: the default value in our code is `client`, `http-pd` is the value he used in the unit tests

https://github.com/pingcap/tidb-operator/blob/7360679e33613e7c5ac16bba97a1a011ddf62d1c/pkg/manager/member/pd_member_manager.go#L446-L453

mikechengwei: If you do this, there will be a problem of backward compatibility. When upgrading the operator, trigger the upgrade of the cluster.

charts/tidb-cluster/values.yaml
mikechengwei: If you do this, there will be a problem of backward compatibility. When upgrading the operator, trigger the upgrade of the cluster.

charts/tidb-cluster/values.yaml
mikechengwei: We set value in yaml file ,but we can't change value in code.

cofyc: I agree. See my previous comments and use actual default values in comments. then people will know port name used if they didn’t set these fields. This is common practice in configuration files.

Yisaer: Thanks for explaining, I understand now.

cofyc:
```suggestion
    # portName: "client"
```

charts/tidb-cluster/values.yaml
cofyc:
```suggestion
    # portName: "mysql-client"
```

mikechengwei:
@Yisaer can you help me re approve this pr?

PR #1831, opened 2020-02-29
Make evict leader scheduler compatitable 

Yisaer:
The result of the `scheduler` pdapi is in the following format as I showed below, the `store id` of the evict-leader-scheduler would be added at the end the of the scheduler name.

evict-2

However, in TiDB 4.0 release, the `store-id` would be omitted as I showed below.

evict-1

In this request, I add new method to make getting evict scheduler store id compatible between the current version and the coming 4.0 version.  

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/manager/member/pd_upgrader.go
+++ b/pkg/manager/member/pd_upgrader.go
@@ -74,6 +74,7 @@ func (pu *pdUpgrader) gracefulUpgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Sta
 
 	if controller.PodWebhookEnabled {
 		setUpgradePartition(newSet, 0)
+		return nil
 	}
 
 	setUpgradePartition(newSet, *oldSet.Spec.UpdateStrategy.RollingUpdate.Partition)
--- a/pkg/manager/member/tikv_upgrader.go
+++ b/pkg/manager/member/tikv_upgrader.go
@@ -89,6 +89,7 @@ func (tku *tikvUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Stateful
 
 	if controller.PodWebhookEnabled {
 		setUpgradePartition(newSet, 0)
+		return nil
 	}
 
 	setUpgradePartition(newSet, *oldSet.Spec.UpdateStrategy.RollingUpdate.Partition)
--- a/pkg/webhook/pod/tikv_creater.go
+++ b/pkg/webhook/pod/tikv_creater.go
@@ -99,10 +99,7 @@ func filterLeaderEvictScheduler(evictLeaderSchedulers []string, pdClient pdapi.P
 		if err != nil {
 			return schedulerIds, err
 		}
-		if c.Schedule != nil {
-			if c.Schedule.SchedulersPayload == nil {
-				c.Schedule.SchedulersPayload = map[string]string{}
-			}
+		if c.Schedule != nil && c.Schedule.SchedulersPayload != nil {
 			v, ok := c.Schedule.SchedulersPayload[evictSchedulerLeader]
 			if ok {
 				payload := &Payload{}
--- a/pkg/webhook/pod/tikv_creater.go
+++ b/pkg/webhook/pod/tikv_creater.go
@@ -90,7 +90,7 @@ func (pc *PodAdmissionControl) admitCreateTiKVPod(pod *core.Pod, tc *v1alpha1.Ti
 	return util.ARSuccess()
 }
 
-// This is method is to make compatible between old pdapi version and 4.0 pdapi version.
+// This method is to make compatible between old pdapi version and 4.0 pdapi version.
 // To get more detail, see: https://github.com/pingcap/tidb-operator/pull/1831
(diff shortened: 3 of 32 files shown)


pkg/manager/member/pd_upgrader.go
```diff
@@ -74,6 +74,7 @@ func (pu *pdUpgrader) gracefulUpgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Sta
 
 	if controller.PodWebhookEnabled {
 		setUpgradePartition(newSet, 0)
+		return nil
```
Yisaer: During my test, I found the bug in upgrading and I fixed it.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -89,6 +89,7 @@ func (tku *tikvUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Stateful
 
 	if controller.PodWebhookEnabled {
 		setUpgradePartition(newSet, 0)
+		return nil
```
Yisaer: ditto.

pkg/webhook/pod/tikv_creater.go
```diff
@@ -84,3 +89,37 @@ func (pc *PodAdmissionControl) admitCreateTiKVPod(pod *core.Pod, tc *v1alpha1.Ti
 
 	return util.ARSuccess()
 }
+
+// This is method is to make compatible between old pdapi version and 4.0 pdapi version.
```
DanielZhangQD:
```suggestion
// This method is to make compatible between old pdapi version and 4.0 pdapi version.
```

pkg/webhook/pod/tikv_creater.go
```diff
@@ -84,3 +89,37 @@ func (pc *PodAdmissionControl) admitCreateTiKVPod(pod *core.Pod, tc *v1alpha1.Ti
...
+// To get more detail, see: https://github.com/pingcap/tidb-operator/pull/1831
+func filterLeaderEvictScheduler(evictLeaderSchedulers []string, pdClient pdapi.PDClient) (sets.String, error) {
+	schedulerIds := sets.String{}
+	if len(evictLeaderSchedulers) == 1 && evictLeaderSchedulers[0] == evictSchedulerLeader {
+		c, err := pdClient.GetConfig()
+		if err != nil {
+			return schedulerIds, err
+		}
+		if c.Schedule != nil {
+			if c.Schedule.SchedulersPayload == nil {
```
DanielZhangQD: If it's nil, why still need to continue to the following codes?

pkg/webhook/pod/tikv_creater.go
```diff
@@ -84,3 +89,37 @@ func (pc *PodAdmissionControl) admitCreateTiKVPod(pod *core.Pod, tc *v1alpha1.Ti
...
+				c.Schedule.SchedulersPayload = map[string]string{}
+			}
+			v, ok := c.Schedule.SchedulersPayload[evictSchedulerLeader]
```
DanielZhangQD: Could you please give an example of the `c`? or any doc link about this new behavior?
I did not find the `c.Schedule.SchedulersPayload` in my cluster.

Yisaer: It's the new payload in pd 4.0

Yisaer: Here is an example for the payload response：
```bash
"balance-hot-region-scheduler": "null",
"balance-leader-scheduler": "{\"name\":\"balance-leader-scheduler\",\"ranges\":[{\"start-key\":\"\",\"end-key\":\"\"}]}",
"balance-region-scheduler": "{\"name\":\"balance-region-scheduler\",\"ranges\":[{\"start-key\":\"\",\"end-key\":\"\"}]}","evict-leader-scheduler": "{\"store-id-ranges\":{\"4\":[{\"start-key\":\"\",\"end-key\":\"\"}]}}",
"label-scheduler": "{\"name\":\"label-scheduler\",\"ranges\":[{\"start-key\":\"\",\"end-key\":\"\"}]}"

```

pkg/webhook/pod/tikv_creater.go
Yisaer: Not necessary,  just solve the nil or empty payload with the same method.

pkg/webhook/pod/tikv_creater.go
```diff
@@ -84,3 +89,37 @@ func (pc *PodAdmissionControl) admitCreateTiKVPod(pod *core.Pod, tc *v1alpha1.Ti
...
+	schedulerIds := sets.String{}
+	if len(evictLeaderSchedulers) == 1 && evictLeaderSchedulers[0] == evictSchedulerLeader {
+		c, err := pdClient.GetConfig()
+		if err != nil {
+			return schedulerIds, err
+		}
+		if c.Schedule != nil {
+			if c.Schedule.SchedulersPayload == nil {
+				c.Schedule.SchedulersPayload = map[string]string{}
+			}
```
DanielZhangQD: You can just change to:
```
if c.Schedule != nil && c.Schedule.SchedulersPayload != nil {
```

Yisaer: updatd, PTAL

PR #1833, opened 2020-02-29
Fix hot region label setting for tikv auto-scaling

Yisaer:

When tikv is scaled-out by auto-scaler, It should be added special hot region label by starting configuration instead of pdapi after started.

This request use mutation pod webhook to add env for the tikv created by auto scaler, which would inject special label durint it starting.

Related changes

```release-note
The TiKV created by the Auto-scaler would only schedule the hot region if `admission.mutation.pods` is enabled.
```

### DIFF
--- a/charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
+++ b/charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
@@ -39,10 +39,7 @@ ARGS="--pd={{ template "cluster.scheme" . }}://${CLUSTER_NAME}-pd:2379 \
 --config=/etc/tikv/tikv.toml
 "
 
-if [ ! -z "${STORE_LABELS:-}" ]; then
-  LABELS=" --labels ${STORE_LABELS} "
-  ARGS="${ARGS}${LABELS}"
-fi
+{{ .Values.tikv.postArgScript }}
 
 echo "starting tikv-server ..."
 echo "/tikv-server ${ARGS}"
--- a/charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
+++ b/charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
@@ -39,6 +39,11 @@ ARGS="--pd={{ template "cluster.scheme" . }}://${CLUSTER_NAME}-pd:2379 \
 --config=/etc/tikv/tikv.toml
 "
 
+if [ ! -z "${STORE_LABELS:-}" ]; then
+  LABELS=" --labels ${STORE_LABELS} "
+  ARGS="${ARGS}${LABELS}"
+fi
+
 echo "starting tikv-server ..."
 echo "/tikv-server ${ARGS}"
 exec /tikv-server ${ARGS}
--- a/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
+++ b/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
@@ -190,5 +190,9 @@ webhooks:
         apiGroups: [ "pingcap.com"]
         apiVersions: ["v1alpha1"]
         resources: ["tidbclusters"]
+      - operations: ["CREATE"]
+        apiGroups: [""]
+        apiVersions: ["v1"]
+        resources: ["pods"]
 {{- end }}
 {{- end }}
--- a/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
+++ b/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
@@ -194,8 +194,8 @@ webhooks:
       caBundle: null
       {{- end }}
     rules:
-      - operations: [ "CREATE" ]
-        apiGroups: [ ""]
+      - operations: ["CREATE"]
+        apiGroups: [""]
         apiVersions: ["v1"]
         resources: ["pods"]
 {{- end }}
--- a/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
+++ b/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
@@ -162,4 +162,41 @@ webhooks:
         apiVersions: ["v1alpha1"]
         resources: ["tidbclusters"]
 {{- end }}
+---
+{{- if .Values.admissionWebhook.mutation.pods }}
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutation-tidb-pod-webhook-cfg
+  labels:
+    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/component: admission-webhook
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+webhooks:
+  - name: podadmission.tidb.pingcap.com
+    {{- if semverCompare ">=1.15-0" .Capabilities.KubeVersion.GitVersion }}
+    objectSelector:
+      matchLabels:
+        "app.kubernetes.io/managed-by": "tidb-operator"
+        "app.kubernetes.io/name": "tidb-cluster"
+    {{- end }}
+    failurePolicy: {{ .Values.admissionWebhook.failurePolicy.mutation | default "Ignore" }}
+    clientConfig:
+      service:
+        name: kubernetes
+        namespace: default
+        path: "/apis/admission.tidb.pingcap.com/v1alpha1/mutatingreviews"
+      {{- if .Values.admissionWebhook.cabundle }}
+      caBundle: {{ .Values.admissionWebhook.cabundle | b64enc }}
+      {{- else }}
+      caBundle: null
+      {{- end }}
+    rules:
+      - operations: [ "CREATE" ]
+        apiGroups: [ ""]
+        apiVersions: ["v1"]
+        resources: ["pods"]
+{{- end }}
 {{- end }}
--- a/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
+++ b/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
@@ -190,9 +190,5 @@ webhooks:
         apiGroups: [ "pingcap.com"]
         apiVersions: ["v1alpha1"]
         resources: ["tidbclusters"]
-      - operations: ["CREATE"]
-        apiGroups: [""]
-        apiVersions: ["v1"]
-        resources: ["pods"]
 {{- end }}
 {{- end }}
--- a/pkg/controller/tidbcluster/tidb_cluster_controller.go
+++ b/pkg/controller/tidbcluster/tidb_cluster_controller.go
@@ -139,7 +139,6 @@ func NewController(
 				pdFailover,
 			),
 			mm.NewTiKVMemberManager(
-				cmInformer.Lister(),
 				pdControl,
 				setControl,
 				svcControl,
--- a/pkg/controller/tidbcluster/tidb_cluster_controller.go
+++ b/pkg/controller/tidbcluster/tidb_cluster_controller.go
@@ -139,6 +139,7 @@ func NewController(
 				pdFailover,
 			),
 			mm.NewTiKVMemberManager(
+				cmInformer.Lister(),
 				pdControl,
 				setControl,
 				svcControl,
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/controller"
+	"github.com/pingcap/tidb-operator/pkg/features"
 	"github.com/pingcap/tidb-operator/pkg/label"
 	"github.com/pingcap/tidb-operator/pkg/manager"
 	"github.com/pingcap/tidb-operator/pkg/pdapi"
@@ -44,6 +45,7 @@ const (
 
 // tikvMemberManager implements manager.Manager.
 type tikvMemberManager struct {
+	cmLister                     corelisters.ConfigMapLister
 	setControl                   controller.StatefulSetControlInterface
 	svcControl                   controller.ServiceControlInterface
 	pdControl                    pdapi.PDControlInterface
@@ -61,7 +63,9 @@ type tikvMemberManager struct {
 }
 
 // NewTiKVMemberManager returns a *tikvMemberManager
-func NewTiKVMemberManager(pdControl pdapi.PDControlInterface,
+func NewTiKVMemberManager(
+	cmLister corelisters.ConfigMapLister,
+	pdControl pdapi.PDControlInterface,
 	setControl controller.StatefulSetControlInterface,
 	svcControl controller.ServiceControlInterface,
 	certControl controller.CertControlInterface,
@@ -75,6 +79,7 @@ func NewTiKVMemberManager(pdControl pdapi.PDControlInterface,
 	tikvScaler Scaler,
 	tikvUpgrader Upgrader) manager.Manager {
 	kvmm := tikvMemberManager{
+		cmLister:     cmLister,
 		pdControl:    pdControl,
 		podLister:    podLister,
 		nodeLister:   nodeLister,
@@ -183,6 +188,12 @@ func (tkmm *tikvMemberManager) syncStatefulSetForTidbCluster(tc *v1alpha1.TidbCl
 		return err
 	}
 
+	if features.DefaultFeatureGate.Enabled(features.AutoScaling) {
+		if err := tkmm.syncTiKVAutoScalingConfigMap(tc); err != nil {
+			return err
+		}
+	}
+
 	newSet, err := getNewTiKVSetForTidbCluster(tc, cm)
 	if err != nil {
 		return err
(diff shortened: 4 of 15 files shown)


pkg/manager/member/tikv_member_manager.go
```diff
@@ -284,6 +295,48 @@ func (tkmm *tikvMemberManager) syncTiKVConfigMap(tc *v1alpha1.TidbCluster, set *
 	return tkmm.typedControl.CreateOrUpdateConfigMap(tc, newCm)
 }
 
+func (tkmm *tikvMemberManager) syncTiKVAutoScalingConfigMap(tc *v1alpha1.TidbCluster) error {
```
Yisaer: If auto-scaling is enabled, we would create and sync a new configmap only for the tikv created by auto-scaler.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -284,6 +295,48 @@ func (tkmm *tikvMemberManager) syncTiKVConfigMap(tc *v1alpha1.TidbCluster, set *
...
+
+	cm, err := tkmm.cmLister.ConfigMaps(tc.Namespace).Get(controller.TiKVMemberName(tc.Name))
+	if err != nil {
+		return err
+	}
+	v, ok := cm.Data["config-file"]
+	if !ok {
+		return fmt.Errorf("")
```
aylei: error msg?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -284,6 +295,48 @@ func (tkmm *tikvMemberManager) syncTiKVConfigMap(tc *v1alpha1.TidbCluster, set *
...
+	}
+	if config.Server == nil {
+		config.Server = &v1alpha1.TiKVServerConfig{}
+	}
+	if config.Server.Labels == nil {
+		config.Server.Labels = map[string]string{}
+	}
+	// TODO: add document to explain the hot region label
+	config.Server.Labels["specialUse"] = "hotRegion"
+	klog.Infof("%v", config)
```
aylei: remove or refine this log

Yisaer: updated.

pkg/manager/member/tikv_member_manager.go
Yisaer: updated.

Yisaer:
> Can we do this in `setStoreLabelsForTiKV`?

I'm afraid we can't, the special hot region label is required to be set in the starting configuration by @nolouch 's advice.

aylei:
The labels about locations are also specified in start args in our official documentation, but it is also possible to label it through the PD API.

pkg/controller/tidbcluster/tidb_cluster_controller.go
```diff
@@ -139,6 +139,7 @@ func NewController(
 				pdFailover,
 			),
 			mm.NewTiKVMemberManager(
+				cmInformer.Lister(),
```
DanielZhangQD: I recall that this would cause tidb operator to use much more memory especially for large clusters, e.g. our IDC. @aylei

Yisaer:
@aylei  The current realization for adding label is using pdapi during tikv syncing. It can't work during our test while the way using start args could.

pkg/manager/member/tikv_scaler.go
```diff
@@ -217,28 +216,6 @@ func (tsd *tikvScaler) SyncAutoScalerAnn(tc *v1alpha1.TidbCluster, actual *apps.
...
-				}
-				if !ok {
-					return fmt.Errorf("tc[%s/%s]'s pod[%s] failed to add special hot region label", tc.Namespace, tc.Name, podName)
-				}
-				break
-			}
-		}
-	}
 	return nil
 }
```
Yisaer: previously we add the special region labels by using pdapi during tikv syncing process, which is successful to add the label we want, but the special scheduling need for the tikv pod created by auto-scaler can't be met. Thats' why we choose the starting args ways and it works. @aylei

pkg/manager/member/tikv_scaler.go
```diff
@@ -217,28 +216,6 @@ func (tsd *tikvScaler) SyncAutoScalerAnn(tc *v1alpha1.TidbCluster, actual *apps.
...
-				if err != nil {
-					return err
-				}
-				if !ok {
-					return fmt.Errorf("tc[%s/%s]'s pod[%s] failed to add special hot region label", tc.Namespace, tc.Name, podName)
-				}
-				break
-			}
-		}
-	}
```
Yisaer: previously we add the special region labels by using pdapi during tikv syncing process, which is successful to add the label we want, but the special scheduling need for the tikv pod created by auto-scaler can't be met. Thats' why we choose the starting args ways and it works. @aylei

cofyc: I don't understand this. the label is added successfully, but does not work?

cofyc:
it's strange to modify tikv config map. if the code goes too complicated, it's hard to test and error-prone. can we ask the tidb team to support setting hotRegion via PD api?

In my understanding, this is an optional feature. I think it's fine to have some tidb version requirements, e.g. supporting hotRegion after tidb 3.0.x or 4.0.0.

Yisaer:
First, the existed tikv configmap is not modified. Operator created a tikv configmap replica only when the auto-scaling in enabled and it's only for the tikv pod created by auto-scaler.

The hot special region label is to notice the pd only scheduling the hot region to the tikv store before scheduling. So if we add the labels by pdapi instead of start args, the special scheduling need wouldn't be met as the scheduling already begin. @cofyc 

Could you give us some details and opnions?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -183,6 +188,12 @@ func (tkmm *tikvMemberManager) syncStatefulSetForTidbCluster(tc *v1alpha1.TidbCl
 		return err
 	}
 
+	if features.DefaultFeatureGate.Enabled(features.AutoScaling) {
+		if err := tkmm.syncTiKVAutoScalingConfigMap(tc); err != nil {
+			return err
+		}
+	}
+
```
Yisaer: > In my understanding, this is an optional feature. I think it's fine to have some tidb version requirements, e.g. supporting setting hotRegion automatically after tidb 3.0.x or 4.0.0.

The replica of tikv configmap only for auto-scaling is controlled by featureGate, I think this met your requirement.
@cofyc

pkg/webhook/pod/pod_mutater.go
```diff
@@ -0,0 +1,96 @@
...
+		if err != nil {
+			return util.ARFail(err)
+		}
+	}
+
+	patch, err := util.CreateJsonPatch(original, pod)
+	if err != nil {
+		return util.ARFail(err)
+	}
+	return util.ARPatch(patch)
```
DanielZhangQD: This block should be covered by `if features.DefaultFeatureGate.Enabled(features.AutoScaling)`?

nolouch:
@Yisaer  In the way of passing the API, TiKV was not labeled for a short period of time. During this time PD does not know the special attributes of this TiKV, and it will produce some unexpected schedules.

pkg/manager/member/tikv_scaler.go
DanielZhangQD: There is a time duration between tikv starts and the label is added via PDAPI, during that time, normal regions would be scheduled to the new tikv.

pkg/webhook/pod/pod_mutater.go
Yisaer: In my design, the mutation of the pod would have a mutation chain and the auto-scaling mutation is part of these.

So the pod spec would still be patched whether there is any change.

Yisaer:
> @Yisaer @cofyc In the way of passing the API, TiKV was not labeled for a short period of time. During this time PD does not know the special attributes of this TiKV, and it will produce some unexpected schedules.

In another word, we should label the store by start args, right? @nolouch

charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
```diff
@@ -162,4 +162,41 @@ webhooks:
...
+        path: "/apis/admission.tidb.pingcap.com/v1alpha1/mutatingreviews"
+      {{- if .Values.admissionWebhook.cabundle }}
+      caBundle: {{ .Values.admissionWebhook.cabundle | b64enc }}
+      {{- else }}
+      caBundle: null
+      {{- end }}
+    rules:
+      - operations: [ "CREATE" ]
+        apiGroups: [ ""]
+        apiVersions: ["v1"]
```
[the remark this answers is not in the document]
Yisaer: updated.

cofyc:
thanks for clarifying this! I understand the root cause now.

cofyc:
my concern is this solution is too complicated, we need to maintain another ConfigMap. IIUC, the problem is we need to tell the pods created by auto scaler to start with `specialUse=hotRegion` label, because setting this label after TiKV starts may be late. how about this?

1) we can tell pods created by auto scaler via ENV, e.g. passing an env in pod mutation
2) instead of use a different configmap, tikv startup script can add cli flag ` --labels` to pass `hotRegion ` label if it founds the pod was created by auto scaler

for 1), tell pods some more information can be a general solution for features in the future.

cofyc:
another concern about hot region is if people increase replicas of tikv on TidbCluster CR, should we remove the hotRegion label from the pod?

the benefit of dynamic configuration is we can adjust this on the fly.

Yisaer:
Currently I didn't figure out the way to add the special region label by start script for this specific case. But I agreed to find a general way to customize the tikv initialization. As the current special configmap is fully strict by featrueGate, I suggest we could use the current method.

And for your another concern, the manually increasing replicas for tikv wouldn't be added special region label. The hot special region label would only be added to the tikv which created by the auto-scaler, this would be ensured by the auto-scaler controller.

For example, if there is 3 tikv already, and the auto-scaler created the forth one.
The tikv-3 would be added the special label. If we removed the auto-scaler and scale-in to 3 replicas and then scale-out to 4 replicas. The newly tikv-3 wouldn't be added sperical region label as it is not created by auto-scaler. @cofyc

cofyc:
I mean passing the labels via tikv-server CLI flags instead of config map.

```
/ # ./tikv-server -h
TiKV 
Release Version:   3.0.8
Git Commit Hash:   0d3168cfd8224fbc48a07796df83ddac0fbcbf46
Git Commit Branch: heads/refs/tags/v3.0.8
UTC Build Time:    2019-12-31 09:15:35
Rust Version:      rustc 1.37.0-nightly (0e4a56b4b 2019-06-13)
The TiKV Authors
A distributed transactional key-value database powered by Rust and Raft

USAGE:
    tikv-server [FLAGS] [OPTIONS]

FLAGS:
        --config-check           Check config file validity and exit
    -h, --help                   Prints help information
        --print-sample-config    Print a sample config to stdout
    -V, --version                Prints version information

OPTIONS:
    -A, --addr <IP:PORT>              Set the listening address
        --advertise-addr <IP:PORT>    Set the advertise listening address for client communication
        --capacity <CAPACITY>         Set the store capacity
    -C, --config <FILE>               Set the configuration file
    -s, --data-dir <PATH>             Set the directory used to store data
        --labels <KEY=VALUE>...       Sets server labels
    -f, --log-file <FILE>             Sets log file
    -L, --log-level <LEVEL>           Set the log level [possible values: trace, debug, info, warn, warning, error,
                                      critical]
        --metrics-addr <IP:PORT>      Sets Prometheus Pushgateway address
        --pd-endpoints <PD_URL>...    Sets PD endpoints
        --status-addr <IP:PORT>       Set the HTTP listening address for the status report service
```

Yisaer:
That method is more helpful, I will change the way to it soom.

cofyc:
> For example, if there is 3 tikv already, and the auto-scaler created the forth one.
The tikv-3 would be added the special label. If we removed the auto-scaler and scale-in to 3 replicas and then scale-out to 4 replicas. The newly tikv-3 wouldn't be added sperical region label as it is not created by auto-scaler.

before we removed the auto-scaler and scale-in to 3 replicas, the people increases the replicas of tikv, e.g.

- tikv replicas 3, pods: 0, 1, 2
- auto scaler increases the replicas to 4, pods: 0, 1, 2, 3(hotRegion)
- the user increases the base replicas of tikv to 4, and auto-scaler think it needs an extra replica: 0, 1, 2, 3 (hotRegion), 4

maybe my example is not right. my concern is I'm not sure if we need to remove the hotRegion label later. if we don't need, I think it's fine to set it via CLI flags or config file. or we can remove it via PD api if the label config in PD will overwrite labels in tikv-server CLI flags or config file.

cofyc:
@nolouch
hi, do label configs setting via PD API overwrite label configs in tikv-server from CLI flags and config file?

tennix:
This is actually heterogeneous deployment for TiKV. It's better handled by multiple StatefulSets, we had the multiple statefulset design before, but it's complicated and requires a lot of work.

I'm wondering if 30s is short enough for PD to avoid too many normal regions scheduled to the newly added tikv pods. @nolouch

Yisaer:
After I tried by tikv cli, I found that if we add labels by tikv cli, we can't load the labels defined in the configuration. 
It seems the tikv cli configuration would overwrite the configuration from configuration file. @cofyc 
```bash
{
  "count": 2,
  "stores": [
    {
      "store": {
        "id": 1,
        "address": "basic-tikv-0.basic-tikv-peer.tidb-admin.svc:20160",
        "labels": [
          {
            "key": "a",
            "value": "b"
          }
        ],
        "version": "4.1.0-alpha",
        "status_address": "0.0.0.0:20180",
        "git_hash": "4761287587c9dcbcec1d86257a9c9ef1eab50f74",
        "start_timestamp": 1583317473,
        "binary_path": "/preset_daemon/tikv/bin/tikv-server",
        "last_heartbeat": 1583317603488211900,
        "state_name": "Up"
      },
      "status": {
        "capacity": "102.2GiB",
        "available": "51.82GiB",
        "used_size": "32.35MiB",
        "leader_count": 21,
        "leader_weight": 1,
        "leader_score": 21,
        "leader_size": 21,
        "region_count": 21,
        "region_weight": 1,
        "region_score": 21,
        "region_size": 21,
        "start_ts": "2020-03-04T10:24:33Z",
        "last_heartbeat_ts": "2020-03-04T10:26:43.4882119Z",
        "uptime": "2m10.4882119s"
      }
    },
    {
      "store": {
        "id": 44,
        "address": "basic-tikv-1.basic-tikv-peer.tidb-admin.svc:20160",
        "labels": [
          {
            "key": "specialUse",
            "value": "hotRegion"
          }
        ],
        "version": "4.1.0-alpha",
        "status_address": "0.0.0.0:20180",
        "git_hash": "4761287587c9dcbcec1d86257a9c9ef1eab50f74",
        "start_timestamp": 1583317536,
        "binary_path": "/preset_daemon/tikv/bin/tikv-server",
        "last_heartbeat": 1583317596473279700,
        "state_name": "Up"
      },
      "status": {
        "capacity": "102.2GiB",
        "available": "51.82GiB",
        "used_size": "31.5MiB",
        "leader_count": 0,
        "leader_weight": 1,
        "leader_score": 0,
        "leader_size": 0,
        "region_count": 0,
        "region_weight": 1,
        "region_score": 0,
        "region_size": 0,
        "start_ts": "2020-03-04T10:25:36Z",
        "last_heartbeat_ts": "2020-03-04T10:26:36.4732797Z",
        "uptime": "1m0.4732797s"
      }
    }
  ]
}
```

Yisaer:
Now webhook would inject ENV instead of change the mounted configmap. PTAL @cofyc @DanielZhangQD

DanielZhangQD:
What about the issue mentioned in this [comment](https://github.com/pingcap/tidb-operator/pull/1833#issuecomment-594375615) @Yisaer

Yisaer:
There existed the way to remove the hot region label manually basicly by passing the label setting by pdapi and removing the mark in the tidbcluster annotation. We would point it out in the released document. @DanielZhangQD @cofyc

nolouch:
@tennix  In the testing, we see produce many unexpected operators.  because the cluster is small, it seems to have a big impact.  
@cofyc  pdctl will overwrite it, but if tikv restart, it will overwrite back from config.

charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl
```diff
@@ -39,6 +39,11 @@ ARGS="--pd={{ template "cluster.scheme" . }}://${CLUSTER_NAME}-pd:2379 \
 --config=/etc/tikv/tikv.toml
 "
 
+if [ ! -z "${STORE_LABELS:-}" ]; then
```
weekface: can we add this script snippets to `.Values.tikv.preStartScript` to avoid modify the `_start_tikv.sh.tpl`?

Yisaer: It's ok to me. I would change to use snippets soon.

Yisaer: updated

Yisaer:
@cofyc @DanielZhangQD @aylei @weekface  PTAL

charts/tidb-operator/templates/admission/admission-webhook-registration.yaml
```diff
@@ -162,4 +162,41 @@ webhooks:
...
+    app.kubernetes.io/component: admission-webhook
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+webhooks:
+  - name: podadmission.tidb.pingcap.com
+    {{- if semverCompare ">=1.15-0" .Capabilities.KubeVersion.GitVersion }}
+    objectSelector:
+      matchLabels:
+        "app.kubernetes.io/managed-by": "tidb-operator"
+        "app.kubernetes.io/name": "tidb-cluster"
+    {{- end }}
```
cofyc: what's the purpose of this selector?

Yisaer: To make admission webhook only check the target with the matched labels. This would narrow the operator admission webhook affecting range, with our tidbcluster pods only.

cofyc:
we may forget to cherry-pick some PRs which conflicts with this PR into 1.1

pkg/webhook/pod/pod_mutater.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return err
+	}
+	sets := operatorUtils.GetAutoScalingOutSlots(tc, v1alpha1.TiKVMemberType)
+	if !sets.Has(ordinal) {
+		return nil
+	}
+
+	cmName := controller.TiKVMemberName(tc.Name)
+	cm, err := pc.kubeCli.CoreV1().ConfigMaps(tc.Namespace).Get(cmName, metav1.GetOptions{})
```
DanielZhangQD: What if the `ConfigUpdateStrategyRollingUpdate` is set for the configmap?

Yisaer:
I would cp this commit manually.

pkg/webhook/pod/pod_mutater.go
Yisaer: fixed in https://github.com/pingcap/tidb-operator/pull/1871

PR #1844, opened 2020-03-02
use cert-manager to create and renew tidb-server certificates (self-signed example)

cofyc:
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


Smana:
Even if self-signed certificate are fine, would that be possible to configure an issuer ? (let's say vault pki)

cofyc:
> Even if self-signed certificate are fine, would that be possible to configure an issuer ? (let's say vault pki)

yes, it's ok to use other Issuer types. I use self-sign ca as an example, then anyone can try it.

PR #1850, opened 2020-03-03
re comment pd service yaml value

mikechengwei:
# What is changed and how does it work?
Original pd clusterIP is none, it will be parsed "None" string. It's an error.
So I comment pd service config，and remind users of this feature.
### Check List 
Tests 
 - Unit test


charts/tidb-cluster/values.yaml
```diff
@@ -76,7 +76,10 @@ pd:
   # pd Service
   # we can only specify clusterIP and loadBalancerIP now
   service:
```
```suggestion
  service: {}
```

?

mikechengwei: Thanks, I have sloved

cofyc:
ah, it conflicts with another PR you created, you need to resolve it.

mikechengwei:
ok.

Issue #1439, opened 2019-12-30
start tidb-server with `-advertise-address`


nolouch:
## Feature Request

**Is your feature request related to a problem? Please describe:**

For let PD  to know the tidb address.
**Describe the feature you'd like:**

That need  `tidb-server` startup with `--advertise-address`, like TiKV:
```
❯ kubectl exec -it hot-new-tikv-1 -n xxx sh
/ # ps
PID   USER     TIME  COMMAND
    1 root     23d18 /tikv-server --pd=hot-new-pd:2379 --advertise-addr=hot-new-tikv-1.hot-new-tikv-peer.csn.svc:20160 --addr=0.0.0.0:20160 --status-addr=0.0
93992 root      0:00 sh
```

tennix:
This argument should be added to the startup script of the configmap.

aylei:
This change will trigger rolling-update of the tidb-servers if the configmap rollout strategy is RollingUpdate

What's the feature related to this change? @nolouch

nolouch:
@aylei  PD wants to know all `TiDB` addresses, and do some proxy worker in TiDB, for example, which table the key belongs to?

tennix:
Yes, this makes PD knows all the cluster component info. So we can retrieve the single source of true from PD for cluster status.

PR #1859, opened 2020-03-04
support starting tidb-server with `-advertise-address` parameter

LinuxGit:
Closes https://github.com/pingcap/tidb-operator/issues/1439

v2.1.0 and later versions have this argument:
https://github.com/pingcap/tidb/blob/v2.1.0/config/config.go#L48

start tidb-server with `-advertise-address` parameter, refer to https://github.com/pingcap/tidb/pull/7078.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test

When `EnableAdvertiseAddress` is set to false:
```
/ # ps
PID   USER     TIME  COMMAND
    1 root      0:00 /tidb-server --store=tikv --host=0.0.0.0 --path=db-pd:2379 --config=/etc/tidb/tidb.toml --log-slow-query=/var/log/tidb/slowlog
```

When `EnableAdvertiseAddress` is set to true:
```
# kubectl exec -it db-tidb-0 sh -n tidb10 -c tidb
/ # ps
PID   USER     TIME  COMMAND
    1 root      0:01 /tidb-server --store=tikv --advertise-address=db-tidb-0.db-tidb-peer.tidb10.svc --host=0.0.0.0 --path=db-pd:2379 --config=/etc/tidb/tidb
```

 - No code

Code changes

 - Has Go code change

Related changes

1.1

```release-note
support starting tidb-server with `-advertise-address` parameter
```

### DIFF
--- a/charts/tidb-cluster/templates/scripts/_start_tidb.sh.tpl
+++ b/charts/tidb-cluster/templates/scripts/_start_tidb.sh.tpl
@@ -26,7 +26,10 @@ then
     tail -f /dev/null
 fi
 
+# Use HOSTNAME if POD_NAME is unset for backward compatibility.
+POD_NAME=${POD_NAME:-$HOSTNAME}
 ARGS="--store=tikv \
+--advertise-address=${POD_NAME}.${HEADLESS_SERVICE_NAME}.${NAMESPACE}.svc \
 --host=0.0.0.0 \
 --path=${CLUSTER_NAME}-pd:2379 \
 --config=/etc/tidb/tidb.toml
--- a/charts/tidb-cluster/templates/scripts/_start_tidb.sh.tpl
+++ b/charts/tidb-cluster/templates/scripts/_start_tidb.sh.tpl
@@ -29,7 +29,9 @@ fi
 # Use HOSTNAME if POD_NAME is unset for backward compatibility.
 POD_NAME=${POD_NAME:-$HOSTNAME}
 ARGS="--store=tikv \
+{{- if .Values.tidb.enableAdvertiseAddress | default false }}
 --advertise-address=${POD_NAME}.${HEADLESS_SERVICE_NAME}.${NAMESPACE}.svc \
+{{- end }}
 --host=0.0.0.0 \
 --path=${CLUSTER_NAME}-pd:2379 \
 --config=/etc/tidb/tidb.toml
--- a/pkg/manager/member/tidb_member_manager.go
+++ b/pkg/manager/member/tidb_member_manager.go
@@ -580,6 +580,7 @@ func getNewTiDBHeadlessServiceForTidbCluster(tc *v1alpha1.TidbCluster) *corev1.S
 func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap) *apps.StatefulSet {
 	ns := tc.GetNamespace()
 	tcName := tc.GetName()
+	headlessSvcName := controller.TiDBPeerMemberName(tcName)
 	baseTiDBSpec := tc.BaseTiDBSpec()
 	instanceName := tc.GetInstanceName()
 	tidbConfigMap := controller.MemberConfigMapName(tc, v1alpha1.TiDBMemberType)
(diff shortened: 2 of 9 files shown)


pkg/manager/member/tidb_member_manager.go
```diff
@@ -712,10 +713,30 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
...
+			},
+		},
 		{
 			Name:  "CLUSTER_NAME",
 			Value: tc.GetName(),
 		},
+		{
+			Name:  "HEADLESS_SERVICE_NAME",
+			Value: headlessSvcName,
+		},
```
cofyc: any changes on the statefulset will trigger rolling upgrading when people upgrade tidb-operator 1.1 from 1.0.  if we require no upgrades in upgrading operator, we should find a new way.

cofyc: like hostNetwork feature, add a switch?

cofyc: what do you think? cc @weekface @Yisaer @aylei

LinuxGit: OK.  I've added side effects in PR description. I'd like to add a switch if we need backward compatibility.

cofyc: another solution is that operator 1.1 doesn't apply these new changes on tidb clusters managed by operator 1.0. to achieve this, we must maintain a version label on tidb cluster CR.

Yisaer: I think we could point it out it in our change log and release document.

DanielZhangQD: Actually, this behavior depends on the tidb version, which is available in the tc, can we check the version of tidb and decide if the sts.spec should be updated? @cofyc @LinuxGit

charts/tidb-cluster/templates/scripts/_start_tidb.sh.tpl
```diff
@@ -26,7 +26,12 @@ then
     tail -f /dev/null
 fi
 
+# Use HOSTNAME if POD_NAME is unset for backward compatibility.
+POD_NAME=${POD_NAME:-$HOSTNAME}
```
aylei: This line should also be protected by trigger

cofyc: this is fine because it works before and after the change, and it does not affect StatefulSet spec.

the new NAMESPACE/HEADLESS_SERVICE_NAME/POD_NAME envs should only be set on TiDB statefulset only when enableAdvertiseAddress is enabled.

pkg/manager/member/tidb_member_manager.go
cofyc: TiDB has this flag since 2.1, I'm not sure if we have clients using the TiDB before 2.1.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -712,10 +714,22 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 		slowLogFileEnvVal = slowQueryLogFile
 	}
 	envs := []corev1.EnvVar{
+		{
```
DanielZhangQD: The newly added ENVs should be controlled by `EnableAdvertiseAddress `, right?

pkg/manager/member/tidb_member_manager.go
```diff
@@ -734,6 +748,20 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 	if tc.IsTLSClusterEnabled() {
 		scheme = corev1.URISchemeHTTPS
 	}
+
+	podSpec := baseTiDBSpec.BuildPodSpec()
+	if baseTiDBSpec.HostNetwork() {
+		podSpec.DNSPolicy = corev1.DNSClusterFirstWithHostNet
+		envs = append(envs, corev1.EnvVar{
```
DanielZhangQD: If `EnableAdvertiseAddress` enabled but `HostNetwork` is not enabled, we do not need to set this env?

LinuxGit: Yes. If HostNetwork is not enabled, $HOSTNAME will be $POD_NAME. 
I've updated code, only set NAMESPACE/HEADLESS_SERVICE_NAME/POD_NAME env when enableAdvertiseAddress is enabled.

pkg/manager/member/tidb_member_manager.go
LinuxGit: @cofyc  I think I could use https://godoc.org/github.com/coreos/go-semver/semver pkg to compare version. `latest` is not a valid semantic versioning, I need to deal with it. I'm not sure if there're other tags that can't compare. 
I could set a env variable. So startup script in Helm doesn't need to compare version.
@tennix @weekface Do you think it need to be compatible with versions before v2.1.0?

tennix: The tidb 2.1 is not actively maintained, so I think it's ok to remove support for tidb 2.1.

LinuxGit: OK, the TiDB 2.1 has this flag. TiDB before 2.1 is not supported.

PR #1871, opened 2020-03-05
Fix wrong method to get tikv configmap in mutation webhook

Yisaer:
Currently the tikv Configmap wouldn't be fetched when we only using TidbCluster CR with `configUpdateStrategy: RollingUpdate`. This request get the configmap from the volume template from the coming tikv pod.

I haven tested all four situations( helm/cr, RollingUpdate/Inplace), all situtations work well
Fix the problem which would let the

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/webhook/pod/pod_mutater.go
+++ b/pkg/webhook/pod/pod_mutater.go
@@ -82,7 +82,7 @@ func (pc *PodAdmissionControl) tikvHotRegionSchedule(tc *v1alpha1.TidbCluster, p
 		return nil
 	}
 
-	cmName := controller.TiKVMemberName(tc.Name)
+	cmName := controller.MemberConfigMapName(tc, v1alpha1.TiKVMemberType)
 	cm, err := pc.kubeCli.CoreV1().ConfigMaps(tc.Namespace).Get(cmName, metav1.GetOptions{})
 	if err != nil {
 		klog.Infof("cm[%s/%s] found error,err %v", tc.Namespace, cmName, err)
--- a/pkg/webhook/pod/pod_mutater.go
+++ b/pkg/webhook/pod/pod_mutater.go
@@ -81,7 +81,7 @@ func (pc *PodAdmissionControl) tikvHotRegionSchedule(tc *v1alpha1.TidbCluster, p
 		return nil
 	}
 
-	cm, err := pc.getTikvConfigMap(tc)
+	cm, err := pc.getTikvConfigMap(tc, pod)
 	if err != nil {
 		klog.Infof("tc[%s/%s]'s tikv configmap not found error,err %v", tc.Namespace, tc.Name, err)
 		return err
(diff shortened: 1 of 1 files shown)


pkg/webhook/pod/pod_mutater.go
```diff
@@ -114,3 +112,18 @@ func (pc *PodAdmissionControl) tikvHotRegionSchedule(tc *v1alpha1.TidbCluster, p
...
+		return nil, err
+	}
+	cmlist, err := pc.kubeCli.CoreV1().ConfigMaps(tc.Namespace).List(metav1.ListOptions{LabelSelector: selector.String()})
+	if err != nil {
+		return nil, err
+	}
+	if len(cmlist.Items) < 1 {
+		return nil, fmt.Errorf("tc[%s/%s]'s tikv configmap failed to find", tc.Namespace, tc.Name)
+	}
+	return &cmlist.Items[0], nil
```
DanielZhangQD: After several updates of the config, there will be several configmaps for tikv, how can you make sure items[0] is the latest valid config?

Yisaer: good catch, currently I get the configmap name by the coming pod template.

pkg/webhook/pod/pod_mutater.go
```diff
@@ -114,3 +112,18 @@ func (pc *PodAdmissionControl) tikvHotRegionSchedule(tc *v1alpha1.TidbCluster, p
...
+	for _, v := range pod.Spec.Volumes {
+		if (v.Name == "config" || v.Name == "startup-script") && v.ConfigMap != nil {
+			cnName = v.ConfigMap.Name
+			break
+		}
+	}
+	if cnName == "" {
+		return nil, fmt.Errorf("tc[%s/%s] 's tikv configmap can't find", tc.Namespace, tc.Name)
+	}
+	return pc.kubeCli.CoreV1().ConfigMaps(tc.Namespace).Get(cnName, metav1.GetOptions{})
```
DanielZhangQD:
```suggestion
	return pc.kubeCli.CoreV1().ConfigMaps(tc.Namespace).Get(cnName, metav1.GetOptions{}), nil
```

pkg/webhook/pod/pod_mutater.go
```diff
@@ -82,10 +81,9 @@ func (pc *PodAdmissionControl) tikvHotRegionSchedule(tc *v1alpha1.TidbCluster, p
 		return nil
 	}
 
-	cmName := controller.TiKVMemberName(tc.Name)
-	cm, err := pc.kubeCli.CoreV1().ConfigMaps(tc.Namespace).Get(cmName, metav1.GetOptions{})
+	cm, err := pc.getTikvConfigMap(tc, pod)
 	if err != nil {
-		klog.Infof("cm[%s/%s] found error,err %v", tc.Namespace, cmName, err)
+		klog.Infof("tc[%s/%s]'s tikv configmap not found error,err %v", tc.Namespace, tc.Name, err)
```
DanielZhangQD:
```suggestion
		klog.Infof("tc[%s/%s]'s tikv %s configmap not found, error: %v", tc.Namespace, tc.Name, pod.Name, err)
```

PR #1876, opened 2020-03-06
Fix defaulting webhook error

Yisaer:
Remove the config defaulting logic in webhook which would cause error when using helm install

Related changes

```release-note
NONE
```


pkg/apis/pingcap/v1alpha1/defaulting/tidbcluster.go
```diff
@@ -51,9 +51,6 @@ func setTidbClusterSpecDefault(tc *v1alpha1.TidbCluster) {
 }
 
 func setTidbSpecDefault(tc *v1alpha1.TidbCluster) {
-	if tc.Spec.TiDB.Config == nil {
-		tc.Spec.TiDB.Config = &v1alpha1.TiDBConfig{}
-	}
```
cofyc: I'm not sure about this. @aylei could you have a look?

PR #1877, opened 2020-03-06
fix operator failover config invalid

mikechengwei:
 the config of operator failover not works. The value in yaml always true.
So we need to fix it .
### Check List 
Tests 
 - Unit test
Code changes


cofyc:
thanks for the fix!

PR #1903, opened 2020-03-11
add aws ami version link

### LINKED ISSUE #1607: How to pin a version of the EKS AMI
We used the terraform modules to provision some EKS worker nodes a bit over a month ago. Recently I ran terraform again with no changes, but terraform wanted changes to the launch configuration. It's due to the use of `most_recent = true` here: https://github.com/pingcap/tidb-operator/blob/master/deploy/modules/aws/tidb-cluster/data.tf#L7

This caused issues because the new version of the AMI lowered ulimits, causing TiKV to stop working. This has already been fixed master here. So it was kind of a pain to track down.

It'd be nice to be able to pin a version. I think by setting `var.worker_ami_name_filter` I should be able to pin a version? But I haven't been able to figure out what exactly I'm supposed to put there. Where do I get this version from? The var help says:

> could be set to a release from amazon-eks-ami, e.g. \"v20190220\

I think it's supposed to be the "Packer version" here https://docs.aws.amazon.com/eks/latest/userguide/eks-linux-ami-versions.html. It'd be nice if the docs were more clear.

DanielZhangQD:
fix #1607
Add aws ami version link to the variable description so that users can know exactly which version to set if they want to.
### Check List 

Tests 

   - `terraform init`, `terraform validate`, `terraform apply` works as expected.

Code changes

Related changes

```release-note
NONE
```


deploy/modules/aws/tidb-cluster/variables.tf
```diff
@@ -131,7 +131,7 @@ variable "worker_group_launch_template_tags" {
 }
 
 variable "worker_ami_name_filter" {
-  description = "Additional name filter for AWS EKS worker AMI. Default behaviour will get latest for the cluster_version but could be set to a release from amazon-eks-ami, e.g. \"v20190220\""
+  description = "Additional name filter for AWS EKS worker AMI. Default behaviour will get the latest AMI for the cluster_version, but it could be set to a specific version, e.g. \"v20190220\", please check the `Packer version` in https://docs.aws.amazon.com/eks/latest/userguide/eks-linux-ami-versions.html for the supported AMI versions."
```
cofyc: @mightyguava what do you think?

PR #1904, opened 2020-03-11
Improve note of failurePolicy and revise IDC configuration

Yisaer:
Improve note of failurePolicy and revise IDC configuration

### Check List 

Tests 

 - No code

Related changes

```release-note
None
```


ci/deploy_tidb_operator_staging.groovy
```diff
@@ -23,12 +23,16 @@ scheduler:
...
 admissionWebhook:
   create: true
+  replicas: 2
   validation:
     statefulSets: true
     pods: true
     pingcapResources: false
   mutation:
     pingcapResources: true
+  failurePolicy:
```
DanielZhangQD: Why change them to `Fail` in IDC where the k8s version is v1.12, which does not match the descriptions in `values.yaml` updated above.

Yisaer: IDC is under our control, it's fine to set fail for failurePolicy in this case. For community customer, we still recommend set failurePolicy as Ignore as I mentioned in the note.

PR #1908, opened 2020-03-11
backup: support kms decryption secret

shuijing198799:
support kms decryption secret in tidb-operator
add KMS field in backup CRD, it should set to be true if the secret has been encrypted by KMS
### Check List 

Tests 

Code changes

 - Has Go code change

```release-note
Add `spec.useKMS` field in `Backup`, `Restore` and `BackupSchedule` custom resources. Set it to `true` if you want to decrypt the secrets encrypted with KMS.
```

### DIFF
--- a/images/tidb-backup-manager/entrypoint.sh
+++ b/images/tidb-backup-manager/entrypoint.sh
@@ -51,33 +51,40 @@ else
 fi
 
 BACKUP_BIN=/tidb-backup-manager
+if [[ -n "${AWS_DEFAULT_REGION}"]]; then
+	EXEC_COMMAND="exec"
+else
+	EXEC_COMMAND="/usr/local/bin/shush exec --"
+fi
+
+cat /tmp/rclone.conf
 
 # exec command
 case "$1" in
     backup)
         shift 1
         echo "$BACKUP_BIN backup $@"
-        exec $BACKUP_BIN backup "$@"
+        $EXEC_COMMAND $BACKUP_BIN backup "$@"
         ;;
     export)
         shift 1
         echo "$BACKUP_BIN export $@"
-        exec $BACKUP_BIN export "$@"
+        $EXEC_COMMAND $BACKUP_BIN export "$@"
         ;;
     restore)
         shift 1
         echo "$BACKUP_BIN restore $@"
-        exec $BACKUP_BIN restore "$@"
+        $EXEC_COMMAND $BACKUP_BIN restore "$@"
         ;;
     import)
         shift 1
         echo "$BACKUP_BIN import $@"
-        exec $BACKUP_BIN import "$@"
+        $EXEC_COMMAND $BACKUP_BIN import "$@"
         ;;
     clean)
         shift 1
         echo "$BACKUP_BIN clean $@"
-        exec $BACKUP_BIN clean "$@"
+        $EXEC_COMMAND $BACKUP_BIN clean "$@"
         ;;
     *)
         echo "Usage: $0 {backup|restore|clean}"
--- a/pkg/backup/constants/constants.go
+++ b/pkg/backup/constants/constants.go
@@ -55,4 +55,7 @@ const (
 
 	// ServiceAccountCAPath is where is CABundle of serviceaccount locates
 	ServiceAccountCAPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+	// KMS secret env prefix
+	KMSSecretPrefix = "KMS_ENCRYPTED"
 )
--- a/pkg/backup/util/util.go
+++ b/pkg/backup/util/util.go
@@ -39,7 +39,7 @@ func CheckAllKeysExistInSecret(secret *corev1.Secret, keys ...string) (string, b
 }
 
 // GenerateS3CertEnvVar generate the env info in order to access S3 compliant storage
-func GenerateS3CertEnvVar(s3 *v1alpha1.S3StorageProvider) ([]corev1.EnvVar, string, error) {
+func GenerateS3CertEnvVar(useKMS bool, s3 *v1alpha1.S3StorageProvider) ([]corev1.EnvVar, string, error) {
 	var envVars []corev1.EnvVar
 
 	switch s3.Provider {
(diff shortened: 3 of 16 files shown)


pkg/backup/util/util.go
```diff
@@ -226,9 +239,16 @@ func GenerateTidbPasswordEnv(ns, name, tidbSecretName string, secretLister corel
...
+	if useKMS {
+		passwordKey = fmt.Sprintf("%s_%s_%s", constants.KMSSecretPrefix, constants.BackupManagerEnvVarPrefix, strings.ToUpper(constants.TidbPasswordKey))
+	} else {
+		passwordKey = fmt.Sprintf("%s_%s", constants.BackupManagerEnvVarPrefix, strings.ToUpper(constants.TidbPasswordKey))
+	}
+
 	certEnv = []corev1.EnvVar{
 		{
-			Name: fmt.Sprintf("%s_%s", constants.BackupManagerEnvVarPrefix, strings.ToUpper(constants.TidbPasswordKey)),
+			Name: passwordKey,
```
DanielZhangQD: backup-manager will get the password from the env, the code there may need to update, for example, https://github.com/pingcap/tidb-operator/blob/master/cmd/backup-manager/app/backup/manager.go#L67

gregwebs:
I think this should be switched to sops (it does have exec-env) so that we can support multiple clouds.

shuijing198799:
> I think this should be switched to sops (it does have exec-env) so that we can support multiple clouds.

It should support by DBaaS first, and then I'll modify the operator code

shuijing198799:
> I think this should be switched to sops (it does have exec-env) so that we can support multiple clouds.

Maybe it will change to sop after support GCP ?

pkg/backup/util/util.go
```diff
@@ -87,6 +87,8 @@ func GenerateS3CertEnvVar(s3 *v1alpha1.S3StorageProvider) ([]corev1.EnvVar, stri
 			Value: s3.StorageClass,
 		},
 	}
+	// KMS has to use with IAM for get credentail to decryption the code
```
DanielZhangQD: Please update the description here.

pkg/backup/util/util.go
```diff
@@ -39,7 +39,7 @@ func CheckAllKeysExistInSecret(secret *corev1.Secret, keys ...string) (string, b
 }
 
 // GenerateS3CertEnvVar generate the env info in order to access S3 compliant storage
-func GenerateS3CertEnvVar(s3 *v1alpha1.S3StorageProvider) ([]corev1.EnvVar, string, error) {
+func GenerateS3CertEnvVar(useKMS bool, s3 *v1alpha1.S3StorageProvider) ([]corev1.EnvVar, string, error) {
```
DanielZhangQD: useKMS is not used in this function.

pkg/backup/util/util.go
```diff
@@ -87,6 +87,8 @@ func GenerateS3CertEnvVar(s3 *v1alpha1.S3StorageProvider) ([]corev1.EnvVar, stri
 			Value: s3.StorageClass,
 		},
 	}
+	// KMS has to use with IAM for get credentail to decryption the code
+	// so if use KMS, the secretKey and accesssKey can be ignore
 	if s3.SecretName != "" {
 		envVars = append(envVars, []corev1.EnvVar{
 			{
```
DanielZhangQD: No need to consider the case where accessKeys are encrypted by KMS?

pkg/backup/util/util.go
shuijing198799: done

pkg/backup/util/util.go
shuijing198799: done

pkg/backup/util/util.go
shuijing198799: though use rarely use like this, we can't don't allow them do like this

pkg/backup/util/util.go
```diff
@@ -152,7 +152,7 @@ func GenerateGcsCertEnvVar(gcs *v1alpha1.GcsStorageProvider) ([]corev1.EnvVar, s
 }
 
 // GenerateStorageCertEnv generate the env info in order to access backend backup storage
-func GenerateStorageCertEnv(ns string, provider v1alpha1.StorageProvider, secretLister corelisters.SecretLister) ([]corev1.EnvVar, string, error) {
+func GenerateStorageCertEnv(ns string, useKMS bool, provider v1alpha1.StorageProvider, secretLister corelisters.SecretLister) ([]corev1.EnvVar, string, error) {
```
DanielZhangQD: useKMS is not used.

shuijing198799: done

pkg/backup/constants/constants.go
```diff
@@ -55,4 +55,7 @@ const (
 
 	// ServiceAccountCAPath is where is CABundle of serviceaccount locates
 	ServiceAccountCAPath = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+	// KMS secret env prefix
+	KMSSecretPrefix = "KMS_ENCRYPTED"
```
cofyc: how this is being used? any docs or links about this env?

images/tidb-backup-manager/entrypoint.sh
```diff
@@ -51,33 +51,40 @@ else
 fi
 
 BACKUP_BIN=/tidb-backup-manager
+if [[ -n "${AWS_DEFAULT_REGION}"]]; then
```
cofyc: this env `AWS_DEFAULT_REGION` is always injected into pods on eks?

pkg/backup/constants/constants.go
shuijing198799: the dbaas encrpytion secret with KMS so the  secret mount to pod should decryption with [shush](https://github.com/realestate-com-au/shush) this prefix means this env should decryption

images/tidb-backup-manager/entrypoint.sh
shuijing198799: this AWS_DEFAULT_REGION only use in useKMS is on

Issue #1917, opened 2020-03-12
TidbMonitor failed to scrape PD/TiKV metrics while TLS is enabled


weekface:
## Bug Report

I created a TiDB Cluster with [TLS enabled between components](https://pingcap.com/docs/stable/how-to/secure/enable-tls-between-components/) with CR:

```
kubectl apply -f example/basic/tidb-cluster.yaml
kubectl apply -f example/basic/tidb-monitor.yaml
```
But there are no metrics data.

Prometheus logs:

```
level=debug ts=2020-03-12T08:22:46.499Z caller=scrape.go:927 component="scrape manager" scrape_pool=pd target=http://10.244.0.23:2379/metrics msg="Scrape failed" err="Get http://10.244.0.23:2379/metrics: EOF"
level=debug ts=2020-03-12T08:22:51.388Z caller=scrape.go:927 component="scrape manager" scrape_pool=tidb target=http://10.244.0.27:10080/metrics msg="Scrape failed" err="Get http://10.244.0.27:10080/metrics: EOF"
```

PD logs:
```
[2020/03/12 08:22:31.532 +00:00] [WARN] [config_logging.go:279] ["rejected connection"] [remote-addr=10.244.0.29:44614] [server-name=] [error="tls: first record does not look like a TLS handshake"]
[2020/03/12 08:22:46.498 +00:00] [WARN] [config_logging.go:279] ["rejected connection"] [remote-addr=10.244.0.29:44710] [server-name=] [error="tls: first record does not look like a TLS handshake"]
[2020/03/12 08:23:01.497 +00:00] [WARN] [config_logging.go:279] ["rejected connection"] [remote-addr=10.244.0.29:44810] [server-name=] [error="tls: first record does not look like a TLS handshake"]
```

TiDB don't have any related logs.

Issue #1855, opened 2020-03-03
TiDB Cluster/Client TLS support implementation in TiDB Operator


weekface:
To support various certificate issuers and methods, for example: user-defined certificates, [K8s builtin CA system](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster) or [cert-manager](https://cert-manager.io), we will refactor the TiDB [Cluster](https://pingcap.com/docs/stable/how-to/secure/enable-tls-between-components/)/[Client](https://pingcap.com/docs/stable/how-to/secure/enable-tls-clients/) TLS feature with new API and new usage.

First, we will change the Cluster/Client TLS API to:

```
tlsCluster:
  enabled: true

tidb:
  tlsClient:
    enabled: true
```

That is all the whole API we needed.

If the user set `tlsCluster.enabled` to `true`, then tidb-operator will assume that there are several Secrets named with: `<cluster-name>-pd-cluster-secret`, `<cluster-name>-tikv-cluster-secret`, `<cluster-name>-tidb-cluster-secret`, `<cluser-name>-cluster-client-secret` and other TiDB components's Secrets are created.

Each Secret data should have three keys: `tls.crt`, `tls.key` and `ca.crt`:

```
apiVersion: v1
kind: Secret
metadata:
  name: <secret-name>
data:
  tls.crt: <base64 decoded certificate data>
  tls.key: <base64 decode key data>
  ca.crt: <base64 decode ca data>
```

These Secrets can be created by the use manually, by K8s builtin CA system or by cert-manager. PD/TiKV/TiDB/... will use these Secrets to start server.

tidb-operator will not supply these certificates automatically.

There are several tasks:

- Remove old auto generated codes
  - [x] remove TLS codes for client #1867
  - [x] remove TLS codes between components https://github.com/pingcap/tidb-operator/pull/1870
- TLS for MySQL Clients
  - [x] TiDB Server/Client #1867
  - [x] TiDB add `require-secure-transport` option to TiDB Server https://github.com/pingcap/tidb/pull/15341
  - [ ] TiDB Operator support `require-secure-transport` option
  - [x] TiDB Initializer https://github.com/pingcap/tidb-operator/pull/1931
  - [x] verify [cert-based authentication](https://pingcap.com/docs/stable/reference/security/cert-based-authentication/) manually
- TLS between TiDB components
  - [x] PD/TiKV/TiDB https://github.com/pingcap/tidb-operator/pull/1870
[...]

PR #1919, opened 2020-03-12
fix prometheus scrape config issue while TLS is enabled

weekface:

This PR fixes https://github.com/pingcap/tidb-operator/issues/1917, and change `scrapeUrl` to the domain of the `Pod`. Then we can set `insecure_skip_verify` to `true` while TLS is enabled.

Part of #1855 

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Fix prometheus scrape config issue while TLS is enabled
```


gregwebs:
will there be a followup to set insecure_skip_verify to false?

weekface:
@gregwebs 

`insecure_skip_verify ` will be set to `false` when TLS is enabled, otherwise `true`.

TiKV is always `true` now, because TiKV metrics API doesn't support TLS yet.  https://github.com/tikv/tikv/issues/5340

PR #1930, opened 2020-03-16
Backup: make tikv support add serviceaccount and switch rclone env_auth to true

shuijing198799:
support tikv specify a serviceAccount name and make rclone can use iam in cloud env
support tikv specify a serviceAccount name and make rclone can use iam in cloud env
### Check List 

Tests 

modify tc and then verify serviceaccount is changed in tikv pod
use rclone in cloud env without set secretkey and accesskey

Code changes

 - Has Go code change

Add `spec.tikv.serviceAccount` field in `TiDBCluster` custom resources. To specify a service account name for tikv pods.
```

### DIFF
--- a/images/tidb-backup-manager/entrypoint.sh
+++ b/images/tidb-backup-manager/entrypoint.sh
@@ -19,7 +19,7 @@ echo "Create rclone.conf file."
 cat <<EOF > /tmp/rclone.conf
 [s3]
 type = s3
-env_auth = false
+env_auth = true
 provider =  ${S3_PROVIDER}
 access_key_id = ${AWS_ACCESS_KEY_ID}
 secret_access_key = ${AWS_SECRET_ACCESS_KEY:-$AWS_SECRET_KEY}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by shuijing198799** [comment]
run-e2e-tests

File: images/tidb-backup-manager/entrypoint.sh [hunk 1]
```diff
@@ -19,7 +19,7 @@ echo "Create rclone.conf file."
 cat <<EOF > /tmp/rclone.conf
 [s3]
 type = s3
-env_auth = false
+env_auth = true
```
DanielZhangQD: We also have to consider the non-IAM role case.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -445,6 +445,7 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 	podSpec.SecurityContext = podSecurityContext
 	podSpec.InitContainers = initContainers
 	podSpec.Containers = []corev1.Container{tikvContainer}
+	podSpec.ServiceAccountName = tc.Spec.TiKV.ServiceAccount
```
DanielZhangQD: Please make sure that this will not trigger a rolling update during tidb-operator upgrade if `ServiceAccount` not configured.

images/tidb-backup-manager/entrypoint.sh
shuijing198799: https://rclone.org/s3/#authentication the secret key and access key is from env AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY, If  access_key and secret_access_key is set, the env have this two parameters. set it to true is a save way for IAM and non-IAM.

PR #1931, opened 2020-03-16
make tidb-initializer support TLS

weekface:

Part of #1855, make TidbInitializer support TLS.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Make tidb-initializer support TLS
```


pkg/manager/member/tidb_init_manager.go
```diff
@@ -15,6 +15,7 @@ package member
 
 import (
 	"fmt"
+	"github.com/pingcap/tidb-operator/pkg/util"
```
Yisaer: plz fmt the goimports.

pkg/manager/member/template.go
```diff
@@ -281,7 +281,11 @@ var tidbInitStartScriptTpl = template.Must(template.New("tidb-init-start-script"
 host = '{{ .ClusterName }}-tidb'
 permit_host = '{{ .PermitHost }}'
 port = 4000
+{{- if .TLS }}
+conn = MySQLdb.connect(host=host, port=port, user='root', connect_timeout=5, ssl={'ca': '{{ .CAPath }}', 'cert': '{{ .CertPath }}', 'key': '{{ .KeyPath }}'})
```
cofyc: as MySQL client, it can connect a TiDB server with tls enabled without client certificate. the communication is SSL encrypted too. tidb server does not verify the client certificate and has no plan to verify it.

weekface: > tidb server does not verify the client certificate and has no plan to verify it.

Not exactly, TiDB Server already supported [Certificate-Based Authentication for Login](https://pingcap.com/docs/stable/reference/security/cert-based-authentication/).

With this method, the TiDB Server verifies client certificates when users log in.

cofyc: if users enabled tidb server-side TLS, they must provide client tls secret for the root user to initialize the tidb cluster? IMO, this is complex and the initialization will fail if the user didn't configure this secret. Maybe we can add a dedicate client secret for the tidb initializer?

cofyc: we can configure a tls client secret like `TidbInitializerSpec.PasswordSecret` for the tidb initializer user when people want to initialize the cluster with the client secret provided. and it's optional because when the server is TLS enabled, people can still connect the server without SSL certificate.

weekface: Yes, when TLS was enabled, the user should prepare two certificates: one server cert and one client cert, we will have **document** to mention that.

The client certificate Secret Name is `<clusterName>-tidb-client-secret`.

Issue #1949, opened 2020-03-17
Issues with TiDB Monitor


DanielZhangQD:
## Bug Report

**What version of Kubernetes are you using?**

1.12.8
**What version of TiDB Operator are you using?**

v1.1.0-beta.2
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

local-storage
**What's the status of the TiDB cluster pods?**

Running
**What did you do?**

Upgrade TiDB Operator from v1.0 to v1.1 and Create TiDB Monitor CR.
**What did you expect to see?**
Monitor works as expected.
**What did you see instead?**

- The existing Monitor deployment is rolling updated after CR created.
- Service NodePorts are changed after CR created.
  Before CR creation:
  ```
  dan10-grafana            NodePort    10.233.22.65    <none>        3000:31558/TCP                   23h
  dan10-monitor-reloader   NodePort    10.233.6.191    <none>        9089:30400/TCP                   23h
  dan10-prometheus         NodePort    10.233.16.45    <none>        9090:32184/TCP                   23h
    ```
  After CR creation:
  ```
  dan10-grafana            NodePort    10.233.22.65    <none>        3000:32287/TCP                   24h
  dan10-monitor-reloader   NodePort    10.233.6.191    <none>        9089:30400/TCP                   24h
  dan10-prometheus         NodePort    10.233.16.45    <none>        9090:30273/TCP                   24h
  dan10-reloader           NodePort    10.233.13.22    <none>        9089:32448/TCP                   40m
  ```
- The reloader service name is changed from `<cluster>-monitor-reloader` to `<cluster>-monitor-reloader`. (Please see detail in above info)
- The monitor deployment rolling updates continuously.
  ```
  dan10-monitor-765b58fcb-bwn5l             0/3     Pending                      0          0s
  dan10-monitor-765b58fcb-bwn5l             0/3     Pending                      0          0s
  dan10-monitor-765b58fcb-bwn5l             0/3     Init:0/1                     0          1s
  dan10-monitor-765b58fcb-bwn5l             0/3     PodInitializing              0          7s
  dan10-monitor-765b58fcb-bwn5l             3/3     Running                      0          9s
  dan10-monitor-765b58fcb-bwn5l             3/3     Terminating                  0          56s
  dan10-monitor-765b58fcb-bwn5l             0/3     Terminating                  0          59s
  dan10-monitor-765b58fcb-bwn5l             0/3     Terminating                  0          60s
[...]

PR #1954, opened 2020-03-17
update permission for tidb-controller-manager and add example for tidb-monitor

DanielZhangQD:

1. After upgrade TiDB Operator from v1.0 to v1.1, then create TiDB Monitor CR, tidb-controller-manager reports below error:
```
Event(v1.ObjectReference{Kind:"TidbMonitor", Namespace:"dan10", Name:"dan10", UID:"c8b9b9ed-67f9-11ea-8886-8a678ff421f7", APIVersion:"pingcap.com/v1alpha1", ResourceVersion:"68470646", FieldPath:""}): type: 'Warning' reason: 'FailedSync' Sync TidbMonitor[dan10/dan10] Deployment failed,err:secrets "dan10-monitor" is forbidden: User "system:serviceaccount:tidb-admin:tidb-controller-manager" cannot update resource "secrets" in API group "" in the namespace "dan10"
 ```
2. No full example for TiDB Monitor.

update permission for tidb-controller-manager and add example for tidb-monitor
### Check List 

Tests 

   TiDB Monitor CR can be created and synced.

Related changes

```release-note
NONE
```

Other issues during test of TiDB Monitor are recorded in #1949.

### DIFF
--- a/charts/tidb-operator/templates/controller-manager-rbac.yaml
+++ b/charts/tidb-operator/templates/controller-manager-rbac.yaml
@@ -135,7 +135,7 @@ rules:
   verbs: ["get", "list", "watch", "create", "update", "delete"]
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["create", "get", "list", "watch"]
+  verbs: ["create", "update", "get", "list", "watch"]
 - apiGroups: [""]
   resources: ["persistentvolumeclaims"]
   verbs: ["get", "list", "watch", "create", "update", "delete"]
--- a/charts/tidb-operator/templates/controller-manager-rbac.yaml
+++ b/charts/tidb-operator/templates/controller-manager-rbac.yaml
@@ -38,7 +38,7 @@ rules:
   verbs: ["get", "list", "watch", "create", "update", "delete"]
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["create", "get", "list", "watch"]
+  verbs: ["create", "update", "get", "list", "watch"]
 - apiGroups: [""]
   resources: ["persistentvolumeclaims"]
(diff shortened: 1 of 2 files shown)


charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -38,7 +38,7 @@ rules:
   verbs: ["get", "list", "watch", "create", "update", "delete"]
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["create", "get", "list", "watch"]
+  verbs: ["create", "update", "get", "list", "watch"]
```
weekface: does this should be consistent with https://github.com/pingcap/tidb-operator/pull/1954/files#diff-ecdc286a1f57077063e4a16696198f7fR138?

manifests/monitor/tidb-monitor.yaml
```diff
@@ -0,0 +1,83 @@
+apiVersion: pingcap.com/v1alpha1
```
weekface: these is already an example of TidbMonitor: `examples/basic/tidb-monitor.yaml`

DanielZhangQD: Those are basic examples and users cannot know how to use the CR with the basic ones.

charts/tidb-operator/templates/controller-manager-rbac.yaml
DanielZhangQD: updated.

DanielZhangQD:
> I recommend to omit empty fields, given that we have a detailed api reference now.

It would be easier for users to know how to use the CR following this example without referring to the API doc.
Comments addressed, PTAL again, thanks! @weekface @aylei

Issue #1701, opened 2020-02-14
UCP: Add toleration support for discovery deployment in tidb-cluster chart


DanielZhangQD:
## Description

Currently, when we deploy tidb cluster with helm chart, it's unable to configure toleration for discovery deployment. We should support it in helm chart.

## Score

* 50

## Mentor(s)

* [@DanielZhangQD](https://github.com/DanielZhangQD)

Contact the mentors: **#tidb-challenge-program** channel in [TiDB-Operator Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Kubernetes
* Helm

## Learning Materials

- [Kubernetes Concepts](https://kubernetes.io/docs/concepts/)
- [Helm chart basic](https://helm.sh/docs/chart_template_guide/getting_started/)

cofyc:
closed via https://github.com/pingcap/tidb-operator/pull/1959.

PR #1959, opened 2020-03-17
Adding tolerations and affinity to the discovery chart template

Smana:
Solve the issue https://github.com/pingcap/tidb-operator/issues/1701

Setting values in the tidb-cluster Helm chart


cofyc:
there is another PR trying to fix the same issue here: https://github.com/pingcap/tidb-operator/issues/1701

Issue #1958, opened 2020-03-17
pod webhook does not work if asts controller not installed in namespace `pingcap`


cofyc:
## Bug Report

https://github.com/pingcap/tidb-operator/blob/master/pkg/webhook/pod/pods.go#L51

our code hard-coded advanced-statefulset controller namespace as `pingcap` and our CI install it in this namespace and works without a problem. but this might not be true in user environments.

Yisaer:
good catch. I will fix it soon.

PR #1963, opened 2020-03-17
Fix some webhook error

Yisaer:

Fix https://github.com/pingcap/tidb-operator/issues/1958

1. Fix the problem that Advanced StatefulSet could not work with Webhook
2. Change the Reaction for the Down State Tikv pod during deleting request in webhook from admit to reject during scale-in which would cause the duplicated id problem.

Related changes

```release-note
1. Fix the problem that Advanced StatefulSet could not work with Webhook
2. Change the Reaction for the Down State Tikv pod during deleting request in webhook from admit to reject.
```


Yisaer:
I suggest rejecting the deleting pod request for tikv when its store state is `Down` instead of passing. If we admit the tikv with store down scaled-in, the next time the tikv cluster scale-out and it would meet duplicated id problem during its starting process as the previous `Down` store is still existed and occupied the address.

What's your opinion? @weekface

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -100,9 +100,12 @@ func (pc *PodAdmissionControl) admitDeleteTiKVPods(payload *admitPayload) *admis
 	case v1alpha1.TiKVStateTombstone:
 		return pc.admitDeleteUselessTiKVPod(payload)
 	case v1alpha1.TiKVStateOffline:
-		return pc.admitDeleteOfflineTiKVPod()
+		return pc.rejectDeleteTiKVPod()
 	case v1alpha1.TiKVStateDown:
-		return pc.admitDeleteUselessTiKVPod(payload)
+		// When the target tikv's store is DOWN, we would not pass the deleting request, otherwise it would cause
+		// the duplicated id problem for the newly tikv pod.
+		// Users should offline the target tikv into tombstone first, then scale-in it.
```
DanielZhangQD: The logic in tikv scaler is to delete the store automatically https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_scaler.go#L110, now we request users to delete the store manually, what do you think? @cofyc @weekface

Yisaer:
I limit the range that the webhook would only reject the deleting pod request for tikv with down store only in Scae-in case. In other case, It would worr as previos. PTAL

PR #1965, opened 2020-03-17
Add tikv store limit pattern

Yisaer:
Add tikv store limit pattern which would avoid the error happen if the external store join the tikv cluster like tiflash.

Related changes

```release-note
NONE
```


pkg/manager/member/tikv_member_manager.go
```diff
@@ -40,7 +40,8 @@ import (
 
 const (
 	// tikvClusterCertPath is where the cert for inter-cluster communication stored (if any)
-	tikvClusterCertPath = "/var/lib/tikv-tls"
+	tikvClusterCertPath   = "/var/lib/tikv-tls"
+	tikvStoreLimitPattern = `%s-tikv-\d+\.%s-tikv-peer\.%s\.svc\:\d+`
```
aylei: it is nearly hardcoded, is there a better way? or can we add a TODO that indicates we need a better way in the future?

Yisaer: This method has been used before in this request : https://github.com/pingcap/tidb-operator/pull/1621
But I'm ok to add comment to indicate find a better way in the future.

Yisaer: updated.

Yisaer:
@aylei @cofyc  PTAL

PR #1975, opened 2020-03-18
lightning: support lightning use IAM

shuijing198799:
add serviceAccount for config
add service Account config to pod template
### Check List 

Tests 

1. install an importer in tidb cluster

2. add annotation in annotation field
```
annotations: {"iam.amazonaws.com/role": "arn:aws:iam::123456789012:role/user"}
```
and set a seviceAccount in values.yaml

3. follow [doc](https://pingcap.com/docs-cn/stable/reference/tools/tidb-lightning/deployment/) to deploy lightning in tidb cluster and use remote S3 backups to restore , ignore accesskey and 
 secret key in field `dataSource.remote.secretName`

4. check wether initContainer can download file from S3 and the serviceAccount is attach to the tidb-lightning pod

Code changes

 - Has Go code change


charts/tidb-lightning/templates/job.yaml
```diff
@@ -125,8 +128,10 @@ spec:
       {{ else }}
       - name: credentials
         secret:
+          {{- if .Values.dataSource.remote.secretName}}
```
DanielZhangQD: If secretName not configured, then this volume is not required, neither the volumeMount.

shuijing198799: done

Issue #1460, opened 2020-01-01
UCP: Validate requests.storage field of PD/TiKV


aylei:
## Description

Currently, we didn't valid `requests.storage` field for PD and TiKV components. However, they are used as storage requirements in PVC and cannot be empty. We should validate them in our validation logic. You can refer to [our current validation code](https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/validation/validation.go) to learn how to implement it.

## Score

* 50

## Mentor(s)

* [@aylei ](https://github.com/aylei)

Contact the mentors: **#tidb-challenge-program** channel in [TiDB-Operator Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang

## Learning Materials

- [Kubernetes Concepts](https://kubernetes.io/docs/concepts/)
- [Kubernetes validation code](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go)

PR #1983, opened 2020-03-18
UCP: add Validate requests.storage field of PD/TiKV

shonge:
UCP: #1460 
Close: https://github.com/pingcap/tidb-operator/issues/1460
add validate code in pkg/apis/pingcap/v1alpha1/validation/validation.go
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


Yisaer:
I think adding the e2e cases to verify the validation and defaulting is necessary.

cofyc:
> I think adding the e2e cases to verify the validation and defaulting is necessary.

actually not necessary, because validation functions are simple functions that take into plain objects and returns `field.ErrorList`. unit tests are enough.

cofyc:
hi, @shonge 
you can add some unit tests in [`validation_test.go`](https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/validation/validation_test.go), it already has some examples.

you can also learn how to write validation tests from [Kubernetes code](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation_test.go) which contains more examples.

Yisaer:
The rest LGTM. Adding some unit tests would be better.

shonge:
@aylei PTAL

pkg/apis/pingcap/v1alpha1/validation/validation_test.go
```diff
@@ -179,3 +181,109 @@ func TestValidateAnnotations(t *testing.T) {
...
+						},
+					},
+					TiDB: v1alpha1.TiDBSpec{
+						BaseImage: "pingcap/tidb",
+						Config:    &v1alpha1.TiDBConfig{},
+					},
+				},
+			},
+		},
+	}
```
Yisaer, line 223: Almost LGTM, but the unit test seems a bit of weird. I recomended to write unit test in table-driven style. FYI https://github.com/golang/go/wiki/TableDrivenTests#parallel-testing

For example, we could define testcase to determine whether the tidbcluster is valid or not, if not, which compont is going to be valid.

Here is one example. https://github.com/pingcap/tidb-operator/blob/723f7c3adda243246faac3790f917bd93b25e00e/pkg/autoscaler/autoscaler/util_test.go#L29-L94

shonge: haha, I will fix it

shonge: Resolved

shonge:
> Almost LGTM

PTAL

shonge:
@Yisaer PTAL

shonge:
@cofyc PTAL

pkg/apis/pingcap/v1alpha1/validation/validation.go
```diff
@@ -168,6 +171,15 @@ func validateComponentSpec(spec *v1alpha1.ComponentSpec, fldPath *field.Path) fi
 	return allErrs
 }
 
+//validateRequestsStorage validates resources requests storage
+func validateRequestsStorage(requests corev1.ResourceList, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if _, ok := requests[corev1.ResourceStorage]; !ok {
+		allErrs = append(allErrs, field.Required(fldPath.Child("requests.storage").Key((string(corev1.ResourceStorage))), "PD or TiKV request storage must not be empty"))
```
aylei, line 178:
```suggestion
		allErrs = append(allErrs, field.Required(fldPath.Child("requests.storage").Key((string(corev1.ResourceStorage))), "storage request must not be empty"))
```

We may invoke this method to verify other components and the error will include the field ref which is self-explanatory

aylei:
Thanks! @shonge

PR #2007, opened 2020-03-23
improve orphan pods clean logic

cofyc:

- check pod has been scheduled or not
- use ResourceVersion precondition

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/manager/member/orphan_pods_cleaner.go
+++ b/pkg/manager/member/orphan_pods_cleaner.go
@@ -17,7 +17,6 @@ import (
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/controller"
 	"github.com/pingcap/tidb-operator/pkg/label"
-	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -88,8 +87,8 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
 			continue
 		}
 
-		if pod.Status.Phase != v1.PodPending {
-			skipReason[podName] = skipReasonOrphanPodsCleanerPodIsNotPending
+		if len(pod.Spec.NodeName) > 0 {
+			skipReason[podName] = skipReasonOrphanPodsCleanerPodHasBeenScheduled
```
cofyc: we should not rely on status information here because the pod may be scheduled but the status may not be updated by the kubelet yet. 

however, if a pod has not been scheduled, it's safe to delete it whether PVCs have been created for it or not.

PR #2015, opened 2020-03-23
Limit Autofailure condition

Yisaer:

1. Default the `spec.tikv.maxfailureOver` and `spec.tidb.maxfailureOver` to 3 when they are no defined
2. Change the action that when `maxfailureOver` is zero, operator would not failover instead of creating limitless instances.
3. Still Checking the `maxfailureOver` whether it is nil in the failover check.

### Check List 

Tests 

 - Unit test

Related changes

```release-note
1. Default the `spec.tikv.maxfailureOver` and `spec.tidb.maxfailureOver` to 3 when they are no defined
2. Change the action that when `maxfailureOver` is zero, operator would not failover instead of creating limitless instances.
```


pkg/manager/member/tidb_member_manager.go
```diff
@@ -198,7 +198,7 @@ func (tmm *tidbMemberManager) syncTiDBStatefulSetForTidbCluster(tc *v1alpha1.Tid
 		}
 	}
 
-	if tmm.autoFailover {
+	if tmm.autoFailover && tc.Spec.TiDB.MaxFailoverCount != nil {
```
cofyc: what's the default value?

cofyc: I seem that we don't have a default value, so the user must specify this field to enable the auto-failover feature? @aylei @weekface what do you think?

Yisaer: Yes, it is ok to me to set the default value to ensure the `MaxFailoverCount` is not nil`.  But I still think it is necessary to check whether the `MaxFailoverCount` is nil here.

pkg/manager/member/tidb_failover.go
```diff
@@ -45,22 +45,22 @@ func (tf *tidbFailover) Failover(tc *v1alpha1.TidbCluster) error {
...
 		}
-	}
-	for _, tidbMember := range tc.Status.TiDB.Members {
-		_, exist := tc.Status.TiDB.FailureMembers[tidbMember.Name]
-		deadline := tidbMember.LastTransitionTime.Add(tf.tidbFailoverPeriod)
-		if !tidbMember.Health && time.Now().After(deadline) && !exist {
-			tc.Status.TiDB.FailureMembers[tidbMember.Name] = v1alpha1.TiDBFailureMember{
-				PodName:   tidbMember.Name,
-				CreatedAt: metav1.Now(),
+		for _, tidbMember := range tc.Status.TiDB.Members {
```
DanielZhangQD: Why move this for loop in the control of failover? We may still need to update the status for the failure members even if the failover is disabled.

pkg/manager/member/tidb_member_manager.go
DanielZhangQD: The `tc.Spec.TiDB.MaxFailoverCount` will never be nil after adding the defaulting function, right?

pkg/manager/member/tikv_failover.go
```diff
@@ -51,18 +51,17 @@ func (tf *tikvFailover) Failover(tc *v1alpha1.TidbCluster) error {
...
 					klog.Warningf("%s/%s failure stores count reached the limit: %d", ns, tcName, tc.Spec.TiKV.MaxFailoverCount)
 					return nil
 				}
-			}
-
-			tc.Status.TiKV.FailureStores[storeID] = v1alpha1.TiKVFailureStore{
-				PodName:   podName,
-				StoreID:   store.ID,
-				CreatedAt: metav1.Now(),
+				tc.Status.TiKV.FailureStores[storeID] = v1alpha1.TiKVFailureStore{
```
DanielZhangQD: Suggest not moving this block under the control of failover count.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -223,7 +223,7 @@ func (tkmm *tikvMemberManager) syncStatefulSetForTidbCluster(tc *v1alpha1.TidbCl
 		return err
 	}
 
-	if tkmm.autoFailover {
+	if tkmm.autoFailover && tc.Spec.TiKV.MaxFailoverCount != nil {
```
DanielZhangQD: The pointer will never be nil after defaulting.

pkg/controller/tidbcluster/tidb_cluster_control.go
```diff
@@ -114,6 +115,8 @@ func (tcc *defaultTidbClusterControl) validate(tc *v1alpha1.TidbCluster) bool {
 }
 
 func (tcc *defaultTidbClusterControl) updateTidbCluster(tc *v1alpha1.TidbCluster) error {
+	// default the tidbcluster
+	defaulting.SetTidbClusterDefault(tc)
```
cofyc: move it to this function, we can have our controller-side validation and defaulting at the same place

https://github.com/pingcap/tidb-operator/blob/52491834b6e0371a4ae9bd862401f4d398d27e61/pkg/controller/tidbcluster/tidb_cluster_control.go#L84-L87

pkg/manager/member/tidb_failover.go
cofyc: yes, the `for _, tidbMember := range tc.Status.TiDB.Members {` loop is part of the status syncing. the user can choose not to failover when a member fails, but the controller should record failure members in the status always.

pkg/manager/member/tikv_member_manager.go
Yisaer: I think the defaulting and checking are 2 different things.

pkg/manager/member/tidb_failover.go
Yisaer: I think there is no updating status here both in the previous or in the current.

pkg/controller/tidbcluster/tidb_cluster_control.go
Yisaer: updated.

cofyc:
https://github.com/pingcap/tidb-operator/blob/52491834b6e0371a4ae9bd862401f4d398d27e61/pkg/apis/pingcap/v1alpha1/types.go#L260-L264

https://github.com/pingcap/tidb-operator/blob/52491834b6e0371a4ae9bd862401f4d398d27e61/pkg/apis/pingcap/v1alpha1/types.go#L307-L311

btw, comments of these two fields must be updated

pkg/manager/member/tikv_failover.go
cofyc: don't record failure tikv store if auto-failover is disabled? I guess this not expected.

cofyc: oh, I see that this is the same as Failmember in tidb status...my mistake.

Issue #1779, opened 2020-02-26
slowLogTailer: rotated slow logs not cleaned 


aylei:
## Bug Report

**What version of TiDB Operator are you using?**

`v1.0.6`

**What's the status of the TiDB cluster pods?**

```
NAME                            READY   STATUS    RESTARTS   AGE
db-discovery-65d54bbb55-zp272   1/1     Running   0          10h
db-monitor-669c96fb55-x2h75     3/3     Running   0          11h
db-pd-0                         1/1     Running   0          11h
db-pd-1                         1/1     Running   0          11h
db-pd-2                         1/1     Running   0          11h
db-tidb-0                       2/2     Running   0          4h28m
db-tidb-1                       2/2     Running   0          4h28m
db-tikv-0                       1/1     Running   0          11h
db-tikv-1                       1/1     Running   0          11h
db-tikv-2                       1/1     Running   0          11h
```

**What did you do?**

Load data to tidb cluster.

**What did you expect to see?**
Load data succeed

**What did you see instead?**
Load data failed because TiDB Pods were evicted to reclaim ephemeral storage:

![image](https://user-images.githubusercontent.com/18556593/75308734-22280280-588a-11ea-88f4-9e71316e0d5f.png)

And it turns out the slow log consumed a lot of ephemeral storage, whereas the tailer does not rotate the slow log:

```
/var/log/tidb # du -sh ./*
70.8M	./slowlog
299.9M	./slowlog-2020-02-26T02-10-16.452
299.9M	./slowlog-2020-02-26T02-11-33.170
299.9M	./slowlog-2020-02-26T02-12-27.762
299.9M	./slowlog-2020-02-26T02-13-44.140
299.9M	./slowlog-2020-02-26T02-16-08.322
299.9M	./slowlog-2020-02-26T02-18-27.989
300.0M	./slowlog-2020-02-26T02-22-49.095
300.0M	./slowlog-2020-02-26T02-26-00.591
299.9M	./slowlog-2020-02-26T02-29-03.809
299.9M	./slowlog-2020-02-26T02-31-16.241
299.9M	./slowlog-2020-02-26T02-32-11.446
300.0M	./slowlog-2020-02-26T02-33-12.131
300.0M	./slowlog-2020-02-26T02-34-16.872
299.9M	./slowlog-2020-02-26T02-35-14.501
299.9M	./slowlog-2020-02-26T02-38-52.075
300.0M	./slowlog-2020-02-26T02-43-13.551
[...]

PR #2045, opened 2020-03-27
Default Tidb Log File Configuration

Yisaer:
Close https://github.com/pingcap/tidb-operator/issues/1779

Defaulting the Tidb Log File configuration on `max-size` and `max-backups`.
For more details about configuration: https://pingcap.com/docs-cn/stable/reference/configuration/tidb-server/configuration-file/#max-backups

### Check List 

Tests 

 - Unit test

Related changes

```release-note
Default the `max-size` and `max-backups` for tidb server configuration with `300` and `3` if the tidb cluster was deployed by CR.
```


Yisaer:
I have test it with sysbench by setting `max-size=1` , `max-backups=3` and `slow-log-threshold=1`. It seems work from the result：

```sh
/var/log/tidb # du -sh ./*
484.0K  ./slowlog
1.0M    ./slowlog-2020-03-27T08-46-38.028
1.0M    ./slowlog-2020-03-27T08-46-41.393
1.0M    ./slowlog-2020-03-27T08-46-45.787
```
@aylei @cofyc

pkg/apis/pingcap/v1alpha1/defaulting/tidbcluster.go
```diff
@@ -26,6 +26,11 @@ const (
 	defaultBinlogImage = "pingcap/tidb-binlog"
 )
 
+var (
+	tidbLogMaxSize    = 300
```
cofyc: `300` means `300MB`? I think we don't need to set it if it's the default value of the `max-size`.

Yisaer: So we only need set the default max-page? I don't know whether 3 is proper for the default `max-backups` in kubernetes cases.

cofyc: yes

there is no perfect value because ephemeral storage is not the same in different environments. I think `3` is ok, about 900 MiB size.

Issue #1314, opened 2019-12-10
Use tidb-lightning to replace the legacy loader for restore


tennix:
## Feature Request

Now the `loader` is used to restore for mydumper backup files. The `loader` is less maintained, and its functionality is fully implemented in [tidb-lightning](https://github.com/pingcap/tidb-lightning) and well maintained. So we should replace `loader` with `tidb-lightning`.

TiDB Lightning supports [two backends](https://pingcap.com/docs/stable/reference/tools/tidb-lightning/tidb-backend/): `importer` and `tidb`. The `importer` backend needs to deploy another process `tikv-importer` while the `tidb` backend can be simply a drop-in replacement of the legacy `loader`.

PR #2068, opened 2020-03-30
use tidb-lightning in restore instead of loader

Yisaer:
Close https://github.com/pingcap/tidb-operator/issues/1314

Use tidb-lightning instead of loader. I have test it with aws and gcs. Both of them works fine.

Related changes

```release-note
Use tidb-lightning in Restore instead of loader.
```


cmd/backup-manager/app/import/restore.go
```diff
@@ -58,14 +58,16 @@ func (ro *Options) loadTidbClusterData(restorePath string) error {
 		return fmt.Errorf("dir %s does not exist or is not a dir", restorePath)
 	}
 	args := []string{
-		fmt.Sprintf("-d=%s", restorePath),
-		fmt.Sprintf("-h=%s", ro.Host),
-		fmt.Sprintf("-P=%d", ro.Port),
-		fmt.Sprintf("-u=%s", ro.User),
-		fmt.Sprintf("-p=%s", ro.Password),
+		"--status-addr=0.0.0.0:8289",
```
DanielZhangQD: We may have to specify the log-file to "" so that the logs of lightning can be captured in L72.

images/tidb-backup-manager/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM pingcap/tidb-enterprise-tools:latest
```
DanielZhangQD: There should be lightning in the base image, please confirm this.

Yisaer: No, pingcap/tidb-enterprise-tools only have syncer and load.

cmd/backup-manager/app/import/restore.go
```diff
@@ -58,14 +58,17 @@ func (ro *Options) loadTidbClusterData(restorePath string) error {
...
+		"--backend=tidb",
+		"--server-mode=false",
+		"–log-file=",
```
DanielZhangQD:
```suggestion
		"–-log-file=",
```

images/tidb-backup-manager/Dockerfile
```diff
@@ -1,6 +1,7 @@
...
 ARG VERSION=v1.51.0
 ARG SHUSH_VERSION=v1.4.0
+ARG TOOLKIT_VERSION=v3.0.9
```
DanielZhangQD:
```suggestion
ARG TOOLKIT_VERSION=v3.0.12
```

images/tidb-backup-manager/Dockerfile
DanielZhangQD: ok.

Issue #1525, opened 2020-01-09
Provide a more convenient way to modify tikv data directory


cwen0:
## Feature Request

**Is your feature request related to a problem? Please describe:**

I want to test TiKV with [Chaos Mesh](https://github.com/pingcap/chaos-mesh), IOChaos requires that the data directory of the application in the target pod should be a subdirectory of `PersistentVolumes`.  
In TiKV, the default data directory is `/var/lib/tikv` not a subdirectory of `PersistentVolumes`. 

**Describe the feature you'd like:**

1. Can we change the default data directory of tikv to "/var/lib/tikv/data" ?
2. Provide a more convenient way to modify tikv data directory.

**Teachability, Documentation, Adoption, Migration Strategy:**
 

1. For some users, they need to modify tikv data directory.
2. Sometimes, we can reuse the `PersistentVolumes` of tikv, eg: store log files in this pv

cofyc:
IIUC, IOChaos requires the data directory of the application to be a subdirectory of PV. I think [the `subPath` feature](https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath) can help.

Is this urgent?

cwen0:
For tikv or pd application, How to use `subPath` ?

cofyc:
as I suggested, if IOChaos requires the data directory of the application to be the subdirectory of a volume, `subPath` can help.

the big problem right now is we hardcoded many volume configurations and assume only one data volume to be configured on PD/TiKV pods.
 
we can add the `subPath` field for primary storage of  PD/TiKV pods or we can use a new, extensive and backward-compatible way to configure storage-related configurations that can solve [this problem](https://github.com/pingcap/tidb-operator/issues/1357) too.

I'll keep you informed on our progress.

benmaoer:
we need to change the default data directory of tikv to "/var/lib/tikv/data"  too.

xialuo1990:
maked, I want support it

cofyc:
we will support it in v1.1.1, any suggestions on CRD change are welcome!

cofyc:
I made a mistake in my previous comments. I guess we need to provide a way to customize tikv data directory (currently is hard coded to `/var/lib/tikv`)t? 

cc @xialuo1990 are you working on this, can you propose a solution? I think this is useful.

cofyc:
[...]

PR #2122, opened 2020-04-03
crd for tiflash

DanielZhangQD:
CRD for TiFlash
### Check List 

Tests 

   manually deploy tidbcluster with CR

Code changes

 - Has Go code change

Related changes

```release-note
Support TiFlash in TidbCluster CR
```


pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -278,6 +282,60 @@ type TiKVSpec struct {
...
+
+	// MaxFailoverCount limit the max replicas could be added in failover, 0 means no failover
+	// Optional: Defaults to 3
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	MaxFailoverCount *int32 `json:"maxFailoverCount,omitempty"`
+
+	// The DiskRequests of the persistent volume for TiFlash data storage.
+	// TiFlash supports multiple disks
+	DiskRequests []DiskRequests `json:"diskRequests"`
```
Yisaer, line 317: I tend to use `StorageRequests` here

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -278,6 +282,60 @@ type TiKVSpec struct {
...
+// +k8s:openapi-gen=true
+// TiDBSpec contains details of TiDB members
+type DiskRequests struct {
+	// The storageClassName of the persistent volume for data storage.
+	// Defaults to Kubernetes default storage class.
+	// +optional
+	StorageClassName *string `json:"storageClassName,omitempty"`
+
+	// StorageSize is the request storage size for TiFlash
+	StorageSize string `json:"storageSize"`
```
Yisaer, line 336: We should keep the expression as same as `pd` and `tikv`, so i tend to use `storage` here.

docs/api-references/config.json
```diff
@@ -1,6 +1,28 @@
...
+    "ProxyConfig",
+    "ClusterManagerPath",
+    "FlashStatus",
+    "FlashQuota",
+    "FlashUser",
+    "FlashProfile",
+    "FlashApplication",
+    "FlashProxy",
+    "FlashRaft",
+    "ClusterLog"
```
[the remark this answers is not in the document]
DanielZhangQD: These are not all of the definitions of the types of TiFlash, just some of them, and I have tried `hideTypePatterns`, but they will still show in the doc.

docs/api-references/config.json
```diff
@@ -26,4 +48,4 @@
     "k8s.io/apimachinery/pkg/apis/": "Kubernetes "
   },
   "markdownDisabled": false
-}
+}
```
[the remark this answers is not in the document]
DanielZhangQD: updated

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -278,6 +282,60 @@ type TiKVSpec struct {
...
+	// The DiskRequests of the persistent volume for TiFlash data storage.
+	// TiFlash supports multiple disks
+	DiskRequests []DiskRequests `json:"diskRequests"`
+
+	// Config is the Configuration of TiFlash
+	// +optional
+	Config *TiFlashConfig `json:"config,omitempty"`
+
+	// Dedicated TiDB cluster spec for TiFlash
+	TiDB *TiDBSpec `json:"tidb,omitempty"`
```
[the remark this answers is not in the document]
DanielZhangQD: Yes, should not mix it here, I will remove it.

pkg/apis/pingcap/v1alpha1/types.go
DanielZhangQD: updated

pkg/apis/pingcap/v1alpha1/types.go
DanielZhangQD: The `storage` for pd and tikv is configured in `corev1.ResourceRequirements` where the other fields are like `cpu`, `memory`, etc. However, `storage` here will not be clear enough.

DanielZhangQD:
@Yisaer comments addressed, PTAL again.

pkg/apis/pingcap/v1alpha1/types.go
Yisaer: Event if assuming the `storage` is not clear enough, I think there is no need to make difference between `pd`/`tikv`/`tiflash` and kubernetes resources expression. `StorageSize` seems like the property to dedicate the byte size of the storage requests which is common used in `TiDB`.

pkg/apis/pingcap/v1alpha1/tiflash_config.go
```diff
@@ -0,0 +1,359 @@
...
+package v1alpha1
+
+// Port from TiFlash configurations till 2020/04/02
+
+// TiFlashConfig is the configuration of TiFlash.
+// +k8s:openapi-gen=true
+type TiFlashConfig struct {
+	// commonConfig is the Configuration of TiFlash process
+	// +optional
+	CommonConfig *CommonConfig `json:"config,omitempty"`
```
weekface, line 23: Do we need add `toml` to the tag?

DanielZhangQD: Good catch, I will update them.

pkg/apis/pingcap/v1alpha1/types.go
cofyc: how about using a more generic and Kubernetes-style to configure volumes starting from tiflash? we can use volumeClaimTemplate to configure volumes for container and volumeMount to configure how to mount them. then we can use all functions they provide, e.g.

- `volumeMode`, `selector` from `volumeClaimTemplate`.
- `subPath` from `volumeMount` (https://github.com/pingcap/tidb-operator/issues/1525#issuecomment-602016020)

Yisaer: Good idea!

DanielZhangQD: Updated.

DanielZhangQD:
@weekface @cofyc @Yisaer Code updated as we have discussed, PTAL again.

Issue #2103, opened 2020-04-02
e2e case for upgrade tidb-operator from 1.0 to 1.1


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Add e2e case for upgrade tidb-operator from 1.0 to 1.1 and verify that only tidb-server will be upgraded because `--advertise-address` flag is enabled by default.

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #2145, opened 2020-04-09
Add e2e test for upgrading from 1.0.x

Yisaer:
Close https://github.com/pingcap/tidb-operator/issues/2103

Add e2e test to verify the upgrading Operator from 1.0.x (default 1.0.6) to current version.

Related changes

```release-note
NONE
```


tests/e2e/tidbcluster/serial.go
```diff
@@ -1008,4 +1010,147 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+	ginkgo.Context("[Verify: Upgrading Operator from 1.0.6", func() {
+		var oa tests.OperatorActions
+		var ocfg *tests.OperatorConfig
+		var version string
+
+		ginkgo.BeforeEach(func() {
+			version = os.Getenv("RELEASED_VERSION")
+			if len(version) < 1 {
+				version = "v1.0.6"
+			}
```
cofyc, line 1023: download v1.0.6 chart in this e2e test

actually, we can run `helm repo add pingcap https://charts.pingcap.org/ && helm install xxx --version v1.0.6` to install releases chart, but this needs some code refactoring. we can improve it in the future

cofyc: `v1.0.6` can be hard-coded in this test

cofyc: `DeployOperator` will clone the git repo if the tag is not `e2e`. If it works well, I think it's fine too.

tests/e2e/tidbcluster/serial.go
```diff
@@ -1008,4 +1010,147 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				framework.ExpectNoError(err, "failed to get pd pods uids")
+			}
+			tikvUids, err := getPodUids(labels.SelectorFromSet(label.New().Instance(tcName).TiKV().Labels()).String())
+			if err != nil {
+				framework.ExpectNoError(err, "failed to get tikv pods uids")
+			}
+			tidbUids, err := getPodUids(labels.SelectorFromSet(label.New().Instance(tcName).TiDB().Labels()).String())
+			if err != nil {
+				framework.ExpectNoError(err, "failed to get tidb pods uids")
+			}
```
cofyc, line 1087: you can cache the pods before the upgrade and use [utilpod.PodsAreChanged](https://github.com/pingcap/tidb-operator/blob/716d8e393944f958365c1cb72c49a2bc82a460f5/tests/e2e/util/pod/pod.go#L30) to check if they are changed or not

https://github.com/pingcap/tidb-operator/blob/716d8e393944f958365c1cb72c49a2bc82a460f5/tests/e2e/tidbcluster/stability.go#L165-L170

check pod by name is good to know which pod changed or not changed

tests/e2e/tidbcluster/serial.go
Yisaer: Actually `helm repo add` is my first choice, but it seems not stable because of the network.

tests/e2e/tidbcluster/serial.go
Yisaer: updated.

tests/e2e/tidbcluster/serial.go
```diff
@@ -1008,4 +1009,121 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
 		})
 	})
+
+	ginkgo.Context("[Verify: Upgrading Operator from 1.0.6", func() {
+		var oa tests.OperatorActions
+		var ocfg *tests.OperatorConfig
+		var version string
+
+		ginkgo.BeforeEach(func() {
+			version = os.Getenv("RELEASED_VERSION")
```
DanielZhangQD, line 1019: Is this necessary as we will set it in the next line?

tests/e2e/tidbcluster/serial.go
```diff
@@ -1008,4 +1009,121 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, fmt.Errorf("tidb should be updated after operator upgrading")
+				}
+
+				// confirm the pd Pod haven't been changed
+				changed, err = utilpod.PodsAreChanged(c, pdPods)()
+				if err != nil {
+					return false, nil
+				}
+				if changed {
+					return false, fmt.Errorf("pd replicas has changed after upgrading operator")
```
DanielZhangQD, line 1112:
```suggestion
					return false, fmt.Errorf("pd pods have been changed after upgrading operator")
```

tests/images/e2e/Dockerfile
```diff
@@ -29,5 +30,21 @@ ADD bin/webhook /usr/local/bin/
...
+    mv tidb-operator /charts/${RELEASED_VERSION}/tidb-operator && \
+    rm -rf tidb-operator && \
+    rm tidb-operator-chart-${RELEASED_VERSION}.tgz
+
+RUN curl -L https://github.com/pingcap/tidb-operator/releases/download/${RELEASED_VERSION}/tidb-cluster-chart-${RELEASED_VERSION}.tgz \
+    -o tidb-cluster-chart-${RELEASED_VERSION}.tgz && \
+    tar -zxvf tidb-cluster-chart-${RELEASED_VERSION}.tgz && \
+    mv tidb-cluster /charts/${RELEASED_VERSION}/tidb-cluster && \
+    rm -rf tidb-cluster && \
+    rm tidb-cluster-chart-${RELEASED_VERSION}.tgz
```
cofyc, line 47: download these charts in `Upgrading Operator from 1.0.6` test

tests/actions.go
```diff
@@ -505,7 +505,7 @@ func (oa *operatorActions) InstallCRDOrDie(info *OperatorConfig) {
 func (oa *operatorActions) DeployOperator(info *OperatorConfig) error {
 	klog.Infof("deploying tidb-operator %s", info.ReleaseName)
 
-	if info.Tag != "e2e" {
+	if info.Tag != "e2e" && !strings.HasPrefix(info.Tag, "v1.0.") {
```
cofyc, line 508: why not using `git clone`?

tests/e2e/tidbcluster/serial.go
Yisaer: updated.

tests/e2e/tidbcluster/serial.go
```diff
@@ -1008,4 +1008,131 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, nil
+				}
+				if changed {
+					return false, fmt.Errorf("tikv pods have been changed after upgrading operator")
+				}
+				klog.Infof("confirm tikv pods haven't been changed")
+
+				return true, nil
+			})
+			framework.ExpectNoError(err, "Failed to check TidbCluster Status After Upgrading Operator")
```
cofyc, line 1135: we should wait for pd/tikv pods are not changed in a certain of time, the expected err should be `wait.ErrWaitTimeout`, example:
https://github.com/pingcap/tidb-operator/blob/a19b72eb7cbf79f3ce77ff00d4a74547e05615d2/tests/e2e/tidbcluster/stability.go#L162-L189

`10*time.Minute` can be changed to `5*time.Minute`.  5 minutes is enough to make sure our logic is correct if pd/tikv pods are not affected in this time.

before checking pd/tikv pods are not affected, you can wait for tidb pods are affected. in this checking, what you expect is `nil` err (return `true, nil` when the tidb pods are different).

Yisaer: I think the current logic is like what you said. We would expect tidb pods updated yet, then we expect the tikv and pd pods haven't been changed. The whole process may need 5 or 10 minutes.

cofyc: you didn't check pd/tikv pods are not affected in a certain time.  after tidb pods are changed, the wait function checks pd/pods once and return `true, nil` if they are not changed. what we expect is they are not changed in a certain time (e.g. 5 minutes).

cofyc: you can have two `wait.Poll` functions:

1) the first waits for the tidb pods are changed
2) the second waits for the pd/tikv pods are not changed in 5 minutes

tests/images/e2e/Dockerfile
Yisaer: updated, using git clone now.

tests/e2e/tidbcluster/serial.go
Yisaer: good idea, updated.

tests/e2e/tidbcluster/serial.go
```diff
@@ -1008,4 +1008,135 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				return true, nil
+			})
+			framework.ExpectNoError(err, "Failed to check Tidb Status After Upgrading Operator")
+
+			err = wait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
+				// confirm the pd Pod haven't been changed
+				changed, err := utilpod.PodsAreChanged(c, pdPods)()
+				if err != nil {
+					klog.Errorf("meet error during verify pd pods, err:%v", err)
+					return true, nil
```
DanielZhangQD, line 1120: `return true` in this case?

tests/images/e2e/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM debian:buster-slim
 
 ENV KUBECTL_VERSION=v1.12.2
 ENV HELM_VERSION=v2.9.1
+ENV RELEASED_VERSION=v1.0.6
```
DanielZhangQD, line 5: Is this used?

tests/e2e/tidbcluster/serial.go
Yisaer: yes, we expect a timeout error here.

tests/images/e2e/Dockerfile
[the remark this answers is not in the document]
Yisaer: updated.

Issue #2096, opened 2020-04-01
Add more event for tidbcluster


Yisaer:
## Feature Request

Currently, we had little method to know what happened for a specific tidbcluster. We have to find logs in Operator. I think we could use event for some case to make user easier to know what happened in their tidbcluster.

For example, we could record the event for failover / scaling.

cofyc:
btw, we may need to record some conditions of tidb cluster objects, then we can emit events on state change (https://github.com/pingcap/tidb-operator/issues/2011).

cofyc:
should this be added into 1.1.0 milestone?

Yisaer:
yes

PR #2150, opened 2020-04-10
Add more events for tidbcluster and autoscaler

Yisaer:
Close https://github.com/pingcap/tidb-operator/issues/2096
Add event for tidbcluster and auto-scaler.
For auto-scaler, we would send event if the auto-scaling is permitted.
For tidbcluster, we would send event for each component (pd/tikv/tidb)'s / failover. 

```release-note
Emit more events for `TidbCluster` and `TidbClusterAutoScaler` to help users know TiDB running status
```


Yisaer:
@cofyc @DanielZhangQD @weekface  PTAL

pkg/manager/member/pd_scaler.go
```diff
@@ -36,12 +38,16 @@ type pdScaler struct {
...
-	pvcControl controller.PVCControlInterface) Scaler {
-	return &pdScaler{generalScaler{pdControl, pvcLister, pvcControl}}
+	pvcControl controller.PVCControlInterface,
+	recorder record.EventRecorder) Scaler {
+	return &pdScaler{generalScaler{pdControl, pvcLister, pvcControl, recorder}}
 }
 
 func (psd *pdScaler) Scale(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet, newSet *apps.StatefulSet) error {
 	scaling, _, _, _ := scaleOne(oldSet, newSet)
+	oldReplicas := *oldSet.Spec.Replicas
```
DanielZhangQD, line 48: It's possible that scaling=0 here.

pkg/manager/member/upgrader.go
```diff
@@ -18,6 +18,11 @@ import (
 	apps "k8s.io/api/apps/v1"
 )
 
+const (
+	upgradingEventReason         = "Upgrading"
+	upgradingEventMessagePattern = "%s is during upgrading"
```
DanielZhangQD, line 23:
```suggestion
	upgradingEventMessagePattern = "%s is upgrading"
```

pkg/manager/member/tikv_scaler.go
```diff
@@ -39,12 +41,16 @@ type tikvScaler struct {
...
-	podLister corelisters.PodLister) Scaler {
-	return &tikvScaler{generalScaler{pdControl, pvcLister, pvcControl}, podLister}
+	podLister corelisters.PodLister,
+	recorder record.EventRecorder) Scaler {
+	return &tikvScaler{generalScaler{pdControl, pvcLister, pvcControl, recorder}, podLister}
 }
 
 func (tsd *tikvScaler) Scale(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet, newSet *apps.StatefulSet) error {
 	scaling, _, _, _ := scaleOne(oldSet, newSet)
+	oldReplicas := *oldSet.Spec.Replicas
```
DanielZhangQD, line 51: ditto

pkg/manager/member/pd_failover.go
```diff
@@ -140,8 +140,8 @@ func (pf *pdFailover) tryToMarkAPeerAsFailure(tc *v1alpha1.TidbCluster) error {
 			return err
 		}
 
-		pf.recorder.Eventf(tc, apiv1.EventTypeWarning, "PDMemberMarkedAsFailure",
-			"%s(%s) marked as a failure member", podName, pdMember.ID)
+		msg := fmt.Sprintf("pd member[%s] is unhealthy", pdMember.ID)
+		pf.recorder.Event(tc, apiv1.EventTypeWarning, failoverEventReason, fmt.Sprintf(failoverEventMsgPattern, "pd", podName, msg))
```
DanielZhangQD, line 144: There is a common issue here, as we know, even if the events emitted in the failover/upgrading/scaling functions, it's possible that the sync logic is broken and returned with error without doing the failover/upgrading/scaling, in this case, the event will mismatch with the actual cluster status. What do you think? @weekface @cofyc

weekface: pd failover method can't be invoked if the status sync is broken.

DanielZhangQD: The event for failover seems acceptable from literal (but actually, it's possible that the cluster is not doing failover), but for scaling and upgrading, it's possible that it does not match with the actual cluster status.

cofyc: yes, events should be reported when something has actually happens, e.g. 

- observes something happened or state changes in the actual world (e.g. a PD member becomes unhealth)
- has done some operations (e.g. tidb-operator itself performed some modification requests)

in this place, we added this PD member into the failure member list, but `tidbcluster` object has not been updated yet and maybe skipped if the later operations fail (e.g. when syncing tidb/tikv). 

we can report a PD becomes unhealthy here (we have observed this), but should not report the PD has been marked as a failure member or during the PD failover (we haven't done this yet).

Yisaer: This sounds more reasonable. updated.

Yisaer:
Currently, I removed upgrading event (which is not important and also hard to maintain) and only leave the scaling and Unhealthy event.

pkg/manager/member/pd_scaler.go
```diff
@@ -36,16 +38,29 @@ type pdScaler struct {
...
+			if err != nil {
+				return err
+			}
+		} else if scaling < 0 {
+			err := psd.ScaleIn(tc, oldSet, newSet)
+			if err != nil {
+				return err
+			}
+		}
+		psd.recorder.Event(tc, corev1.EventTypeNormal, scalingEventReason, fmt.Sprintf(scalingEventMsgPattern, "pd", oldReplicas, targetReplicas))
```
DanielZhangQD, line 62: The updates to the statefulset will only be done [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/pd_member_manager.go#L267), any change before that line is possible to be skipped until next sync as stated in https://github.com/pingcap/tidb-operator/pull/2150#discussion_r407356912.

pkg/manager/member/tikv_scaler.go
```diff
@@ -39,16 +41,29 @@ type tikvScaler struct {
...
+			if err != nil {
+				return err
+			}
+		} else if scaling < 0 {
+			err := tsd.ScaleIn(tc, oldSet, newSet)
+			if err != nil {
+				return err
+			}
+		}
+		tsd.recorder.Event(tc, corev1.EventTypeNormal, scalingEventReason, fmt.Sprintf(scalingEventMsgPattern, "tikv", oldReplicas, targetReplicas))
```
DanielZhangQD, line 65: ditto

pkg/manager/member/pd_scaler.go
Yisaer: updated, scaling event removed for now.

pkg/controller/tidbcluster/tidb_cluster_controller.go
```diff
@@ -135,6 +135,7 @@ func NewController(
 				pdUpgrader,
 				autoFailover,
 				pdFailover,
+				recorder,
```
DanielZhangQD, line 138: This recorder and the ones in L154, L168 are not used.

pkg/manager/member/tikv_scaler_test.go
```diff
@@ -435,7 +435,6 @@ func newFakeTiKVScaler() (*tikvScaler, *pdapi.FakePDControl, cache.Indexer, cach
 	podInformer := kubeInformerFactory.Core().V1().Pods()
 	pdControl := pdapi.NewFakePDControl(kubeCli)
 	pvcControl := controller.NewFakePVCControl(pvcInformer)
-
```
DanielZhangQD, line 438: Please revert this.

Yisaer:
I would manually cp this.

cofyc:
it failed because of conflicts because we missed some PRs. we can cherry-pick them first, then the bot can cherry-pick automatically for us.

PR #2163, opened 2020-04-13
Add prometheus operator monitor example

mikechengwei:
Add prometheus opertaor  example.
Add prometheus opertaor 
### Check List 
 - Unit test
 - E2E test

Side effects

 - None


Yisaer:
We recommended to use `TidbMonitor` to monitor the TidbCluster. I don't think it is necessary to import promtheus-operator example into Tidb Operator repo.

mikechengwei:
@aylei  What do you think?

aylei:
I think this is a great and valuable PR, which is useful for users who has already ran prometheus-operator and want to integrate the monitoring of TiDB cluster to their own monitoring system.

BTW, I think the `Prometheus` object is not necessary, because it is managed by user and outside the scope of tidb-operator.

aylei:
Could you please add a README file about this example?

mikechengwei:
ok

examples/prometheus-operator/README.md
```diff
@@ -0,0 +1,51 @@
+# If you are use prometheus operator to monitor your service,you can see this example.
```
aylei, line 1: I think we could use a brief title and use this the first line of this doc

aylei: For example: how to monitor TiDB using prometheus-operator

examples/prometheus-operator/tikv-serviceMonitor.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_instance]
+      action: replace
+      targetLabel: cluster
+  namespaceSelector:
+    matchNames:
+      - default
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: tikv
+      app.kubernetes.io/instance: basic
```
Yisaer, line 55: plz add a blank line here.

Yisaer:
> BTW, I think the Prometheus object is not necessary, because it is managed by user and outside the scope of tidb-operator.

I'm thinking maybe we need to add tidbcluster and prometheus object here to build a complete case to show how to use prometheus-operator to monitor the tidbcluster? WDYT? @aylei @mikechengwei

mikechengwei:
I have write prometheus object and TIDB, I can add it. @Yisaer

examples/prometheus-operator/README.md
```diff
@@ -0,0 +1,55 @@
+# How to monitor TiDB using prometheus-operator
+
+> **Note:**
+>
+> This example describe ServiceMonitor and PrometheusRule yaml demo , you can use existing Prometheus CRD instance to load .
```
Yisaer, line 5:
```suggestion
This example describe how to monitor TidbCluster with existing Prometheus managed by [prometheus-operator](https://github.com/coreos/prometheus-operator) in the Kubernetes
```

examples/prometheus-operator/README.md
```diff
@@ -0,0 +1,55 @@
...
+basic-tidb   24h
+basic-tikv   24h
+```
+
+```bash
+$ kubectl get PrometheusRule -n <namespace>|grep basic
+NAME         AGE
+basic-prometheus-rules   23h
+```
+
```
Yisaer, line 49: I think we should also to add some steps to tell users how to see if prometheus has successfully monitor the tidbcluster ( like visit the endpoints pages)

examples/prometheus-operator/README.md
```diff
@@ -0,0 +1,55 @@
...
+    alertmanagers.monitoring.coreos.com            2020-04-12T14:22:49Z
+    podmonitors.monitoring.coreos.com              2020-04-12T14:22:50Z
+    prometheuses.monitoring.coreos.com             2020-04-12T14:22:50Z
+    prometheusrules.monitoring.coreos.com          2020-04-12T14:22:50Z
+    servicemonitors.monitoring.coreos.com          2020-04-12T14:22:51Z
+  ```
+  
+## Initialize
+
+Initialize serviceMonitor with label ` app: tidb cluster: basic` and prometheus rules.
```
Yisaer, line 29: We should tell users that this case is only used to demo and not suggest to use in the prod env.

examples/prometheus-operator/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+> **Note:**
+>
+> this example describe how to monitor TidbCluster with existing Prometheus managed by [prometheus-operator](https://github.com/coreos/prometheus-operator) in the Kubernetes
+> this case is only used to demo and not suggest to use in the prod env.
+
+**Prerequisites**: 
+- Has Prometheus operator installed. [Doc](https://github.com/coreos/kube-prometheus)
+    
+  This could by verified by the following command:
```
yeya24, line 11: s/by/be/g

examples/prometheus-operator/prometheus.yaml
```diff
@@ -0,0 +1,49 @@
...
+    prometheus: basic
+  name: basic
+  namespace: monitoring
+spec:
+  alerting:
+    alertmanagers:
+    - name: alertmanager-main
+      namespace: monitoring
+      port: web
+  thanos:
```
yeya24, line 14: I don't think we should add thanos in this example. Since you don't set up other things like thanos query, obj store, etc.

examples/prometheus-operator/prometheus-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+    app: prometheus
+    cluster: basic
+  name: basic-pometheus
+  namespace: monitoring
+spec:
+  ports:
+  - name: web
+    port: 9090
+    targetPort: 9090
+  - name: sidecar
```
yeya24, line 14: As I mentioned below, we could remove the sidecar port

examples/prometheus-operator/prometheus.yaml
mikechengwei: More companies use thanos , so this is neccessary. And this's  just a example.

mikechengwei: The purpose of this example is to tell the user can integrate tidb monitoring with the existing monitoring system.

yeya24: The demo is just a guide to integrate tidbmonitor with Prometheus operator, right? So it is unnecessary to have thanos, this is just part of feature that prom operator has, but not the core feature we need in this example. If you really want to mention thanos, you can add a link to the thanos example in the prometheus-operator repo

mikechengwei: This demo is a guide to teach user integrate tidbmonitor with existing monitor system，which maybe contain thanos framework.

Yisaer: I agreed with @yeya24 , thanos here is not necessary. This guide is to show how to monitor tidbcluster with prometheus-operator, so I think it should provide the same ability as tidbmonitor.

examples/prometheus-operator/prometheus.yaml
```diff
@@ -0,0 +1,46 @@
...
+  nodeSelector:
+    kubernetes.io/os: linux
+  podMonitorSelector: {}
+  externalLabels: 
+    monitor: basic-tidb
+  replicaExternalLabelName: replica
+  replicas: 1
+  resources:
+    requests:
+      memory: 25Gi
```
yeya24, line 25: Why 25Gi? Do we need that large memory?

Yisaer:
@mikechengwei  Hi, could you update the base branch?

mikechengwei:
@Yisaer CI is always failed.

PR #2166, opened 2020-04-13
Add unit test for restore controller

onlymellb:
Add unit tests for restore controller
Run `make test` directly
### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

```release-note
NONE
```


pkg/controller/restore/restore_controller_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+func TestRestoreControllerSync(t *testing.T) {
+	g := NewGomegaWithT(t)
+	type testcase struct {
+		name                 string
+		addRestoreToIndexer  bool
+		errWhenUpdateRestore bool
+		invalidKeyFn         func(restore *v1alpha1.Restore) string
+		errExpectFn          func(*GomegaWithT, error)
+	}
```
Yisaer, line 150: We recommend writing unit test in following style:

https://github.com/pingcap/tidb-operator/blob/da683864aef544ea6e6fa921c727d52c3d1bcfcd/pkg/autoscaler/autoscaler/util_test.go#L29-L94

there is no need to define testFn and we can use `go test -run <xxx/sub-xxx>` to test a sub case.

FYI: https://github.com/golang/go/wiki/TableDrivenTests#parallel-testing

PR #2168, opened 2020-04-13
add AGE column for all CRDs

cofyc:

Add the `AGE` column to show creation timestamp for all CRDs. This is convenient for users to know when the object was created. By the way, almost all Kubernetes built-in objects have this column too. `AGE` column is the default column in Kubernetes if `additionalPrinterColumns` is empty.

example:
```
$ kubectl  get tc
NAME    PD                  STORAGE   READY   DESIRE   TIKV                  STORAGE   READY   DESIRE   TIDB                  READY   DESIRE   AGE
basic   pingcap/pd:v3.0.8   1Gi       3       3        pingcap/tikv:v3.0.8   1Gi       3       3        pingcap/tidb:v3.0.8   2       2        5h8m
```

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Add the `AGE` column to show creation timestamp for all CRDs.
```

### DIFF
--- a/manifests/crd.yaml
+++ b/manifests/crd.yaml
@@ -50,6 +50,9 @@ spec:
     description: The desired replicas number of TiDB cluster
     name: Desire
     type: integer
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
   group: pingcap.com
   names:
     kind: TidbCluster
@@ -7748,6 +7751,9 @@ spec:
     name: Completed
     priority: 1
     type: date
+  - JSONPath: .metadata.creationTimestamp
+    name: Age
+    type: date
   group: pingcap.com
   names:
     kind: Backup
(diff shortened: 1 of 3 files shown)


manifests/crd.yaml
```diff
@@ -50,6 +50,9 @@ spec:
     description: The desired replicas number of TiDB cluster
     name: Desire
     type: integer
+  - JSONPath: .metadata.creationTimestamp
```
DanielZhangQD, line 53: TidbMonitor is missed?

cofyc: no `additionalPrinterColumns` for TidbMonitor, so it will use the default (`NAME` and `AGE`)

```
$ kubectl get tidbmonitor
NAME    AGE
basic   23s
```

Maybe we can set it explicitly?

cofyc: by the way, we can configure some useful columns for TidbMonitor CR. I've created an issue for this. https://github.com/pingcap/tidb-operator/issues/2169

PR #2151, opened 2020-04-10
security: tikv encryption kms config

shuijing198799:
### What problem does this PR solve? 
Add static encrytion of tikv data config in tikv config 
### What is changed and how does it work?
Add static encrytion of tikv data config in tikv config , generate the openapi, crd and docs
### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change

```release-note
add filed `TiKVEncryptionConfig` struct in `TiKVConfig` , it defines how to encrytion data key and raw data in TiKV, and what backup store the master key, details see the descryption in tikv_config.go
```

PR #2198, opened 2020-04-15
Automated cherry pick of #2151: security: tikv encryption kms config

cofyc:
Cherry pick of #2151 on release-1.1.

#2151: security: tikv encryption kms config


cofyc:
@shuijing198799 fixed the generated file, PTAL

PR #2213, opened 2020-04-16
fix br log issue and include both 3.1 and 4.0 br in the tidb-backup-manager image

DanielZhangQD:
…anager image

- The stderr logs from BR are missed
- The backup-manager cannot backup or restore v3.1 TiDB Cluster
- Add the stderr logs from BR
- Include both 3.1 and 4.0 br in the tidb-backup-manager image
- Backup controller gets the tikv version from tc and passes it to backup-manager
- backup-manager chooses the correct BR version to do backup and restore according to the tikv version
### Check List 

Tests 

   - Backup and Restore for TiDB Cluster v3.1.0-rc
   - Backup and Restore for TiDB Cluster v4.0.0-rc
   - Backup and Restore for TiDB Cluster nightly

Code changes

 - Has Go code change

Related changes

```release-note
Add stderr logs from BR to the backup-manager logs
```

### DIFF
--- a/cmd/backup-manager/app/backup/backup.go
+++ b/cmd/backup-manager/app/backup/backup.go
@@ -14,9 +14,11 @@
 package backup
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os/exec"
 	"path"
 	"strings"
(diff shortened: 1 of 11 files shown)


pkg/backup/backup/backup_manager.go
```diff
@@ -280,12 +288,19 @@ func (bm *backupManager) makeBackupJob(backup *v1alpha1.Backup) (*batchv1.Job, s
 	}
 
 	envVars = append(envVars, storageEnv...)
+	envVars = append(envVars, corev1.EnvVar{
```
DanielZhangQD, line 291: New version BR needs to set this env to print logs to stdout.

images/tidb-backup-manager/Dockerfile
```diff
@@ -2,19 +2,15 @@ FROM pingcap/tidb-enterprise-tools:latest
 ARG VERSION=v1.51.0
 ARG SHUSH_VERSION=v1.4.0
 ARG TOOLKIT_VERSION=v3.0.12
+ARG TOOLKIT_V31=v3.1.0-rc
+ARG TOOLKIT_V40=v4.0.0-rc
```
cofyc, line 6: `s/TOOLKIT/BR/g`

DanielZhangQD: But they are actually the toolkit version, BR binaries are included in the toolkit packages.
And also we may need to include other tools later in the toolkit packages.

cmd/backup-manager/app/backup/backup.go
```diff
@@ -64,29 +66,39 @@ func (bo *Options) backupData(backup *v1alpha1.Backup) (string, error) {
 	}
 	fullArgs = append(fullArgs, args...)
 	klog.Infof("Running br command with args: %v", fullArgs)
-	cmd := exec.Command("br", fullArgs...)
-	cmd.Stderr = cmd.Stdout
+	bin := "br" + backupUtil.Suffix(bo.TiKVVersion)
+	cmd := exec.Command(bin, fullArgs...)
```
cofyc, line 70: if `BR_LOG_TO_TERM` should be always set, set it here.

images/tidb-backup-manager/Dockerfile
shuijing198799: TOOLKIT has alot of tools I think TOOLKIT is all right.

cofyc: yes, we extract `br` from toolkit tarball but it's the implementation details. Your purpose is to specify the version of BR you wanted.

I can't understand the differences between these similar arguments.

```
ARG TOOLKIT_VERSION=v3.0.12
ARG TOOLKIT_V31=v3.1.0-rc
ARG TOOLKIT_V40=v4.0.0-rc
```

`s/TOOLKIT/TOOLKIT_BR/g` ?

cmd/backup-manager/app/backup/backup.go
DanielZhangQD: Yes, this is a way, I think the current setting is OK because we can see the env from the job clearly, if set it here, we cannot know it without going through the code or getting into the Pod.

images/tidb-backup-manager/Dockerfile
DanielZhangQD: If with `TOOLKIT_BR`, what if we need to retrieve other tools, e.g. Lightning from this tarball later, as you said, `br` is implementation details, here the toolkit version is what I want.

PR #2225, opened 2020-04-17
update the function to get the image tag

DanielZhangQD:
The original way to get the image tag did not consider the "127.0.0.1:30002/pingcap/tidb-operator:v3.1.0" case.
Update the function to get the image tag.
### Check List 

Tests 

   - Backup and Restore with v3.1.0-rc
   - Backup and Restore with nightly

Code changes

 - Has Go code change

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/backup/util/util.go
+++ b/pkg/backup/util/util.go
@@ -410,3 +410,13 @@ func ValidateRestore(restore *v1alpha1.Restore) error {
 	}
 	return nil
 }
+
+// GetImageTag get the image tag
+func GetImageTag(image string) string {
+	var tag string
+	colonIdx := strings.LastIndexByte(image, ':')
+	if colonIdx >= 0 {
+		tag = image[colonIdx+1:]
+	}
+	return tag
+}
--- a/pkg/backup/util/util.go
+++ b/pkg/backup/util/util.go
@@ -411,12 +411,15 @@ func ValidateRestore(restore *v1alpha1.Restore) error {
 	return nil
 }
 
-// GetImageTag get the image tag
-func GetImageTag(image string) string {
-	var tag string
+// ParseImage returns the image name and the tag from the input image string
+func ParseImage(image string) (string, string) {
+	var name, tag string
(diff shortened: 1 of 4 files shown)


pkg/backup/util/util.go
```diff
@@ -410,3 +410,13 @@ func ValidateRestore(restore *v1alpha1.Restore) error {
...
+
+// GetImageTag get the image tag
+func GetImageTag(image string) string {
+	var tag string
+	colonIdx := strings.LastIndexByte(image, ':')
+	if colonIdx >= 0 {
+		tag = image[colonIdx+1:]
+	}
+	return tag
+}
```
Yisaer, line 422: add a unit test.

pkg/backup/util/util.go
```diff
@@ -410,3 +410,13 @@ func ValidateRestore(restore *v1alpha1.Restore) error {
 	}
 	return nil
 }
+
+// GetImageTag get the image tag
```
cofyc, line 414: `gets` or `returns`

pkg/backup/util/util.go
```diff
@@ -410,3 +410,13 @@ func ValidateRestore(restore *v1alpha1.Restore) error {
...
+func GetImageTag(image string) string {
```
cofyc, line 415: update it to return `image` and `tag`, e.g. `func PaseImageTag(image string) (string, string)`.
in your case, discard the first result, e.g. `_, tikvVersion := PaseImageTag(...)`.
make it generic, then we can use it in more places.

DanielZhangQD:
@Yisaer @cofyc comments addressed, PTAL.

pkg/backup/util/util.go
DanielZhangQD: Done.

pkg/backup/util/util.go
DanielZhangQD: Done.

pkg/backup/util/util.go
DanielZhangQD: Done.

PR #2235, opened 2020-04-20
Add ai_elastic_scheduling in tidb-operator

Sullivan12138:
Add a new feature of ai_elastic_scheduling in tidb-operator.
Add an ai url in the  pkg/autoscaler/autoscaler/autoscaler_manager.go. From this url we will get predict data generated by another python process. I define two structures, beacuse the python process send data in this structure. I temporarily comment the syncTiDB function beacuse now we cannot predict the TiDB replicas. Now the syncTiKV function will directly use the predicted replicas from ai url, instead of getting data from Prometheus and calculating it. I change the scaleInSeconds and scaleOutSeconds for later test.

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

None

Related changes

None

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -70,6 +70,7 @@ require (
 	github.com/ugorji/go/codec v0.0.0-20190204201341-e444a5086c43
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
+	go.uber.org/zap v1.9.1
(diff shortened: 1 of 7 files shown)


Yisaer:
Hi @Sullivan12138 , I think you change is only worked in our internal test that the community can't use it. However, instead of implementing the specific server, I think it would be quite useful to make tidb-operator auto-scaler have extended ability by querying the external service to decide whether the target tidbcluster should auto-scale.

I think this would be much useful to community and it still could work in your case.

Issue #2219, opened 2020-04-16
PD placement rules should be enabled if using TiFlash


yeya24:
## Bug Report

**What version of TiDB Operator are you using?**

latest release

**What did you do?**

Start TiDBCluster with TiFlash

**What did you expect to see?**

TiFlash and TiDBcluster running normally

**What did you see instead?**

TiFlash failed to start because the PD placement-rules is not enabled, this is a prerequisite for TiFlash. Error log of TiFlash

```
klo yeya24-tiflash-0 -c proxylog
[2020/04/16 19:27:56.857 +00:00] [ERROR] [util.rs:327] ["request failed"] [err="Grpc(RpcFailure(RpcStatus { status: FailedPrecondition, details: Some(\"placement rules is disabled\") }))"]
[2020/04/16 19:27:56.857 +00:00] [ERROR] [util.rs:327] ["request failed"] [err="Grpc(RpcFailure(RpcStatus { status: FailedPrecondition, details: Some(\"placement rules is disabled\") }))"]
[2020/04/16 19:27:56.857 +00:00] [FATAL] [server.rs:287] ["failed to start node: Other(\"[src/pd/util.rs:335]: fail to request\")"]
```

PD config status:

```
pdctl config show | grep enable-placement
    "enable-placement-rules": "false",
```

If I manually enabled replacement rules in PD, then TiFlash starts perfectly

DanielZhangQD:
@yeya24 Thanks for reporting this issue, the integration with TiFlash is still in progress, we will enable the placement rules config automatically.

yeya24:
This can be closed now.

Yisaer:
I think we could using pdapi to configure placement rule before syncing tiflash instead of setting it in the config.toml in this PR https://github.com/pingcap/tidb-operator/pull/2328/files

PR #2237, opened 2020-04-20
scaling for tiflash

### LINKED ISSUE #2218: Cannot start tiflash if TimeZone is not set
## Bug Report

v1.15.2

The latest release

**What did you do?**

Deploy a TiDB cluster with TiFlash. However, I didn't set the TimeZone field in TiDBClusterSpec, because I thought the default value is `UTC`. However, TiFlash cannot start if TimeZone is not set explicitly. The error is

```
➜  tipocket git:(add-tiflash-crd) ✗ klo yeya24-tiflash-0 -c tiflash
Logging information to /data0/logs/server.log
Logging errors to /data0/logs/error.log
Poco::Exception. Code: 1000, e.code() = 0, e.displayText() = Exception: Could not determine time zone from TZ variable value: '': boost::filesystem::file_size: Operation not permitted: "/usr/share/zoneinfo", e.what() = Exception
```

If I set the TimeZone in the spec manually, then TiFlash can start correctly.

If I didn't set the TimeZone field, TiFlash can still start normally, since it will use the default time zone `UTC`.

**What did you see instead?**

The time zone env var in TiFlash is "". 
The problem is in the code below, I think we should use `tc.TimeZone()` instead of using `tc.Spec.Timezone`

``` Go
		{
			Name:  "TZ",
			Value: tc.Spec.Timezone,
		},
```

DanielZhangQD:
Scaling for TiFlash 
Add new function to support scaling TiFlash
### Check List 

Tests 

 - Unit test
    - Scale-out TiFlash successfully, Scale in TiFlash successfully, the store is deleted and the PVCs are marked as defer deleting, Scale-out TiFlash again and the old PVCs are removed and the scaling out succeeds.
    - Scale-out PD successfully, Scale in PD successfully, the member is deleted and the PVC is marked as defer deleting, Scale-out PD again and the old PVC is removed and the scaling out succeeds.

Code changes

 - Has Go code change

```release-note
Support scaling for TiFlash
```


pkg/controller/pod_control.go
```diff
@@ -305,7 +305,6 @@ func (fpc *FakePodControl) UpdateMetaInfo(_ *v1alpha1.TidbCluster, pod *corev1.P
 	}
 
 	setIfNotEmpty(pod.Labels, label.NameLabelKey, TestName)
-	setIfNotEmpty(pod.Labels, label.ComponentLabelKey, TestComponentName)
```
DanielZhangQD, line 308: No need to set it here, the caller function already set it and this will overwrite the setting in the cases.

pkg/controller/pvc_control.go
```diff
@@ -127,6 +127,7 @@ func (rpc *realPVCControl) UpdateMetaInfo(tc *v1alpha1.TidbCluster, pvc *corev1.
 	if pvc.Labels[label.ClusterIDLabelKey] == clusterID &&
 		pvc.Labels[label.MemberIDLabelKey] == memberID &&
 		pvc.Labels[label.StoreIDLabelKey] == storeID &&
+		pvc.Labels[label.AnnPodNameKey] == podName &&
```
DanielZhangQD, line 130: Add pod name label so we can list all the PVCs of one Pod.

pkg/manager/member/pd_member_manager.go
```diff
@@ -713,7 +713,15 @@ func getPDConfigMap(tc *v1alpha1.TidbCluster) (*corev1.ConfigMap, error) {
 		config.Dashboard.TiDBCertPath = path.Join(tidbClientCertPath, corev1.TLSCertKey)
 		config.Dashboard.TiDBKeyPath = path.Join(tidbClientCertPath, corev1.TLSPrivateKeyKey)
 	}
-
+	if tc.Spec.TiFlash != nil {
```
DanielZhangQD, line 716: If TiFlash is enabled and EnablePlacementRules is not set explicitly, enable it.

pkg/manager/member/orphan_pods_cleaner.go
```diff
@@ -92,39 +92,46 @@ func (opc *orphanPodsCleaner) Clean(tc *v1alpha1.TidbCluster) (map[string]string
 			continue
 		}
 
-		// TODO support multiple pvcs case?
-		var pvcName string
+		var pvcName []string
```
weekface, line 95:
```suggestion
		var pvcNames []string
```

DanielZhangQD: updated.

DanielZhangQD:
@weekface comments addressed, PTAL.

yeya24:
Thanks for the fix. Is there a rc release for this?

pkg/manager/member/pd_member_manager.go
Yisaer: Can we add a document to explain why EnablePlacementRules is necessary for tiflash

pkg/manager/member/tiflash_scaler.go
```diff
@@ -47,13 +53,114 @@ func (tfs *tiflashScaler) Scale(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulS
...
-	return nil
+	klog.Infof("scaling in tiflash statefulset %s/%s, ordinal: %d (replicas: %d, delete slots: %v)", oldSet.Namespace, oldSet.Name, ordinal, replicas, deleteSlots.List())
+	// We need delete store from cluster before decreasing the statefulset replicas
+	podName := ordinalPodName(v1alpha1.TiFlashMemberType, tcName, ordinal)
+	pod, err := tfs.podLister.Pods(ns).Get(podName)
+	if err != nil {
+		return err
+	}
+
+	if controller.PodWebhookEnabled {
```
Yisaer, line 94: I think we can manage tiflash scaling in the webhook  instead of in the controller as webhook would be the default choice to manage the lifecyle of each component?

pkg/manager/member/scaler.go
```diff
@@ -54,38 +54,94 @@ type generalScaler struct {
 }
 
 func (gs *generalScaler) deleteDeferDeletingPVC(tc *v1alpha1.TidbCluster,
-	setName string, memberType v1alpha1.MemberType, ordinal int32) (map[int32]string, error) {
+	setName string, memberType v1alpha1.MemberType, ordinal int32) (map[string]string, error) {
 	ns := tc.GetNamespace()
 	// for unit test
-	skipReason := map[int32]string{}
+	skipReason := map[string]string{}
```
Yisaer, line 60: Why change the skipReason type here, is it for the legibility？

DanielZhangQD: Now there will be multiple PVCs for one Pod, the `string` will be podName if no PVC found for the Pod and will be the pvcName if any PVC found, and then we can verify the result in the UT cases.

pkg/manager/member/tiflash_scaler.go
DanielZhangQD: When will Webhook be the default choice? I think it's not decided yet, btw, before v1.15, all of the requests for Pod will be sent to the Webhook if enabled, if something wrong with the Webhook, it will block the whole cluster, so I would like to support it in controller first.

pkg/manager/member/pd_member_manager.go
DanielZhangQD: Updated

DanielZhangQD:
@Yisaer comments addressed, PTAL.

DanielZhangQD:
> Thanks for the fix. Is there a rc release for this?

The fixes for #2218 and #2219 are moved to https://github.com/pingcap/tidb-operator/pull/2242 and it will be cherry-picked to release-1.1, whether to release another rc is not determined yet.

Issue #2173, opened 2020-04-14
Install newer EKS version with the terraform scripts on AWS


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, the default eks version is 1.12 which will not be supported soon.
**Describe the feature you'd like:**

Install newer EKS version with the terraform scripts on AWS
**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

DanielZhangQD:
With 1.14, the following error occurs:
```
Warning  FailedCreate  3m16s (x480 over 3h48m)  statefulset-controller  create Pod kolbe-test-cluster-tidb-0 in StatefulSet kolbe-test-cluster-tidb failed error: pods "kolbe-test-cluster-tidb-0" is forbidden: unable to validate against any pod security policy: [pod.spec.securityContext.sysctls[0]: Forbidden: unsafe sysctl "net.ipv4.tcp_keepalive_time" is not allowed pod.spec.securityContext.sysctls[1]: Forbidden: unsafe sysctl "net.ipv4.tcp_keepalive_intvl" is not allowed pod.spec.securityContext.sysctls[2]: Forbidden: unsafe sysctl "net.core.somaxconn" is not allowed]
```

PR #2238, opened 2020-04-20
Update eks version in deploy scripts

Yisaer:
Close https://github.com/pingcap/tidb-operator/issues/2173

Update the default eks version in terraform scripts as 1.15

As psp is enabled after eks 1.13, I add init annotation for tidb to initialize `sysctl` to avoiding pod create failed due to psp. FYI: https://docs.aws.amazon.com/eks/latest/userguide/pod-security-policy.html

```release-note
Update the default eks version in terraform scripts as 1.15
```


deploy/aws/variables.tf
```diff
@@ -14,12 +14,12 @@ variable "eks_name" {
 
 variable "eks_version" {
   description = "Kubernetes version to use for the EKS cluster."
-  default     = "1.12"
+  default     = "1.15"
```
cofyc, line 17: is it possible to use the platform default value? if users do not specify it explicitly, the default kubernetes version in AWS will be used.

Yisaer: IMO，using `default` or `latest` value is not stable. It may work well current time but encounter some problem when aws platform decided to update the version like this time( psp problem).

cofyc: make senses

PR #2250, opened 2020-04-21
Fix the config error in restore

Yisaer:
Fix the config error in restore

Related changes

```release-note
Fix the config error in restore
```


cmd/backup-manager/app/import/restore.go
```diff
@@ -58,14 +58,14 @@ func (ro *Options) loadTidbClusterData(restorePath string) error {
...
 	args := []string{
-		"--status-addr=0.0.0.0:8289",
-		"--backend=tidb",
-		"--server-mode=false",
-		"–-log-file=",
-		fmt.Sprintf("--tidb-user=%s", ro.User),
-		fmt.Sprintf("--tidb-password=%s", ro.Password),
-		fmt.Sprintf("--tidb-host=%s", ro.Host),
-		fmt.Sprintf("--d=%s", restorePath),
+		"-status-addr=0.0.0.0:8289",
```
DanielZhangQD, line 61: The options in the tidb-lightning chart are also with `--` prefix, could you please double confirm the format? 
https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl#L28

cmd/backup-manager/app/import/restore.go
```diff
@@ -61,7 +61,8 @@ func (ro *Options) loadTidbClusterData(restorePath string) error {
 		"--status-addr=0.0.0.0:8289",
 		"--backend=tidb",
 		"--server-mode=false",
-		"–-log-file=",
+		"--log-file=",
+		"--tidb-port=4000",
```
DanielZhangQD, line 65: Please use `ro.Port`.

Yisaer: updated.

Issue #2264, opened 2020-04-22
no `install-up-operator.sh` in `hack`


longfangsong:
## Bug Report

**What version of Kubernetes are you using?**

NA.
**What version of TiDB Operator are you using?**

NA.
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

NA.
**What's the status of the TiDB cluster pods?**

NA.
**What did you do?**

Run `hack/install-up-operator.sh`, as `CONTRIBUTING.md` describes.
**What did you expect to see?**
The script exists and do its job.
**What did you see instead?**
There's no such scirpt. Do you mean `local-up-operator.sh`?

I can create a pr to fix this, but shall I change the doc or the filename?

longfangsong:
BTW, the document [here](https://pingcap.com/docs-cn/tidb-in-kubernetes/stable/deploy-tidb-from-kubernetes-kind/#第-1-步通过-kind-部署-kubernetes-集群) said
```
export KUBECONFIG="$(kind get kubeconfig-path)"
```
but kubeconfig-path is not a valid command for the kind 0.7.0, which `hack/kind-cluster-build.sh` installs.

cofyc:
> BTW, the document [here](https://pingcap.com/docs-cn/tidb-in-kubernetes/stable/deploy-tidb-from-kubernetes-kind/#%E7%AC%AC-1-%E6%AD%A5%E9%80%9A%E8%BF%87-kind-%E9%83%A8%E7%BD%B2-kubernetes-%E9%9B%86%E7%BE%A4) said
> 
> ```
> export KUBECONFIG="$(kind get kubeconfig-path)"
> ```
> 
> but `get kubeconfig-path` is not a valid command for the kind 0.7.0, which `local-up-operator.sh -i` or `hack/kind-cluster-build.sh` installs.

yes, this doc is out of date now and can be simplified (https://github.com/pingcap/docs-tidb-operator/issues/149).

PR #2265, opened 2020-04-22
fix the wrong script name in `CONTRIBUTING.md`

longfangsong:

This fix the wrong script name in `CONTRIBUTING.md`,  which mentioned in #2264 .

Just change the `install-up-operator.sh` into `local-up-operator.sh`.

```release-note
NONE
```


cofyc:
Thanks for the fix!

Issue #2253, opened 2020-04-22
tidb-controller-manager does not get new CA/certs for a newly-deployed cluster


kolbe:
## Bug Report

**What version of Kubernetes are you using?**

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:42:50Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-eks-f459c0", GitCommit:"f459c0672169dd35e77af56c24556530a05e9ab1", GitTreeState:"clean", BuildDate:"2020-03-18T04:24:17Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of TiDB Operator are you using?**

a6bbe4c4bec08003797465fd6d575f3fde41408c

**What did you do?**

I created a TidbCluster with cert-manager controlled Cluster TLS enabled. I then deleted that TidbCluster and its namespace (and all secrets associated with the TidbCluster). Then I deployed a new TidbCluster with a new CA and certificates.

The tidb-controller-manager pod cannot connect to PD:
```
E0422 04:52:32.054065       1 pd_member_manager.go:206] failed to sync TidbCluster: [tidb-cluster/kolbe-test-cluster]'s status, error: Get https://kolbe-test-cluster-pd.tidb-cluster:2379/pd/health: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "TiDB")
E0422 04:52:32.065516       1 tidb_cluster_controller.go:282] TidbCluster: tidb-cluster/kolbe-test-cluster, sync failed TidbCluster: tidb-cluster/kolbe-test-cluster's pd status sync failed, can't failover, requeuing
```

**What did you expect to see?**

I think tidb-controller-manager should be able to manage this automatically.

**What did you see instead?**

If I deleted the tidb-controller-manager pod, it was recreated with the correct secrets and it could connect successfully to the PD service.

DanielZhangQD:
@weekface Please help check this issue, Thanks!

weekface:
[...]

PR #2275, opened 2020-04-23
Fix pdapi cache when cluster TLS enabled

weekface:

fixes: https://github.com/pingcap/tidb-operator/issues/2253

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Fix pdapi cache issue while cluster TLS is enabled
```

### DIFF
--- a/pkg/controller/tidb_control.go
+++ b/pkg/controller/tidb_control.go
@@ -27,6 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"net/http"
+	"sync"
 	"time"
 )
 
@@ -53,6 +54,10 @@ type TiDBControlInterface interface {
 
 // defaultTiDBControl is default implementation of TiDBControlInterface.
 type defaultTiDBControl struct {
+	// Now, all controllers use this single tidbConrol instance,
+	// we should add a mutex to avoid two controllers update the httpClient.Transport at same time.
+	mutex sync.Mutex
+	// TODO use httpClients cache instead, like pkg/pdapi/pdapi.go does.
 	httpClient *http.Client
 	kubeCli    kubernetes.Interface
 }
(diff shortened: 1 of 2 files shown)


pkg/pdapi/pdapi.go
```diff
@@ -67,6 +67,7 @@ func NewDefaultPDControl(kubeCli kubernetes.Interface) PDControlInterface {
 // It loads in-cluster root ca if caCert is empty.
 func GetTLSConfig(kubeCli kubernetes.Interface, namespace Namespace, tcName string, caCert []byte) (*tls.Config, error) {
 	secretName := util.ClusterClientTLSSecretName(tcName)
+	// TODO use secretLister instead of kubeCli.CoreV1().Secrets().Get() to optimize the performance
```
cofyc, line 70: this is intended, see https://github.com/pingcap/tidb-operator/issues/1327

weekface: got it.

pkg/controller/tidb_control.go
```diff
@@ -53,6 +54,10 @@ type TiDBControlInterface interface {
 
 // defaultTiDBControl is default implementation of TiDBControlInterface.
 type defaultTiDBControl struct {
+	// Now, all controllers use this single tidbConrol instance,
+	// we should add a mutex to avoid two controllers update the httpClient.Transport at same time.
+	mutex sync.Mutex
+	// TODO use httpClients cache instead, like pkg/pdapi/pdapi.go does.
```
cofyc, line 60: the cache mechanism in `pdapi` has an issue that stale clients will not be freed when the TC cluster is gone. If we want to optimize this, we can cache PD/TiDB clients (per TiDB cluster CR) in the tidb cluster controller instead and free on the deletion of the tidb cluster.

weekface: Cache the whole client per TiDB cluster in tidb cluster controller is a good idea.

PR #2281, opened 2020-04-23
add some useful packages to the bastion

kolbe:

I noticed that the bastion could be more useful if it had a few additional packages.

This PR adds screen, git, vim, bind-utils to bastion-userdata.

### Check List 

Tests 

I deployed with these changes in place and the new packages were available on the bastion host.

Related changes

```release-note
NONE
```


gregwebs:
We should stop using a bastion for the purpose of SSH and instead utilize [AWS SSM](https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html). It works really well, reduces cost, and increases security.

Then we need to make sure it is easy to modify security group rules to connect to MySQL from an instance of your choice where you can install whatever you want.

tennix:
> We should stop using a bastion for the purpose of SSH and instead utilize [AWS SSM](https://docs.aws.amazon.com/systems-manager/latest/userguide/ssm-agent.html). It works really well, reduces cost, and increases security.
> 
> Then we need to make sure it is easy to modify security group rules to connect to MySQL from an instance of your choice where you can install whatever you want.

I agree. But this PR looks fine, we should do SSM switching in other PRs. You can create an issue as a reminder for SSM integration.

PR #2284, opened 2020-04-24
Fix service update which would changed the nodeport

Yisaer:
currently if we make tidb service type as nodeport, it would change the nodeport in every reconcilation. This pr fix this bug.

### Check List 

Tests 

 - E2E test

Related changes

```release-note
Fix the bug when `TidbCluster` service type is `NodePort`, the nodeport would change frequently.
```


pkg/controller/service_control.go
```diff
@@ -83,6 +84,26 @@ func (sc *realServiceControl) UpdateService(tc *v1alpha1.TidbCluster, svc *corev
...
+						// Because the portName could be edited,
+						// we use Port number to link the desired Service Port and the existed Service Port in the nested loop
+						if dport.Port == eport.Port && dport.Protocol == eport.Protocol {
+							dport.NodePort = eport.NodePort
+							existingSvc.Spec.Ports[i] = dport
+							break
+						}
+					}
+				}
+			}
```
cofyc, line 106: why the node port may change? because we empty the `NodePort` field each time? what's your solution to fix this?

`UpdateService` is a generic function to update service and resolve conflicts if necessary. Trying to implement this logic in this function is strange. I think we should generate correct desired service specs before calling the API update. And generating the correct desired service can help avoid unnecessary API updates (skip if the desired spec matches existing spec).

Yisaer: This sounds make sense, updated.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -528,6 +528,50 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+		framework.ExpectNoError(err, "Expected TiDB cluster created")
+		err = oa.WaitForTidbClusterReady(tc, 30*time.Minute, 15*time.Second)
+		framework.ExpectNoError(err, "Expected TiDB cluster ready")
+		s, err := c.CoreV1().Services(ns).Get("nodePort-tidb", metav1.GetOptions{})
+		framework.ExpectNoError(err)
+		if s.Spec.Type != corev1.ServiceTypeNodePort {
+			framework.Failf("nodePort tidbcluster tidb service type isn't NodePort")
+		}
+		ports := s.Spec.Ports
+		// check node port unchanged for 5mins
```
DanielZhangQD, line 552: Update some attributes of the service and then check no change for nodeport?

Yisaer: updated

Yisaer: updated

DanielZhangQD:
@Yisaer please also check why the nodePort changed even if there is no update to the service configuration, maybe some checks are not correct which causes the service to be updated during each sync.

Yisaer:
> o check why the nodePort changed even if there is no update to the service configuration, maybe some checks are not correct which causes the service to b

I think the previous disscusion have already answered the question. The nodeport changed because we left nodeport empty during each syncing.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -528,6 +528,92 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+				for _, eport := range ports {
+					if dport.Port == eport.Port && dport.NodePort != eport.NodePort {
+						klog.Error("nodePort tidbcluster tidb service NodePort changed")
+						return true, nil
+					}
+				}
+			}
+			return false, nil
+		})
+		framework.ExpectError(err)
```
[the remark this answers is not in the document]
cofyc: actually, you can return these errors in this wait function. this will make this function shorter and errors will be printed if the test fails because what you expect is `Timeout` error.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -528,6 +528,92 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+		err = genericCli.Create(context.TODO(), nodeTc)
+		framework.ExpectNoError(err, "Expected TiDB cluster created")
+		err = oa.WaitForTidbClusterReady(tc, 30*time.Minute, 15*time.Second)
+		framework.ExpectNoError(err, "Expected TiDB cluster ready")
+
+		var s *corev1.Service
+		err = wait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
+			s, err = c.CoreV1().Services(ns).Get("nodeport-tidb", metav1.GetOptions{})
+			if err != nil {
+				klog.Errorf(err.Error())
```
cofyc, line 551: better to use `framework.Logf` which writes to ginkgo writer and the logs can be printed into ginkgo test report if failed.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -528,6 +528,92 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+			if s.Spec.Type != corev1.ServiceTypeNodePort {
+				return false, fmt.Errorf("nodePort tidbcluster tidb service type isn't NodePort")
+			}
+			return true, nil
+		})
+		framework.ExpectNoError(err)
+		ports := s.Spec.Ports
+
+		// check node port unchanged for 5mins
+		err = wait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
```
cofyc, line 563: 1 minute IMO is enough because our reconciliation period is 30 seconds.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -528,6 +528,92 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+		framework.ExpectError(err)
+		klog.Info("nodePort tidbcluster tidb service NodePort haven't changed")
+
+		tc, err = cli.PingcapV1alpha1().TidbClusters(ns).Get("nodeport", metav1.GetOptions{})
+		framework.ExpectNoError(err)
+		tc.Spec.TiDB.Service.Annotations = map[string]string{
+			"foo": "bar",
+		}
+		_, err = cli.PingcapV1alpha1().TidbClusters(ns).Update(tc)
+		framework.ExpectNoError(err)
```
cofyc, line 592: use previous `nodeTC` and `controller.GuaranteedUpdate`, e.g.

```
        err = controller.GuaranteedUpdate(genericCli, nodeTc, func() error {
            nodeTc.Spec.TiDB.Service.Annotations = map[string]string{
			    "foo": "bar",
    	    }  
            return nil
        }) 
```

`controller.GuaranteedUpdate(` can help resolve conflicts (status of `TidbCluster` objects are updated by controller frequently) and make our test stable.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -528,6 +528,92 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+			for _, oport := range ports {
+				for _, eport := range s.Spec.Ports {
+					if oport.Port == eport.Port && oport.Protocol == eport.Protocol && oport.NodePort != eport.NodePort {
+						return false, fmt.Errorf("node port has been changed")
+					}
+				}
+			}
+			return true, nil
+		})
+		framework.ExpectNoError(err)
```
cofyc, line 615: you should wait for `Timeout` error to happen. IMO you can reuse previous wait function. Testing the update of service succeeds or not is out of scope of this test.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -528,6 +528,92 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+		ports := s.Spec.Ports
+
+		// check node port unchanged for 5mins
+		err = wait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
+			s, err := c.CoreV1().Services(ns).Get("nodeport-tidb", metav1.GetOptions{})
+			if err != nil {
+				klog.Errorf(err.Error())
```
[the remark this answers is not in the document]
cofyc: actually, you can return these errors in this wait function. this will make this function shorter and errors will be printed if the test fails because what you expect is `Timeout` error.

DanielZhangQD:
> > o check why the nodePort changed even if there is no update to the service configuration, maybe some checks are not correct which causes the service to b
> 
> I think the previous disscusion have already answered the question. The nodeport changed because we left nodeport empty during each syncing.

I mean what why the service is updated even if no change to the service config in the tc, I have checked this and found that `ClusterIP` is included in the ServiceLastAppliedConfigAnnotation, but there is no `ClusterIP` in the new service spec, so the quality check always returns `false` [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/equality.go#L80), please fix this issue as it's the cause of `it would change the nodeport in every reconcilation` in the issue description.

Yisaer:
> > > o check why the nodePort changed even if there is no update to the service configuration, maybe some checks are not correct which causes the service to b
> > 
> > 
> > I think the previous disscusion have already answered the question. The nodeport changed because we left nodeport empty during each syncing.
> 
> I mean what why the service is updated even if no change to the service config in the tc, I have checked this and found that `ClusterIP` is included in the ServiceLastAppliedConfigAnnotation, but there is no `ClusterIP` in the new service spec, so the quality check always returns `false` [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/equality.go#L80), please fix this issue as it's the cause of `it would change the nodeport in every reconcilation` in the issue description.

I suggest to create a new issue to solve this problem.

tests/e2e/tidbcluster/tidbcluster.go
Yisaer: updated.

tests/e2e/tidbcluster/tidbcluster.go
Yisaer: updated.

tests/e2e/tidbcluster/tidbcluster.go
Yisaer: updated.

tests/e2e/tidbcluster/tidbcluster.go
Yisaer: updated

PR #2289, opened 2020-04-24
Fix auto scaler status syncing error

Yisaer:
Fix the error when syncing tidbclusterautoscaler status, add check whether the target sts is existed.

```release-note
Fix the error in syncing tidbclusterautoscaler status would cause panic when target tidbcluster is not existed.
```


cofyc:
suggested release notes:

```
Fix a panic in syncing `TidbClusterAutoScaler` status when the target `TidbCluster` does not exist
```

PR #2291, opened 2020-04-25
Fix sts webhook error

Yisaer:
Fix the error which would cause panic for sts webhook when sts updated strategy is not rolling updated.

```release-note
Fix a bug which would cause panic in statefulset webhook when update strategy of `StatefulSet` is not `RollingUpdate`
```

### DIFF
--- a/pkg/webhook/statefulset/statefulset.go
+++ b/pkg/webhook/statefulset/statefulset.go
@@ -110,12 +110,17 @@ func (sc *StatefulSetAdmissionControl) AdmitStatefulSets(ar *admission.Admission
 		return util.ARFail(err)
 	}
 
-	if stsPartition != nil && *stsPartition > 0 && *stsPartition <= int32(partition) {
-		klog.Infof("statefulset %s/%s has been protect by partition %s annotations", namespace, name, partitionStr)
-		return util.ARFail(errors.New("protect by partition annotation"))
+	if stsPartition != nil {
+		if *stsPartition > 0 && *stsPartition <= int32(partition) {
+			klog.Infof("statefulset %s/%s has been protect by partition %s annotations", namespace, name, partitionStr)
+			return util.ARFail(errors.New("protect by partition annotation"))
+		} else {
+			klog.Infof("admit statefulset %s/%s update partition to %d, protect partition is %d", namespace, name, *stsPartition, partition)
+			return util.ARSuccess()
+		}
+	} else {
+		return util.ARSuccess()
 	}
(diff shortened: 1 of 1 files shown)


pkg/webhook/statefulset/statefulset.go
```diff
@@ -110,12 +110,11 @@ func (sc *StatefulSetAdmissionControl) AdmitStatefulSets(ar *admission.Admission
...
 	}
 
-	setPartition := *(stsPartition)
-	if setPartition > 0 && setPartition <= int32(partition) {
+	if stsPartition != nil && *stsPartition > 0 && *stsPartition <= int32(partition) {
 		klog.Infof("statefulset %s/%s has been protect by partition %s annotations", namespace, name, partitionStr)
 		return util.ARFail(errors.New("protect by partition annotation"))
 	}
-	klog.Infof("admit statefulset %s/%s update partition to %d, protect partition is %d", namespace, name, setPartition, partition)
+	klog.Infof("admit statefulset %s/%s update partition to %d, protect partition is %d", namespace, name, *stsPartition, partition)
```
DanielZhangQD, line 117: `stsPartition` is possible to be nil here.

pkg/webhook/statefulset/statefulset.go
```diff
@@ -125,12 +124,18 @@ func getStsAttributes(data []byte, apiVersion string) (*metav1.ObjectMeta, *int3
...
 
 	set := apps.StatefulSet{}
 	if _, _, err := deserializer.Decode(data, nil, &set); err != nil {
 		return nil, nil, err
 	}
-	return &(set.ObjectMeta), set.Spec.UpdateStrategy.RollingUpdate.Partition, nil
+	if set.Spec.UpdateStrategy.RollingUpdate != nil {
+		return &(set.ObjectMeta), set.Spec.UpdateStrategy.RollingUpdate.Partition, nil
+	}
+	return &(set.ObjectMeta), nil, nil
```
cofyc, line 140: the code seems identical no matter what apiVersion is?

cofyc:
Suggested release note:

```
Fix a bug which would cause panic in statefulset webhook when update strategy of `StatefulSet` is not `RollingUpdate`
```

https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md

pkg/webhook/statefulset/statefulset.go
Yisaer: updated

pkg/webhook/statefulset/statefulset.go
Yisaer: updated.

PR #2292, opened 2020-04-25
Delete Occupied store when tikv scale-out

Yisaer:
Delete store if its address occupied the address during tikv scale-out. The store which occupied the address could be caused in following case:
1. tidbcluster with 2 tikv up
2. tidbcluster scale tikv to 3
3. During tikv-2 start process, the unexpected termination kill the tikv-2. The store is registered but the data lost.
4. The new tikv-2 restarted and encounter the duplicate address error and never start success.

To solve this problem, we would delete the store during tikv scale-out if the store occupied the target tikv store address.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Fix the error when newly tikv created failed, it would never started success.
```

### DIFF
--- a/pkg/manager/member/tikv_scaler.go
+++ b/pkg/manager/member/tikv_scaler.go
@@ -16,6 +16,7 @@ package member
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pingcap/advanced-statefulset/pkg/apis/apps/v1/helper"
@@ -82,7 +83,7 @@ func (tsd *tikvScaler) ScaleOut(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulS
 	// In that case, the address might been registered but the data couldn't be persisted which would occupied the address until
 	// be deleted manually.
 	for _, store := range storesInfo.Stores {
-		if store.Store.Address == scaleOutTikvAddresss {
+		if strings.HasPrefix(store.Store.Address, scaleOutTikvAddresss) {
 			err := pdClient.DeleteStore(store.Store.Id)
 			if err == nil {
 				err = fmt.Errorf("tc[%s/%s]'s tikv found occupied store address before scale-out, start to delete store[%d/%s]",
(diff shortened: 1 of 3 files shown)


pkg/util/util.go
```diff
@@ -226,3 +226,7 @@ func IsOwnedByTidbCluster(obj metav1.Object) (bool, *metav1.OwnerReference) {
 	}
 	return ref.Kind == v1alpha1.TiDBClusterKind && gv.Group == v1alpha1.SchemeGroupVersion.Group, ref
 }
+
+func GetPeerServiceName(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType) string {
```
DanielZhangQD, line 230: There is a `TiKVPeerMemberName` in controller_utils.go, should we unify them?

pkg/manager/member/tikv_scaler.go
```diff
@@ -67,6 +72,33 @@ func (tsd *tikvScaler) ScaleOut(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulS
...
+		return err
+	}
+	// check whether there existed store which have occupied the address before scale-out. If existed, we would delete the store first
+	// and wait the next round.
+	// The existed store which have occupied the address could be caused by the tikv which start failed due to the unexpected terminal
+	// In that case, the address might been registered but the data couldn't be persisted which would occupied the address until
+	// be deleted manually.
+	for _, store := range storesInfo.Stores {
+		if strings.HasPrefix(store.Store.Address, scaleOutTikvAddresss) {
+			err := pdClient.DeleteStore(store.Store.Id)
```
DanielZhangQD, line 91: No need to delete store for each sync, you can refer to the logic in `scaleIn`.

cofyc: I don't understand the problem we are trying to solve. Before `setReplicasAndDeleteSlots`, the new tikv pod will not be created, so in which case the address will be occupied?

cofyc:
```
- During tikv-2 start process, the unexpected termination kill the tikv-2. The store is registered but the data lost.
- The new tikv-2 restarted and encounter the duplicate address error and never start success.
```

after the tikv-2 pod is created, the work of tidb-operator scaling out is done. I don't think this can solve the problem. 

if the TikV may lose its registered store information in PD, I think it's a critical bug which should be solved in TiKV itself.

PR #2299, opened 2020-04-26
services are updated during every each sync

DanielZhangQD:
Fix https://github.com/pingcap/tidb-operator/pull/2284#issuecomment-619485792
And also the `last-applied-configuration` annotation is not set correctly during update for headless services.
Set the `last-applied-configuration` annotation correctly for services.
### Check List 

Tests 

   - `kubectl get svc -A --watch` confirms that services are not updated without any change to service spec.
   - check the log of tidb-controller-manager that no service update log without any change to service spec.

Code changes

 - Has Go code change

Related changes

```release-note
Fix the issue that services are updated during each sync even if no changes to the service configuration
```

### DIFF
--- a/pkg/manager/member/pd_member_manager.go
+++ b/pkg/manager/member/pd_member_manager.go
@@ -142,9 +142,7 @@ func (pmm *pdMemberManager) syncPDServiceForTidbCluster(tc *v1alpha1.TidbCluster
 		if err != nil {
 			return err
 		}
-		if newSvc.Spec.ClusterIP == "" {
-			svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP
-		}
+		svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP
 		_, err = pmm.svcControl.UpdateService(tc, &svc)
 		return err
 	}
--- a/pkg/manager/member/pd_member_manager.go
+++ b/pkg/manager/member/pd_member_manager.go
@@ -138,10 +138,12 @@ func (pmm *pdMemberManager) syncPDServiceForTidbCluster(tc *v1alpha1.TidbCluster
 		svc := *oldSvc
 		svc.Spec = newSvc.Spec
 		// TODO add unit test
-		svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP
 		err = controller.SetServiceLastAppliedConfigAnnotation(&svc)
 		if err != nil {
 			return err
+		}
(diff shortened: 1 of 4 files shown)


pkg/manager/member/tiflash_member_manager.go
```diff
@@ -146,7 +146,7 @@ func (tfmm *tiflashMemberManager) syncHeadlessService(tc *v1alpha1.TidbCluster)
 	if !equal {
 		svc := *oldSvc
 		svc.Spec = newSvc.Spec
-		err = controller.SetServiceLastAppliedConfigAnnotation(newSvc)
+		err = controller.SetServiceLastAppliedConfigAnnotation(&svc)
```
Yisaer, line 149: Do we need to set clusterIP here?

DanielZhangQD: We support setting clusterIP in the `service` of tc, so only need to set the clusterIP if it's not set in the `service` of the tc.

DanielZhangQD:
> I think There is no difference in logic for pd/tikv/tidb/tiflash to sync the service except the svc spec while I see the different logic in each controller. Can we extract the common logic and call the same function to handle this logic?

There are slight differences between the logics now, we can refactor them later.

pkg/manager/member/pd_member_manager.go
```diff
@@ -138,11 +138,13 @@ func (pmm *pdMemberManager) syncPDServiceForTidbCluster(tc *v1alpha1.TidbCluster
 		svc := *oldSvc
 		svc.Spec = newSvc.Spec
 		// TODO add unit test
-		svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP
 		err = controller.SetServiceLastAppliedConfigAnnotation(&svc)
 		if err != nil {
 			return err
 		}
+		if newSvc.Spec.ClusterIP == "" {
```
weekface, line 145: I think it makes no sense for users to update `clusterIP`.

Yisaer: If user fulfill the ClusterIP in the tidbcluster spec, shall we update the Service ClusterIP as user provided?

DanielZhangQD: ClusterIP cannot be updated after the initial set.
If users set ClusterIP during the cluster creation, the clusterIP will be set in the `CreateService` function.

DanielZhangQD:
@weekface @Yisaer comments addressed, PTAL again.

Issue #2241, opened 2020-04-21
UCP: Support the noise reduction for auto-scaling


Yisaer:
## Description

Currently, the auto-scaling doesn't have the  noise reduction design. It means in each reconciliation in auto-scaling controller, the result would be updated into tidbcluster immediately when the result indicates to scale-out or scale-in.

Now we want to support the noise reduction for auto-scaling and design detailed is designed in following:

For `TidbClusterAutoScaler`, the status would have 3 choices: Normal / ReadyToScaleOut / ReadyToScaleIn.

1. The initialial status is `Normal`
2. If the result in the reconciliation in auto-scaling controller is lager/smaller than the current replicas, the status would become `ReadyToScaleOut`/`ReadyToScaleIn` and record the timestamp in annotation.
3. If the status is remained as `ReadyToScaleOut` / `ReadyToScaleIn` for a certain time (which could also be configured in the `TidbClusterAutoScaler` spec), the auto-scaling controller would finally updated its result in the tidbcluster.
4. When auto-scaling updated the tidbcluster or the result in the reconciliation in auto-scaling controller is equal to the current replicas, the status should become `Normal` again.

I think the PR for this issue not only contains the logic of the noise reduction but also should have a e2e case to cover it. You could look the previous logic for auto-scaling in e2e here:
https://github.com/pingcap/tidb-operator/blob/5341babebc265c79bab978211493434f47dfe0d5/tests/e2e/tidbcluster/serial.go#L653

## Score

* 2500

## Mentor(s)

* [@yisaer](https://github.com/Yisaer)

Contact the mentors: **#tidb-challenge-program** channel in [TiDB-Operator Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace
As the logic is not simple and the detail is still needed to be discussed, we are welcome for every challenger to contact the mentor if you have any questions.

## Recommended Skills
* Golang
* Kubernetes

Yisaer:
@vincent178  Hi, Is there any progress now?

vincent178:
@Yisaer yeah, I'm still reading the code, will reach out to you maybe tomorrow or the day after to discuss some details, thanks.

[...]

PR #2307, opened 2020-04-26
[UCP] Noise reduction support when cluster auto scaling

vincent178:

UCP #2241 

Add Phrase for `TidbClusterAutoScaler`, it has three optional values: Normal, ReadyToScaleOut, ReadyToScaleIn.
Add user defined autoscale threshold.

use tidb as an example, tikv should be the same.

1. check phase
after autoscaler calculate the target replica, if target replica equals to current replica, set auto scaler phase to normal and return, otherwise check if the phase equal to scaleOut or scaleIn, if not, update the phase and record timestamp. go to 2.

2. check timestamp threshold
check the record timestamp remains longer than threshold which user defined. If not, return, otherwise go to 3.

3. do autoscale
set phase to normal and do normal autoscale.

**Design for e2e tests:**
After mock response from Prometheus, without noise reduction, it will start auto scaling after maximum 30s, but with noise reduction, we expect at least in 300s (ReadyToScaleThresholdSeconds), 
cluster should remain the replica number. And after `ReadyToScaleThresholdSeconds` time, the  cluster should start normal auto scaling and auto scale phase should be back to normal.

### Check List

Tests

 - Unit test
 - E2E test

Code changes

 - Has Go code change

Side effects

n/a

Related changes

```release-note
User now can specify `ReadyToScaleThresholdSeconds` in `TidbClusterAutoScaler`, until cluster keep in ready to auto scale status for `ReadyToScaleThresholdSeconds` seconds, then cluster start auto scaling.
```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
+++ b/pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
@@ -18,6 +18,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type AutoScalerPhase string
+
+const (
+	NormalAutoScalerPhase          AutoScalerPhase = "Normal"
+	ReadyToScaleOutAutoScalerPhase AutoScalerPhase = "ReadyToScaleOut"
+	ReadyToScaleInAutoScalerPhase  AutoScalerPhase = "ReadyToScaleIn"
+)
+
 // +genclient
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
@@ -76,6 +84,12 @@ type TidbClusterAutoScalerSpec struct {
 // TikvAutoScalerSpec describes the spec for tikv auto-scaling
 type TikvAutoScalerSpec struct {
 	BasicAutoScalerSpec `json:",inline"`
+
+	// ReadyToScaleThresholdSeconds represents duration that the ReadyToScale phase
+	// should last for before auto scaling.
+	// If not set, the default ReadyToScaleThresholdSeconds will be set to 30.
+	// +optional
+	ReadyToScaleThresholdSeconds *int32 `json:"readyToScaleThresholdSeconds,omitempty"`
 }
 
 // +k8s:openapi-gen=true
@@ -184,6 +198,8 @@ type TikvAutoScalerStatus struct {
 // +k8s:openapi-gen=true
 // BasicAutoScalerStatus describe the basic auto-scaling status
 type BasicAutoScalerStatus struct {
+	// Phase describes cluster auto scaling phase
+	Phase AutoScalerPhase `json:"phase,omitempty"`
 	// MetricsStatusList describes the metrics status in the last auto-scaling reconciliation
 	// +optional
 	MetricsStatusList []MetricsStatus `json:"metrics,omitempty"`
--- a/pkg/autoscaler/autoscaler/autoscaler_manager.go
+++ b/pkg/autoscaler/autoscaler/autoscaler_manager.go
@@ -151,6 +151,8 @@ func (am *autoScalerManager) updateAutoScaling(oldTc *v1alpha1.TidbCluster,
 		return nil, nil
 	}
 
+	tac.Annotations[label.AnnLastSyncingTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
+
 	if tac.Spec.TiKV != nil {
 		if oldTc.Status.TiKV.StatefulSet != nil {
 			tac.Status.TiKV.CurrentReplicas = oldTc.Status.TiKV.StatefulSet.CurrentReplicas
--- a/pkg/autoscaler/autoscaler/tikv_autoscaler.go
+++ b/pkg/autoscaler/autoscaler/tikv_autoscaler.go
@@ -70,12 +70,24 @@ func (am *autoScalerManager) syncTiKV(tc *v1alpha1.TidbCluster, tac *v1alpha1.Ti
 // The currentReplicas of TiKV calculated in auto-scaling is the count of the StateUp TiKV instance, so we need to
 // add the number of other state tikv instance replicas when we update the TidbCluster.Spec.TiKV.Replicas
 func syncTiKVAfterCalculated(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler, currentReplicas, recommendedReplicas int32, sts *appsv1.StatefulSet) error {
-
 	intervalSeconds := tac.Spec.TiKV.ScaleInIntervalSeconds
-	if recommendedReplicas > tc.Spec.TiKV.Replicas {
-		intervalSeconds = tac.Spec.TiKV.ScaleOutIntervalSeconds
+	if recommendedReplicas > currentReplicas {
+		if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleOutAutoScalerPhase {
+			tac.Status.TiKV.Phase = v1alpha1.ReadyToScaleOutAutoScalerPhase
+			// phase could change from Normal to ReadyToScaleOut, ReadyToScaleIn to ReadyToScaleOut,
+			// reset timestamp in both cases.
+			tac.Annotations[label.AnnTiKVReadyToScaleTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
+		}
+	} else {
+		if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleInAutoScalerPhase {
+			tac.Status.TiKV.Phase = v1alpha1.ReadyToScaleInAutoScalerPhase
+			// phase could change from Normal to ReadyToScaleIn, ReadyToScaleOut to ReadyToScaleIn,
+			// reset timestamp in both cases.
+			tac.Annotations[label.AnnTiKVReadyToScaleTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
+		}
 	}
-	ableToScale, err := checkStsAutoScalingInterval(tac, *intervalSeconds, v1alpha1.TiKVMemberType)
+
+	ableToScale, err := checkStsAutoScaling(tac, *tac.Spec.TiKV.ReadyToScaleThresholdSeconds, *intervalSeconds, v1alpha1.TiKVMemberType)
 	if err != nil {
 		return err
 	}
@@ -98,7 +110,6 @@ func filterTiKVInstances(tc *v1alpha1.TidbCluster) []string {
 
 // we record the auto-scaling out slot for tikv, in order to add special hot labels when they are created
 func updateTcTiKVIfScale(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler, currentReplicas, recommendedReplicas int32, sts *appsv1.StatefulSet) error {
-	tac.Annotations[label.AnnTiKVLastAutoScalingTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
 	if recommendedReplicas > currentReplicas {
 		newlyScaleOutOrdinalSets := helper.GetPodOrdinals(recommendedReplicas, sts).Difference(helper.GetPodOrdinals(currentReplicas, sts))
 		if newlyScaleOutOrdinalSets.Len() > 0 {
@@ -113,6 +124,8 @@ func updateTcTiKVIfScale(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAuto
 			tc.Annotations[label.AnnTiKVAutoScalingOutOrdinals] = v
 		}
 	}
+	tac.Status.TiKV.Phase = v1alpha1.NormalAutoScalerPhase
+	tac.Annotations[label.AnnTiKVLastAutoScalingTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
 	tc.Spec.TiKV.Replicas = recommendedReplicas
 	tac.Status.TiKV.RecommendedReplicas = &recommendedReplicas
 	return nil
--- a/pkg/autoscaler/autoscaler/util.go
+++ b/pkg/autoscaler/autoscaler/util.go
@@ -18,7 +18,9 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/jonboulle/clockwork"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+	"github.com/pingcap/tidb-operator/pkg/controller"
 	"github.com/pingcap/tidb-operator/pkg/label"
 	operatorUtils "github.com/pingcap/tidb-operator/pkg/util"
 	appsv1 "k8s.io/api/apps/v1"
@@ -57,11 +59,81 @@ func checkStsAutoScalingPrerequisites(set *appsv1.StatefulSet) bool {
 	return true
 }
 
-// checkStsAutoScalingInterval would check whether there is enough interval duration between every two auto-scaling
-func checkStsAutoScalingInterval(tac *v1alpha1.TidbClusterAutoScaler, intervalSeconds int32, memberType v1alpha1.MemberType) (bool, error) {
+func checkStsAutoScaling(tac *v1alpha1.TidbClusterAutoScaler, thresholdSeconds, intervalSeconds int32, memberType v1alpha1.MemberType) (bool, error) {
+	realClock := clockwork.NewRealClock()
 	if tac.Annotations == nil {
 		tac.Annotations = map[string]string{}
 	}
+	// 3*controller.ResyncDuration is maximum time allowed before reset phase status
+	ableToScale, err := checkLastSyncingTimestamp(tac, 3*controller.ResyncDuration, realClock)
+	if err != nil {
+		return false, err
+	}
+	if !ableToScale {
+		return false, nil
+	}
+	ableToScale, err = checkStsReadyAutoScalingTimestamp(tac, thresholdSeconds, realClock)
+	if err != nil {
+		return false, err
+	}
+	if !ableToScale {
+		return false, nil
+	}
+	ableToScale, err = checkStsAutoScalingInterval(tac, intervalSeconds, memberType)
+	if err != nil {
+		return false, err
+	}
+	if !ableToScale {
+		return false, nil
+	}
+	return true, nil
+}
+
+// checkLastSyncingTimestamp reset TiKV phase if last auto scaling timestamp is longer than thresholdSec
+func checkLastSyncingTimestamp(tac *v1alpha1.TidbClusterAutoScaler, thresholdSec time.Duration, clock clockwork.Clock) (bool, error) {
+	if tac.Annotations == nil {
+		tac.Annotations = map[string]string{}
+	}
+
+	lastAutoScalingTimestamp, existed := tac.Annotations[label.AnnLastSyncingTimestamp]
+	if !existed {
+		// NOTE: because record autoscaler sync timestamp happens after check auto scale,
+		// label will not exist during first sync, return allow auto scale in this case.
+		return true, nil
+	}
+	t, err := strconv.ParseInt(lastAutoScalingTimestamp, 10, 64)
+	if err != nil {
+		return false, err
+	}
+	// if there's no resync within thresholdSec, reset TiKV phase to Normal
+	if clock.Now().After(time.Unix(t, 0).Add(thresholdSec)) {
+		tac.Status.TiKV.Phase = v1alpha1.NormalAutoScalerPhase
+		return false, nil
+	}
+	return true, nil
+}
+
+// checkStsReadyAutoScalingTimestamp would check whether there is enough time window after ready to scale
+func checkStsReadyAutoScalingTimestamp(tac *v1alpha1.TidbClusterAutoScaler, thresholdSeconds int32, clock clockwork.Clock) (bool, error) {
+	readyAutoScalingTimestamp, existed := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+
+	if !existed {
+		tac.Annotations[label.AnnTiKVReadyToScaleTimestamp] = fmt.Sprintf("%d", clock.Now().Unix())
+		return false, nil
+	}
+	t, err := strconv.ParseInt(readyAutoScalingTimestamp, 10, 32)
+	if err != nil {
+		return false, err
+	}
+	readyAutoScalingSec := int32(clock.Now().Sub(time.Unix(t, 0)).Seconds())
+	if thresholdSeconds > readyAutoScalingSec {
+		return false, nil
+	}
+	return true, nil
+}
+
+// checkStsAutoScalingInterval would check whether there is enough interval duration between every two auto-scaling
+func checkStsAutoScalingInterval(tac *v1alpha1.TidbClusterAutoScaler, intervalSeconds int32, memberType v1alpha1.MemberType) (bool, error) {
 	lastAutoScalingTimestamp, existed := tac.Annotations[label.AnnTiDBLastAutoScalingTimestamp]
 	if memberType == v1alpha1.TiKVMemberType {
 		lastAutoScalingTimestamp, existed = tac.Annotations[label.AnnTiKVLastAutoScalingTimestamp]
@@ -146,6 +218,9 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
 			if tac.Spec.TiKV.MetricsTimeDuration == nil {
 				tac.Spec.TiKV.MetricsTimeDuration = pointer.StringPtr("3m")
 			}
+		}
+		if tac.Spec.TiKV.ReadyToScaleThresholdSeconds == nil {
+			tac.Spec.TiKV.ReadyToScaleThresholdSeconds = pointer.Int32Ptr(30)
 		}
 	}
 
--- a/pkg/label/label.go
+++ b/pkg/label/label.go
@@ -109,6 +109,12 @@ const (
 	// AnnTiKVLastAutoScalingTimestamp is annotation key of tidbclusterto which ordinal is created by tikv auto-scaling
 	AnnTiKVLastAutoScalingTimestamp = "tikv.tidb.pingcap.com/last-autoscaling-timestamp"
 
+	// AnnTiKVReadyToScaleTimestamp records timestamp when tikv ready to scale
+	AnnTiKVReadyToScaleTimestamp = "tikv.tidb.pingcap.com/ready-to-scale-timestamp"
+
+	// AnnLastSyncingTimestamp records last sync timestamp
+	AnnLastSyncingTimestamp = "tidb.pingcap.com/last-syncing-timestamp"
+
 	// AnnTiDBConsecutiveScaleOutCount describes the least consecutive count to scale-out for tidb
 	AnnTiDBConsecutiveScaleOutCount = "tidb.tidb.pingcap.com/consecutive-scale-out-count"
 	// AnnTiDBConsecutiveScaleInCount describes the least consecutive count to scale-in for tidb
--- a/tests/e2e/tidbcluster/serial.go
+++ b/tests/e2e/tidbcluster/serial.go
@@ -435,6 +435,7 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
 					MetricsTimeDuration:    &duration,
 					ScaleInIntervalSeconds: pointer.Int32Ptr(100),
 				},
+				ReadyToScaleThresholdSeconds: pointer.Int32Ptr(40),
 			}
 			tac.Spec.TiKV.Metrics = []autoscalingv2beta2.MetricSpec{}
 			tac.Spec.TiKV.Metrics = append(tac.Spec.TiKV.Metrics, defaultMetricSpec)
@@ -445,23 +446,49 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
 			framework.ExpectNoError(err, "create pdapi error")
 			defer cancel()
 			var firstScaleTimestamp int64
-			err = wait.Poll(10*time.Second, 10*time.Minute, func() (done bool, err error) {
+			var readyToScaleTimestamp int64
+			err = wait.Poll(10*time.Second, 5*time.Minute, func() (done bool, err error) {
+				tac, err = cli.PingcapV1alpha1().TidbClusterAutoScalers(ns).Get(tac.Name, metav1.GetOptions{})
+				if err != nil {
+					return false, nil
+				}
+				if tac.Annotations == nil || len(tac.Annotations) < 1 {
+					framework.Logf("tac haven't marked any annotation")
+					return false, nil
+				}
+				t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+				if !ok {
+					framework.Logf("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+					return false, nil
+				}
+				readyToScaleTimestamp, err = strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					return false, err
+				}
+				if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleOutAutoScalerPhase {
+					framework.Logf("tac dont' have the right ReadyToScale phase, expect: %s, got %s", v1alpha1.ReadyToScaleOutAutoScalerPhase, tac.Status.TiKV.Phase)
+					return false, nil
+				}
+				return true, nil
+			})
+			framework.ExpectNoError(err, "check tikv has ready-to-scale-timestamp")
+			err = wait.Poll(10*time.Second, 5*time.Minute, func() (done bool, err error) {
 				tc, err := cli.PingcapV1alpha1().TidbClusters(tc.Namespace).Get(tc.Name, metav1.GetOptions{})
 				if err != nil {
 					return false, nil
 				}
 				// check replicas
 				if tc.Spec.TiKV.Replicas != int32(4) {
-					klog.Infof("tikv haven't auto-scale to 4 replicas")
+					framework.Logf("tikv haven't auto-scale to 4 replicas")
 					return false, nil
 				}
 				if len(tc.Status.TiKV.Stores) != 4 {
-					klog.Infof("tikv's stores haven't auto-scale to 4")
+					framework.Logf("tikv's stores haven't auto-scale to 4")
 					return false, nil
 				}
 				// check annotations
 				if tc.Annotations == nil || len(tc.Annotations) < 1 {
-					klog.Infof("tc haven't marked any annotation")
+					framework.Logf("tc haven't marked any annotation")
 					return false, nil
 				}
 				tac, err = cli.PingcapV1alpha1().TidbClusterAutoScalers(ns).Get(tac.Name, metav1.GetOptions{})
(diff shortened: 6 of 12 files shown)


pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -175,6 +175,7 @@ type TikvAutoScalerStatus struct {
 // +k8s:openapi-gen=true
 // BasicAutoScalerStatus describe the basic auto-scaling status
 type BasicAutoScalerStatus struct {
+	Phase MemberPhase `json:"phase,omitempty"`
```
Yisaer, line 178: I tend to define a new struct as `AutoScalerPhase`.

pkg/label/label.go
```diff
@@ -104,6 +104,10 @@ const (
 	// AnnTiKVLastAutoScalingTimestamp is annotation key of tidbclusterto which ordinal is created by tikv auto-scaling
 	AnnTiKVLastAutoScalingTimestamp = "tikv.tidb.pingcap.com/last-autoscaling-timestamp"
 
+	AnnTiDBReadyToScaleTimestamp = "tidb.tidb.pingcap.com/ready-to-scale-timestamp"
+
+	AnnTiKVReadyToScaleTimestamp = "tikv.tidb.pingcap.com/ready-to-scale-timestamp"
```
Yisaer, line 109: please add note to explain the annotation

pkg/autoscaler/autoscaler/util.go
```diff
@@ -57,6 +57,25 @@ func checkStsAutoScalingPrerequisites(set *appsv1.StatefulSet) bool {
 	return true
 }
 
+func checkStsReadyAutoScalingTimestamp(tac *v1alpha1.TidbClusterAutoScaler, threshold int32) (bool, error) {
+	if tac.Annotations == nil {
```
Yisaer, line 61: I think this function should add `v1alpha1.MemberType` as param thus the tikv could also use this function.

Yisaer: And also please add the note to explain the function usage and add an unit-test to cover it.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -59,8 +60,28 @@ func syncTiDBAfterCalculated(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbCluster
...
+	}
+	ableToScale, err := checkStsReadyAutoScalingTimestamp(tac, 123)
+	if err != nil {
+		return err
+	}
+	if !ableToScale {
+		return nil
 	}
-	ableToScale, err := checkStsAutoScalingInterval(tac, *intervalSeconds, v1alpha1.TiDBMemberType)
+	ableToScale, err = checkStsAutoScalingInterval(tac, *intervalSeconds, v1alpha1.TiDBMemberType)
```
Yisaer, line 84: I think we could extract `checkStsReadyAutoScalingTimestamp` and `checkStsAutoScalingInterval` into one function as `checkStsAutoScaling`, then we would call `checkStsReadyAutoScalingTimestamp` and `checkStsAutoScalingInterval` in this function.

pkg/autoscaler/autoscaler/tidb_autoscaler.go
```diff
@@ -59,8 +60,28 @@ func syncTiDBAfterCalculated(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbCluster
...
+		}
+	} else {
+		if tac.Status.TiDB.Phase != v1alpha1.ReadyToScaleIn {
+			tac.Status.TiDB.Phase = v1alpha1.ReadyToScaleIn
+			// phase could change from Normal to ReadyToScaleIn, ReadyToScaleOut to ReadyToScaleIn,
+			// reset timestamp in both cases.
+			tac.Annotations[label.AnnTiDBReadyToScaleTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
+		}
+	}
+	ableToScale, err := checkStsReadyAutoScalingTimestamp(tac, 123)
```
Yisaer, line 77: We should expose the time window in `Spec`

DanielZhangQD:
@Yisaer As we have discussed before, we also have to add the step control for each scaling, for example, we do not want to scale out/in 10 Pods at one time even if the calculated replicas is that we need to scale out/in 10 replicas to reduce the jitter.

Yisaer:
> @Yisaer As we have discussed before, we also have to add the step control for each scaling, for example, we do not want to scale out/in 10 Pods at one time even if the calculated replicas is that we need to scale out/in 10 replicas to reduce the jitter.

We would relieaze it later in a new pr.

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -70,6 +78,13 @@ type TidbClusterAutoScalerSpec struct {
 	// TiDB represents the auto-scaling spec for tidb
 	// +optional
 	TiDB *TidbAutoScalerSpec `json:"tidb,omitempty"`
+
+	// ReadyToScaleThresholdSeconds represents cluster needs to keep ReadyToScale phase
+	// for ReadyToScaleThresholdSeconds time to start auto scaling.
+	// This is only applied to TiKV.
+	// If not set, the default ReadyToScaleThresholdSeconds will be set to 300.
+	// +optional
+	ReadyToScaleThresholdSeconds *int32 `json:"readyToScaleThresholdSeconds,omitempty"`
```
Yisaer, line 87: I suggest move this config to `TiKVSpec` as it only works for tikv.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -33,6 +34,10 @@ const (
 
 	invalidMemberTypeErrorMsg    = "tac[%s/%s] invalid set MemberType:%s"
 	invalidTacAnnotationErrorMsg = "tac[%s/%s] annotation invalid set,err:%v"
+
+	// lastSyncThresholdSec is maximum time allowed before reset phase status
+	// 30 is the resync during, 3 is a factor
+	lastSyncThresholdSec = 30 * 3
```
Yisaer, line 40: Replace 30 as `controller.ResyncDuration`, so the lastSyncThreshold would be 3 times ResyncDuration

pkg/autoscaler/autoscaler/util.go
```diff
@@ -174,6 +248,10 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
 			tac.Spec.Monitor.Namespace = tac.Namespace
 		}
 	}
+
+	if tac.Spec.ReadyToScaleThresholdSeconds == nil {
+		tac.Spec.ReadyToScaleThresholdSeconds = pointer.Int32Ptr(300)
+	}
```
Yisaer, line 254: suggest move this in `defaultTac` function

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -180,6 +194,7 @@ type TikvAutoScalerStatus struct {
 // +k8s:openapi-gen=true
 // BasicAutoScalerStatus describe the basic auto-scaling status
 type BasicAutoScalerStatus struct {
+	Phase AutoScalerPhase `json:"phase,omitempty"`
```
Yisaer, line 197: Please Add comment to explain this Atrribute.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -147,6 +219,9 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
 				tac.Spec.TiKV.MetricsTimeDuration = pointer.StringPtr("3m")
 			}
 		}
+		if tac.Spec.TiKV.ReadyToScaleThresholdSeconds == nil {
+			tac.Spec.TiKV.ReadyToScaleThresholdSeconds = pointer.Int32Ptr(300)
+		}
```
Yisaer, line 224: Move this logic in `defaultTac` function.

tests/e2e/tidbcluster/serial.go
```diff
@@ -22,6 +22,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-openapi/errors"
```
Yisaer, line 25: make goimports fmt.

tests/e2e/tidbcluster/serial.go
```diff
@@ -985,6 +994,308 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				if err != nil {
+					return false, nil
+				}
+				for _, label := range info.Store.Labels {
+					if label.Key == "specialUse" && label.Value == "hotRegion" {
+						return true, nil
+					}
+				}
+				klog.Infof("tikv auto-scale out haven't find the special label")
+				return false, nil
```
Yisaer, line 1137: What about merge noise reduction test case into the existed tikv auto-scaling e2e test ? We don't have to create a new e2e test case.

tests/e2e/tidbcluster/serial.go
```diff
@@ -746,6 +753,39 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				framework.ExpectNoError(err, "get tidb cluster auto scaler error")
+			}
+			if tac.Annotations == nil || len(tac.Annotations) < 1 {
+				klog.Infof("tac haven't marked any annotation")
+				framework.ExpectNoError(err, "tac no annotation error")
+			}
+			t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+			if !ok {
+				klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+				framework.ExpectNoError(errors.NotFound(""), "tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation error")
```
cofyc, line 776: `go-openapi/errors` is not suitable here (it generates HTTP not found error). you can simply use `framework.Failf(...)`

tests/e2e/tidbcluster/serial.go
```diff
@@ -825,6 +865,38 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				framework.ExpectNoError(err, "get tidb cluster auto scaler error")
+			}
+			if tac.Annotations == nil || len(tac.Annotations) < 1 {
+				klog.Infof("tac haven't marked any annotation")
+				framework.ExpectNoError(err, "tac no annotation error")
+			}
+			t, ok = tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+			if !ok {
+				klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+				framework.ExpectNoError(errors.NotFound(""), "tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation error")
```
cofyc, line 887: ditto

tests/e2e/tidbcluster/serial.go
```diff
@@ -746,6 +753,39 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
 			pdClient, cancel, err := proxiedpdclient.NewProxiedPDClient(c, fw, ns, clusterName, false, nil)
 			framework.ExpectNoError(err, "create pdapi error")
 			defer cancel()
+
+			// wait for ReadyToScaleThresholdSeconds seconds
+			klog.Info("wait 300 seconds, auto scale should not happen with noise reduction")
+			time.Sleep(300 * time.Second)
```
cofyc, line 759: too much sleep time in these two e2e tests, it will take a long time to finish. can sleep time be eliminated? is it enough to cover this feature in unit tests?

cofyc:
hi, @vincent178

thanks for your patience!

I have a few concerns about e2e tests. How about merging API/Controller change first? cc @Yisaer What do you think?

E2E tests can be submitted and reviewed in a separate follow-up PR.

tests/e2e/tidbcluster/serial.go
vincent178: yeah, agree, one optimized way is to set ReadyToScaleThresholdSeconds to 30s when deploy operator, only only wait 30s here should already show noise reduction capability. what do you think?

Yisaer:
I think we can make the operator enable Auto-scaling in Basic ginkgo context and move the auto-scaling e2e test case to basic ginkgo context from serial. Currently I think changing `ReadyToScaleThresholdSeconds` to 30s is also ok for me and don't need to add new e2e case.

tests/e2e/tidbcluster/serial.go
```diff
@@ -752,6 +753,16 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+				if !ok {
+					klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+					return false, nil
+				}
+				readyToScaleTimestamp, err := strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					framework.ExpectNoError(err, "tikv.tidb.pingcap.com/ready-to-scale-timestamp parse int error")
+					return false, nil
+				}
```
Yisaer, line 765: I think we shouldn't check `AnnTiKVReadyToScaleTimestamp` and whether tikv replicas is 4 in the same `Wait.Pull`, we should first check `AnnTiKVReadyToScaleTimestamp` annotation and tidbclusterAutoScaler status in one `Wait.Pull`, then we would check whether `tidbclusterAutoScaler` become ready Normal and whether tikv replicas become 4.

tests/e2e/tidbcluster/serial.go
```diff
@@ -830,6 +844,16 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
```
Yisaer, line 856: ditto.

tests/e2e/tidbcluster/serial.go
```diff
@@ -747,7 +748,32 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				if err != nil {
+					return true, err
+				}
+				if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleOutAutoScalerPhase {
+					return true, fmt.Errorf("tikv don't have right ReadyToScale phase")
+				}
+				return true, nil
+			})
+			framework.ExpectNoError(err, "check tikv has ready-to-scale-timestamp")
+			err = wait.Poll(10*time.Second, 5*time.Minute, func() (done bool, err error) {
```
vincent178, line 776: update from 10mins to 5mins because the cluster is already running for a while.

tests/e2e/tidbcluster/serial.go
```diff
@@ -825,6 +858,29 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, nil
+				}
+				t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+				if !ok {
+					klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+					return false, nil
+				}
+				readyToScaleTimestamp, err = strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					return true, err
```
vincent178, line 877: return `true` because we should end the poll loop as early as possible once found error to save time.

tests/e2e/tidbcluster/serial.go
```diff
@@ -825,6 +858,30 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
 
+			err = wait.Poll(10*time.Second, 5*time.Minute, func() (done bool, err error) {
+				tac, err = cli.PingcapV1alpha1().TidbClusterAutoScalers(ns).Get(tac.Name, metav1.GetOptions{})
+				if err != nil {
+					return false, nil
+				}
+				if tc.Annotations == nil || len(tc.Annotations) < 1 {
+					klog.Infof("tc haven't marked any annotation")
+					return false, nil
+				}
```
Yisaer, line 869: Why we need to check the tc annotations here? It seems the e2e tesd failed due to this.

tests/e2e/tidbcluster/serial.go
```diff
@@ -747,7 +748,32 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
 			var firstScaleTimestamp int64
-			err = wait.Poll(10*time.Second, 10*time.Minute, func() (done bool, err error) {
+			var readyToScaleTimestamp int64
+			err = wait.Poll(10*time.Second, 5*time.Minute, func() (done bool, err error) {
+				tac, err = cli.PingcapV1alpha1().TidbClusterAutoScalers(ns).Get(tac.Name, metav1.GetOptions{})
+				if err != nil {
+					return false, nil
+				}
+				if tac.Annotations == nil || len(tac.Annotations) < 1 {
+					klog.Infof("tc haven't marked any annotation")
```
Yisaer, line 758: use `tac haven't marked any annotation` instead and use `framework.Logf` to record the log.

tests/e2e/tidbcluster/serial.go
```diff
@@ -747,7 +748,32 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, nil
+				}
+				t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+				if !ok {
+					klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
```
Yisaer, line 763: ditto.

tests/e2e/tidbcluster/serial.go
```diff
@@ -747,7 +748,32 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, nil
+				}
+				readyToScaleTimestamp, err = strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					return true, err
+				}
+				if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleOutAutoScalerPhase {
+					return true, fmt.Errorf("tikv don't have right ReadyToScale phase")
```
Yisaer, line 771: return false, fmt.Errorf("tikv don't have right ReadyToScale phase")

tests/e2e/tidbcluster/serial.go
```diff
@@ -747,7 +748,32 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, nil
+				}
+				t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+				if !ok {
+					klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+					return false, nil
+				}
+				readyToScaleTimestamp, err = strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					return true, err
```
Yisaer, line 768: return false,err

tests/e2e/tidbcluster/serial.go
```diff
@@ -783,6 +809,13 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
 				if err != nil {
 					return false, err
 				}
+				// check readyToScaleTimestamp
+				if time.Now().Sub(time.Unix(readyToScaleTimestamp, 0)).Seconds() < 40 {
+					return true, fmt.Errorf("tikv seconds scale doesn't meet the ReadyToScale threshold")
```
Yisaer, line 814: return false, fmt.Errorf("tikv seconds scale doesn't meet the ReadyToScale threshold")

tests/e2e/tidbcluster/serial.go
```diff
@@ -783,6 +809,13 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				}
+				if tac.Status.TiKV.Phase != v1alpha1.NormalAutoScalerPhase {
+					return true, fmt.Errorf("tikv don't have right ReadyToScale phase")
```
Yisaer, line 817: return false, fmt.Errorf("tikv don't have right ReadyToScale phase")

tests/e2e/tidbcluster/serial.go
```diff
@@ -825,6 +858,30 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
 
+			err = wait.Poll(10*time.Second, 5*time.Minute, func() (done bool, err error) {
+				tac, err = cli.PingcapV1alpha1().TidbClusterAutoScalers(ns).Get(tac.Name, metav1.GetOptions{})
+				if err != nil {
+					return false, nil
+				}
+				if tac.Annotations == nil || len(tac.Annotations) < 1 {
+					klog.Infof("tc haven't marked any annotation")
+					return false, nil
+				}
```
Yisaer, line 869: ditto.

tests/e2e/tidbcluster/serial.go
```diff
@@ -825,6 +858,30 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+				if !ok {
+					klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
```
Yisaer, line 872: ditto.

tests/e2e/tidbcluster/serial.go
```diff
@@ -825,6 +858,30 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, nil
+				}
+				readyToScaleTimestamp, err = strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					return true, err
+				}
+				if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleInAutoScalerPhase {
+					return true, fmt.Errorf("tikv don't have right ReadyToScale phase")
```
Yisaer, line 880: ditto.

tests/e2e/tidbcluster/serial.go
```diff
@@ -825,6 +858,30 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+					return false, nil
+				}
+				t, ok := tac.Annotations[label.AnnTiKVReadyToScaleTimestamp]
+				if !ok {
+					klog.Info("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+					return false, nil
+				}
+				readyToScaleTimestamp, err = strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					return true, err
```
Yisaer, line 877: ditto.

tests/e2e/tidbcluster/serial.go
```diff
@@ -747,42 +748,75 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				if !ok {
+					framework.Logf("tac has no tikv.tidb.pingcap.com/ready-to-scale-timestamp annotation")
+					return false, nil
+				}
+				readyToScaleTimestamp, err = strconv.ParseInt(t, 10, 64)
+				if err != nil {
+					return false, err
+				}
+				if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleOutAutoScalerPhase {
+					framework.Logf("tac dont' have the right ReadyToScale phase, expect: %s, got %s", v1alpha1.ReadyToScaleOutAutoScalerPhase, tac.Status.TiKV.Phase)
```
vincent178, line 771: use log instead of return error because we only reset phase but not timestamp, so in second scale, the timestamp is there, but the tac haven’t change phase.

pkg/autoscaler/autoscaler/autoscaler_manager.go
```diff
@@ -151,6 +151,9 @@ func (am *autoScalerManager) updateAutoScaling(oldTc *v1alpha1.TidbCluster,
 		return nil, nil
 	}
 
+	// record autoscaler sync timestamp
+	tac.Annotations[label.AnnStsLastSyncTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
```
DanielZhangQD, line 155: Suggest moving this code to the [Sync](https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L67) function, the sync may return before it reaches to this function.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -70,12 +70,24 @@ func (am *autoScalerManager) syncTiKV(tc *v1alpha1.TidbCluster, tac *v1alpha1.Ti
...
-	if recommendedReplicas > tc.Spec.TiKV.Replicas {
-		intervalSeconds = tac.Spec.TiKV.ScaleOutIntervalSeconds
+	if recommendedReplicas > currentReplicas {
+		if tac.Status.TiKV.Phase != v1alpha1.ReadyToScaleOutAutoScalerPhase {
+			tac.Status.TiKV.Phase = v1alpha1.ReadyToScaleOutAutoScalerPhase
+			// phase could change from Normal to ReadyToScaleOut, ReadyToScaleIn to ReadyToScaleOut,
+			// reset timestamp in both cases.
+			tac.Annotations[label.AnnTiKVReadyToScaleTimestamp] = fmt.Sprintf("%d", time.Now().Unix())
+		}
+	} else {
```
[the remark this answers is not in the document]
vincent178: good question, in L61, if will early exit if `recommendedReplicas == currentReplicas`.

pkg/autoscaler/autoscaler/autoscaler_manager.go
Yisaer: I think the current logic is ok. If the sync ended before it reaches this code, we shouldn't update the tac annotation[AnnStsLastSyncTimestamp]

vincent178: agree with @Yisaer and add one more comment that this code needs happen after caculate last sync time in auto scaler, if this code happens before, we can’t get a real last sync time (always close to when the auto scale happen).

DanielZhangQD: I do not understand here, what's this annotation exactly for? What's the standard to update this annotation?

pkg/autoscaler/autoscaler/tikv_autoscaler.go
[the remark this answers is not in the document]
DanielZhangQD: OK

pkg/autoscaler/autoscaler/autoscaler_manager.go
vincent178: This annotation is for tracking last sync/auto scale happen, so that if operator is down for sometime, we can reset phase after operator is up instead of right after operator is up, if we found out the ReadyToScale statisfy threshold and start auto scaling.

vincent178:
@Yisaer @DanielZhangQD PTAL

pkg/autoscaler/autoscaler/autoscaler_manager.go
Yisaer: @vincent178 I think `AnnStsLastSyncTimestamp` is not suitable in this case as this annotation is used to resync the statefulset in periodicity controller. What about using a new annotation named as `AnnLastTikvAutoScaleingSyncTimestamp`

Yisaer, line 155: @vincent178 I think AnnStsLastSyncTimestamp is not suitable in this case as this annotation is used to resync the statefulset in periodicity controller. What about using a new annotation named as AnnLastTikvAutoScaleingSyncTimestamp

pkg/autoscaler/autoscaler/util.go
```diff
@@ -57,11 +59,81 @@ func checkStsAutoScalingPrerequisites(set *appsv1.StatefulSet) bool {
...
+	return true, nil
+}
+
+// checkStsLastAutoScalingTimestamp reset TiKV phase if last auto scaling timestamp is longer than thresholdSec
+func checkStsLastAutoScalingTimestamp(tac *v1alpha1.TidbClusterAutoScaler, thresholdSec time.Duration, clock clockwork.Clock) (bool, error) {
 	if tac.Annotations == nil {
 		tac.Annotations = map[string]string{}
 	}
+
+	lastAutoScalingTimestamp, existed := tac.Annotations[label.AnnTiKVLastAutoScalingTimestamp]
```
Yisaer, line 98: `AnnTiKVLastAutoScalingTimestamp` is used to check the interval duration between each 2 auto-scaling, not syncing. so I tend to use annotation like `last-syncing` to record the timestamp for each syncing, and check the interval duration between each 2 syncing whether is meet the requirement.

pkg/autoscaler/autoscaler/autoscaler_manager.go
DanielZhangQD: Oh, it's useless to update the annotation in Sync function if it cannot reach this function, actually, even if you set the annotation here it may not be updated to the tac because the logic may return before the tac really updated.

DanielZhangQD:
@vincent178 Please check the two rewording comments and close or address them with your preference.

DanielZhangQD:
Sorry, two more concerns here (scale in has the same logic):
1. if at one sync, the phase is updated to readyToScaleout, then after that, during each sync, the result is scale-out, however, it always fails [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L121), after more than  `3*controller.ResyncDuration`, the phase is set to Normal in `checkStsLastAutoScalingTimestamp` and scale-out rejected even if it succeeds [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L121).
2. if at one sync, the tikv is scaled out, but the function [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L160) always fails, in this case, for the following syncs, the scale-out will always be accepted and the `ScaleOutIntervalSeconds` and `ReadyToScaleThresholdSeconds` cannot affect the result.

@Yisaer @vincent178 Please evaluate if the above concerns are valid.

vincent178:
Hi @DanielZhangQD,
> if at one sync, the phase is updated to readyToScaleout, then after that, during each sync, the result is scale-out, however, it always fails here, after more than 3*controller.ResyncDuration, the phase is set to Normal in checkStsLastAutoScalingTimestamp and scale-out rejected even if it succeeds here.

I think this is a valid concern, since we check last sync threshold seconds in method `pkg/autoscaler/autoscaler/autoscaler_manager.go:L89` but record last sync in method `pkg/autoscaler/autoscaler/autoscaler_manager.go:L95`, anything error in the middle might break the recording which leads to next time the ReadyToScale phase might reset to normal. To address this issue, I would suggest to record last sync timestmap in `pkg/autoscaler/autoscaler/autoscaler_manager.go:L89`.

> if at one sync, the tikv is scaled out, but the function here always fails, in this case, for the following syncs, the scale-out will always be accepted and the ScaleOutIntervalSeconds and ReadyToScaleThresholdSeconds cannot affect the result.

Try to understand your concern here, you’re saying that the AutoScaler do pass all the check and record a target replica number, but when do the real update in `pkg/autoscaler/autoscaler/autoscaler_manager.go:L95 `, it failed, so next time, the checks will pass as always. If this is the concern, my understanding is ok to have this behavior, becuase the first time all checks passed, and the update logic failure should not affect the check. For example, a web app has validation for user input, and it passed, but somehow it failed to save to database (due to db issue or network issue), next time, user has the same input, the validation should still pass. Please correct me if I’m wrong.

DanielZhangQD:
> I think this is a valid concern, since we check last sync threshold seconds in method `pkg/autoscaler/autoscaler/autoscaler_manager.go:L89` but record last sync in method `pkg/autoscaler/autoscaler/autoscaler_manager.go:L95`, anything error in the middle might break the recording which leads to next time the ReadyToScale phase might reset to normal. To address this issue, I would suggest to record last sync timestmap in `pkg/autoscaler/autoscaler/autoscaler_manager.go:L89`.
> 
Actually, recording the last sync timestmap in `pkg/autoscaler/autoscaler/autoscaler_manager.go:L89` is not gonna help, because the timestamp will not be updated to the tac until this [function](https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L185) succeeds.

> Try to understand your concern here, you’re saying that the AutoScaler do pass all the check and record a target replica number, but when do the real update in `pkg/autoscaler/autoscaler/autoscaler_manager.go:L95 `, it failed, so next time, the checks will pass as always. If this is the concern, my understanding is ok to have this behavior, becuase the first time all checks passed, and the update logic failure should not affect the check. For example, a web app has validation for user input, and it passed, but somehow it failed to save to database (due to db issue or network issue), next time, user has the same input, the validation should still pass. Please correct me if I’m wrong.

Your understanding is correct, my point is that tc is updated in `pkg/autoscaler/autoscaler/autoscaler_manager.go#L92`, but if tac update failed in `pkg/autoscaler/autoscaler/autoscaler_manager.go#L95`, then the current noise reduction config `ReadyToScaleThresholdSeconds` and `ScaleOutIntervalSeconds` would not take effect any more, right? for example, at one sync, scale to 3, then the next sync, scale to 4....

vincent178:
> Your understanding is correct, my point is that tc is updated in pkg/autoscaler/autoscaler/autoscaler_manager.go#L92, but if tac update failed in pkg/autoscaler/autoscaler/autoscaler_manager.go#L95, then the current noise reduction config ReadyToScaleThresholdSeconds and ScaleOutIntervalSeconds would not take effect any more, right? for example, at one sync, scale to 3, then the next sync, scale to 4....

My understanding is because we choose to update tc controller first and then update autoscaler after, there's a chance that the cluster update successfully but autoscaler doesn't. To address this issue, we can record the annotation on tc controller instead of tac controller, or we can move update tac controller before update tc controller. What do you think?

Yisaer:
As we can't provide the transaction ability for both update tidbcluster and tidbclusteAutoscaler. I think is ok for me that the current logic might update tidbcluster spec successfully but the tidbclusterautoscaler's status updated failed.
For @DanielZhangQD 's concern, I think currently it is not much important but still have a way to solve it. What about reversing the ordinal to update both API Object.

We can update tidbclusterAutoscaler's status first, if success, then we will try to update the tidbcluster spec. if both updating operations is success, it behaves like previous. If updating autoscaler's status failed, it just leaves to the next syncing round. If updating tidbcluster'spec failed, as we only persist the autoscaler's status, the auto-scaling of tidbcluster would end and leave to the next round. WDYT @DanielZhangQD  @vincent178

DanielZhangQD:
It's correct that we do not have transaction ability to update multiple objects.
Just some thoughts, is it good enough that we maintain the timestamps in the memory and update them in the memory when the TidbCluster syncs succeed.
If the operator is down and then up again, just start all over again, for the current solution, if the operator is down more than `3*controller.ResyncDuration`, it will start all over again too.
WDYT @vincent178 @Yisaer

Yisaer:
@DanielZhangQD  I think your concern is existed but out of this PR's range. As you have mentioned that the `ScaleOutIntervalDuration` also can't meet your requirement as following:

> if at one sync, the tikv is scaled out, but the function here always fails, in this case, for the following syncs, the scale-out will always be accepted and the ScaleOutIntervalSeconds and ReadyToScaleThresholdSeconds cannot affect the result.

I don't think this pr should solve both noise reduction and your concern.  In your latest comment, I don't ensure whether record in memory is a good choice. But it you think that recording in memory could solve your concert, it's ok for me to create a new issue to track this problem. You can finish this issue before the auto-scaling feature GA.

For this PR, it only should do the noise reduction and leave this problem be as this problem not belongs here.

DanielZhangQD:
> If updating tidbcluster'spec failed, as we only persist the autoscaler's status, the auto-scaling of tidbcluster would end and leave to the next round. WDYT @DanielZhangQD @vincent178

In this case, the timestamps are already updated, so during the next sync, the scale cannot be done due to the check of `ReadyToScaleThresholdSeconds` and `ScaleOutIntervalSeconds`, we have to wait for another round of duration check.

Let's involve other maintainers to review this PR. @cofyc @weekface Could you please help review this PR? Thanks!

vincent178:
@cofyc @Yisaer @DanielZhangQD PTAL

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -76,6 +84,12 @@ type TidbClusterAutoScalerSpec struct {
 // TikvAutoScalerSpec describes the spec for tikv auto-scaling
 type TikvAutoScalerSpec struct {
 	BasicAutoScalerSpec `json:",inline"`
+
+	// ReadyToScaleThresholdSeconds represents duration that the ReadyToScale phase
+	// should last for before auto scaling.
+	// If not set, the default ReadyToScaleThresholdSeconds will be set to 300.
+	// +optional
+	ReadyToScaleThresholdSeconds *int32 `json:"readyToScaleThresholdSeconds,omitempty"`
```
Yisaer, line 92: 300 is about 10 times re-sync duration, I think it might be too long for a default value. What about 60 as it only require 2 re-sync duration.

pkg/autoscaler/autoscaler/util.go
Yisaer, line 224: 300 is too large. ditto.

pkg/label/label.go
```diff
@@ -109,6 +109,12 @@ const (
 	// AnnTiKVLastAutoScalingTimestamp is annotation key of tidbclusterto which ordinal is created by tikv auto-scaling
 	AnnTiKVLastAutoScalingTimestamp = "tikv.tidb.pingcap.com/last-autoscaling-timestamp"
 
+	// AnnTiKVReadyToScaleTimestamp records timestamp when tikv ready to scale
+	AnnTiKVReadyToScaleTimestamp = "tikv.tidb.pingcap.com/ready-to-scale-timestamp"
+
+	// AnnLastSyncingTimestamp records last sync timestamp
+	AnnLastSyncingTimestamp = "auto-scaling.tidb.pingcap.com/last-syncing-timestamp"
```
Yisaer, line 116: I think `auto-scaling` here is unnecessary, this annotation could be common used for each API Object.

PR #2312, opened 2020-04-27
Add Ingress spec for Dashboard

Yisaer:
Add Ingress Spec For Dashboard and add manager to sync dashboard ingress.  

Note that the the dashboard ingress and TlsCluster couldn't be both enabled currently as the Ingress assumes the TLS termination. FYI: https://kubernetes.io/docs/concepts/services-networking/ingress/#tls 

### Check List 

 - E2E test

Related changes

```release-note
Support Ingress for Dashboard in TidbCluster Spec
```


pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1161,3 +1166,29 @@ type RestoreStatus struct {
 	TimeCompleted metav1.Time        `json:"timeCompleted"`
 	Conditions    []RestoreCondition `json:"conditions"`
 }
+
+// +k8s:openapi-gen=true
+// DashBoard describe the TiDB Dashboard desired state
+type DashBoardSpec struct {
```
cofyc, line 1172: `DashBoardSpec` -> `DashboardSpec`

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1161,3 +1166,29 @@ type RestoreStatus struct {
...
+	// PDDashboardIngress describe the ingress of the TiDB Dashboard
+	// +optional
+	Ingress *IngressSpec `json:"ingress,omitempty"`
```
cofyc, line 1175: it is possible to use Kubernetes builtin ingress types directly?

Yisaer: I suggest not as we should control the path that the ingress would only expose the dashboard api excluding pdapi.

pkg/apis/pingcap/v1alpha1/types.go
Yisaer: updated.

PR #2328, opened 2020-04-29
do not enable placement rules automatically

### LINKED ISSUE #2219: PD placement rules should be enabled if using TiFlash

DanielZhangQD:
Fix #2219 

If users enable TiFlash for the existing TiDB Cluster, enable placement rules automatically with pdapi.
### Check List 

Tests 

   - Enable TiFlash for the existing TiDB Cluster, enable-placement-rules is set automatically and PD cluster will not be rolling updated.
   - Create cluster with enable-placement-rules set, tiflash will start successfully and the operator will not update the enable-placement-rules config.

Code changes

 - Has Go code change

Related changes

```release-note
Set `enable-placement-rules` to `true` for PD if TiFlash is enabled in the cluster
```

### DIFF
--- a/pkg/manager/member/pd_member_manager.go
+++ b/pkg/manager/member/pd_member_manager.go
@@ -741,17 +741,7 @@ func getPDConfigMap(tc *v1alpha1.TidbCluster) (*corev1.ConfigMap, error) {
 		config.Dashboard.TiDBCertPath = path.Join(tidbClientCertPath, corev1.TLSCertKey)
 		config.Dashboard.TiDBKeyPath = path.Join(tidbClientCertPath, corev1.TLSPrivateKeyKey)
 	}
-	// TiFlash requires PD to enable the `replication.enable-placement-rules`
-	// Check detail in https://pingcap.com/docs/stable/reference/tiflash/deploy/
-	if tc.Spec.TiFlash != nil {
-		if config.Replication == nil {
-			config.Replication = &v1alpha1.PDReplicationConfig{}
-		}
-		if config.Replication.EnablePlacementRules == nil {
-			enable := true
-			config.Replication.EnablePlacementRules = &enable
-		}
-	}
+
 	confText, err := MarshalTOML(config)
 	if err != nil {
 		return nil, err
--- a/pkg/pdapi/pdapi.go
+++ b/pkg/pdapi/pdapi.go
@@ -154,8 +154,8 @@ type PDClient interface {
 	// storeLabelsEqualNodeLabels compares store labels with node labels
 	// for historic reasons, PD stores TiKV labels as []*StoreLabel which is a key-value pair slice
 	SetStoreLabels(storeID uint64, labels map[string]string) (bool, error)
-	// UpdateEnablePlacementRules updates the enable-placement-rules config
-	UpdateEnablePlacementRules(rule PDReplicationConfig) error
+	// UpdateReplicationConfig updates the replication config
+	UpdateReplicationConfig(config PDReplicationConfig) error
 	// DeleteStore deletes a TiKV store from cluster
 	DeleteStore(storeID uint64) error
 	// SetStoreState sets store to specified state.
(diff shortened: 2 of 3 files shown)


yeya24:
So we still have to enable this on our own? Instead of managing by operator?

DanielZhangQD:
> So we still have to enable this on our own? Instead of be managed by the operator?

For an existing cluster, users have to enable it manually via pd-ctl, update the configuration file does not take effect.

pkg/manager/member/pd_member_manager.go
```diff
@@ -741,17 +741,7 @@ func getPDConfigMap(tc *v1alpha1.TidbCluster) (*corev1.ConfigMap, error) {
 		config.Dashboard.TiDBCertPath = path.Join(tidbClientCertPath, corev1.TLSCertKey)
 		config.Dashboard.TiDBKeyPath = path.Join(tidbClientCertPath, corev1.TLSPrivateKeyKey)
 	}
-	// TiFlash requires PD to enable the `replication.enable-placement-rules`
-	// Check detail in https://pingcap.com/docs/stable/reference/tiflash/deploy/
-	if tc.Spec.TiFlash != nil {
```
weekface, line 746: Enable placement rules if we create the cluster?

DanielZhangQD: When creating the configmap, it does not know whether the cluster is new or not, and it should not care about it.
I think we can make it clear in the doc, after all, users also have to add the TiFlash configurations, the TiFlash docs also request users to enable it manually.

Yisaer: Currently there are 2 steps for user to enable Tiflash. First, enable rule-placement by pd-ctl. Second configure Tiflash Table Replicas by sql. I think only SQL configuration should be exposed to the users excluding pdapi. We have some seniors that users couldn't get in touch with pdapi. 

As TiFlash is [supprted to be enabled in an existed cluster](https://pingcap.com/docs-cn/stable/reference/tiflash/deploy/#%E5%9C%A8%E5%8E%9F%E6%9C%89-tidb-%E9%9B%86%E7%BE%A4%E4%B8%8A%E6%96%B0%E5%A2%9E-tiflash-%E7%BB%84%E4%BB%B6), I think we could enabled it `placement rule` by pdapi before syncing tiflash.

DanielZhangQD: updated

DanielZhangQD:
@Yisaer @weekface comments addressed, PTAL.

pkg/pdapi/pdapi.go
```diff
@@ -512,6 +515,24 @@ func (pc *pdClient) SetStoreLabels(storeID uint64, labels map[string]string) (bo
...
+	}
+	res, err := pc.httpClient.Post(apiURL, "application/json", bytes.NewBuffer(data))
+	if err != nil {
+		return err
+	}
+	defer httputil.DeferClose(res.Body)
+	if res.StatusCode == http.StatusOK {
+		return nil
+	}
+	err2 := httputil.ReadErrorBody(res.Body)
```
Yisaer, line 532: why not directly use err here

pkg/pdapi/pdapi.go
```diff
@@ -697,6 +718,7 @@ const (
 	DeleteMemberByIDActionType         ActionType = "DeleteMemberByID"
 	DeleteMemberActionType             ActionType = "DeleteMember "
 	SetStoreLabelsActionType           ActionType = "SetStoreLabels"
+	UpdatePlacementRulesActionType     ActionType = "UpdateEnablePlacementRules"
```
Yisaer, line 721: I think PlacementRules is part of the `ReplicationConfig`, We should name it as `SetReplicationConfig`

DanielZhangQD: There are other fields in this struct, here I only update this one field, if need to update other fields, we should define another function, they may be used in different scenarios.

DanielZhangQD: updated.

pkg/pdapi/pdapi.go
DanielZhangQD: updated

PR #2332, opened 2020-04-29
Use patch for sts in periodicity controller 

Yisaer:
periodicity controller use update to inject timestamp annotation for the tidbcluster sts. However, the data in the Lister might be out-of-date which causing the tikv/tidb sts scale-in again after they get scale-out by auto-scaler.
In this request, we use patch to merge the target sts annotation instead of updating.

Related changes

```release-note
Using `Patch` in `Periodicity` to avoid `StatefulSet` be updated to wrong state.
```


pkg/controller/periodicity/periodicity_controller.go
```diff
@@ -94,29 +86,35 @@ func (c *Controller) syncStatefulSetTimeStamp() error {
...
-		ok, tcRef := util.IsOwnedByTidbCluster(sts)
+		ok, _ := util.IsOwnedByTidbCluster(sts)
 		if !ok {
 			continue
 		}
-		tc, err := c.tcLister.TidbClusters(sts.Namespace).Get(tcRef.Name)
-		if err != nil {
-			errs = append(errs, err)
-			continue
-		}
```
cofyc, line 109: why remove this? there is no need to update statefulsets not owned by tidb clusters.

pkg/controller/periodicity/periodicity_controller.go
```diff
@@ -94,29 +86,35 @@ func (c *Controller) syncStatefulSetTimeStamp() error {
...
 			sts.Annotations = map[string]string{}
 		}
 		sts.Annotations[label.AnnStsLastSyncTimestamp] = time.Now().Format(time.RFC3339)
-		newSts, err := c.statefulSetControl.UpdateStatefulSet(tc, sts)
+		var mergePatch []byte
+		mergePatch, err = json.Marshal(map[string]interface{}{
+			"metadata": map[string]interface{}{
+				"annotations": sts.Annotations,
+			},
+		})
```
cofyc, line 107: I think using the Patch method is a good idea. It can avoid API conflicts in kube-apiserver.

But the root cause is we should copy before modifying the statefulset.

pkg/controller/periodicity/periodicity_controller.go
```diff
@@ -94,29 +86,35 @@ func (c *Controller) syncStatefulSetTimeStamp() error {
...
 		}
-		tc, err := c.tcLister.TidbClusters(sts.Namespace).Get(tcRef.Name)
-		if err != nil {
-			errs = append(errs, err)
-			continue
-		}
 		if sts.Annotations == nil {
 			sts.Annotations = map[string]string{}
 		}
 		sts.Annotations[label.AnnStsLastSyncTimestamp] = time.Now().Format(time.RFC3339)
```
cofyc, line 101: create a new annotations map, otherwise, cache in informer will be modified.

Yisaer: updated.

pkg/controller/periodicity/periodicity_controller.go
Yisaer: updated.

pkg/controller/periodicity/periodicity_controller.go
```diff
@@ -102,21 +95,40 @@ func (c *Controller) syncStatefulSetTimeStamp() error {
 		if !ok {
 			continue
 		}
-		tc, err := c.tcLister.TidbClusters(sts.Namespace).Get(tcRef.Name)
+		if sts.Annotations == nil {
+			sts.Annotations = map[string]string{}
+		}
```
cofyc, line 100:
```suggestion
```

Yisaer: updated.

PR #2334, opened 2020-04-29
KMS: configuration is change via tikv

shuijing198799:
 configuration is change via tikv

### Check List 

Tests 

 - Manual test

Code changes

 - Has Go code change

```release-note
Modify the json tag and toml tag of the field `spec.tikv.config.encrpytion.method` , from `method` to `data-encryption-method`
```


DanielZhangQD:
@shuijing198799 Please update `release notes`.

Issue #2329, opened 2020-04-29
monitor PVC will be deleted if TidbMonitor CR is deleted


DanielZhangQD:
## Bug Report

**What version of Kubernetes are you using?**

1.12.8
**What version of TiDB Operator are you using?**

v1.1.0-rc.2
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

local-storage
**What's the status of the TiDB cluster pods?**

Running
**What did you do?**

Delete TidbMonitor CR
**What did you expect to see?**
monitor PVC remained
**What did you see instead?**
monitor PVC deleted

```
# k get pvc -n demo
NAME               STATUS   VOLUME              CAPACITY   ACCESS MODES   STORAGECLASS    AGE
demo-monitor       Bound    local-pv-45d26c43   440Gi      RWO            local-storage   37s
pd-demo-pd-0       Bound    local-pv-d097f77b   440Gi      RWO            local-storage   32s
tikv-demo-tikv-0   Bound    local-pv-3ac89acb   440Gi      RWO            local-storage   7s

# k delete tidbmonitor -n demo demo
tidbmonitor.pingcap.com "demo" deleted
# k get pvc -n demo           
NAME                   STATUS        VOLUME              CAPACITY   ACCESS MODES   STORAGECLASS    AGE
data0-demo-tiflash-0   Pending                                                     local-storage   1s
demo-monitor           Terminating   local-pv-45d26c43   440Gi      RWO            local-storage   56s
pd-demo-pd-0           Bound         local-pv-d097f77b   440Gi      RWO            local-storage   51s
tikv-demo-tikv-0       Bound         local-pv-3ac89acb   440Gi      RWO            local-storage   26s

# k get pvc -n demo 
NAME                   STATUS   VOLUME              CAPACITY   ACCESS MODES   STORAGECLASS    AGE
data0-demo-tiflash-0   Bound    local-pv-7855f795   440Gi      RWO            local-storage   78s
pd-demo-pd-0           Bound    local-pv-d097f77b   440Gi      RWO            local-storage   2m8s
tikv-demo-tikv-0       Bound    local-pv-3ac89acb   440Gi      RWO            local-storage   103s
```

DanielZhangQD:
`ownerReferences` is set for the PVC:
```
  ownerReferences:
  - apiVersion: pingcap.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: TidbMonitor
    name: demo
[...]

PR #2374, opened 2020-05-06
Remain PVC when TidbMonitor get deleted.

Yisaer:
close https://github.com/pingcap/tidb-operator/issues/2329

Currently we would delete pvc when tidbmonitor is deleted. In this request, we would remain PVC if TidbMonitor is deleted.

### Check List 

Tests 

Related changes

```release-note
Don't set owner references to keep `PVC` for `TidbMonitor` when it gets deleted
```


cofyc:
why not set `reclaimPolicy` to `Retain`?

Yisaer:
> why not set `reclaimPolicy` to `Retain`?

We would keep the same strategy in PV/PVC between TidbMonitor and TidbCluster:
1. Remain PVC when the object get deleted.
2. expose ReclaimPlicy for user.

cofyc:
`Support to remain `PVC` for `TidbMonitor` when it get deleted.`
->
`Don't set owner references to keep `PVC` for `TidbMonitor` when it gets deleted`

describe what's changed.

PR #2400, opened 2020-05-08
backup: fix gcs get commit without prefix error

shuijing198799:
fix gcs get commit without prefix error
return a bucket with prefix when get new storage
### Check List 

Tests 

 - E2E test
run backup in dbaas

Code changes

 - Has Go code change


DanielZhangQD:
@shuijing198799 Did you test both cases with and without Prefix set?

PR #2417, opened 2020-05-10
add storageClassName field from documentation

ica10888:
# What is changed and how does it work?

From this documentation  [deploy-on-general-kubernetes](https://pingcap.com/docs/tidb-in-kubernetes/stable/deploy-on-general-kubernetes/#storage-class)

> You can set the storage class by modifying `storageClassName` of each component in 
 `${cluster_name}/tidb-cluster.yaml` and `${cluster_name}/tidb-monitor.yaml` .

if not seting this field , the storageClassName will use defult sc what k8s cluster support . obviously, this setting is better for showing in TidbCluster example.

Tests 
 - No code


Yisaer:
Thanks for your contribution. It seems the `storageClassName` is placed into wrong location. The `storageClassName` should under `PdSpec` or `TiKVSpec`. You can find more detail in our api references. https://pingcap.com/docs-cn/tidb-in-kubernetes/stable/api-references/#pdspec

examples/basic/tidb-cluster.yaml
```diff
@@ -12,12 +12,14 @@ spec:
     requests:
       storage: "1Gi"
     config: {}
+    storageClassName: local-storage
```
Yisaer, line 15: Please note this attribute as we would use default storageClass in basic example.

cofyc:
it's intended not to specify `storageClassName` and use the default storage class of Kubernetes in our examples. this is because different clusters may have different storage classes configured and our example should work in any cluster with a builtin default storage class (most managed clusters in cloud and minikube/kind).

if you refer to [this doc](https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/) to set the default storage class in your Kubernetes cluster.

examples/basic/tidb-cluster.yaml
ica10888: ok , I have wrote annotation about this

examples/basic/tidb-cluster.yaml
```diff
@@ -9,12 +9,16 @@ spec:
   pd:
     baseImage: pingcap/pd
     replicas: 3
+    # if storageClassName does not set, storageClassName will be chosen what kubernetes cluster support automatically
```
cofyc, line 12:
```suggestion
    # if storageClassName is not set, the default Storage Class of the Kubernetes cluster will be used
```

examples/basic/tidb-cluster.yaml
```diff
@@ -9,12 +9,16 @@ spec:
...
+    # storageClassName: local-storage
     requests:
       storage: "1Gi"
     config: {}
   tikv:
     baseImage: pingcap/tikv
     replicas: 3
+    # if storageClassName does not set, storageClassName will be chosen what kubernetes cluster support automatically
```
cofyc, line 20:
```suggestion
    # if storageClassName is not set, the default Storage Class of the Kubernetes cluster will be used
```

cofyc:
Thanks! It's good to have a comment here.

ica10888:
Thanks for guiding to commit this PR . have a good day.

Issue #2331, opened 2020-04-29
Auto configure pd-server metric-storage


tennix:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currenlty, TiDB dashboard will report the following errors if users don't configure `metric-store` for pd-server.
```
Error 1105: query metric error: [domain:9009]Prometheus address is not set in PD and etcd, Currently, the PD config `pd-server.metric-storage` value should be prometheus address, please check whether the config value is correct, you can use below sql check the value: select * from information_schema.cluster_config where type='pd' and `key` ='pd-server.metric-storage'; , you can use this shell command to set the config: `curl -X POST -d '{"metric-storage":"http://{PROMETHEUS_ADDRESS}"}' http://{PD_ADDRESS}/pd/api/v1/config`, PROMETHEUS_ADDRESS is the prometheus address, It's used for query metric data; PD_ADDRESS is the HTTP API address of PD server, all PD servers need to set this config. Here is an example: `curl -X POST -d '{"metric-storage":"http://127.0.0.1:9090"}' http://127.0.0.1:2379/pd/api/v1/config
```

**Describe the feature you'd like:**

Now tidb-dashbord is included in pd, it requires following [configurations](https://github.com/pingcap/pd/blob/master/conf/config.toml#L57-L61):

```toml
[pd-server]
## the metric storage is the cluster metric storage. This is use for query metric data.
## Currently we use prometheus as metric storage, we may use PD/TiKV as metric storage later.
## For usability, recommended to temporarily set it to the prometheus address, eg: http://127.0.0.1:9090
metric-storage = ""
```

Most of the users will use tidb-operator managed monitor, so the prometheus address is known in advance. It would be better if tidb-operator can automatically configure pd if a user does not provide an explicit `metric-storage` configuration.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

breeswish:
Hint: in order to be compatible with Dashboard, an etcd key `/topology/prometheus` must exist (and no need to set `metric-storage`) containing Prometheus host and port. For the value of the key, see https://github.com/pingcap-incubator/tiup-cluster/pull/295/files

PR #2424, opened 2020-05-11
Add MonitorRef into tidbcluster status

Yisaer:

To support Dashboard query metric ability, we should deliver the monitor information (host/port) to PD.
In this request, if `TidbMonitor` is deployed to monitor the `TidbCluster`, the tidbcluster status would show the monitorRef.

1. The `MonitorRef` would be added only if the target tidbcluster status monitorRef is nil.
2. When the `TidbMonitor` is deleted, the MonitorRef in tidbcluster status would be cleaned.

Related changes

```release-note
Support `TidbMonitorRef` in `TidbCluster` Status
```


pkg/monitor/monitor/monitor_manager.go
```diff
@@ -82,7 +86,26 @@ func NewMonitorManager(
 func (mm *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
 
 	if monitor.DeletionTimestamp != nil {
-		return nil
+		return mm.patchTidbClusterStatus(nil, monitor)
+	}
+	if monitor.Spec.Clusters == nil || len(monitor.Spec.Clusters) < 1 {
+		err := fmt.Errorf("tm[%s/%s] haven't point the target tidbcluster", monitor.Namespace, monitor.Name)
```
DanielZhangQD, line 92:
```suggestion
		err := fmt.Errorf("tm[%s/%s] does not configure the target tidbcluster", monitor.Namespace, monitor.Name)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -134,6 +157,12 @@ func (mm *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
 	}
 	klog.V(4).Infof("tm[%s/%s]'s ingress synced", monitor.Namespace, monitor.Name)
 
+	if err := mm.patchTidbClusterStatus(&tcRef, monitor); err != nil {
+		message := fmt.Sprintf("Sync TidbMonitorRef into targetCluster status failed, err:%v", err)
```
DanielZhangQD, line 161: Suggest adding the tc info.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -338,3 +353,32 @@ func (mm *MonitorManager) removeIngressIfExist(monitor *v1alpha1.TidbMonitor, na
...
+		mergePatch, err = json.Marshal(map[string]interface{}{
+			"status": map[string]interface{}{
+				"monitor": nil,
+			},
+		})
+	}
+	if err != nil {
+		return err
+	}
+	_, err = mm.cli.PingcapV1alpha1().TidbClusters(tc.Namespace).Patch(tc.Name, types.MergePatchType, mergePatch)
```
DanielZhangQD, line 382: Now the tc is synced frequently, suggest adding some retry here.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -82,7 +86,26 @@ func NewMonitorManager(
...
+		err := fmt.Errorf("tm[%s/%s] haven't point the target tidbcluster", monitor.Namespace, monitor.Name)
+		return err
+	}
+	tcRef := monitor.Spec.Clusters[0]
+	if len(tcRef.Namespace) < 1 {
+		tcRef.Namespace = monitor.Namespace
+	}
+	tc, err := mm.cli.PingcapV1alpha1().TidbClusters(tcRef.Namespace).Get(tcRef.Name, metav1.GetOptions{})
+	if err != nil {
+		rerr := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] checked failed, err: %v", monitor.Namespace, monitor.Name, tcRef.Namespace, tcRef.Name, err)
```
DanielZhangQD, line 101:
```suggestion
		rerr := fmt.Errorf("get tm[%s/%s]'s target tc[%s/%s] failed, err: %v", monitor.Namespace, monitor.Name, tcRef.Namespace, tcRef.Name, err)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -82,7 +86,26 @@ func NewMonitorManager(
...
+		return rerr
+	}
+	if tc.Status.Monitor != nil {
+		if tc.Status.Monitor.Name != monitor.Name || tc.Status.Monitor.Namespace != monitor.Namespace {
+			err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor yet", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name)
```
DanielZhangQD, line 106:
```suggestion
			err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor [%s/%s]", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name, tc.Status.Monitor.Namespace, tc.Status.Monitor.Name)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -82,7 +86,26 @@ func NewMonitorManager(
 func (mm *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
 
 	if monitor.DeletionTimestamp != nil {
-		return nil
+		return mm.patchTidbClusterStatus(nil, monitor)
```
DanielZhangQD, line 89: Is this necessary since you have added the status sync in tidblcuster?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -82,7 +86,26 @@ func NewMonitorManager(
...
+	}
+	if monitor.Spec.Clusters == nil || len(monitor.Spec.Clusters) < 1 {
+		err := fmt.Errorf("tm[%s/%s] haven't point the target tidbcluster", monitor.Namespace, monitor.Name)
+		return err
+	}
+	tcRef := monitor.Spec.Clusters[0]
+	if len(tcRef.Namespace) < 1 {
+		tcRef.Namespace = monitor.Namespace
+	}
+	tc, err := mm.cli.PingcapV1alpha1().TidbClusters(tcRef.Namespace).Get(tcRef.Name, metav1.GetOptions{})
```
DanielZhangQD, line 99: Is it necessary to call kube-apiserver directly here?
What's the drawback of the lister?
We should avoid call kube-apiserver directly especially when too many clusters managed by tidb-operator.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -206,12 +206,13 @@ type TidbClusterSpec struct {
...
-	TiDB      TiDBStatus    `json:"tidb,omitempty"`
-	Pump      PumpStatus    `josn:"pump,omitempty"`
-	TiFlash   TiFlashStatus `json:"tiflash,omitempty"`
+	ClusterID string          `json:"clusterID,omitempty"`
+	PD        PDStatus        `json:"pd,omitempty"`
+	TiKV      TiKVStatus      `json:"tikv,omitempty"`
+	TiDB      TiDBStatus      `json:"tidb,omitempty"`
+	Pump      PumpStatus      `josn:"pump,omitempty"`
+	TiFlash   TiFlashStatus   `json:"tiflash,omitempty"`
+	Monitor   *TidbMonitorRef `json:"monitor,omitempty"`
```
DanielZhangQD, line 215: Where is this reference used?
How does this PR fix #2331?

Yisaer: Here is the first step to realize #2331

pkg/monitor/monitor/monitor_manager.go
Yisaer: I guess that's ok as we only sync tidbmonitor 30s once in default.

pkg/monitor/monitor/monitor_manager.go
Yisaer: removed

pkg/monitor/monitor/monitor_manager.go
Yisaer: Retrying is not necessary here. If error happens, I tend to end this reconciliation and wait for the next round.

pkg/monitor/monitor/monitor_manager.go
DanielZhangQD: What if we have 30 tidbmonitors in the cluster, btw, we also have other codes call kube-apiserver directly, which will cause pressure to the kube-apiserver, so I think if not required, we can use the lister, and for the two tidbmonitor to monitor the same tidbcluster case, one of the request to update will fail anyway, right?

pkg/monitor/monitor/monitor_manager.go
Yisaer: updated.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -84,6 +88,34 @@ func (mm *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+			return err
+		}
+	}
+
+	// TODO: support Validating hook for updating tidbcluster that forbid update monitorRef with a different reference
+	// Patch tidbcluster status first in order to avoiding let multi tidbmonitor monitor the same tidbcluster
+	if err := mm.patchTidbClusterStatus(&tcRef, monitor); err != nil {
+		message := fmt.Sprintf("Sync TidbMonitorRef into targetCluster[%s/%s] status failed, err:%v", tc.Namespace, tc.Name, err)
+		mm.recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, message)
+		return err
```
DanielZhangQD, line 117: Is it OK that we just emit event and print error log here and continue to sync the resources without returning error?

DanielZhangQD: I think we should move the status update code block to the end of the sync function, we should update the resource first then update its reference.

Yisaer: I think the monitorRef in Tidbcluster Ref could also seem as a mutex contention that tidbmonitor only success syncing its own information to the target tidbcluster status, then the tidbmonitor would start to sync the resources. That is to say, tidbmonitor would try to patch the tidbcluster first.

Yisaer: > Is it OK that we just emit event and print error log here and continue to sync the resources without returning error?

updated, using log instead of event.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -84,6 +88,34 @@ func (mm *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+	if tc.Status.Monitor != nil {
+		if tc.Status.Monitor.Name != monitor.Name || tc.Status.Monitor.Namespace != monitor.Namespace {
+			err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor [%s/%s]", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name, tc.Status.Monitor.Namespace, tc.Status.Monitor.Name)
+			mm.recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, err.Error())
+			return err
+		}
+	}
+
+	// TODO: support Validating hook for updating tidbcluster that forbid update monitorRef with a different reference
+	// Patch tidbcluster status first in order to avoiding let multi tidbmonitor monitor the same tidbcluster
```
DanielZhangQD, line 113:
```suggestion
	// Patch tidbcluster status first to avoid multi tidbmonitor monitoring the same tidbcluster
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -84,6 +88,36 @@ func (mm *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+	}
+	if tc.Status.Monitor != nil {
+		if tc.Status.Monitor.Name != monitor.Name || tc.Status.Monitor.Namespace != monitor.Namespace {
+			err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor [%s/%s]", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name, tc.Status.Monitor.Namespace, tc.Status.Monitor.Name)
+			mm.recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, err.Error())
+			return err
+		}
+	}
+
+	// TODO: support Validating hook for updating tidbmonitor that forbidding update clusterRef whose monitorRef has already
```
DanielZhangQD, line 112:
```suggestion
	// TODO: Support validating webhook that forbids the tidbmonitor to update the monitorRef for the tidbcluster whose monitorRef has already
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -84,6 +88,36 @@ func (mm *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+	// been occupied by other TidbMonitor.
```
DanielZhangQD, line 113:
```suggestion
	// been set by another TidbMonitor.
```

PR #2442, opened 2020-05-12
Support etcd-client in pdapi

Yisaer:

To support Dashboard query metric ability, we should deliver the monitor information (host/port) to PD etcd.

This request add etcdClient in PDClient in order to Put/Delete Key/Value in PD Etcd.

```release-note
NONE
```


cofyc:
I think we can have a separate `PDEtcdClient` implementation.

Yisaer:
> PDEtcdClient

That's a good idea.

PR #2443, opened 2020-05-12
Support PD Etcd Client

Yisaer:

To support Dashboard query metric ability, we should deliver the monitor information (host/port) to PD etcd.

This request add etcdClient in PDClient in order to Put/Delete Key/Value in PD Etcd.

```release-note
NONE
```


pkg/pdapi/pdetcd.go
```diff
@@ -0,0 +1,62 @@
...
+	return &pdEtcdClient{
+		etcdClient: etcdClient,
+		timeout:    timeout,
+	}, nil
+}
+
+func (pec *pdEtcdClient) PutKey(key, value string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), pec.timeout)
+	defer cancel()
+	_, err := pec.etcdClient.Put(ctx, key, value)
```
weekface, line 53:
```suggestion
	return pec.etcdClient.Put(ctx, key, value)
```

why not return the original result?

Yisaer: I think we don't need to handle the etcd response.  we only need to ensure the `PutKey` is success by check whether error is nil.

pkg/pdapi/pdetcd.go
```diff
@@ -0,0 +1,62 @@
...
+
+	DeleteKey(key string) error
+}
+
+type pdEtcdClient struct {
+	timeout    time.Duration
+	etcdClient *etcdclientv3.Client
+}
+
+func NewPdEtcdClient(url string, timeout time.Duration, tlsConfig *tls.Config) (PDEtcdApi, error) {
```
DanielZhangQD, line 35: Why do we need etcd client? Can we use the pdclient?

Yisaer: We use etcdClient to set etcd key in PD in order to support dashboard.

Issue #2410, opened 2020-05-09
No named port for tikv metrics port


tirsen:
There is no container port set up for the tikv metrics port 20180:

```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubernetes.io/psp: eks.privileged
    prometheus.io/path: /metrics
    prometheus.io/port: "20180"
...
    ports:
    - containerPort: 20160
      name: server
      protocol: TCP
```

DanielZhangQD:
@tirsen Yes, it's not set explicitly in the yaml, but the metrics can be pulled by Prometheus, do you have any issues with the current yaml settings?

tirsen:
No I realized I can use `targetPort` instead so it's all good.

DanielZhangQD:
OK, feel free to let us know if any more issues, thanks!

tirsen:
Actually using`targetPort` doesn't look like it's working with prometheus-operator. It needs a named port.

Yisaer:
Recently the community have contributed the example of prometheus-operator to monitor tidbcluster.  It seems ok for this community user in his situation. @tirsen  Could you provide your prometheus-operator version that I can test whether it is necessary.

https://github.com/pingcap/tidb-operator/tree/master/examples/prometheus-operator

yeya24:
@tirsen Hello, you can achieve that without adding status port by relabeling. Example `ServiceMonitor`. I just tested in my environment and it works well. Could you please try it?

``` yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: basic-tikv
spec:
  endpoints:
  - interval: 15s
    scrapeTimeout: 10s
    path: /metrics
    relabelings:
    - sourceLabels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
      regex: ([^:]+)(?::\d+)?;(\d+)
      targetLabel: __address__
      replacement: $1:$2
      action: replace
    metricRelabelings:
    - sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_instance]
      action: keep
      regex: basic
    - sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_component]
      action: keep
      regex: tikv
    - sourceLabels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
      action: keep
      regex: 'true'
    - sourceLabels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
[...]

PR #2456, opened 2020-05-13
[DNM] Create named port for tikv status 20180 (#2410)

tirsen:
#2410

### DIFF
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -435,6 +435,11 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 				ContainerPort: int32(20160),
 				Protocol:      corev1.ProtocolTCP,
 			},
+			{
+				Name:          "status",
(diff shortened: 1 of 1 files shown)


tirsen:
I have tested this locally and I can confirm that Prometheus can now scrape metrics after adding that port. Prometheus could not scrape the metrics even if using an integer `targetPort` because Prometheus auto discovery can't discover ports without names.

Yisaer:
This changes is ok for me but I think it would upgrade tikv if user starts upgrade operator from v1.0.x. I think #2466 could also solve the problem so I tend to receive this request but not port to the v1.1

cofyc:
hi, @yeya24 does your suggested solution in #2466 solves this problem too?

@tirsen we try to avoid change TiKV default statefulset spec as possible as we can because it will trigger a rolling update.

tirsen:
It would be cool to have changes that are applied after the diffing so that you can make "optional" changes that are only applied after other changes that does require a restart (such as an image upgrade).

DanielZhangQD:
@tirsen We will fix this in another PR to add a parameter to control this to avoid rolling update TiKV by default, thanks!

PR #2466, opened 2020-05-14
Update prometheus operator tikv example

yeya24:

The current tikv service monitor example doesn't work with prometheus operator because tikv doesn't open 20180 explicitly.

This example works for me. Verification

image

### DIFF
--- a/examples/prometheus-operator/tikv-serviceMonitor.yaml
+++ b/examples/prometheus-operator/tikv-serviceMonitor.yaml
@@ -10,7 +10,12 @@ spec:
   - interval: 15s
     scrapeTimeout: 10s
     path: /metrics
-    targetPort: 20180
+    relabelings:
+    - sourceLabels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
+      regex: ([^:]+)(?::\d+)?;(\d+)
+      targetLabel: __address__
+      replacement: $1:$2
+      action: replace
     metricRelabelings:
     - sourceLabels: [__meta_kubernetes_pod_label_app_kubernetes_io_instance]
       action: keep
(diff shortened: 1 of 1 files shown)


yeya24:
@mikechengwei  It would be appreciated if you could take a look 😄

Yisaer:
That's great ! What about also adding a new label to show the tikv instance friendly for user. For example:
```
- sourceLabels:[__meta_kubernetes_pod_name,__meta_kubernetes_pod_label_app_kubernetes_io_instance,__meta_kubernetes_pod_annotation_prometheus_io_port]
regex: (.+);(.+);(.+)
targetLabel: tikv_instance
replacement: $1.$2-tikv-peer:$3
```

yeya24:
> That's great ! What about also adding a new label to show the tikv instance friendly for user. For example:
> 
> ```
> - sourceLabels:[__meta_kubernetes_pod_name,__meta_kubernetes_pod_label_app_kubernetes_io_instance,__meta_kubernetes_pod_annotation_prometheus_io_port]
> regex: (.+);(.+);(.+)
> targetLabel: tikv_instance
> replacement: $1.$2-tikv-peer:$3
> ```

I am ok with that. But there is already a `pod` label. Do you think it is good enough?

Yisaer:
@yeya yes, it is good enough. sry I missed this.

mikechengwei:
I have no problem.

cofyc:
how about waiting for @tirsen to check if this solves the problem in their environments?

DanielZhangQD:
> how about waiting for @tirsen to check if this solves the problem in their environments?

@tirsen has confirmed that this can fix his issue.

DanielZhangQD:
@yeya24 Thanks a lot for the contribution!

tirsen:
So this doesn't actually work for us out of the box (possibly because we are on prometheus-operator 0.33). We have to explicitly specify `port: server` in order for prometheus to auto discover those targets.

yeya:
> @yeya yes, it is good enough. sry I missed this.

I've got nothing to do with it, but it looks great! ;-)

yeya24:
> So this doesn't actually work for us out of the box (possibly because we are on prometheus-operator 0.33). We have to explicitly specify `port: server` in order for prometheus to auto discover the tikv targets, maybe later versions of the prometheus-operator automatically discovers all ports if you don't specify one?

Thanks for the test. I am not very sure about it but it seems nothing to do with the version of prometheus operator. May I ask what version of Prometheus you are using?

PR #2483, opened 2020-05-18
Support Dashboard Metrics Storage Key

Yisaer:
close https://github.com/pingcap/tidb-operator/issues/2331

This request sync prometheus and grafana information into pd etcd with tidb nightly verison. I leave alertmanager into ucp. Here is the demo:

image

image

```release-note
Support `Dashboard` metrics ability for `TidbCluster` when `TidbMonitor` deployed 
```


pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -14,37 +14,180 @@
...
+		err = cleanGrafanaKey(pdEtcdClient)
+		if err != nil {
+			klog.Error(err.Error())
+			return err
 		}
 	}
 	return nil
 }
 
+func putGrafanaKey(etcdClient pdapi.PDEtcdClient, value string) error {
```
DanielZhangQD, line 150: What's the benefit of wrapping up the etcdClient functions here?
The functions are just with different names only.

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -14,37 +14,180 @@
...
 func (tcsm *TidbClusterStatusManager) Sync(tc *v1alpha1.TidbCluster) error {
+	return tcsm.syncTidbMonitorRefAndKey(tc)
+}
+
+func (tcsm *TidbClusterStatusManager) syncTidbMonitorRefAndKey(tc *v1alpha1.TidbCluster) error {
+	tm, err := tcsm.syncTidbMonitorRef(tc)
+	if err != nil {
+		return err
+	}
+	return tcsm.syncDashboardMetricStorage(tc, tm)
```
DanielZhangQD, line 57: Status manager should focus on the status update, can we move the etcd key operation to a separate function out of the status manager?

pkg/manager/member/tidbcluster_status_manager.go
Yisaer: Just want to make things clear and isolated.

pkg/manager/member/tidbcluster_status_manager.go
Yisaer: Currently, i think it is not necessary as it seems we would only solve the `MonitorRef` and `AutoSclaerRef` (in future) here. If we movet the etcd operation out of statusManager, we will try to fetch `TidbMonitor` again in its scope.

pkg/manager/member/tidbcluster_status_manager.go
Yisaer: updated.

pkg/pdapi/pdetcd.go
```diff
@@ -19,11 +19,13 @@ import (
 	"time"
 
 	etcdclientv3 "github.com/coreos/etcd/clientv3"
+	etcdclientv3util "github.com/coreos/etcd/clientv3/clientv3util"
 )
 
-type PDEtcdApi interface {
+type PDEtcdClient interface {
+	// PutKey would put key to the target pd etcd cluster
```
cofyc, line 26: why use `would` here? At the end of the function, the put/delete operation may not finish?

pkg/pdapi/pdetcd.go
```diff
@@ -19,11 +19,13 @@ import (
...
 	PutKey(key, value string) error
-
+	// DeleteKey would delete key from the target pd etcd cluster
```
cofyc, line 28: ditto

pkg/pdapi/pdetcd.go
Yisaer: I think the comment is misleading, I should use `will` here.

pkg/pdapi/pdapi.go
```diff
@@ -72,6 +75,32 @@ func GetTLSConfig(kubeCli kubernetes.Interface, namespace Namespace, tcName stri
 	return crypto.LoadTlsConfigFromSecret(secret, caCert)
 }
 
+func (pdc *defaultPDControl) GetPDEtcdClient(namespace Namespace, tcName string, tlsEnabled bool) (PDEtcdClient, error) {
+	pdc.mutex.Lock()
```
DanielZhangQD, line 79: Suggest define separate mutex here, it's an independent client with existing PDClient.

Yisaer: updated.

Issue #2467, opened 2020-05-14
Change config field of components to pointer


aylei:
## Bug Report

**What did you do?**

configure 
```
server:
 request-batch-enable-cross-command: false
```
for tikv
**What did you expect to see?**
tikv start with request-batch-enable-cross-command disabled

**What did you see instead?**
tikv start with:

```
[server]
```
the `request-batch-enable-cross-command` is not specified

Issue #1661, opened 2020-02-10
Integrate with TiDB v4.0 for configuration parameters update


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

There are configuration parameters update, e.g. add, remove, type change, default value modification, etc. for the components of TiDB v4.0.
https://github.com/pingcap/tidb/pull/12853 is an example.
**Describe the feature you'd like:**

TiDB Operator needs to update for the configuration changes to support TiDB v4.0.
**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

DanielZhangQD:
@Yisaer Please also double-check the parameters in tidb release v3.1, which will release RC version by 03/19, normally they should already be included in the parameters of v4.0.
Thanks!

PR #2488, opened 2020-05-18
Update TiDB and TiKV configuration to 4.0.0-rc.2

Yisaer:
Close https://github.com/pingcap/tidb-operator/issues/2467
Close https://github.com/pingcap/tidb-operator/issues/1661

```release-note
Support TiDB and TiKV Configuration to v4.0.0-rc.2
```


pkg/apis/pingcap/v1alpha1/tiflash_config.go
```diff
@@ -32,9 +32,9 @@ type TiFlashConfig struct {
 // +k8s:openapi-gen=false
 type FlashServerConfig struct {
 	// +optional
-	EngineAddr string `json:"engine-addr,omitempty" toml:"engine-addr,omitempty"`
+	EngineAddr *string `json:"engine-addr,omitempty" toml:"engine-addr,omitempty"`
```
DanielZhangQD, line 35: Please also update the default config logic in https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tiflash_util.go#L122.
And also make sure that TiFlash can start up successfully, it requires some configs to exist even if they are empty strings.

Yisaer: updated.

PR #2545, opened 2020-05-25
Directly use CRD in e2e test instead of helm

Yisaer:

Currently, our test code in e2e/stability is based on helm in managing tidbcluster. As CRD is the only tool recommended to manage tidbcluster after operator 1.1.0 released, I think it is necessary to use CRD instead of helm to manage tidbcluster in test. However, the current OperatorActions is based on the helm and it's hard to revise it as it would both influence the e2e test and stabilty test. 

As @cofyc commented, the XXXAction is overdesigned and in this request we only using function to manage tidbcluster here.

```release-note
NONE
```


cofyc:
I agree to deprecate helm operations, but suggest using plain utility functions. it should not be hard to write logic around kubernetes and our interface or generic interface from the controller-runtime library.

cofyc:
IMO current XXXActions are over-designed and not portable.

Yisaer:
Updated, now we will directly use function to manage crd.

PR #2560, opened 2020-05-26
remove records of deleted pods from failure stores automatically and add an e2e test

cofyc:
### What problem does this PR solve? 
fixes #2459 
fixes #1464 
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has Go code change
 - Has CI related scripts change
 - Has Terraform scripts change

Side effects

 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation

### Does this PR introduce a user-facing change?:

```release-note
NONE
```

PR #2570, opened 2020-05-27
Fix PD failover

Yisaer:
Limit the PD Member in checking PD failover. If the PD Member didn't manage by the operator, the operator wouldn't failover if the unmanaged pd member failed.

```release-note
`Operator` won't failover failed `PD` Member if it didn't be managed by `Operator`.
```

### DIFF
--- a/pkg/manager/member/pd_failover.go
+++ b/pkg/manager/member/pd_failover.go
@@ -18,7 +18,6 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/pingcap/advanced-statefulset/client/apis/apps/v1/helper"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/client/clientset/versioned"
 	"github.com/pingcap/tidb-operator/pkg/controller"
@@ -27,7 +26,6 @@ import (
 	apiv1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 	apps "k8s.io/client-go/listers/apps/v1"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/record"
@@ -43,7 +41,6 @@ type pdFailover struct {
 	pvcLister        corelisters.PersistentVolumeClaimLister
 	pvcControl       controller.PVCControlInterface
 	pvLister         corelisters.PersistentVolumeLister
-	setLister        apps.StatefulSetLister
 	recorder         record.EventRecorder
 }
 
@@ -67,7 +64,6 @@ func NewPDFailover(cli versioned.Interface,
 		pvcLister,
 		pvcControl,
 		pvLister,
-		setLister,
 		recorder}
 }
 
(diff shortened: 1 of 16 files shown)


pkg/manager/member/pd_failover.go
```diff
@@ -77,7 +83,18 @@ func (pf *pdFailover) Failover(tc *v1alpha1.TidbCluster) error {
...
+	pdsts, err := pf.setLister.StatefulSets(ns).Get(controller.PDMemberName(tcName))
+	if err != nil {
+		return err
+	}
+	podNames := sets.String{}
+	for _, ordinal := range helper.GetPodOrdinals(*pdsts.Spec.Replicas, pdsts).List() {
+		podNames.Insert(util.GetPodName(tc, v1alpha1.PDMemberType, ordinal))
+	}
 	for podName, pdMember := range tc.Status.PD.Members {
+		if !podNames.Has(podName) {
```
DanielZhangQD, line 95: I think we should update `tryToMarkAPeerAsFailure` instead of here, why ignore the non-managed members in the cluster health check?

Yisaer: It's ok for me. updated.

pkg/manager/member/pd_failover.go
```diff
@@ -121,7 +127,20 @@ func (pf *pdFailover) tryToMarkAPeerAsFailure(tc *v1alpha1.TidbCluster) error {
...
+	if err != nil {
+		return err
+	}
+	podNames := sets.String{}
+	for _, ordinal := range helper.GetPodOrdinals(*pdsts.Spec.Replicas, pdsts).List() {
+		podNames.Insert(util.GetPodName(tc, v1alpha1.PDMemberType, ordinal))
+	}
+
 	for podName, pdMember := range tc.Status.PD.Members {
+		if !podNames.Has(podName) {
```
DanielZhangQD, line 140: Should we keep consistent with tikv handling [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_member_manager.go#L605)?

Yisaer: updated.

pkg/manager/member/pd_failover.go
```diff
@@ -125,6 +125,12 @@ func (pf *pdFailover) tryToMarkAPeerAsFailure(tc *v1alpha1.TidbCluster) error {
 		if pdMember.LastTransitionTime.IsZero() {
 			continue
 		}
+		if !pf.isPodDesired(tc, podName) {
```
DanielZhangQD, line 128: Should we also delete the non desired members in the failure members as done in https://github.com/pingcap/tidb-operator/pull/2560/files#diff-cff8f7143431f3d6302182e300a81909R58?

pkg/manager/member/pd_failover.go
```diff
@@ -125,6 +125,12 @@ func (pf *pdFailover) tryToMarkAPeerAsFailure(tc *v1alpha1.TidbCluster) error {
...
+			// we should ignore the store record of deleted pod, otherwise the
+			// record of deleted pod may be added back to failure stores
+			// (before it enters into Offline/Tombstone state)
```
cofyc, line 131:
```suggestion
```

this comment is for tikv/tiflash stores, not correct here

pkg/manager/member/pd_failover.go
cofyc: I'm not sure if it's necessary right now since all failure members will be cleared. anyway, we can revise it later.

PR #2582, opened 2020-05-28
Revise defaulting and validating

Yisaer:
Currently, the controller-manager will directly default and validate the pingcapresource(tidbcluster). So I think the webhook for validate and default the pingcap resources is not necessary and removed in this request.

The e2e test for defaulting and validating is removed as I think the current unit test is enough to cover them.

```release-note
Remove the pingcap resource webhook configuration in Operator charts.
```


tests/e2e/tidbcluster/serial.go
```diff
@@ -153,197 +150,6 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
 		})
 	})
 
-	ginkgo.Context("[Feature: Defaulting and Validating]", func() {
-		var ocfg *tests.OperatorConfig
-		var oa tests.OperatorActions
-
-		ginkgo.BeforeEach(func() {
```
Yisaer, line 160: In my view, there is unnecessary to add e2e test to cover default and validate templates. Unit test is enought.

Issue #2512, opened 2020-05-20
Split e2e-test into independent jobs


Yisaer:
## Feature Request

Currently, if the e2e-test failed by 1 job like serial or advancedsts. we have to rerun the e2e-test and the total 4 jobs need to rerun again. Maybe we could split the e2e-test into 4 independent jobs. This would reduce the duplicated test.  WDYT?  @cofyc

cofyc:
I think it's a good idea.

cofyc:
- [ ] add new Jenkins file for general e2e kind job: https://github.com/pingcap/tidb-operator/pull/2872
- [ ] update Github settings for master and release-1.1 branches

PR #2590, opened 2020-05-29
Make some e2e jobs into periodicity job

Yisaer:
Currently, I think some test cases in basic context is not necessary or important for each Pull request, so I suggest moving some e2e jobs into a periodicity job with a specific frequency (maybe once a day)

Related changes

```release-note
NONE
```

### DIFF
--- a/tests/e2e/tidbcluster/stability-periodicity.go
+++ b/tests/e2e/tidbcluster/stability-periodicity.go
@@ -58,7 +58,7 @@ import (
 )
 
 var _ = ginkgo.Describe("[tidb-operator][Stability]", func() {
-	f := e2eframework.NewDefaultFramework("br")
+	f := e2eframework.NewDefaultFramework("periodicity")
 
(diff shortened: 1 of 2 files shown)


Yisaer:
@cofyc Do you think this is a proper way to reduce the consuming time for e2e test?

cofyc:
Actually, it's best we can run all tests on pull request.
After https://github.com/pingcap/tidb-operator/issues/2512 is done, we can add some optional jobs (running on the PR but not required to merge the PR).

btw, some e2e tests are used to verify controller logic only which we can try to verify in unit tests by using fake clientset.

Yisaer:
> Actually, it's best we can run all tests on pull request.
> After #2512 is done, we can add some optional jobs (running on the PR but not required to merge the PR).
> 
> btw, some e2e tests are used to verify controller logic only which we can rewrite in unit tests by using fake clientset.

That's great.

PR #2647, opened 2020-06-08
optionally set resources and priorityclass

zjj2wry:
### Check List 

Tests 

```
helm template charts/tidb-operator -f charts/tidb-operator/values.yaml | kubectl apply -f -
```

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/charts/tidb-operator/values.yaml
+++ b/charts/tidb-operator/values.yaml
@@ -50,6 +50,9 @@ controllerManager:
     requests:
       cpu: 80m
       memory: 50Mi
+#  # REF: https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
+#  priorityClassName: system-cluster-critical
+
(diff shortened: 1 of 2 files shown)


charts/tidb-operator/values.yaml
```diff
@@ -44,12 +44,12 @@ controllerManager:
   logLevel: 2
   replicas: 1
   resources:
-    limits:
-      cpu: 250m
-      memory: 150Mi
```
cofyc, line 49: is this intended? cc @zjj2wry @aylei

zjj2wry: yes, now dbaas is set to a large value to avoid oom. In the future, resources may be controlled by vpa. https://github.com/pingcap/dbaas/pull/1757

cofyc: got

PR #2656, opened 2020-06-08
[DNM] Use secret as Grafana credentials

### LINKED ISSUE #2637: tidb-monitor grafana should set username and password in a Secret
## Feature Request

From the example https://github.com/pingcap/tidb-operator/blob/master/manifests/monitor/tidb-monitor.yaml, seems the grafana in tidb-monitor can only set username and password in plain text. This is not good and it is better to use `Secret` to specify these fields.

yeya24:

Fixes #2637 

```release-note
Use secretRef for `Grafana` credentials in `TidbMonitor` instead of plain text
```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
+++ b/pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
@@ -121,8 +121,18 @@ type GrafanaSpec struct {
 
 	LogLevel string      `json:"logLevel,omitempty"`
 	Service  ServiceSpec `json:"service,omitempty"`
-	Username string      `json:"username,omitempty"`
-	Password string      `json:"password,omitempty"`
+
+	// +optional
+	// if passwordSecret is omitted, Grafana will use `admin` as its password by default.
+	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty"`
+
+	Username string `json:"username,omitempty"`
+
+	// +optional
+	// Deprecated
+	// +k8s:openapi-gen=false
+	Password string `json:"password,omitempty"`
+
 	// +optional
 	Envs map[string]string `json:"envs,omitempty"`
 
--- a/pkg/monitor/monitor/util.go
+++ b/pkg/monitor/monitor/util.go
@@ -93,21 +93,6 @@ func getMonitorConfigMap(tc *v1alpha1.TidbCluster, monitor *v1alpha1.TidbMonitor
 	return cm, nil
 }
 
-func getMonitorSecret(monitor *v1alpha1.TidbMonitor) *core.Secret {
-	return &core.Secret{
-		ObjectMeta: meta.ObjectMeta{
-			Name:            GetMonitorObjectName(monitor),
-			Namespace:       monitor.Namespace,
-			Labels:          buildTidbMonitorLabel(monitor.Name),
-			OwnerReferences: []meta.OwnerReference{controller.GetTiDBMonitorOwnerRef(monitor)},
-		},
-		Data: map[string][]byte{
-			"username": []byte(monitor.Spec.Grafana.Username),
-			"password": []byte(monitor.Spec.Grafana.Password),
-		},
-	}
-}
-
 func getMonitorServiceAccount(monitor *v1alpha1.TidbMonitor) *core.ServiceAccount {
 	sa := &core.ServiceAccount{
 		ObjectMeta: meta.ObjectMeta{
@@ -192,7 +177,7 @@ func getMonitorRoleBinding(sa *core.ServiceAccount, role *rbac.Role, monitor *v1
 	}
 }
 
-func getMonitorDeployment(sa *core.ServiceAccount, config *core.ConfigMap, secret *core.Secret, monitor *v1alpha1.TidbMonitor, tc *v1alpha1.TidbCluster) (*apps.Deployment, error) {
+func getMonitorDeployment(sa *core.ServiceAccount, config *core.ConfigMap, monitor *v1alpha1.TidbMonitor, tc *v1alpha1.TidbCluster) (*apps.Deployment, error) {
 	deployment := getMonitorDeploymentSkeleton(sa, monitor)
 	initContainer := getMonitorInitContainer(monitor, tc)
 	deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, initContainer)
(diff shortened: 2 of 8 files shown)


pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -118,8 +118,15 @@ type GrafanaSpec struct {
...
-	Username string      `json:"username,omitempty"`
-	Password string      `json:"password,omitempty"`
+
+	// +optional
+	// if username is omitted, Grafana will use `admin` as its username by default.
+	Username *corev1.SecretKeySelector `json:"username,omitempty"`
+
+	// +optional
+	// if password is omitted, Grafana will use `admin` as its password by default.
+	Password *corev1.SecretKeySelector `json:"password,omitempty"`
```
Yisaer, line 128: Does it be Forward compatible to the previous TidbMonitor object? What about using a new property to indicate the grafana's password secretRef and make the current Password as Deprecated?

yeya24: Great, I think that makes sense. Will update this week.

Yisaer:
@yeya24  Hi,any progress now?

yeya24:
> @yeya24 Hi,any progress now?

Will work on it today

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -118,8 +118,25 @@ type GrafanaSpec struct {
 
 	LogLevel string      `json:"logLevel,omitempty"`
 	Service  ServiceSpec `json:"service,omitempty"`
-	Username string      `json:"username,omitempty"`
-	Password string      `json:"password,omitempty"`
+
+	// +optional
+	// if usernameSecret is omitted, Grafana will use `admin` as its username by default.
+	UsernameSecret *corev1.SecretKeySelector `json:"usernameSecret,omitempty"`
```
Yisaer, line 124: I think we can use `Username` and `UsernameSecret` is unnecessary.

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -118,8 +118,25 @@ type GrafanaSpec struct {
...
+
+	// +optional
+	// if passwordSecret is omitted, Grafana will use `admin` as its password by default.
+	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty"`
+
+	// +optional
+	// Deprecated
+	// +k8s:openapi-gen=false
+	Username string `json:"username,omitempty"`
```
Yisaer, line 133: We don't need to set `Username` as `Deprecated`.

pkg/monitor/monitor/util.go
```diff
@@ -448,32 +433,14 @@ func getMonitorGrafanaContainer(secret *core.Secret, monitor *v1alpha1.TidbMonit
...
-				Name: "GF_SECURITY_ADMIN_PASSWORD",
-				ValueFrom: &core.EnvVarSource{
-					SecretKeyRef: &core.SecretKeySelector{
-						LocalObjectReference: core.LocalObjectReference{
-							Name: secret.Name,
-						},
-						Key: "password",
-					},
-				},
-			},
```
Yisaer, line 472: I suggest remain these logic for the backward compatibility for 1.1.x user. And add the comment that. We will finally drop these logic in 1.2.0

yeya24: I am not very sure. This is a serious security problem, do we really need to prioritize backward compatibility over security in this case?

Yisaer: This is a good concern. I think we can break the backward compatibility here. Just mention this the in the request's release note.

cofyc: Especially in patch versions, we should not introduce break changes. The security problem is not unavoidable, we can encourage people to use the secret to store password after this.

We can remove `password` field in next minor version v1.12.

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -121,8 +121,18 @@ type GrafanaSpec struct {
...
-	Password string      `json:"password,omitempty"`
+
+	// +optional
+	// if passwordSecret is omitted, Grafana will use `admin` as its password by default.
+	PasswordSecret *corev1.SecretKeySelector `json:"passwordSecret,omitempty"`
+
+	Username string `json:"username,omitempty"`
+
+	// +optional
+	// Deprecated
```
cofyc, line 132:
```suggestion
	// Deprecated in v1.1.2 for security concerns, planned for removal in v1.2. Use `passwordSecret` instead.
```

pkg/monitor/monitor/util.go
DanielZhangQD: Agree with @cofyc , we should maintain the backward compatibility here, and we can make it clear in the doc & comments that the secret should be used.

Yisaer:
@yeya24 Hi, after discussion with other maintainers, we think it's better not to introduce break changes in the minor version. We could remove the password in the next major version.

yeya24:
> @yeya24 Hi, after discussion with other maintainers, we think it's better not to introduce break changes in the minor version. We could remove the password in the next major version.

Understand. Sorry for the late response last week. I will update it today.

DanielZhangQD:
> > @yeya24 Hi, after discussion with other maintainers, we think it's better not to introduce break changes in the minor version. We could remove the password in the next major version.
> 
> Understand. Sorry for the late response last week. I will update it today.

@yeya24 Just a friendly ping.

yeya24:
> > > @yeya24 Hi, after discussion with other maintainers, we think it's better not to introduce break changes in the minor version. We could remove the password in the next major version.
> > 
> > 
> > Understand. Sorry for the late response last week. I will update it today.
> 
> @yeya24 Just a friendly ping.

Got it, will try to update it this week.

DanielZhangQD:
@yeya24 Are you still working on this?

yeya24:
Sorry I don't have time to work on this recently. This is stale and would be good if someone else can pick this up.

DanielZhangQD:
> Sorry I don't have time to work on this recently. This is stale and would be good if someone else can pick this up.

OK. Thanks!

DanielZhangQD:
Close this temporarily, we may have someone to continue the work later.

Issue #2395, opened 2020-05-08
Procedure to access TiDB dashboard with TLS enabled


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Currently, it's not easy to access TiDB dashboard with TLS enabled.
**Describe the feature you'd like:**

A Procedure to access TiDB dashboard with TLS enabled.
**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

DanielZhangQD:
Nginx ingress controller supports TLS with backend service:
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#backend-certificate-authentication

Yisaer:
I tried some annotation list on the previous document and it seems not work. Maybe the nginx controller we use have a lower version or I use it in a wrong way. 

I think providing a more common way to expose the dashboard service instead of using the advance ability in nginx controller (besides, nginx is not the only way for the ingress controller). I'd like to use a proxy server(eg: discovery service) to proxy the dashboard service.

The discovery service would expose the dashboard service with http and hide the logic between the proxy server and the dashboard service (enable tls or not). In this way, we could use the basic ability of the Ingress to expose the dashboard service and user will use dashboard with ignoring whether tls enabled or not.

PR #2684, opened 2020-06-11
Add proxy server for Dashboard

Yisaer:

Close https://github.com/pingcap/tidb-operator/issues/2395

When `tlsCluster` is enabled,  dashboard should be visited after loaded pd client cert in browser which didn't make sense.

This pr make `discovery` service as a proxy-server for the pd dashboard service like following.
When `tlsCluster` is enabled, the proxy-server would automatically loaded cert from secret and expose the http api. Also, proxy-server ensured that only requestURI started with `/dashboard` would be redirected. No pdapi would be exposed by proxy-server.

image

Related changes

```release-note
Expose `Dashboard` service with `HTTP` endpoint whether `tlsCluster` is enabled.
```


pkg/discovery/server/proxy.go
```diff
@@ -0,0 +1,72 @@
...
+			klog.Fatal(err)
+		}
+		proxy.Transport = &http.Transport{TLSClientConfig: tlsConfig}
+	}
+	director := proxy.Director
+	proxy.Director = func(req *http.Request) {
+		if strings.HasPrefix(req.RequestURI, "/dashboard") {
+			director(req)
+			req.Host = req.URL.Host
+		}
```
Yisaer, line 63: Ensuring no pdapi would be exposed.

cofyc:
how about passing the part、backend service、tls configs to the discovery service? then this discovery service can be a generic proxy server.

Yisaer:
> how about passing the part、backend service、tls configs to the discovery service? then this discovery service can be a generic proxy server.

That's ok for me but I think it is not necessary currently as I can't find out second service needed proxy to be exposed.

pkg/discovery/server/proxy.go
```diff
@@ -0,0 +1,73 @@
...
+	url := &url.URL{
+		Host:   fmt.Sprintf("%s-pd:2379", tcName),
+		Scheme: "http",
+	}
+	tc, err := cli.PingcapV1alpha1().TidbClusters(namespace).Get(tcName, metav1.GetOptions{})
+	if err != nil {
+		klog.Error(err)
+		return nil, err
+	}
+	if tc.Spec.TLSCluster != nil && tc.Spec.TLSCluster.Enabled {
```
weekface, line 40:
```suggestion
	if tc.IsTLSClusterEnabled() {
```

pkg/discovery/server/proxy.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func buildProxy(cli versioned.Interface, kubeCli kubernetes.Interface, tcName, namespace string) *httputil.ReverseProxy {
+	url, err := buildUrl(cli, tcName, namespace)
+	if err != nil {
+		klog.Fatal(err)
+	}
+	proxy := httputil.NewSingleHostReverseProxy(url)
+	if url.Scheme == "https" {
+		tlsConfig, err := pdapi.GetTLSConfig(kubeCli, pdapi.Namespace(namespace), tcName, nil)
```
cofyc, line 53: the client secret may be invalid or expired, we need to find a way to refresh HTTP server when the secret is updated
we can mount the secrets into the pod, and reload the TLS when necessary like PD, tidb

cmd/discovery/main.go
```diff
@@ -33,12 +33,14 @@ import (
...
 	printVersion bool
 	port         int
+	proxyPort    int
 )
 
 func init() {
 	flag.BoolVar(&printVersion, "V", false, "Show version and quit")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.IntVar(&port, "port", 10261, "The port that the tidb discovery's http service runs on (default 10261)")
+	flag.IntVar(&proxyPort, "proxy-port", 10262, "The port that proxy the tidbcluster's pd dashboard service")
```
cofyc, line 43:
```suggestion
	flag.IntVar(&port, "port", 10261, "The port that the tidb discovery's proxy service runs on (default 10262)")
```

same as `port` flag

pkg/discovery/server/proxy.go
Yisaer: good catch

Yisaer: updated.

pkg/discovery/server/proxy.go
```diff
@@ -0,0 +1,91 @@
...
+		if err != nil {
+			klog.Fatal(err)
+		}
+		rootCAs.AppendCertsFromPEM(caByte)
+		tlsConfig := &tls.Config{
+			Certificates: []tls.Certificate{cert},
+			RootCAs:      rootCAs,
+		}
+		proxy.Transport = &http.Transport{TLSClientConfig: tlsConfig}
+	}
```
cofyc, line 76: certs are loaded once at startup, I guess we should reload TLS certificates for every client request

cofyc: this is a simple mechanism to make sure this proxy will not fail when the certs are expired
we can also have some unit tests to verify this
ref: https://github.com/etcd-io/etcd/pull/7829

Yisaer: > certs are loaded once at startup, I guess we should reload TLS certificates for every client request

Not sure whether this would be quite time-consuming but is this necessary?

cofyc: this is how PD/TiDB refresh TLS certificates too
it's possible to refresh on expiration (see etcd TODO), but I don't think it's necessary. just read certificates every time
actually these small disk files are cached in buffer in most time

Yisaer: make sense. updated. now the proxy would be initialed each time

Issue #2631, opened 2020-06-05
upgrade and scaling in tikv at the same time will cause upgrade failure


DanielZhangQD:
## Bug Report

**What version of Kubernetes are you using?**

1.12.8
**What version of TiDB Operator are you using?**

v1.1.0
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

local-storage
**What's the status of the TiDB cluster pods?**

Running
**What did you do?**

Update spec.version from 3.1.1 to 3.1.2
Update spec.tikv.replicas from 6 to 3
**What did you expect to see?**
Cluster is upgraded successfully and tikv is scaled in to 3.
**What did you see instead?**
tikv-5 is scaled in before tikv upgrade and the tikv upgrade hangs after tikv-5 is upgraded because tikv-5 can never be ready again since it's in offline state.

DanielZhangQD:
```
I0605 08:37:46.246529       1 stateful_set_control.go:81] TidbCluster: [dan106/dan106]'s StatefulSet: [dan106/dan106-pd] updated successfully
I0605 08:37:50.454438       1 tikv_scaler.go:89] scaling in tikv statefulset dan106/dan106-tikv, ordinal: 5 (replicas: 5, delete slots: [])
I0605 08:37:51.785155       1 tikv_scaler.go:114] tikv scale in: delete store 154 for tikv dan106/dan106-tikv-5 successfully
I0605 08:37:51.792941       1 tidbcluster_control.go:68] TidbCluster: [dan106/dan106] updated successfully
I0605 08:37:51.792974       1 tidb_cluster_controller.go:295] TidbCluster: dan106/dan106, still need sync: TiKV dan106/dan106-tikv-5 store 154  still in cluster, state: Up, requeuing
I0605 08:38:00.038326       1 utils.go:149] set dan106/dan106-pd partition to 1
I0605 08:38:00.038372       1 utils.go:149] set dan106/dan106-pd partition to 0
I0605 08:38:00.045450       1 stateful_set_control.go:81] TidbCluster: [dan106/dan106]'s StatefulSet: [dan106/dan106-pd] updated successfully
I0605 08:38:06.991063       1 tidbcluster_control.go:68] TidbCluster: [dan106/dan106] updated successfully
[...]

PR #2705, opened 2020-06-15
do not scale or upgrade tikv at the same time

DanielZhangQD:
Fix https://github.com/pingcap/tidb-operator/issues/2631
Cross-check during upgrading and scaling and make sure that upgrading and scaling for tikv do not occur at the same time.
### Check List 

Tests 

 - Unit test
 - E2E test
   - Scale out TiKV and then upgrade TiKV
   - Scale in TiKV and then upgrade TiKV
   - Scale in and upgrade TiKV at the same time
   - Upgrade TidbCluster to newer version and scale in TiKV at the same time

Code changes

 - Has Go code change

Related changes

```release-note
Do not scale or upgrade tikv at the same time
```

### DIFF
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -633,7 +633,7 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 	}
 	if upgrading && tc.Status.PD.Phase != v1alpha1.UpgradePhase {
 		tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
-	} else {
+	} else if tc.Status.TiKV.Phase != v1alpha1.ScaleOutPhase && tc.Status.TiKV.Phase != v1alpha1.ScaleInPhase {
 		tc.Status.TiKV.Phase = v1alpha1.NormalPhase
 	}
 
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -633,7 +633,7 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 	}
 	if upgrading && tc.Status.PD.Phase != v1alpha1.UpgradePhase {
 		tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
-	} else if tc.Status.TiKV.Phase != v1alpha1.ScaleOutPhase && tc.Status.TiKV.Phase != v1alpha1.ScaleInPhase {
+	} else if !tc.TiKVScaling() {
 		tc.Status.TiKV.Phase = v1alpha1.NormalPhase
 	}
 
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -633,7 +633,9 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 	}
 	if upgrading && tc.Status.PD.Phase != v1alpha1.UpgradePhase {
 		tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
-	} else if !tc.TiKVScaling() {
+	} else if tc.TiKVStsDesiredReplicas() != *set.Spec.Replicas {
+		tc.Status.TiKV.Phase = v1alpha1.ScalePhase
+	} else {
 		tc.Status.TiKV.Phase = v1alpha1.NormalPhase
 	}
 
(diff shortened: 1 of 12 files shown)


Yisaer:
I think PD should have same logic to avoid the same error.

pkg/manager/member/tikv_scaler.go
```diff
@@ -144,7 +173,7 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
 			}
 			klog.Infof("tikv scale in: set pvc %s/%s annotation: %s to %s",
 				ns, pvcName, label.AnnPVCDeferDeleting, now)
-
+			tc.Status.TiKV.Phase = v1alpha1.ScaleInPhase
```
Yisaer, line 176: I think we can't directly set TiKV phase as `ScaleInPhase` here. During scaling, we would update its `statefulset` and `tidbcluster`. What if  the updating of statefulset is success but the updating of tidbcluster is failed? This would cause the scaling truly happened, but we didn't record the state.

I think we should judge the status in `syncTidbClusterStatus` function as upgrading do.

I

DanielZhangQD: We cannot exactly know whether TiKV is scaling in the `syncTidbClusterStatus` function, because the operator may be deleting the store at the time we're checking and we also cannot check the store state because if a store is in tombstone state we cannot make sure it's the store for the current Pod and or a previous Pod that had been deleted before.
We have a retry mechanism for the tidbcluster update and yes, it still can fail, but actually we cannot fix all of the corner cases, this PR will mitigate this issue in most cases.

DanielZhangQD:
> I think PD should have same logic to avoid the same error.

PD is the first component to be handled, if the spec change requires both upgrade and scaling at the same time, operator will always upgrade it first.

Yisaer:
I think we need a mutex variable to ensure the upgrading and scaling won't happen in the same time. Only when the upgrading or scaling have fetch the mutex successfully, then we could do the action.

pkg/manager/member/tikv_scaler.go
```diff
@@ -85,6 +102,16 @@ func (tsd *tikvScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSe
...
+	// During TidbCluster upgrade, if TiKV is scaled at the same time, since
+	// TiKV cannot be upgraded during PD upgrade, the TiKV scaling will occur
+	// before the TiKV upgrade, in this case, the Pump, TiDB, TiFlash, TiCDC, etc.
+	// will be upgraded before TiKV upgrade.
+	// To avoid this case, we skip the scaling in during PD upgrade.
+	if tc.PDUpgrading() {
+		klog.Infof("the TidbCluster: [%s/%s]'s pd is upgrading, can not scale in until the upgrade completed",
+			ns, tcName)
+		return nil
+	}
```
cofyc, line 114: I don't think this is necessary.

Yisaer:
@DanielZhangQD I'm wondering If you first scale pd from 5 to 3 and it might take about 10 sec for pd-4 to transfer leader and for pd-3 to be deleted member.

```sh
0:00 Scaling PD from 5-3
0:02 Upgrading PD Image
```

Will the controller logic prevent pd from upgrading as it is under scaling?

pkg/manager/member/tikv_scaler.go
DanielZhangQD: updated

DanielZhangQD:
> @DanielZhangQD I'm wondering If you first scale pd from 5 to 3 and it might take about 10 sec for pd-4 to transfer leader and for pd-3 to be deleted member.
> 
> Then we follow the action below
> 
> ```shell
> 0:00 Scaling PD from 5-3
> 0:02 Upgrading PD Image
> ```
> 
> Will the controller logic prevent pd from upgrading as it is under scaling?

I have tested this at the beginning, PD can be upgraded and scaled successfully, but yes, in theory, it's possible to hit the similar issue here, I have updated code for PD.

DanielZhangQD:
@Yisaer @cofyc Code updated, PTAL.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -633,6 +633,8 @@ func (tkmm *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 	}
 	if upgrading && tc.Status.PD.Phase != v1alpha1.UpgradePhase {
 		tc.Status.TiKV.Phase = v1alpha1.UpgradePhase
+	} else if tc.TiKVStsDesiredReplicas() != *set.Spec.Replicas {
+		tc.Status.TiKV.Phase = v1alpha1.ScalePhase
```
Yisaer, line 637: If we update `tikvSpec` with some upgrading and scaling in one updating. I think the logic here will think tikv is under Scaling. That is to say, we will start scaling tikv before upgrading tikv. WDYT @DanielZhangQD @cofyc

DanielZhangQD: Yes, it is.

Yisaer: I think this also apply to the PD logic.

DanielZhangQD: I think whether to scaling or upgrading first is OK, the key is not to do both of them at the same time.

Issue #2697, opened 2020-06-14
`patch PV` is printed continuously in the log of tidb-controller-manager


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Create TidbCluster and TidbMonitor, check the logs of tidb-controller-manager, the following logs are printed continuously:
```
I0614 11:55:01.488168       1 event.go:255] Event(v1.ObjectReference{Kind:"TidbMonitor", Namespace:"dan106", Name:"dan106", UID:"04f1836b-ae25-11ea-be5f-8a678ff421f7", APIVersion:"pingcap.com/v1alpha1", ResourceVersion:"140939280", FieldPath:""}): type: 'Normal' reason: 'SuccessfulPatch' patch PV local-pv-2aa7c888 in TidbCluster dan106 successful
I0614 11:55:02.088359       1 event.go:255] Event(v1.ObjectReference{Kind:"TidbCluster", Namespace:"dan106", Name:"dan106", UID:"00ddb676-ae25-11ea-be5f-8a678ff421f7", APIVersion:"pingcap.com/v1alpha1", ResourceVersion:"141009081", FieldPath:""}): type: 'Normal' reason: 'SuccessfulPatch' patch PV local-pv-7033d850 in TidbCluster dan106 successful
I0614 11:55:02.488288       1 event.go:255] Event(v1.ObjectReference{Kind:"TidbMonitor", Namespace:"dan106", Name:"dan106", UID:"04f1836b-ae25-11ea-be5f-8a678ff421f7", APIVersion:"pingcap.com/v1alpha1", ResourceVersion:"140939280", FieldPath:""}): type: 'Normal' reason: 'SuccessfulPatch' patch PV local-pv-92e152f3 in TidbCluster dan106 successful
I0614 11:55:03.289057       1 event.go:255] Event(v1.ObjectReference{Kind:"TidbCluster", Namespace:"dan106", Name:"dan106", UID:"00ddb676-ae25-11ea-be5f-8a678ff421f7", APIVersion:"pingcap.com/v1alpha1", ResourceVersion:"141009081", FieldPath:""}): type: 'Normal' reason: 'SuccessfulPatch' patch PV local-pv-2aa7c888 in TidbCluster dan106 successful
```
**Describe the feature you'd like:**

The logs should not be printed continuously
**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

[...]

PR #2707, opened 2020-06-15
Fix labelselector error in reclaimPolicyManager

Yisaer:
The label selector logic is wrong in `reclaimPolicyManager`'s `sync` function which would cause TidbMonitor covered TidbCluster's reclaimPolicy, this request fix it.

fix https://github.com/pingcap/tidb-operator/issues/2697

```release-note
Fix the pv syncing error when `TidbMonitor` and `TidbCluster` have different value in `reclaimPolicy`
```

### DIFF
--- a/pkg/controller/pv_control.go
+++ b/pkg/controller/pv_control.go
@@ -69,9 +69,17 @@ func (rpc *realPVControl) PatchPVReclaimPolicy(obj runtime.Object, pv *corev1.Pe
 
 	name := metaObj.GetName()
 	pvName := pv.GetName()
-	patchBytes := []byte(fmt.Sprintf(`{"spec":{"persistentVolumeReclaimPolicy":"%s"}}`, reclaimPolicy))
 
-	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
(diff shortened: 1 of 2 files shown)


pkg/controller/pv_control.go
```diff
@@ -69,9 +69,17 @@ func (rpc *realPVControl) PatchPVReclaimPolicy(obj runtime.Object, pv *corev1.Pe
...
-	patchBytes := []byte(fmt.Sprintf(`{"spec":{"persistentVolumeReclaimPolicy":"%s"}}`, reclaimPolicy))
 
-	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+	latestPV, err := rpc.kubeCli.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	if string(latestPV.Spec.PersistentVolumeReclaimPolicy) == string(reclaimPolicy) {
+		return nil
+	}
```
cofyc, line 79: don't check here

Yisaer: updated.

PR #2754, opened 2020-06-17
add a example that uses images in China

cofyc:
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


examples/basic-china/tidb-cluster.yaml
```diff
@@ -0,0 +1,34 @@
+# IT IS NOT SUITABLE FOR PRODUCTION USE.
```
DanielZhangQD, line 1: Is it better to replace the `china` with something like `ali-registry`? It's strange to use `china` in the repo.

cofyc: how about `cn`? 

images we used in this example are located in China, I would suggest using a region name here.

cofyc: `ali-registry` is too specific

DanielZhangQD: ok

examples/basic-cn/tidb-monitor.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: pingcap.com/v1alpha1
+kind: TidbMonitor
+metadata:
+  name: basic
+spec:
+  clusters:
+  - name: basic
+  prometheus:
+    baseImage: prom/prometheus
```
DanielZhangQD, line 9: Should we change the images of prom and grafana to the ones in `registry.cn-beijing.aliyuncs.com/tidb` too?

cofyc: I suggest not, because it will break once users change Prometheus version unless we sync all Prometheus images to our registry

tidb monitor is not optional

examples/basic-cn/tidb-cluster.yaml
```diff
@@ -0,0 +1,34 @@
+# IT IS NOT SUITABLE FOR PRODUCTION USE.
+# This YAML describes a basic TiDB cluster with minimum resource requirements,
+# which should be able to run in any Kubernetes cluster with storage support.
+apiVersion: pingcap.com/v1alpha1
+kind: TidbCluster
+metadata:
+  name: basic
+spec:
+  version: v4.0.0
```
tennix, line 9:
```suggestion
  version: v4.0.1
```

tests/examples/003-basic-cn.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
tennix, line 1:
```suggestion
#!/usr/bin/env bash
```

examples/basic-cn/tidb-monitor.yaml
```diff
@@ -0,0 +1,20 @@
...
+    version: v2.18.1
+  grafana:
+    baseImage: grafana/grafana
+    version: 6.0.1
+  initializer:
+    baseImage: registry.cn-beijing.aliyuncs.com/tidb/tidb-monitor-initializer
+    version: v4.0.0
```
DanielZhangQD, line 16:
```suggestion
    version: v4.0.1
```

PR #2758, opened 2020-06-17
Added "path" to S3StorageProvider API doc

kolbe:

The "path" item was missing from the API docs for S3StorageProvider.

### Check List 

Tests 

- No code

Related changes

```release-note
NONE
```

### DIFF
--- a/docs/api-references/docs.md
+++ b/docs/api-references/docs.md
@@ -7883,6 +7883,17 @@ string
 </tr>
 <tr>
 <td>
+<code>path</code></br>
+<em>
+string
+</em>
+</td>
+<td>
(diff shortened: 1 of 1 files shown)


Yisaer:
The `Path` was hiddened because of document configuration. I think it is only for the `TiFlash` Configuration But it seems affect other areas. @DanielZhangQD  Could you solve this problem and make sure the hidden configuration won't affect any areas except for `TiFlash`?

kolbe:
`path` is used for restore using Lightning: https://pingcap.com/docs/tidb-in-kubernetes/stable/restore-from-s3/#restoration-process

DanielZhangQD:
OK, I will check this issue.

DanielZhangQD:
Fixed in https://github.com/pingcap/tidb-operator/issues/2794.

PR #2769, opened 2020-06-18
Fix pv patch continuously  problem

Yisaer:
Currently, pv would be continuously patched which is not we want. I found that there are mainly 2 problems in the current logic:

1. the selector for the tidbcluster pvc would select all the pvc including `tidbmonitor`' pvc.

```golang
case v1alpha1.TiDBClusterKind:	
     l, err = label.New().Instance(instanceName).Selector()	
     if err != nil {	
         return err	
     }
```

2.  If we didn't define the `PersistentVolumeReclaimPolicy` in the Spec, the comparing here will be not equal forever.

```golang
if pv.Spec.PersistentVolumeReclaimPolicy == policy {
    continue
}
```

Make `PersistentVolumeReclaimPolicy` as pointer and fix the selector logic.

```release-note
Fix the bug that the `PV` managed by Operator would be patched continuously
```


pkg/apis/pingcap/v1alpha1/defaulting/tidbcluster.go
```diff
@@ -60,6 +60,10 @@ func setTidbClusterSpecDefault(tc *v1alpha1.TidbCluster) {
 		d := false
 		tc.Spec.EnablePVReclaim = &d
 	}
+	recyclePVP := corev1.PersistentVolumeReclaimRecycle
```
DanielZhangQD, line 63: Suggest using retain as the default policy. WDYT @cofyc 
Retain is what we suggested in our docs and only few PVs support recycle.

pkg/manager/meta/reclaim_policy_manager.go
```diff
@@ -90,6 +74,20 @@ func (rpm *reclaimPolicyManager) sync(kind, ns, instanceName string, isPVReclaim
 		if pvc.Spec.VolumeName == "" {
 			continue
 		}
+		l := label.Label(pvc.Labels)
+		switch kind {
+		case v1alpha1.TiDBClusterKind:
+			if !l.IsPD() && !l.IsTiDB() && !l.IsTiKV() && !l.IsTiFlash() {
```
DanielZhangQD, line 80: Pump also uses PV.

Yisaer: it's ok for me.

pkg/apis/pingcap/v1alpha1/defaulting/tidbcluster.go
Yisaer: updated.

Issue #2767, opened 2020-06-18
PD Member haven't been deleted after pd scaling-in somtimes


Yisaer:
## Bug Report

Operator v1.1.0  / PD v4.0.0

In recent stability test, We found that the pd member might haven't been deleted after pd scaling-in.

However, from the stability test logs, the logs actually showed that the member has been deleted.

```sh
pd_scaler.go:143] pd scale in: delete member cluster1-pd-5 successfully
``` 

https://github.com/pingcap/tidb-operator/blob/9208698ea418d1f1037b65b05d8c0d0cb4af9b14/pkg/manager/member/pd_scaler.go#L138-L143

Though I haven't known why deleteMember API success but the member still exist. I think we could double check after deleteMember to ensured the deleted Member is not existed during pd scale-in.

PR #2793, opened 2020-06-22
Strict condition for PD Scale-In

Yisaer:
Fix https://github.com/pingcap/tidb-operator/issues/2767

Double check whether the pd member existed after it have been deleted during pd scale-in

Related changes

```release-note
Fix the bug that `PD` Member may still be remained after scaling-in
```


pkg/manager/member/pd_scaler.go
```diff
@@ -142,6 +142,26 @@ func (psd *pdScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet,
...
+
+	existed := false
+	for _, member := range membersInfo.Members {
+		if member.Name == memberName {
+			existed = true
+			break
+		}
+	}
+	if existed {
+		err = fmt.Errorf("pd scale in: failed to check member %s deleted after being deleted", memberName)
```
weekface, line 160:
```suggestion
		err = fmt.Errorf("pd scale in: member %s still exist after being deleted", memberName)
```
?

pkg/manager/member/pd_scaler.go
```diff
@@ -142,6 +142,26 @@ func (psd *pdScaler) ScaleIn(tc *v1alpha1.TidbCluster, oldSet *apps.StatefulSet,
 	}
 	klog.Infof("pd scale in: delete member %s successfully", memberName)
 
+	// double check whether member deleted after delete member
```
cofyc, line 145: is [this](https://github.com/pingcap/tidb-operator/issues/2767) a bug in PD? if it is, add a link to the bug issue and comment it's a workaround solution. if it is not a bug, add a comment describing why the member may not be deleted after the deleting API request succeeds (deleted asynchronously?)

Yisaer: I haven't found the bug yet and have no idea why the member haven't been deleted after deleting API success as it should be.

cofyc: "it should be after deleting API success"

if this is true, then it's a bug in PD, add the bug issue here. It will explain why this is required, otherwise, it's confusing

Yisaer: sure https://github.com/pingcap/pd/issues/2541 updated.

cofyc: great!

Issue #2597, opened 2020-06-01
Gracefully stop tidb pod


tennix:
## Feature Request

When stopping a tidb pod, kubelet will first send `TERM` signal to tidb-server, tidb-server will wait for 15s before closing all SQL connections and then do all the cleanup. If it fails to finish all the cleanup, then kubelet will send `KILL` signal to tidb-server, and tidb-server will exit immediately.

When tidb-server fails to respond to the client, the client will timeout or fail, thus increasing latency.

We might consider using a readiness probe to let the load balancer (internal or external) to remove the backend first before making the server fail to respond to the client. This would help to reduce the latency during tidb pods shutdown.

The key point is letting readiness probe fail but still be able to receive requests.

ref: https://github.com/pingcap/tidb/blob/v4.0.1/server/server.go#L553-L557

zjj2wry:
maybe we can add a prestop lifeclycle to guarantee tidb-server have enough time to closing all SQL connections gracefully
```
lifeCycle:
    preStop:
      exec:
        command: ["/bin/bash", "-c", "sleep 15"]
```

cofyc:
> maybe we can add a prestop lifeclycle to guarantee tidb-server have enough time to closing all SQL connections gracefully
> 
> ```yaml
> lifeCycle:
>     preStop:
>       exec:
>         command: ["/bin/bash", "-c", "sleep 15"]
> ```

It seems a good solution. In addition to builtin 15 seconds wait time, users can configure extra wait time in `preStop` hook.

> The key point is letting readiness probe fail but still be able to receive requests.

When a pod is going to be deleted (DeletionTimestamp != nil),  its IP [is ignored in endpoints controller](https://github.com/kubernetes/kubernetes/blob/2bd9643cee5b3b3a5ecbd3af49d09018f0773c77/pkg/controller/endpoint/endpoints_controller.go#L419-L422) and will be completely removed from endpoints.

weekface:
When we start to update TiDB, the StatefulSet controller sends command to delete Pod. The pod is [removed](https://github.com/pingcap/tidb-operator/issues/2597#issuecomment-644533144) from [endpoints](https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods) list for service, and are no longer considered part of the set of running Pods for replication controllers. Then the LB(internal or external) can remove the backend.

Next, we should wait for the connections to become 0 and then stop the pod:

[...]

PR #2810, opened 2020-06-24
Add support to stop TiDB gracefully

weekface:

This pr add `terminationGracePeriodSeconds` and `lifecycle` to `TidbCluster` `spec.tidb` attribute.

User can use these two attributes to stop TiDB server gracefully:

1. set `spec.tidb.terminationGracePeriodSeconds` to a proper value, for exmaple: `600` (10 minutes).
2. set `spec.tidb.lifecycle` to: [`kill -QUIT 1`](https://github.com/pingcap/tidb/blob/8cbacf0d7c4612e6b5fa89836ca170af11eb81b2/util/signal/signal_posix.go#L54);

```
# IT IS NOT SUITABLE FOR PRODUCTION USE.
# This YAML describes a basic TiDB cluster with minimum resource requirements,
# which should be able to run in any Kubernetes cluster with storage support.
kind: TidbCluster
metadata:
  name: basic
spec:
  version: v4.0.0
  timezone: UTC
  pvReclaimPolicy: Delete
  discovery: {}
  pd:
    baseImage: pingcap/pd
    replicas: 1
    # if storageClassName is not set, the default Storage Class of the Kubernetes cluster will be used
    # storageClassName: local-storage
    requests:
      storage: "1Gi"
    config: {}
  tikv:
    baseImage: pingcap/tikv
    replicas: 1
    # if storageClassName is not set, the default Storage Class of the Kubernetes cluster will be used
    # storageClassName: local-storage
    requests:
      storage: "1Gi"
    config: {}
  tidb:
    baseImage: pingcap/tidb
    replicas: 1
    service:
      type: ClusterIP
    config: {}
    terminationGracePeriodSeconds: 600
    lifecycle:
      preStop:
        exec:
          command: ["kill", "-QUIT", "1"]
```

The TiDB server will go [graceful shutting down](https://github.com/pingcap/tidb/blob/v4.0.1/server/server.go#L572) to wait for all clients to close the connection until k8s force kill the TiDB container(10 minutes later).

fixes: https://github.com/pingcap/tidb-operator/issues/2597

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Support configuring container lifecycle hooks and `terminationGracePeriodSeconds` in TiDB spec
```


tennix:
Can we make sure that TiDB still receives connection before the LB removes the backend?

tennix:
In #2597, we want the tidb-server to continue to receive connections before it's removed from LB backend. For example, fail the readiness probe first but don't send `KILL` signal to tidb-server, so tidb-server can continue to receive connections, after 30s, send `KILL` signal to tidb-server and wait tidb-server gracefully terminate the existing connections.

weekface:
> Can we make sure that TiDB still receives connection before the LB removes the backend?

I think so.

When a pod is going to be deleted (DeletionTimestamp != nil), its IP is ignored in endpoints controller and will be completely removed from endpoints. ref: https://github.com/pingcap/tidb-operator/issues/2597#issuecomment-644533144

So we don't need `readiness probe`.

Before removed from endpoints, TiDB can receive new connections. After removed from endpoints, TiDB can't receive new connections, but the old connections work well as expected.

The `kill -QUIT 1` and a longer `terminationGracePeriodSeconds ` will wait for all these old connections closed by client.

DanielZhangQD:
> > Can we make sure that TiDB still receives connection before the LB removes the backend?
> 
> I think so.
> 
> When a pod is going to be deleted (DeletionTimestamp != nil), its IP is ignored in endpoints controller and will be completely removed from endpoints. ref: [#2597 (comment)](https://github.com/pingcap/tidb-operator/issues/2597#issuecomment-644533144)
> 
> So we don't need `readiness probe`.
> 
> Before removed from endpoints, TiDB can receive new connections. After removed from endpoints, TiDB can't receive new connections, but the old connections work well as expected.
> 
> The `kill -QUIT 1` and a longer `terminationGracePeriodSeconds ` will wait for all these old connections closed by client.

I think the LB that @tennix mentioned is the real load balancer, e.g. the NLB on AWS, it has its own probe for the target backend, which is also for port 4000 with our current configurations, however, with this PR, I think the new connections will still be sent to the TiDB Pod that is being terminated, which will cause request failure because the NLB will not remove it from its backend at that time.

I think the key point is:

1. The LB removes the Pod from its backend first
  1. Before that, the Pod can still handle new connections
  2. After that, no new connections will come to the Pod
2. The Pod can be terminated with SIGQUIT and the existing connections are handled gracefully

Is my understanding correct? @tennix

weekface:
> however, with this PR, I think the new connections will still be sent to the TiDB Pod that is being terminated, which will cause request failure because the NLB will not remove it from its backend at that time.

I don't know why this request failed? TiDB will not close the connection until the user closes the connection himself.

DanielZhangQD:
> > however, with this PR, I think the new connections will still be sent to the TiDB Pod that is being terminated, which will cause request failure because the NLB will not remove it from its backend at that time.
> 
> I don't know why this request failed? TiDB will not close the connection until the user closes the connection himself.

For this description, `Before removed from endpoints, TiDB can receive new connections. After removed from endpoints, TiDB can't receive new connections, but the old connections work well as expected.`, please verify when the pod is removed from the endpoint, will it also be removed from the load balancer backend?
If yes, I think this PR is good.
If no, the LB will still send new connection request to this Pod, but this Pod does not handle new connections at that time.

weekface:
> please verify when the pod is removed from the endpoint, will it also be removed from the load balancer backend?

About endpoints remove please this comment: https://github.com/pingcap/tidb-operator/issues/2597#issuecomment-644533144

We can't control the LB behavior in TiDB Operator, all we can do is wait the connections to close gracefully by client. 

> If no, the LB will still send new connection request to this Pod, but this Pod does not handle new connections at that time.

The request flow from LB to Pod is: LB -> NodePort -> Service -> Pod. After the pod was removed from Service, no new connection can be sent to the Pod anymore.

tennix:
The event happens in the following order:
1. Users or controller delete the pod
2. kube-apiserver adds a deletion marker on the pod object
3. kube-proxy removes the pod from endpoints and kubelet send `KILL` to tidb-server process (these two events happen in random order I guess)
4. tidb-server stops receiving new connections after it got `KILL` signal but still handles old connections, so if a new connection comes in, it will fail.

The node will be removed from the external LB if using `Local` external traffic policy. This is not guaranteed to happen before tidb-server stop receiving new connections. So the client will get timeout errors if there is no retry logic (for example some PHP frameworks).

cofyc:
as the tidb-server closes the listener immediately after receiving termination signals (including graceful QUIT), does it help to wait a few seconds before sending signals to the tidb-server process?

weekface:
> The node will be removed from the external LB if using `Local` external traffic policy. This is not guaranteed to happen before tidb-server stop receiving new connections. So the client will get timeout errors if there is no retry logic (for example some PHP frameworks).

For service that is `LoadBalacer` and `Local` external traffic policy, k8s will start a [ServiceHealth](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/proxy/healthcheck/service_health.go#L35) http server for each Service. The LB will health check this server to remove or add node to the external LB backends. This server will return a [`503`](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/proxy/healthcheck/service_health.go#L162) if the endpoints are empty on this node. Then the external LB can remove this node from LB backends after retrying a few times.

This is not guaranteed to happen before stopping tidb-server. But we can sleep a few seconds before send `QUIT` signal to tidb-server as @cofyc said to wait the external LB remove this node from its backends.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -557,6 +557,22 @@ type TiDBSpec struct {
...
+	// Set this value longer than the expected cleanup time for your process.
+	// Defaults to 30 seconds.
+	// +optional
+	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
+
+	// Lifecycle describes actions that the management system should take in response to container lifecycle
+	// events. For the PostStart and PreStop lifecycle handlers, management of the container blocks
+	// until the action is complete, unless the container process fails, in which case the handler is aborted.
+	// +optional
+	Lifecycle *corev1.Lifecycle `json:"lifecycle,omitempty"`
```
cofyc, line 575: how about moving these two fields to `ComponentSpec` struct? thought it may not normally be necessary to configure these fields for other components, it would be strange for users that only TiDB has these pod fields.

weekface: good idea

pkg/manager/member/tidb_member_manager.go
```diff
@@ -759,6 +764,11 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 			},
 		},
 	}
+
+	if tc.Spec.TiDB.TerminationGracePeriodSeconds != nil {
+		tidbSet.Spec.Template.Spec.TerminationGracePeriodSeconds = tc.Spec.TiDB.TerminationGracePeriodSeconds
+	}
```
cofyc, line 770: didn't update other components' controller logic
implement in a separate PR?

cofyc:
I suggest changing the release note to a more user faced one: `Support configuring container lifecycle hooks and `terminationGracePeriodSeconds` in TiDB spec`

stopping TiDB gracefully is the main goal, but it's not the actual change in this PR. any actions can be injected to achieve different purposes.

pkg/manager/member/tidb_member_manager.go
weekface: ok, https://github.com/pingcap/tidb-operator/issues/2847

PR #2825, opened 2020-06-28
Skip duplicated e2e case

Yisaer:
Skip the e2e case `upgrading tidb-operator in the same minor series should not trigger rolling-update` as it have already been covered by the `[Verify: Upgrading Operator from 1.1.0]`

Related changes

```release-note
NONE
```


tests/e2e/tidbcluster/serial.go
```diff
@@ -1033,6 +1033,7 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
 		})
 
 		ginkgo.It("basic deployment", func() {
+			framework.Skipf("duplicated test")
```
cofyc, line 1036: if it's duplicated, remove it?

PR #2846, opened 2020-07-01
Make pod webhook able to manage tikv group pod

Yisaer:
Make Pod Webhook could manage both tikvgroup Pod and tidbcluster tikv Pod.

```release-note
NONE
```

### DIFF
--- a/cmd/admission-webhook/main.go
+++ b/cmd/admission-webhook/main.go
@@ -32,6 +32,7 @@ var (
 	printVersion             bool
 	extraServiceAccounts     string
 	evictRegionLeaderTimeout time.Duration
+	resyncDuration           time.Duration
 )
 
 func init() {
@@ -39,6 +40,7 @@ func init() {
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "comma-separated, extra Service Accounts the Webhook should control. The full pattern for each common service account is system:serviceaccount:<namespace>:<serviceaccount-name>")
 	flag.DurationVar(&evictRegionLeaderTimeout, "evictRegionLeaderTimeout", 3*time.Minute, "TiKV evict region leader timeout period, default 3 min")
+	flag.DurationVar(&resyncDuration, "resync-duration", 30*time.Second, "Resync time of informer")
 	features.DefaultFeatureGate.AddFlag(flag.CommandLine)
 }
 
@@ -60,6 +62,7 @@ func main() {
 	ah := &webhook.AdmissionHook{
 		ExtraServiceAccounts:     extraServiceAccounts,
 		EvictRegionLeaderTimeout: evictRegionLeaderTimeout,
+		ResyncDuration:           resyncDuration,
 	}
 	ns := os.Getenv("NAMESPACE")
 	if len(ns) < 1 {
--- a/cmd/admission-webhook/main.go
+++ b/cmd/admission-webhook/main.go
@@ -16,6 +16,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"math/rand"
 	"os"
 	"time"
 
@@ -32,7 +33,7 @@ var (
 	printVersion             bool
 	extraServiceAccounts     string
 	evictRegionLeaderTimeout time.Duration
-	resyncDuration           time.Duration
+	minResyncDuration        time.Duration
 )
 
 func init() {
@@ -40,7 +41,7 @@ func init() {
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "comma-separated, extra Service Accounts the Webhook should control. The full pattern for each common service account is system:serviceaccount:<namespace>:<serviceaccount-name>")
 	flag.DurationVar(&evictRegionLeaderTimeout, "evictRegionLeaderTimeout", 3*time.Minute, "TiKV evict region leader timeout period, default 3 min")
-	flag.DurationVar(&resyncDuration, "resync-duration", 12*time.Hour, "Resync time of informer")
+	flag.DurationVar(&minResyncDuration, "min-resync-duration", 12*time.Hour, "The resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod.")
 	features.DefaultFeatureGate.AddFlag(flag.CommandLine)
 }
 
@@ -58,6 +59,9 @@ func main() {
 	flag.CommandLine.VisitAll(func(flag *flag.Flag) {
 		klog.V(1).Infof("FLAG: --%s=%q", flag.Name, flag.Value)
 	})
+	// We choose a random resync period between MinResyncPeriod and 2 *
+	// MinResyncPeriod, so that our pods started at the same time don't list the apiserver simultaneously.
+	resyncDuration := time.Duration(minResyncDuration.Seconds()*(1+rand.Float64())) * time.Second
 
 	ah := &webhook.AdmissionHook{
 		ExtraServiceAccounts:     extraServiceAccounts,
--- a/cmd/admission-webhook/main.go
+++ b/cmd/admission-webhook/main.go
@@ -40,7 +40,7 @@ func init() {
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "comma-separated, extra Service Accounts the Webhook should control. The full pattern for each common service account is system:serviceaccount:<namespace>:<serviceaccount-name>")
 	flag.DurationVar(&evictRegionLeaderTimeout, "evictRegionLeaderTimeout", 3*time.Minute, "TiKV evict region leader timeout period, default 3 min")
-	flag.DurationVar(&resyncDuration, "resync-duration", 30*time.Second, "Resync time of informer")
+	flag.DurationVar(&resyncDuration, "resync-duration", 12*time.Hour, "Resync time of informer")
 	features.DefaultFeatureGate.AddFlag(flag.CommandLine)
 }
 
--- a/pkg/webhook/admission_hooks.go
+++ b/pkg/webhook/admission_hooks.go
@@ -103,7 +103,7 @@ func (a *AdmissionHook) Admit(ar *admission.AdmissionRequest) *admission.Admissi
 	name := ar.Name
 	namespace := ar.Namespace
 	kind := ar.Kind.Kind
-	klog.Infof("receive mutation request for %s[%s/%s]", kind, namespace, name)
+	klog.V(4).Infof("receive mutation request for %s[%s/%s]", kind, namespace, name)
 
 	resp := a.strategyAC.Mutate(ar)
 	if !resp.Allowed {
--- a/pkg/webhook/admission_hooks.go
+++ b/pkg/webhook/admission_hooks.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/pingcap/advanced-statefulset/client/apis/apps/v1/helper"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+	informers "github.com/pingcap/tidb-operator/pkg/client/informers/externalversions"
 	"github.com/pingcap/tidb-operator/pkg/features"
 	"github.com/pingcap/tidb-operator/pkg/pdapi"
 	"github.com/pingcap/tidb-operator/pkg/webhook/pod"
@@ -46,6 +47,7 @@ type AdmissionHook struct {
 	podAC                    *pod.PodAdmissionControl
 	stsAC                    *statefulset.StatefulSetAdmissionControl
 	strategyAC               *strategy.AdmissionWebhook
+	ResyncDuration           time.Duration
 	ExtraServiceAccounts     string
 	EvictRegionLeaderTimeout time.Duration
 }
@@ -160,7 +162,11 @@ func (a *AdmissionHook) Initialize(cfg *rest.Config, stopCh <-chan struct{}) err
 		Interface: eventv1.New(kubeCli.CoreV1().RESTClient()).Events("")})
 	recorder := eventBroadcaster.NewRecorder(v1alpha1.Scheme, corev1.EventSource{Component: "tidb-admission-controller"})
 
-	pc := pod.NewPodAdmissionControl(kubeCli, cli, pdControl, strings.Split(a.ExtraServiceAccounts, ","), a.EvictRegionLeaderTimeout, recorder)
+	var informerFactory informers.SharedInformerFactory
+	var options []informers.SharedInformerOption
+	informerFactory = informers.NewSharedInformerFactoryWithOptions(cli, a.ResyncDuration, options...)
+
+	pc := pod.NewPodAdmissionControl(kubeCli, cli, pdControl, strings.Split(a.ExtraServiceAccounts, ","), a.EvictRegionLeaderTimeout, informerFactory, recorder)
 	a.podAC = pc
 	klog.Info("pod admission webhook initialized successfully")
 	a.stsAC = statefulset.NewStatefulSetAdmissionControl(cli)
--- a/pkg/webhook/admission_hooks.go
+++ b/pkg/webhook/admission_hooks.go
@@ -167,6 +167,7 @@ func (a *AdmissionHook) Initialize(cfg *rest.Config, stopCh <-chan struct{}) err
 	informerFactory = informers.NewSharedInformerFactoryWithOptions(cli, a.ResyncDuration, options...)
 
 	pc := pod.NewPodAdmissionControl(kubeCli, cli, pdControl, strings.Split(a.ExtraServiceAccounts, ","), a.EvictRegionLeaderTimeout, informerFactory, recorder)
+	informerFactory.Start(stopCh)
 	a.podAC = pc
 	klog.Info("pod admission webhook initialized successfully")
 	a.stsAC = statefulset.NewStatefulSetAdmissionControl(cli)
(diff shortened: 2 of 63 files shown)


pkg/webhook/pod/tikv_deleter.go
```diff
@@ -76,24 +70,29 @@ func (pc *PodAdmissionControl) admitDeleteTiKVPods(payload *admitPayload) *admis
...
-			klog.Infof("tc[%s/%s]'s tikv pod[%s/%s] failed to delete,%v", namespace, tcName, namespace, name, err)
-			return util.ARFail(err)
+	switch controllerKind {
+	case v1alpha1.TiDBClusterKind:
+		expectedAddress = fmt.Sprintf("%s.%s-tikv-peer.%s.svc:20160", name, controllerName, namespace)
+	case v1alpha1.TiKVGroupKind:
+		expectedAddress = fmt.Sprintf("%s.%s-tikv-group-peer.%s.svc:20160", name, controllerName, namespace)
+	default:
+		err := fmt.Errorf("tikv pod[%s/%s] controlled by unknown controllerKind[%s], forbid to delete", namespace, name, controllerKind)
+		return util.ARFail(err)
```
cofyc, line 82: this is unreachable because all possible values of `controllerKind` must be handled in this function
you can add a comment `// unreachable` or just use `panic("unreachable")`

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -127,24 +125,25 @@ func (pc *PodAdmissionControl) admitDeleteUselessTiKVPod(payload *admitPayload)
...
 
 	if !isInOrdinal {
 		pvcName := operatorUtils.OrdinalPVCName(v1alpha1.TiKVMemberType, payload.ownerStatefulSet.Name, ordinal)
 		pvc, err := pc.kubeCli.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, meta.GetOptions{})
 		if err != nil {
 			if errors.IsNotFound(err) {
-				pc.recorder.Event(payload.tc, corev1.EventTypeNormal, tikvScaleInReason, podDeleteEventMessage(name))
-				return util.ARSuccess()
+				pc.recorder.Event(payload.controller, corev1.EventTypeNormal, tikvScaleInReason, podDeleteEventMessage(name))
+				return util.ARFail(err)
```
cofyc, line 137: why rejecting here?

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -168,24 +166,44 @@ func (pc *PodAdmissionControl) admitDeleteUpTiKVPod(payload *admitPayload, store
...
+		}
+		specReplicas = tc.Spec.TiKV.Replicas
+	} else if controllerKind == v1alpha1.TiKVGroupKind {
+		tg, ok := payload.controller.(*v1alpha1.TiKVGroup)
+		if !ok {
+			err := fmt.Errorf("tikv pod[%s/%s]'s controller is not tikvgroup,forbid to be deleted", namespace, name)
+			return util.ARFail(err)
+		}
+		specReplicas = tg.Spec.Replicas
+	} else {
```
cofyc, line 198: ditto

pkg/webhook/pod/tikv_util.go
```diff
@@ -29,38 +29,52 @@ import (
...
 		}
-		podName := memberUtil.TikvPodName(tcName, i)
+		var podName string
+		switch controllerKind {
+		case v1alpha1.TiDBClusterKind:
+			podName = memberUtil.TikvPodName(controllerName, i)
+		case v1alpha1.TiKVGroupKind:
+			podName = memberUtil.TiKVGroupPodName(controllerName, i)
+		default:
+			return fmt.Errorf("")
```
cofyc, line 57: `controllerDesc.kind` is set in the code and has only two possible values, so this is unreachable too. add an unreachable comment or use `panic("unreachable")`, then people don't need to consider this part of code as it's unreachable

pkg/webhook/pod/tikv_deleter.go
Yisaer: updated.

pkg/webhook/admission_hooks.go
```diff
@@ -103,7 +103,7 @@ func (a *AdmissionHook) Admit(ar *admission.AdmissionRequest) *admission.Admissi
 	name := ar.Name
 	namespace := ar.Namespace
 	kind := ar.Kind.Kind
-	klog.Infof("receive mutation request for %s[%s/%s]", kind, namespace, name)
+	klog.V(4).Infof("receive mutation request for %s[%s/%s]", kind, namespace, name)
```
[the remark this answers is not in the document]
Yisaer: updated.

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -76,24 +70,30 @@ func (pc *PodAdmissionControl) admitDeleteTiKVPods(payload *admitPayload) *admis
...
-			return util.ARFail(err)
+	switch controllerKind {
+	case v1alpha1.TiDBClusterKind:
+		expectedAddress = fmt.Sprintf("%s.%s-tikv-peer.%s.svc:20160", name, controllerName, namespace)
+	case v1alpha1.TiKVGroupKind:
+		expectedAddress = fmt.Sprintf("%s.%s-tikv-group-peer.%s.svc:20160", name, controllerName, namespace)
+	default:
+		// unreachable
+		err := fmt.Errorf("tikv pod[%s/%s] controlled by unknown controllerKind[%s], forbid to delete", namespace, name, controllerKind)
+		return util.ARFail(err)
```
DanielZhangQD, line 83: If not controlled by the controllers we want to handle, I think we should admit it.

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -168,24 +167,45 @@ func (pc *PodAdmissionControl) admitDeleteUpTiKVPod(payload *admitPayload, store
...
+		tg, ok := payload.controller.(*v1alpha1.TiKVGroup)
+		if !ok {
+			err := fmt.Errorf("tikv pod[%s/%s]'s controller is not tikvgroup,forbid to be deleted", namespace, name)
+			return util.ARFail(err)
+		}
+		specReplicas = tg.Spec.Replicas
+	} else {
+		// unreachable
+		err := fmt.Errorf("tikv pod[%s/%s] has unknown controller[%s], forbid to be deleted", namespace, name, controllerKind)
+		return util.ARFail(err)
```
DanielZhangQD, line 202: ditto

pkg/webhook/pod/tikv_deleter.go
Yisaer: updated

pkg/webhook/pod/pd_deleter.go
```diff
@@ -35,6 +37,11 @@ func (pc *PodAdmissionControl) admitDeletePdPods(payload *admitPayload) *admissi
 	if err != nil {
 		return util.ARFail(err)
 	}
+	tc, ok := payload.controller.(*v1alpha1.TidbCluster)
+	if !ok {
+		err := fmt.Errorf("pd pod[%s/%s]'s controller is not tidbcluster, forbit to be deleted", namespace, name)
+		return util.ARFail(err)
```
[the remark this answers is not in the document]

pkg/webhook/pod/tikv_deleter.go
cofyc: this is an unreachable branch, the logic does not matter actually, a common practice is to panic explicitly or clarify in the comment. of course, we should remove these in code refactoring as possible as we can.

pkg/webhook/pod/pd_deleter.go
```diff
@@ -116,7 +123,12 @@ func (pc *PodAdmissionControl) admitDeleteNonPDMemberPod(payload *admitPayload)
 	if err != nil {
 		return util.ARFail(err)
 	}
-	tcName := payload.tc.Name
+	tc, ok := payload.controller.(*v1alpha1.TidbCluster)
+	if !ok {
+		err := fmt.Errorf("pd pod[%s/%s]'s controller is not tidbcluster, forbid to be deleted", namespace, name)
+		return util.ARFail(err)
```
DanielZhangQD, line 129: Change to admit?

pkg/webhook/pod/pd_deleter.go
```diff
@@ -166,7 +178,12 @@ func (pc *PodAdmissionControl) admitDeleteExceedReplicasPDPod(payload *admitPayl
 
 	name := payload.pod.Name
 	namespace := payload.pod.Namespace
-	tcName := payload.tc.Name
+	tc, ok := payload.controller.(*v1alpha1.TidbCluster)
+	if !ok {
+		err := fmt.Errorf("pd pod[%s/%s]'s controller is not tidbcluster, forbid to be deleted", namespace, name)
+		return util.ARFail(err)
```
DanielZhangQD, line 184: ditto

pkg/webhook/pod/pd_deleter.go
```diff
@@ -197,7 +214,12 @@ func (pc *PodAdmissionControl) transferPDLeader(payload *admitPayload) *admissio
 	if err != nil {
 		return util.ARFail(err)
 	}
-	tcName := payload.tc.Name
+	tc, ok := payload.controller.(*v1alpha1.TidbCluster)
+	if !ok {
+		err := fmt.Errorf("pd pod[%s/%s]'s controller is not tidbcluster, forbid to be deleted", namespace, name)
+		return util.ARFail(err)
```
DanielZhangQD, line 220: ditto

pkg/webhook/pod/pods.go
```diff
@@ -169,44 +202,82 @@ func (pc *PodAdmissionControl) admitDeletePods(name, namespace string) *admissio
...
+			klog.Errorf("failed get tikvgroup[%s/%s],refuse to delete pod[%s/%s]", namespace, tgName, namespace, name)
+			return util.ARFail(err)
+		}
+		ownerTcName := tg.Spec.ClusterName
+		tc, err := pc.operatorCli.PingcapV1alpha1().TidbClusters(namespace).Get(ownerTcName, metav1.GetOptions{})
+		if err != nil {
+			// Event if the ownerTC is deleted, we won't delete the tikvgroup pod unless its owner controller is deleted
+			klog.Errorf("failed get tc[%s/%s],refuse to delete pod[%s/%s]", namespace, ownerTcName, namespace, name)
+			return util.ARFail(err)
+		}
```
DanielZhangQD, line 269: If we do not care about tc, why do we check the tc here? Is it OK to remove this block?

pkg/webhook/pod/pods.go
```diff
@@ -169,44 +202,82 @@ func (pc *PodAdmissionControl) admitDeletePods(name, namespace string) *admissio
...
+		payload.pdClient = pc.pdControl.GetPDClient(pdapi.Namespace(namespace), ownerTcName, tc.IsTLSClusterEnabled())
+		payload.controller = tc
```
DanielZhangQD, line 271: The controller should be `tg`?

pkg/webhook/pod/pods.go
```diff
@@ -237,28 +308,42 @@ func (pc *PodAdmissionControl) AdmitCreatePods(ar *admission.AdmissionRequest) *
...
+		if err != nil {
+			klog.Errorf("failed get tikvgroup[%s/%s],refuse to create pod[%s/%s],%v", namespace, tgName, namespace, name, err)
+			return util.ARFail(err)
+		}
+		tcName := tg.Spec.ClusterName
+		ownerTc, err = pc.operatorCli.PingcapV1alpha1().TidbClusters(namespace).Get(tcName, metav1.GetOptions{})
+		if err != nil {
+			klog.Errorf("failed get tc[%s/%s],refuse to create pod[%s/%s],%v", namespace, tcName, namespace, name, err)
+			return util.ARFail(err)
+		}
```
DanielZhangQD, line 338: I think there is no need to check tc for the pod of tikvgroup creation, no need to mix them up, even if tc is good here and the tikvs are created, it is still possible that at some time the PD cluster is down, which is similar with the case that we start the tikvs here even if we get tc failed.

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -76,24 +70,30 @@ func (pc *PodAdmissionControl) admitDeleteTiKVPods(payload *admitPayload) *admis
...
-			klog.Infof("tc[%s/%s]'s tikv pod[%s/%s] failed to delete,%v", namespace, tcName, namespace, name, err)
-			return util.ARFail(err)
+	switch controllerKind {
+	case v1alpha1.TiDBClusterKind:
+		expectedAddress = fmt.Sprintf("%s.%s-tikv-peer.%s.svc:20160", name, controllerName, namespace)
+	case v1alpha1.TiKVGroupKind:
+		expectedAddress = fmt.Sprintf("%s.%s-tikv-group-peer.%s.svc:20160", name, controllerName, namespace)
+	default:
+		// unreachable
+		klog.V(4).Infof("tikv pod[%s/%s] controlled by unknown controllerKind[%s], forbid to delete", namespace, name, controllerKind)
```
DanielZhangQD, line 82:
```suggestion
		klog.V(4).Infof("tikv pod[%s/%s] controlled by unknown controllerKind[%s], allow to delete", namespace, name, controllerKind)
```

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -168,24 +167,45 @@ func (pc *PodAdmissionControl) admitDeleteUpTiKVPod(payload *admitPayload, store
...
+	} else if controllerKind == v1alpha1.TiKVGroupKind {
+		tg, ok := payload.controller.(*v1alpha1.TiKVGroup)
+		if !ok {
+			err := fmt.Errorf("tikv pod[%s/%s]'s controller is not tikvgroup,forbid to be deleted", namespace, name)
+			return util.ARFail(err)
+		}
+		specReplicas = tg.Spec.Replicas
+	} else {
+		// unreachable
+		klog.V(4).Infof("tikv pod[%s/%s] has unknown controller[%s], forbid to be deleted", namespace, name, controllerKind)
```
DanielZhangQD, line 201:
```suggestion
		klog.V(4).Infof("tikv pod[%s/%s] has unknown controller[%s], allow to be deleted", namespace, name, controllerKind)
```

pkg/webhook/pod/pods.go
Yisaer: I think we can't. we need to fetch some cluster level information in tidbcluster spec.

pkg/webhook/pod/pods.go
Yisaer: We didn't "check" tc here. we are going to fetch cluster level spec information here.

pkg/webhook/pod/pd_deleter.go

pkg/webhook/pod/pods.go
```diff
@@ -237,28 +308,42 @@ func (pc *PodAdmissionControl) AdmitCreatePods(ar *admission.AdmissionRequest) *
...
+		if !exist {
 			return util.ARSuccess()
 		}
-		klog.Errorf("failed get tc[%s/%s],refuse to create pod[%s/%s],%v", namespace, tcName, namespace, name, err)
-		return util.ARFail(err)
-	}
-
-	if memberUtils.NeedForceUpgrade(tc) {
-		klog.Infof("tc[%s/%s] is force upgraded, admit to create pod[%s/%s]", namespace, tcName, namespace, name)
+		ownerTc, err = pc.operatorCli.PingcapV1alpha1().TidbClusters(namespace).Get(tcName, metav1.GetOptions{})
```
DanielZhangQD, line 318: Suggest retrieving from lister.

Yisaer: updated

cmd/admission-webhook/main.go
```diff
@@ -32,13 +32,15 @@ var (
...
 	evictRegionLeaderTimeout time.Duration
+	resyncDuration           time.Duration
 )
 
 func init() {
 	flag.BoolVar(&printVersion, "V", false, "Show version and quit")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "comma-separated, extra Service Accounts the Webhook should control. The full pattern for each common service account is system:serviceaccount:<namespace>:<serviceaccount-name>")
 	flag.DurationVar(&evictRegionLeaderTimeout, "evictRegionLeaderTimeout", 3*time.Minute, "TiKV evict region leader timeout period, default 3 min")
+	flag.DurationVar(&resyncDuration, "resync-duration", 30*time.Second, "Resync time of informer")
```
cofyc, line 43: use a long period, e.g. 12 hours

image

Yisaer: updated.

cmd/admission-webhook/main.go
```diff
@@ -32,13 +32,15 @@ var (
...
 	evictRegionLeaderTimeout time.Duration
+	resyncDuration           time.Duration
 )
 
 func init() {
 	flag.BoolVar(&printVersion, "V", false, "Show version and quit")
 	flag.BoolVar(&printVersion, "version", false, "Show version and quit")
 	flag.StringVar(&extraServiceAccounts, "extraServiceAccounts", "", "comma-separated, extra Service Accounts the Webhook should control. The full pattern for each common service account is system:serviceaccount:<namespace>:<serviceaccount-name>")
 	flag.DurationVar(&evictRegionLeaderTimeout, "evictRegionLeaderTimeout", 3*time.Minute, "TiKV evict region leader timeout period, default 3 min")
+	flag.DurationVar(&resyncDuration, "resync-duration", 12*time.Hour, "Resync time of informer")
```
cofyc, line 43:
```suggestion
	flag.DurationVar(&minResyncDuration, "min-resync-duration", 12*time.Hour, "The resync period in reflectors will be random between MinResyncPeriod and 2*MinResyncPeriod.")
```

and add this to choose  a random resync period:

```
	// We choose a random resync period between MinResyncPeriod and 2 *
	// MinResyncPeriod, so that our pods starting at the same time don't list the apiserver simultaneously.
	resyncPeriod := time.Duration(minResyncPeriod.Seconds()*(1+rand.Float64())) * time.Second
```

cofyc: This is a common practice to avoid generate the load simultaneously.

Issue #1936, opened 2020-03-16
UCP: Better controller error message


aylei:
## Feature Request

If the control-loop returns an error, the only log is:

```
E0316 06:57:23.127936       1 tidb_cluster_controller.go:275] TidbCluster: tiflash31003/tiflash31003-cluster, sync failed pod "tiflash-0" not found, requeuing
```

There's no stack of other information that shows where the error happens, and this is bad for diagnosis.

## What do you want?

Better error msg with more context informations.

## Score

* 300

## Mentor(s)

* [@aylei ](https://github.com/aylei)

Contact the mentors: **#tidb-challenge-program** channel in [TiDB-Operator Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang
* Kubernetes

aylei:
Use `pingcap/errors.Annotatef` should help

PR #2851, opened 2020-07-01
add more error info for controllers

lichunzhu:
Fix https://github.com/pingcap/tidb-operator/issues/1936

Replace all `fmt.Errorf` in `utilruntime.HandleError` with `pingacp/errors.Errorf`.

### Check List 

Tests 

 - Unit test
 - E2E test

Related changes

```release-note
NONE
```


lichunzhu:
@aylei Does this PR meet your expectations?

cofyc:
can you paste some examples?

lichunzhu:
The error stack info will be shown like this:
```log
E0702 03:52:43.608254       1 tidb_cluster_controller.go:300] TidbCluster: default/basic, sync failed TidbCluster: [default/basic], waiting for TiKV cluster running, requeuing
github.com/pingcap/tidb-operator/pkg/controller/tidbcluster.(*Controller).processNextWorkItem
        github.com/pingcap/tidb-operator@/pkg/controller/tidbcluster/tidb_cluster_controller.go:300
github.com/pingcap/tidb-operator/pkg/controller/tidbcluster.(*Controller).worker
        github.com/pingcap/tidb-operator@/pkg/controller/tidbcluster/tidb_cluster_controller.go:283
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
        k8s.io/apimachinery@v0.0.0/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
        k8s.io/apimachinery@v0.0.0/pkg/util/wait/wait.go:153
k8s.io/apimachinery/pkg/util/wait.Until
        k8s.io/apimachinery@v0.0.0/pkg/util/wait/wait.go:88
runtime.goexit
        runtime/asm_amd64.s:1357
I0702 03:52:43.743274       1 tidb_cluster_controller.go:313] Finished syncing TidbCluster "default/basic" (130.691ms)
I0702 03:52:43.743326       1 tidb_cluster_controller.go:298] TidbCluster: default/basic, still need sync: TidbCluster: [default/basic], waiting for TiKV cluster running, requeuing
```

cofyc:
the stack shows the stack of function calls, but it's not useful in our scenarios because it does not show where the error happens (see the goal of the issue #1936).
I think it's not possible to know where the error happens unless the returned error contains the context information.
Instead of using the `pingacp/errors.Errorf` function, maybe we should improve the error message in the controller, e.g. adding the function prefix (`xxxx: failed to ....`)

cofyc:
#1936 does have specific goals, we can improve the error messages in our core logic first:

- pd_member_manager.go
- tikv_member_manager.go
- tidb_member_manager.go
- tiflash_member_manager.go

For example:

the error in https://github.com/pingcap/tidb-operator/issues/1936#issue-582115776 may be returned here:

https://github.com/pingcap/tidb-operator/blob/60cf7d493eb056fd7668a042f46ee7b974b73480/pkg/manager/member/tiflash_member_manager.go#L784-L787

the raw error returned by `client-go` is propagated to the outside controller and printed. we don't know where the error is generated, however, we can change this code to

```
        pod, err := tfmm.podLister.Pods(ns).Get(podName)
        if err != nil {
            return setCount, fmt.Errorf("setStoreLabelsForTiFlash: failed to get pod for store %d, error: %v ", status.ID, err)
        }
```

aylei:
Can we annotate the error at the first place (the bottom of error stack in our code), and print the stack trace in the `processNextWorkItem` logic?

cofyc:
> Can we annotate the error at the first place (the bottom of error stack in our code), and print the stack trace in the `processNextWorkItem` logic?

this will work but the stack spans multiple lines and errors are common in our operator. I'm afraid that might be a lot of log output.

cofyc:
To sum up, there are two options:

1. add a specific descriptive message and an optional prefix in the error message, then users can know where the error is generated by searching the message/function in the code
2. annotate with the stack information when the error is generated, then users can know where the error is generated from the stack

I would prefer option 1 for most errors as it prints only one line message. Because the reconciliation controller retries on the error, and API errors or some objects are not synced yet are common in the operator.

option 2 is suitable for fatal errors and the trace in the stack is important in diagnosing the logic bug.

What do you think?

aylei:
> To sum up, there are two options:
> 
> 1. add a specific descriptive message and an optional function prefix in the error message, then users can know where the error is generated by searching the message/function in the code
> 2. annotate with the stack information when the error is generated, then users can know where the error is generated from the stack
> 
> I would prefer option 1 for most errors in the operator as it prints only one line message. Because the reconciliation controller retries on the error, and API errors or some objects are not synced yet are common in the operator.
> 
> option 2 is suitable for fatal errors and when the trace in the stack is important in diagnosing the logic bug.
> 
> What do you think?

Thanks for the conclusion. IMO option 1 would be generally better.

cofyc:
@lichunzhu can you update this PR to use option 1?

lichunzhu:
> @lichunzhu can you update this PR to use option 1?

OK

cofyc:
cc @aylei @Yisaer @DanielZhangQD @weekface PTAL

cofyc:
@lichunzhu can you also update the title of the PR?

lichunzhu:
emmm, I have more info for this PR, I will open another one.

PR #2857, opened 2020-07-02
fix drainer initialCommitTs issue

weekface:

`initialCommitTs` should be a `string`, itherwise failed,  for example: `initialCommitTs: 417774109500375041` will be parsed as:

image

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Fix drainer values.yaml `initialCommitTs` bug
```


charts/tidb-drainer/values.yaml
```diff
@@ -28,7 +28,7 @@ storage: 10Gi
 # disbale detect causality
 disableDetect: false
 # if drainer donesn't have checkpoint, use initial commitTS to initial checkpoint
-initialCommitTs: 0
+initialCommitTs: "0"
```
DanielZhangQD, line 31: From the comment of drainer dev, operator should not set the default value or set it to "-1", which is the default value in drainer, so we can set it to `# initialCommitTs: "-1"`?
BTW, the `_start_drainer.sh.tpl` also needs to update?

weekface: _start_drainer.sh.tpl don't need to update.

weekface, line 31:
```suggestion
initialCommitTs: "-1"
```

Do you means change it to "-1"? @DanielZhangQD

Yisaer: actuallly

DanielZhangQD: We'd better set the default value in `_start_drainer.sh.tpl` to `-1`.

DanielZhangQD:
@cofyc PTAL.

PR #2862, opened 2020-07-03
Refactor TiKV auto-scaling

Yisaer:

1. refactor the tikv auto-scaling controller to be ready for the storage auto-scaling
2. deprecated the de-nosie strategy due to it make auto-scaling in operator too complicated, and we will finally let pd to handle the auto-scaling strategy layer, and auto-scaler controller become the execution layer.

Related changes

```release-note
Remove  `readyToScaleThresholdSeconds` in `TidbClusterAutoScaler`, Operator won't support de-noise in `TidbClusterAutoScaler`
```


cofyc:
Suggested release notes: `Deprecate `readyToScaleThresholdSeconds` in `TidbClusterAutoScaler` as the operator will no longer support de-noise strategy`

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -85,6 +85,7 @@ type TidbClusterAutoScalerSpec struct {
 type TikvAutoScalerSpec struct {
 	BasicAutoScalerSpec `json:",inline"`
 
+	// Deprecated
```
cofyc, line 88: in which version you plan to remove these fields? you can add a note about this, e.g. `Deprecated in v1.1.3, planned for removal in v1.x.`

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -183,6 +184,7 @@ type TikvAutoScalerStatus struct {
 // +k8s:openapi-gen=true
 // BasicAutoScalerStatus describe the basic auto-scaling status
 type BasicAutoScalerStatus struct {
+	// Deprecated
```
cofyc, line 187: ditto

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -193,6 +195,7 @@ type BasicAutoScalerStatus struct {
 	// RecommendedReplicas describes the calculated replicas in the last auto-scaling reconciliation for the component(tidb/tikv)
 	// +optional
 	RecommendedReplicas int32 `json:"recommendedReplicas,omitempty"`
+	// Deprecated
```
cofyc, line 198: ditto

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
Yisaer: updated

pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go
```diff
@@ -85,6 +85,7 @@ type TidbClusterAutoScalerSpec struct {
 type TikvAutoScalerSpec struct {
 	BasicAutoScalerSpec `json:",inline"`
 
+	// Deprecated in v1.1.3, planned for removal in v1.3
```
DanielZhangQD, line 88: Why remove this?
It's not reasonable to scale as soon as the metric exceeds the threshold.
BTW, even if the PD will handle the metrics later, we can remove this when the similar function is implemented in PD. WDYT @cofyc

cofyc: I believe this is in @Yisaer's plan, replaced by PD before v1.3.

Yisaer: I think the current de-noise design in operator is too complicated and haven't been verified whether it is necessary. If we decided let PD make the auto-scaling plan, I think operator could only expose api and do the execution. That will be quite simple and clear, easy to know the status and debug.

cofyc:
please update the release note to reflect the latest change

Issue #2436, opened 2020-05-11
Supporting dumpling in Backup CRD


kolbe:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Support of `mydumper` is being deprecated in favor of dumpling (https://github.com/pingcap/dumpling).

The Backup CRD should support dumpling (maybe instead of mydumper?).

DanielZhangQD:
dumpling will be GAed in v4.0.2.

PR #2870, opened 2020-07-06
Replace mydumper with dumpling in tidbBackupManager

lichunzhu:
Resolve issue https://github.com/pingcap/tidb-operator/issues/2436

- Replace mydumper image with dumpling.
- Delete mydumper related configs.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
  * Test upgrade from old `back-schedule` job. After the image is updated, the regular job can still work correctly, but won't use the options from `mydumper`. Besides, the pod will keep logging the infomation below:
```log
I0707 07:48:48.351964       1 manager.go:234] set cluster demo/basic-backup-2020-07-07t07-48-00 tikv_gc_life_time to 72h success
W0707 07:48:48.352683       1 export.go:69] deprecated mydumper config {"tableRegex":"^(?!(mysql|test|INFORMATION_SCHEMA|PERFORMANCE_SCHEMA|METRICS_SCHEMA|INSPECTION_SCHEMA|lcz))"} which won't take effect is still in use. please transfer to dumpling config
I0707 07:48:48.352759       1 export.go:82] the dump process is ready, command "/dumpling --output=/backup/chunzhu/backup-2020-07-07T07:48:48Z --host=basic-tidb --port=4000 --user=root --password= --threads=16 --rows=10000 --filter *.* --filter !/^(mysql|test|INFORMATION_SCHEMA|PERFORMANCE_SCHEMA|METRICS_SCHEMA|INSPECTION_SCHEMA)$/.*"
I0707 07:48:49.139288       1 manager.go:252] reset cluster demo/basic-backup-2020-07-07t07-48-00 tikv_gc_life_time to 10m0s success
```
  * Test dump with dumpling's filter and options, it can work normally.
```yaml
spec:
  ...
  dumpling:
    options:
    - --threads=16
    - --rows=10000
    tableFilter:
    - "pingcap.*"
  ...
```

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Replace `mydumper` with `dumpling` for backup.
ACTION REQUIRED: If `spec.mydumper` is configured in the Backup CR or `spec.backupTemplate.mydumper` is configured in the BackupSchedule CR, please migrate it to `spec.dumpling` or `spec.backupTemplate.dumpling` after you upgrade TiDB Operator to v1.1.3, please note that the `spec.mydumper` or `spec.backupTemplate.mydumper` will be lost after the upgrade.
```


cmd/backup-manager/app/export/export.go
```diff
@@ -62,18 +62,21 @@ func (bo *Options) dumpTidbClusterData(backup *v1alpha1.Backup) (string, error)
 	if err != nil {
 		return "", err
 	}
+	if backup.Spec.Mydumper.TableRegex != nil && len(*backup.Spec.Mydumper.TableRegex) > 0 {
+		return "", fmt.Errorf("dumpling doesn't support tableRegex option, please use tableFilter instead")
+	}
```
lichunzhu, line 67: Should we delete `TableRegex` directly?

charts/tidb-cluster/templates/scripts/_start_scheduled_backup.sh.tpl
```diff
@@ -22,15 +22,14 @@ echo "Increase TiKV GC life time to {{ .Values.scheduledBackup.tikvGCLifeTime |
...
+  --output=${backupPath} \
   --host=${host} \
   --port=4000 \
   --user=${TIDB_USER} \
   --password=${TIDB_PASSWORD} \
-  --long-query-guard=3600 \
-  --tidb-force-priority=LOW_PRIORITY \
-  --regex '^(?!(mysql\.))' \
+  --filter '*.*' \
+  --filter '!mysql.*' \
```
DanielZhangQD, line 32: We should keep consistent with https://github.com/pingcap/tidb-operator/blob/master/cmd/backup-manager/app/constants/constants.go#L67

cmd/backup-manager/app/export/export.go
DanielZhangQD: No need to return an error here, I think just a warning log that this is not supported anymore is ok.

DanielZhangQD:
@lichunzhu Please add action required to the release notes, and request users to update the mydumper configuration if they have set before, and please also update the doc.

lichunzhu:
> @lichunzhu Please add action required to the release notes, and request users to update the mydumper configuration if they have set before, and please also update the doc.

OK

lichunzhu:
> @lichunzhu Please add action required to the release notes, and request users to update the mydumper configuration if they have set before, and please also update the doc.

ACTION REQUIRED is updated

cofyc:
what if we have a new backup tool in the future, do we need to break backward compatible again? 
is it possible to support multiple backup tools together? for example, add a field to specify the backup tool.

lichunzhu:
> what if we have a new backup tool in the future, do we need to break backward compatible again?
> is it possible to support multiple backup tools together? for example, add a field to specify the backup tool.

I think it's okay to support multiple backup tools together, but these backup tools should have distinct functions. For me, dumpling and mydumper are similar. What's more, in the future mydumper will possibly replaced by dumpling. So I think we should get rid of mydumper. It won't happen frequently.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1124,7 +1124,11 @@ type BackupSpec struct {
 	// BRConfig is the configs for BR
 	BR *BRConfig `json:"br,omitempty"`
 	// MydumperConfig is the configs for mydumper
+	// Deprecated
```
cofyc, line 1127: It seems that we don't support `mydumper` anymore. I guess we need to say it explicitly in the comment that this field does not take effect anymore, or simply remove this field.

cofyc:
I'm fine with removing `mydumper` support (if it's going to replaced by dumping). But the old CRs will not take effect immediately, I'm not sure if our users can accept this.  @DanielZhangQD what do you think?

cmd/backup-manager/app/util/util.go
```diff
@@ -182,10 +178,10 @@ func ConstructBRGlobalOptionsForBackup(backup *v1alpha1.Backup) ([]string, error
 // ConstructMydumperOptionsForBackup constructs mydumper options for backup
 func ConstructMydumperOptionsForBackup(backup *v1alpha1.Backup) []string {
```
cofyc, line 179: update the function name too

cofyc:
> I'm fine with removing `mydumper` support (if it's going to be replaced by dumping). But the old CRs will not take effect immediately, I'm not sure if our users can accept this. @DanielZhangQD what do you think?

Or we can provide a document for the migrating?

DanielZhangQD:
> > I'm fine with removing `mydumper` support (if it's going to be replaced by dumping). But the old CRs will not take effect immediately, I'm not sure if our users can accept this. @DanielZhangQD what do you think?
> 
> Or we can provide a document for the migration?

I think it's OK to remove mydumper support and it only affects the scheduled backup.
And we have to make it clear in both the release notes and document to tell users to migrate the configuration from mydumper to dumpling.
For the mydumper field in the CR, I think it's OK to remove them.

pkg/apis/pingcap/v1alpha1/types.go
DanielZhangQD: I think we can remove  this field.

DanielZhangQD:
> I'm fine with removing `mydumper` support (if it's going to be replaced by dumping). But the old CRs will not take effect immediately, I'm not sure if our users can accept this. @DanielZhangQD what do you think?

The old CRs will go with default dumpling configurations. 
From the current information I have, the backup CR with `mydumper` is rarely used.

images/tidb-backup-manager/Dockerfile
```diff
@@ -37,8 +37,8 @@ RUN \
   wget -nv https://download.pingcap.org/tidb-toolkit-${TOOLKIT_V40}-linux-amd64.tar.gz \
   && tar -xzf tidb-toolkit-${TOOLKIT_V40}-linux-amd64.tar.gz \
   && mv tidb-toolkit-${TOOLKIT_V40}-linux-amd64/bin/br /usr/local/bin/br40 \
-  && mv tidb-toolkit-${TOOLKIT_V40}-linux-amd64/bin/mydumper /mydumper \
-  && chmod 755 /usr/local/bin/br40 /mydumper \
+  && mv tidb-toolkit-${TOOLKIT_V40}-linux-amd64/bin/dumpling /dumpling \
```
DanielZhangQD, line 40: Can we combine the L19-L26 block with this block?

cmd/backup-manager/app/util/util.go
lichunzhu: addressed in https://github.com/pingcap/tidb-operator/pull/2870/commits/47d6d41707c75f880b1e31f410574272cf4ed3a4

images/tidb-backup-manager/Dockerfile
lichunzhu: addressed in https://github.com/pingcap/tidb-operator/pull/2870/commits/aa532c90288690796c5079d59a7cc3f817705b4f

lichunzhu:
@DanielZhangQD @cofyc PTAL again

PR #2886, opened 2020-07-08
tikv scaling takes precedence over upgrading

### LINKED ISSUE #2739: failover and scaling are blocked if one Pod failed during rolling update
## Feature Request

Take TiKV for example, during TiKV rolling update, if one TiKV Pod failed, e.g. something wrong with the node it was running and it cannot be scheduled to any node, then the upgrade will be stuck in waiting this Pod ready and its store UP, however, in this case, if the failover cannot occur because it's blocked by the logic https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_upgrader.go#L110-L115, and if users want to scale out a new TiKV to increase the replicas, it still is impossible due to the same reason.

cofyc:
fixes #2739 
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Support scaling and auto-failover even if a TiKV store fails in upgrading
```


tests/e2e/tidbcluster/stability.go
```diff
@@ -923,13 +986,12 @@ var _ = ginkgo.Describe("[tidb-operator][Stability]", func() {
...
 			err = utiltidbcluster.WaitForTidbClusterCondition(cli, tc.Namespace, tc.Name, time.Minute*5, func(tc *v1alpha1.TidbCluster) (bool, error) {
-				exist := false
 				for _, failureStore := range tc.Status.TiKV.FailureStores {
 					if failureStore.PodName == podName {
-						exist = true
+						return true, nil
 					}
 				}
-				return exist, nil
+				return false, nil
```
cofyc, line 994: simplify the code

pkg/manager/member/tikv_scaler_test.go
```diff
@@ -252,19 +252,6 @@ func TestTiKVScalerScaleIn(t *testing.T) {
 			errExpectFn:   errExpectRequeue,
 			changed:       false,
 		},
-		{
-			name:          "tikv is upgrading",
```
DanielZhangQD, line 256: Can we modify the case to cover the scenario that scaling can be done during upgrading?

cofyc: updated

Issue #2286, opened 2020-04-24
do not remove the backup data when remove the backup CR


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Now when we delete the backup CR, the backup data in S3 or GCS will be purged too, it should be possible that users can remove the backup CR with the backup data preserved.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

tennix:
We should make this optional, users can choose the deletion policy whether to delete the backup files on S3/GCS.

kolbe:
I think the most natural state is for the backup "artifact" to be decoupled from the Backup resource by default. A backup job should produce an artifact that is placed on external storage and deleting the Backup object should not have any effect on the artifact.

If we want a configuration option for the Backup job to delete the artifact, that may be useful, but it should not be enabled by default.

The ramifications of automatically deleting all backups when a cluster is deleted could be disastrous.

tennix:
> I think the most natural state is for the backup "artifact" to be decoupled from the Backup resource by default. A backup job should produce an artifact that is placed on external storage and deleting the Backup object should not have any effect on the artifact.
> 
> If we want a configuration option for the Backup job to delete the artifact, that may be useful, but it should not be enabled by default.
> 
> The ramifications of automatically deleting all backups when a cluster is deleted could be disastrous.

Yes, I agree this should not be enabled by default.

DanielZhangQD:
For this issue:

- We can add a field, e.g. `cleanData`, with default value `false`, so that the backup data will not be removed when CR is removed by default.
- If `cleanData` is false, I think no need to add finalized any more [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/backup/backup_control.go#L71) WDYT @cofyc @tennix @kolbe

tennix:
I remember the main reason for the finalizer is when the `Backup` CR is deleted, even we have the backup files on S3, we cannot do the restore with incremental backup & restore.

DanielZhangQD:
> I remember the main reason for the finalizer is when the `Backup` CR is deleted, even we have the backup files on S3, we cannot do the restore with incremental backup & restore.

I don't understand, what's the relationship between the finalizer and the incremental backup & restore?

tennix:
When we do full backup, we need to record the tso. It's kept in the Backup object status field.

[...]

PR #2892, opened 2020-07-09
Make remove backup data optional

### LINKED ISSUE #2888: backups block namespace deletion
## Bug Report

v1.1.0

**What did you do?**
I deleted the namespace the `TidbCluster` was in.

The namespace eventually gets deleted

**What did you see instead?**

The namespace got stuck in the `Terminating` state. Here's what I think happened:

There were `Backups` for the cluster. To delete them, the operator first runs a `Job` to delete the data from S3. However, since the namespace is `Terminating`, Kubernetes refuses to create the `Job`. The operator will keep retrying, and the finalizer on the `Backup` prevents it from being deleted. So the namespace just ends up stuck.

The only workaround seems to be to delete the finalizers.

[operator-logs.txt](https://github.com/pingcap/tidb-operator/files/4891954/operator-logs.txt)

lichunzhu:
Fix https://github.com/pingcap/tidb-operator/issues/2286 Fix #2888 

Add `cleanData` argument in backup spec(default: false). When it's enabled, the dumped data will be deleted after backup process is deleted.

### Check List 

Tests 

 - Unit test
 - E2E test
   * Test backup to S3 when `cleanData` is false. After deletion dumped data won't be deleted.
   * Test backup to S3 when `cleanData` is true. After deletion dumped data will be deleted.

Related changes

```release-note
ACTION REQUIRED: Add a field `cleanData` in `BackupSpec` to denote whether to clean backup data before the object is deleted from the cluster (defaults to false). If you want to clean backup data as before, please set `spec.cleanData` in `Backup` yaml or `spec.backupTemplate.cleanData` in `BackupSchedule` yaml to true.
```


pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1138,6 +1138,8 @@ type BackupSpec struct {
 	UseKMS bool `json:"useKMS,omitempty"`
 	// Specify service account of backup
 	ServiceAccount string `json:"serviceAccount,omitempty"`
+	// CleanData denotes whether to clean backuped data after backup job is deleted
```
cofyc, line 1141:
```suggestion
	// CleanData denotes whether to clean backup data before the object is deleted from the cluster
```

the finalizer implements `pre-delete` hooks.

lichunzhu:
@DanielZhangQD @cofyc PTAL

pkg/controller/backup/backup_control.go
```diff
@@ -53,12 +53,14 @@ type defaultBackupControl struct {
 // UpdateBackup executes the core logic loop for a Backup.
 func (bc *defaultBackupControl) UpdateBackup(backup *v1alpha1.Backup) error {
 	backup.SetGroupVersionKind(controller.BackupControllerKind)
-	if err := bc.addProtectionFinalizer(backup); err != nil {
-		return err
-	}
+	if backup.Spec.CleanData {
```
cofyc, line 56: this check seems not necessary and IMO we should always try to add/remove finalizer, for example, users may change this field from `true` to `false`

cofyc:
a simplified version of release note, what do you think?

```
Add a field `spec.cleanData` in `Backup` to denote whether to clean backup data before the object is deleted from the cluster  (defaults to false)
```

lichunzhu:
> a simplified version of release note, what do you think?
> 
> ```
> Add a field `spec.cleanData` in `Backup` to denote whether to clean backup data before the object is deleted from the cluster  (defaults to false)
> ```

Revised. It's much better.

cofyc:
oh, this is an incompatible change. The old Backup will always try to clean the data.

I think we need an `ACTION REQUIRED` release note and telling people to set the `cleanData` field to `true` if they want to keep the old behavior.

cofyc:
cc @DanielZhangQD do we need to cherry-pick this into 1.1?

lichunzhu:
> oh, this is an incompatible change. The old Backup will always try to clean the data.
> 
> I think we need an `ACTION REQUIRED` release note and telling people to set the `cleanData` field to `true` if they want to keep the old behavior.

Yes, you're right.

DanielZhangQD:
> cc @DanielZhangQD do we need to cherry-pick this into 1.1?

Yes, I have added the label.

lichunzhu:
> oh, this is an incompatible change. The old Backup will always try to clean the data.
> 
> I think we need an `ACTION REQUIRED` release note and telling people to set the `cleanData` field to `true` if they want to keep the old behavior.

Added in release note.

cofyc:
@lichunzhu please also set `cleanData` to `true` in `GetBackupCRDWithS3` (`tests/pkg/fixture/fixture.g`). We should clean backup data in our testing.

pkg/controller/backup/backup_control.go
```diff
@@ -86,7 +86,7 @@ func (bc *defaultBackupControl) removeProtectionFinalizer(backup *v1alpha1.Backu
 	ns := backup.GetNamespace()
 	name := backup.GetName()
 
-	if isDeletionCandidate(backup) && v1alpha1.IsBackupClean(backup) {
+	if !backup.Spec.CleanData || (isDeletionCandidate(backup) && v1alpha1.IsBackupClean(backup)) {
```
DanielZhangQD, line 89:
```suggestion
	if backup.Spec.CleanData && isDeletionCandidate(backup) && v1alpha1.IsBackupClean(backup) {
```

DanielZhangQD: Is my suggestion correct?

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1138,6 +1138,8 @@ type BackupSpec struct {
 	UseKMS bool `json:"useKMS,omitempty"`
 	// Specify service account of backup
 	ServiceAccount string `json:"serviceAccount,omitempty"`
+	// CleanData denotes whether to clean backup data before the object is deleted from the cluster
```
DanielZhangQD, line 1141: Please help update the examples in `manifests/backup/`.

lichunzhu: addressed in https://github.com/pingcap/tidb-operator/pull/2892/commits/ee29d324b6fc2e46c4587c2bb4ecd9a83947ad4d

pkg/controller/backup/backup_control.go
lichunzhu: addressed in https://github.com/pingcap/tidb-operator/pull/2892/commits/156e00d9240831138b91865b96c2dcea9a600d10

manifests/backup/backup-aws-s3-br.yaml
```diff
@@ -10,6 +10,7 @@ spec:
   # backupType: full
   # useKMS: false
   # serviceAccount: myServiceAccount
+  # cleanData: true
```
DanielZhangQD, line 13: No need for BR yamls.

DanielZhangQD: sorry, it also needs to change for BR.

DanielZhangQD:
@lichunzhu The release notes also have to mention `BackupSchedule`CR.

lichunzhu:
> @lichunzhu The release notes also have to mention `BackupSchedule`C

> @lichunzhu The release notes also have to mention `BackupSchedule`CR.

Release note has been revised. Conflicts have also been resolved.

Issue #2734, opened 2020-06-16
Restore reuses PVC with too-small storage


kolbe:
## Bug Report

**What version of Kubernetes are you using?**

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.1", GitCommit:"7879fc12a63337efff607952a323df90cdc7a335", GitTreeState:"clean", BuildDate:"2020-04-08T17:38:50Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.11-eks-af3caf", GitCommit:"af3caf6136cd355f467083651cc1010a499f59b1", GitTreeState:"clean", BuildDate:"2020-03-27T21:51:36Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of TiDB Operator are you using?**

```
TiDB Operator Version: version.Info{GitVersion:"v1.1.0", GitCommit:"bc9ca21adb990ca01f41c0ea9b7f6bf5d2efc16b", GitTreeState:"clean", BuildDate:"2020-05-28T11:31:09Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

```
# kubectl get sc
NAME            PROVISIONER                    AGE
ebs-gp2         kubernetes.io/aws-ebs          3h46m
gp2 (default)   kubernetes.io/aws-ebs          3h58m
local-storage   kubernetes.io/no-provisioner   3h46m

# kubectl get pvc -n "$namespace"
NAME                     STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS    AGE
pd-my-cluster-pd-0       Bound    pvc-cb89be5e-5229-46ef-97f7-e4b296e35268   1Gi        RWO            ebs-gp2         3h39m
pd-my-cluster-pd-1       Bound    pvc-287d93c8-820e-49cb-964e-06f0cadc3e58   1Gi        RWO            ebs-gp2         3h39m
pd-my-cluster-pd-2       Bound    pvc-84bd5211-f608-4d65-894f-fe658b3f6bb2   1Gi        RWO            ebs-gp2         3h39m
[...]

PR #2894, opened 2020-07-09
check pvc size before running

lichunzhu:
Try to fix https://github.com/pingcap/tidb-operator/issues/2734.

The pvc name for a restore job is fixed now. So when the pvc exists but the requested storage size is smaller than we set, it's better to lead users to delete the pvc.

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

Related changes

```release-note
Fail the backup job if existing PVC size is smaller than the storage request in the backup job
```


pkg/backup/backup/backup_manager.go
```diff
@@ -407,9 +407,12 @@ func (bm *backupManager) ensureBackupPVCExist(backup *v1alpha1.Backup) (string,
 		return "ParseStorageSizeFailed", errMsg
 	}
 	backupPVCName := backup.GetBackupPVCName()
-	_, err = bm.pvcLister.PersistentVolumeClaims(ns).Get(backupPVCName)
+	pvc, err := bm.pvcLister.PersistentVolumeClaims(ns).Get(backupPVCName)
 
 	if err == nil {
+		if pvcRs := pvc.Spec.Resources.Requests[corev1.ResourceStorage]; pvcRs.Cmp(rs) == -1 {
+			return "", fmt.Errorf("%s/%s's backup pvc %s's storage size %s is less than expected storage size %s, please delete old pvc to continue", ns, name, pvc.GetName(), pvcRs.String(), rs.String())
```
DanielZhangQD, line 414: Add a short `reason` here.

pkg/backup/restore/restore_manager.go
```diff
@@ -417,6 +417,8 @@ func (rm *restoreManager) ensureRestorePVCExist(restore *v1alpha1.Restore) (stri
 			errMsg := fmt.Errorf(" %s/%s create restore pvc %s failed, err: %v", ns, name, pvc.GetName(), err)
 			return "CreatePVCFailed", errMsg
 		}
+	} else if pvcRs := pvc.Spec.Resources.Requests[corev1.ResourceStorage]; pvcRs.Cmp(rs) == -1 {
+		return "", fmt.Errorf("%s/%s's restore pvc %s's storage size %s is less than expected storage size %s, please delete old pvc to continue", ns, name, pvc.GetName(), pvcRs.String(), rs.String())
```
DanielZhangQD, line 421: ditto

DanielZhangQD:
> in addition to this, I think we should emit an event too, then users can know what happened by describing the object

The backup-manager now stores various errors in the .status.conditions, I think it's OK to keep consistent with the current behavior, WDYT? @cofyc

pkg/backup/backup/backup_manager.go
lichunzhu: addressed in https://github.com/pingcap/tidb-operator/pull/2894/commits/f656885822777a2cc56b35542e0e13d0e539e710

cofyc:
remove the period at the end of release note: https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md

DanielZhangQD:
What about `Fail the backup job if existing PVC size is smaller than the storage request in the backup job` for release notes? @lichunzhu

lichunzhu:
> Fail the backup job if existing PVC size is smaller than the storage request in the backup job

SGTM, revised.

PR #2913, opened 2020-07-13
Support TiKVGroup Upgrading and Scaling

Yisaer:

Support Scaling and Upgrading for a Single TiKVGroup by Pod webhook.

Note:
1. this request haven't supported multi TiKVGroup Upgrading and Scaling control yet.
2. I let default scheduler to control the tikvgroup statefulset until tidb-scheduler support tikvgroup.

```release-note
Support Upgrading and Scaling for Single TiKVGroup
```


pkg/manager/member/tikv_group_scaler.go
```diff
@@ -0,0 +1,92 @@
...
+	} else if scaling < 0 {
+		return tsd.ScaleIn(tc, oldSet, newSet)
+	}
+	return nil
+}
+
+func (tsd *tikvGroupScaler) ScaleOut(tg *v1alpha1.TiKVGroup, oldSet *apps.StatefulSet, newSet *apps.StatefulSet) error {
+	_, ordinal, replicas, deleteSlots := scaleOne(oldSet, newSet)
+	resetReplicas(newSet, oldSet)
+	fmt.Println("here")
```
cofyc: remove debug code

pkg/manager/member/tikv_group_member_manager.go
```diff
@@ -608,6 +631,8 @@ func getNewTiKVSetForTiKVGroup(tg *v1alpha1.TiKVGroup, tc *v1alpha1.TidbCluster,
 		Resources:    controller.ContainerResource(tc.Spec.TiKV.ResourceRequirements),
 	}
 	podSpec := baseTiKVSpec.BuildPodSpec()
+	// TODO: make tidb-scheduler support TiKVGroup
+	podSpec.SchedulerName = "default-scheduler"
```
cofyc, line 635: create a track issue?

pkg/manager/member/tikv_group_member_manager.go
```diff
@@ -426,9 +440,18 @@ func getTikVConfigMapForTiKVGroup(tg *v1alpha1.TiKVGroup, tc *v1alpha1.TidbClust
...
+	}
+	enableAdvertiseStatusAddr := true
+	// For the version less than v4.0.1, don't use advertise-status-addr start arg
+	if version.LessThan(tikvV401Version) {
+		enableAdvertiseStatusAddr = false
+	}
 	scriptModel := &TiKVStartScriptModel{
 		Scheme:                    tc.Scheme(),
-		EnableAdvertiseStatusAddr: true,
+		EnableAdvertiseStatusAddr: enableAdvertiseStatusAddr,
```
cofyc, line 454: what if `tc.spec.EnableDynamicConfiguration` is set to `false`? will it cause problems?

pkg/manager/member/scaler.go
```diff
@@ -53,29 +55,45 @@ type generalScaler struct {
...
 	// for unit test
 	skipReason := map[string]string{}
-
-	// pvcName := ordinalPVCName(memberType, setName, ordinal)
-	podName := ordinalPodName(memberType, tc.Name, ordinal)
-	l := label.New().Instance(tc.GetInstanceName())
-	l[label.AnnPodNameKey] = podName
+	var podName, kind string
+	var l label.Label
+	_, ok := controller.(*v1alpha1.TidbCluster)
```
cofyc, line 66: maybe it's better to use type switches here, example: https://tour.golang.org/methods/16

pkg/manager/member/scaler.go
```diff
@@ -53,29 +55,45 @@ type generalScaler struct {
...
+		podName = ordinalPodName(memberType, meta.GetName(), ordinal)
+		l = label.New().Instance(meta.GetName())
+		l[label.AnnPodNameKey] = podName
+		kind = v1alpha1.TiDBClusterKind
+	} else {
+		_, ok := controller.(*v1alpha1.TiKVGroup)
+		if ok {
+			podName = fmt.Sprintf("%s-%s-group-%d", meta.GetName(), memberType, ordinal)
+			l = label.NewGroup().Instance(meta.GetName())
+			// TODO: support sync meta info into TiKVGroup resources (pod/pvc)
```
cofyc, line 77: create a track issue? (ignore me if already tracked)

tests/e2e/tidbcluster/serial.go
```diff
@@ -209,6 +226,81 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+
+		ginkgo.BeforeEach(func() {
+			ocfg = &tests.OperatorConfig{
+				Namespace:         ns,
+				ReleaseName:       "operator",
+				Image:             cfg.OperatorImage,
+				Tag:               cfg.OperatorTag,
+				SchedulerImage:    "k8s.gcr.io/kube-scheduler",
+				LogLevel:          "4",
+				ImagePullPolicy:   v1.PullIfNotPresent,
```
cofyc, line 241: `ImagePullPolicy` and `SchedulerImage` can be removed as they are using the default values

tests/pkg/fixture/fixture.go
```diff
@@ -155,6 +155,43 @@ func GetTidbCluster(ns, name, version string) *v1alpha1.TidbCluster {
...
+	if v, err := semver.NewVersion(version); err == nil && v.LessThan(tikvV4Beta) {
+		tikvStorageConfig = nil
+	}
+	tg := &v1alpha1.TiKVGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: ns,
+		},
+		Spec: v1alpha1.TiKVGroupSpec{
+			TiKVSpec: v1alpha1.TiKVSpec{
```
cofyc, line 175: do tidb cluster and tikv group share the same `v1alpha1.TiKVSpec`? if it is, we can have a utility function to return `v1alpha1.TiKVSpec`. this avoids duplicated code, and it's easy to update the default tikv spec (for example, adjust the configuration for testing)

Yisaer: No, they don't. They only share the same image spec currently.

cofyc: I see.

how about sharing the default TiKV config in GetTiKVGroup/GetTiKVCluster functions? if in the future, we need to update the default testing configuration for TiKV, we only need to update in one place.

pkg/manager/member/tikv_group_member_manager.go
Yisaer: If tikv version is under v4.0.1, we should not set advertise-status-addr otherwise it will cause start failed.

cofyc: I mean it's different with here

https://github.com/pingcap/tidb-operator/blob/6d9a84c179eb60c23587943eae790327b95051e3/pkg/manager/member/tikv_member_manager.go#L573-L581

If `tc.spec.EnableDynamicConfiguration` is `false`, tikv replicas deployed by `TidbCluster` will not report to PD, but replicas deployed by `TikvGroup` will report to PD.

I'm fine with this if it's your intention. I guess this will not cause problems.

Yisaer: TiKV will always be reported to PD in version v4.0.0, `EnableDynamicConfiguration` is used to control whether send `advertise-status-addr` or `advertise-addr` to PD.

PR #2918, opened 2020-07-13
Fix externalendpoint judgement in tikv auto-scaling

sunzhuohang:
Fix externalendpoint judgement in tikv auto-scaling.
image

### DIFF
--- a/pkg/autoscaler/autoscaler/tikv_autoscaler.go
+++ b/pkg/autoscaler/autoscaler/tikv_autoscaler.go
@@ -80,16 +80,7 @@ func calculateTiKVMetrics(tac *v1alpha1.TidbClusterAutoScaler, tc *v1alpha1.Tidb
 
 	// check externalEndpoint
 	if tac.Spec.TiKV.ExternalEndpoint != nil {
-		targetReplicas, err := query.ExternalService(tc, v1alpha1.TiKVMemberType, tac.Spec.TiKV.ExternalEndpoint, kubecli)
-		if err != nil {
-			klog.Errorf("tac[%s/%s] 's query to the external endpoint got error: %v", tac.Namespace, tac.Name, err)
-			return err
-		}
-		targetReplicas = limitTargetReplicas(targetReplicas, tac, v1alpha1.TiKVMemberType)
-		if targetReplicas == tc.Spec.TiKV.Replicas {
-			return nil
-		}
-		return syncTiKVAfterCalculated(tc, tac, tc.Spec.TiKV.Replicas, targetReplicas)
+		return calculateTiKVExternalService(tc, tac, sts, kubecli)
 	}
 
 	// check CPU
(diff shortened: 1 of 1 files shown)


pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -146,11 +148,21 @@ func calculateTiKVStorageMetrics(tac *v1alpha1.TidbClusterAutoScaler, tc *v1alph
...
+		targetReplicas, err = calculate.CalculateRecomendedReplicasByCpuCosts(tac, sq, sts, client, v1alpha1.TiKVMemberType, duration, metric.MetricSpec)
+		if err != nil {
+			return err
+		}
+	} else {
+		targetReplicas, err = query.ExternalService(tc, v1alpha1.TiKVMemberType, tac.Spec.TiKV.ExternalEndpoint, kubecli)
+		if err != nil {
+			klog.Errorf("tac[%s/%s] 's query to the external endpoint got error: %v", tac.Namespace, tac.Name, err)
+			return err
+		}
```
Yisaer, line 165: I think we should check external before cpu case  instead of in cpu function.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -111,6 +127,21 @@ func calculateTiKVMetrics(tac *v1alpha1.TidbClusterAutoScaler, tc *v1alpha1.Tidb
...
+	}
+	ableToScale, err := checkStsAutoScalingInterval(tac, *intervalSeconds, v1alpha1.TiKVMemberType)
+	if err != nil {
+		return err
+	}
+	if !ableToScale {
+		return nil
+	}
+	return updateTacIfTiKVScale(tc, tac, recommendedReplicas)
+}
```
Yisaer, line 143: use `syncTiKVAfterCalculated` directly.

pkg/autoscaler/autoscaler/tikv_autoscaler.go
```diff
@@ -111,6 +118,35 @@ func calculateTiKVMetrics(tac *v1alpha1.TidbClusterAutoScaler, tc *v1alpha1.Tidb
...
+		return err
+	}
+	if !ableToScale {
+		return nil
+	}
+	err = updateTacIfTiKVScale(tc, tac, targetReplicas)
+	if err != nil {
+		return err
+	}
+	return addAnnotationMarkIfScaleOutDueToCPUMetrics(tc, currentReplicas, targetReplicas, sts)
```
Yisaer, line 147: do we need to mark the scale-out pod in external service case? This would add label `specialUse:Hot` store label for it.

Yisaer: the rest LGTM.

sunzhuohang: Yes, I think it‘s required in external service case,  the scale-out pod should only balance hot region too.

Issue #2692, opened 2020-06-12
Provide way to set store labels for TiKV during initialization 


Yisaer:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Sometimes we need TiKV Initialize with some store labels in order to schedule the data they need instead of setting labels by pd-api after tikv started.

**Describe the feature you'd like:**

`TiKVSpec` should provide a way to let users setting the initialization labels.

Yisaer:
@mahjonp  Does this feature meet your requirement?

mahjonp:
Cool. /cc @BusyJay @cosven

cosven:
My requirement is that different tikv pod can have different labels. 

To my understanding,  if we set labels in TiKVSpec, all tikv pods will have the same labels?

Yisaer:
@cosven  Your understanding is right. For your requirement, I think mixing the [heterogeneous design](https://github.com/pingcap/tidb-operator/issues/2240) and this feature might work.

DanielZhangQD:
We can set labels via config [here](https://github.com/tikv/tikv/blob/master/etc/config-template.toml#L174) and as @Yisaer has explained, when the heterogeneous deployment is supported, we can deploy TiKVs with different labels.

PR #2945, opened 2020-07-15
Support initialization store labels

Yisaer:
Close https://github.com/pingcap/tidb-operator/issues/2692

Support initialization store labels for TiKV

Related changes

```release-note
Support initialize store labels for TiKVGroup and TidbCluster
```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -394,9 +394,9 @@ type TiKVSpec struct {
 	// +optional
 	Config *TiKVConfig `json:"config,omitempty"`
 
-	// InitializeStoreLabel indicates the store label set during initialization
+	// InitializeStoreLabels indicates the store label set during initialization
 	// +optional
-	InitializeStoreLabel map[string]string `json:"initializeStoreLabel,omitempty"`
+	InitializeStoreLabels map[string]string `json:"initializeStoreLabels,omitempty"`
 }
 
 // TiFlashSpec contains details of TiFlash members
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -393,6 +393,10 @@ type TiKVSpec struct {
 	// Config is the Configuration of tikv-servers
 	// +optional
 	Config *TiKVConfig `json:"config,omitempty"`
+
+	// InitializeStoreLabel indicates the store label set during initialization
+	// +optional
+	InitializeStoreLabel map[string]string `json:"initializeStoreLabel,omitempty"`
 }
 
 // TiFlashSpec contains details of TiFlash members
--- a/pkg/manager/member/tikv_group_member_manager.go
+++ b/pkg/manager/member/tikv_group_member_manager.go
@@ -645,7 +645,7 @@ func getNewTiKVSetForTiKVGroup(tg *v1alpha1.TiKVGroup, tc *v1alpha1.TidbCluster,
 		})
 	}
 	tikvContainer.Env = util.AppendEnv(env, baseTiKVSpec.Env())
-	storeLabels := tg.Spec.InitializeStoreLabel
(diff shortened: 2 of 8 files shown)


pkg/manager/member/tikv_group_member_manager.go
```diff
@@ -645,6 +645,20 @@ func getNewTiKVSetForTiKVGroup(tg *v1alpha1.TiKVGroup, tc *v1alpha1.TidbCluster,
 		})
 	}
 	tikvContainer.Env = util.AppendEnv(env, baseTiKVSpec.Env())
+	var storeLabels map[string]string
+	if tg.Spec.InitializeStoreLabels != nil {
+		storeLabels = tg.Spec.InitializeStoreLabels
+	} else {
+		storeLabels = map[string]string{}
+	}
+	storeLabels["groupName"] = tg.Name
```
Yisaer, line 654: For each TiKVGroup instance, we will set `groupName`: <tikvgroup-name> as its store label. @rleungx

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -393,6 +393,10 @@ type TiKVSpec struct {
 	// Config is the Configuration of tikv-servers
 	// +optional
 	Config *TiKVConfig `json:"config,omitempty"`
+
+	// InitializeStoreLabels indicates the store label set during initialization
+	// +optional
+	InitializeStoreLabels map[string]string `json:"initializeStoreLabels,omitempty"`
```
cofyc, line 399: I would prefer a shorter name, `storeLabels` or `initStorelabels`

pkg/manager/member/utils.go
```diff
@@ -386,3 +387,24 @@ func getTikVConfigMapForTiKVSpec(tikvSpec *v1alpha1.TiKVSpec, tc *v1alpha1.TidbC
...
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		v, ok := labels[key]
+		if ok {
+			s = fmt.Sprintf("%s,%s", s, fmt.Sprintf("%s=%s", key, v))
+		}
+	}
+	return s[1:]
```
cofyc, line 409:
```suggestion
    var envs []string
    for k, v := range labels {
        envs = append(envs, fmt.Sprintf("%s=%s", k, v)) 
    }   
    sort.Strings(envs)
    return strings.Join(envs, ",")
```

we should always try to use built-in functions to simplify code

btw, please add some basic units 1) nil map 2) 1 element map 3) more than 1 elements map

pkg/apis/pingcap/v1alpha1/types.go
cofyc: maybe we should this generic way to configure environments for containers

https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/types.go#L711

Yisaer: Setting store label by ENV is the way operator pass the store labels for tikv but not obvious and convenient for users, I still suggest exposing store labels by a certain atrributes.

cofyc: I don't think it's necessary to provide a special way for this. It's simple as configuring environments for a normal pod in Kubernetes.

cofyc: We already have too many special configurations. Anyone knowing the standard way to configure environments for containers can configure this env and any other environments for TiKV and other components. 

If we use this special field for `STORE_LABELS`, we have more things to consider, e.g. what if users set both `STORE_LABELS` env and `storeLabels` field. Should we merge them? If not, which one has a higher priority, etc. If we have a special field for this env, should we add fields for other envs if users ask?

Yisaer: ok, I agree with setting store labels by env.

Issue #2169, opened 2020-04-13
Add some useful additional printer columns for TidbMonitor


cofyc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Now it uses the default columns.

```
$ kubectl get tidbmonitor
NAME    AGE
basic   23s
```

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

cofyc:
@mikechengwei 
in production, what information is useful and should be printed?

mikechengwei:
TidbMonitor have grafana ,prometheus,reloader and additional container. Do we need to show their status like TidbCluster.
```
NAME                PD                                              STORAGE   READY   DESIRE   TIKV                                              STORAGE   READY   DESIRE   TIDB   READY   DESIRE
```

cofyc:
maybe we can show their running status? then we can quickly check if components have problems or not

mikechengwei:
Do you mean `ready` and `desire` info?  Tidbmonitor is running in a pod with multiple containers, i think we can't use `ready` and `desire` columns. We can use `ready` column the same as in pod columns.

mikechengwei:
We can inherit tidbmonitor pod info.  `READY` ,  `STATUS` and  `RESTARTS` columns . What do you think?

cofyc:
I think it will be useful.

mikechengwei:
Ok, i can do it and do it well soon.

cofyc:
thanks! for users who do not know technical details, they may check the status `TidbMonitor` to see if the monitor is ready or not. I guess it will be useful if we can print more information by default.

Issue #2968, opened 2020-07-16
Support tidbMonitor multiple replicas to ensure high availability


mikechengwei:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Should we support multiple replicas of TidbMonitor to ensure high availability?  Now ,the replica is one, hard-coded .

https://github.com/pingcap/tidb-operator/blob/e84a3d4fd007a62dfb3fd61681636357ebd787d3/pkg/monitor/monitor/util.go#L229
**Describe the feature you'd like:**

I think we can support replica field in tidbMonitor spec.

mikechengwei:
And i can do it .

DanielZhangQD:
@mikechengwei Thanks a lot!
BTW, usually, we would have monitoring data persistent, how will the data be handled if multiple replicas are created?

mikechengwei:
Currently we use deployment with specified pvc to deploy tidbmonitor. Can we change to use statefulset object? Monitoring is about storage-related services. @DanielZhangQD

DanielZhangQD:
> Currently we use deployment with specified pvc to deploy tidbmonitor. Can we change to use statefulset object? Monitoring is about storage-related services. @DanielZhangQD

Yes, it makes sense, please help evaluate it.
BTW, if we decide to change to sts, we have to consider the migration:

- The resource migration from deployment to sts
- The data migration from the previous Pod

If we just create sts and new PVC in the new version and keep the deployment and PVC of the previous versions, is there a way to provide procedure to guide users about how to access the old data? If yes, then I think it's OK to not consider the migration and after several days, when users do not need the old data, they can just remove the deployment and the PVC.
WDYT @mikechengwei @cofyc

mikechengwei:
When user upgrade  operator , we can keep deployment monitor and create  new sts monitor. In the code, we can determine whether the k8s environment has monitor deployment, it can continue running with old way until the user thinks there is no problem, manauly delete deployment and pvc.

DanielZhangQD:
> When user upgrade operator , we can keep deployment monitor and create new sts monitor. 
Yes.
> In the code, we can determine whether the k8s environment has monitor deployment, 
If we can provide a procedure to guide users to access the old data, I think no need to care about the deployment in the code.

mikechengwei:
I will provide a guide as soon.

cofyc:
[...]

PR #2967, opened 2020-07-16
Add some useful additional printer columns for TidbMonitor 

mikechengwei:
### What problem does this PR solve? 
#2169 

### What is changed and how does it work?
1.  add `ready`, `updatedReplicas` and `availableReplicas` columns  to tidbmonitor status and sync .
2. kubectl support printing `READY`  ,`UP-TO-DATE`  and AVAILABLE columns 

We use  deployment to deploy tidbmonitor , so we can print same column as the deployment.
### Check List 

 - Unit test
 - E2E test
 - Stability test
 - Manual test (add detailed scripts or steps below)

Code changes

 - Has Go code change

Side effects

 - None

PR #2982, opened 2020-07-17
Add some useful additional printer columns for TidbMonitor 

mikechengwei:
#2169 

1.  add `ready`, `updatedReplicas` and `availableReplicas` columns  to tidbmonitor status and sync .
2. kubectl support printing `READY`  ,`UP-TO-DATE`  and AVAILABLE columns 

We use  deployment to deploy tidbmonitor , so we can print same column as the deployment.
### Check List 

 - Unit test
 - E2E test
 - Stability test

Code changes

 - Has Go code change

Side effects

 - None


mikechengwei:
@cofyc Hi, we want to  use sts deploy tidbmonitor in  #2968 . So  using deploymentStatus directly is in vain. What do you think?

cofyc:
> @cofyc Hi, we want to use sts deploy tidbmonitor in #2968 . So using deploymentStatus directly is in vain. What do you think?

I think it's good to use sts, then we don't need to create PVCs manually in our controller. The most important thing is the migration. We need to provide a smooth migration experience, otherwise, users may lose data. 

Do you have a detailed design for this change?

cofyc:
btw, I left a few comments on your original PR https://github.com/pingcap/tidb-operator/pull/2967 PTAL

mikechengwei:
@cofyc  I have see you comments  . I have a question whether we should use READY ,UP-TO-DATE and AVAILABLE  fields directly to avoid migration caused by sts deploy.

I am currently referring to the code in kubernetes below:
https://github.com/kubernetes/kubernetes/blob/23b66eaabd3a535dbee6474638c5bf51e78fbcfa/pkg/printers/internalversion/printers.go#L1940-L1956

cofyc:
I'm fine not adding any printer columns until we migrated to the StatefulSet.

mikechengwei:
ok

DanielZhangQD:
@mikechengwei When do you plan to work on the issue that deploys monitor with sts, if you're going to work on it in a short time, I think you can wait until then to add the printer columns, WDYT?

mikechengwei:
No problem. I will finish sts design in next week and do it at weekend in this week.

Issue #2969, opened 2020-07-16
Support TLS to tidb-binlog drainer downstream


kolbe:
## Feature Request

**Is your feature request related to a problem? Please describe:**
https://github.com/pingcap/tidb-operator/issues/1605 should have added support for TLS between the drainer and the downstream MySQL/TiDB instance, but it looks like that was overlooked.

**Describe the feature you'd like:**
TiDB-Operator charts should support TLS between the TiDB-Binlog drainer and the downstream target.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

kolbe:
Part of https://github.com/pingcap/tidb-operator/issues/1855

PR #2993, opened 2020-07-20
add tls support between drainer and downstream database server

lichunzhu:
resolve https://github.com/pingcap/tidb-operator/issues/2969
add tls support between drainer and downstream database server.

add tls syncer part for drainer.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
    * Write to downstream database without tls.
    * Write to downstream database with only downstream tls specified.
    * Write to downstream but checkpoint to another database with/without tls enabled.

Related changes

```release-note
Add TLS support between drainer and downstream database server
```

### DIFF
--- a/charts/tidb-drainer/values.yaml
+++ b/charts/tidb-drainer/values.yaml
@@ -48,20 +48,34 @@ tlsCluster:
   certAllowedCN: []
   #  - TiDB
 
-# The tls config between drainer and the downstream database server (MySQL/TiDB)
-tlsSyncer:
-  tlsClientSecretName:
+# The TLS config between drainer and the downstream database server (MySQL/TiDB)
+tlsSyncer: {}
+  # The steps to enable this feature:
+  #   1. Generate MySQL Client certificate.
+  #      There are documents help to generate these certificates:
+  #        - enable TLS for the MySQL Client through cfssl or cert-manager: https://docs.pingcap.com/tidb-in-kubernetes/stable/enable-tls-for-mysql-client
+  #        - check and enble TLS for the MySQL Client: https://docs.pingcap.com/tidb/stable/enable-tls-between-clients-and-servers
+  #   2. Create one secret object for Drainer which contains the certificates created above.
+  #      Suggest the name of this Secret is <downstream_database_secret_name>.
+  #        For Drainer: kubectl create secret generic ${downstream_database_secret_name} --namespace=${namespace} --from-file=tls.crt=client.pem --from-file=tls.key=client-key.pem --from-file=ca.crt=ca.pem
+  #   3. Then create the Drainer cluster with `tlsSyncer.tlsClientSecretName` set to `<downstream_database_secret_name>`.
+  # tlsClientSecretName: demo-tidb-client-secret
 
   # certAllowedCN is the Common Name that allowed
-  certAllowedCN: []
+  # certAllowedCN: []
   #  - TiDB
 
-  # checkpoint is the tls config for the database we save binlog checkpoint
-  # Omit this part if you just want to save checkpoint in downstream database
-  checkpoint:
-    tlsClientSecretName:
+  # checkpoint is the TLS config for the database we save binlog checkpoint
+  # In most cases, you needn't to specify [syncer.to.checkpoint.type].
+  # Drainer will use downstream to save binlog checkpoint and you can just left this field commented.
+  # You have to fill this field only if you want to save binlog checkpoint to ** another database which has enabled TLS **.
+  # The steps to enable this feature is similar like enable tlsSyncer.tlsClientSecretName,
+  # which means you need to Generate MySQL Client certificate, create one secret object and then set tlsClientSecretName
+  # to your checkpoint database secret name like above.
+  # checkpoint: {}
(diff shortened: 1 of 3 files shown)


charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
   certAllowedCN: []
   #  - TiDB
 
+# The tls config between drainer and the downstream database server (MySQL/TiDB)
```
DanielZhangQD, line 51:
```suggestion
# The TLS config between drainer and the downstream database server (MySQL/TiDB)
```

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+tlsSyncer:
+  tlsClientSecretName:
```
DanielZhangQD, line 53: Please add comments to explain what's this field is for and what keys should be included in the secret or add an example command as in L41-L43.

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+
+  # certAllowedCN is the Common Name that allowed
```
DanielZhangQD, line 55: What's the difference between this one and the one in L64, please make it clear.

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+  certAllowedCN: []
+  #  - TiDB
+
+  # checkpoint is the tls config for the database we save binlog checkpoint
```
DanielZhangQD, line 59:
```suggestion
  # checkpoint is the TLS config for the database we save binlog checkpoint
```

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+  # Omit this part if you just want to save checkpoint in downstream database
```
DanielZhangQD, line 60: What does this mean?

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+  checkpoint:
+    tlsClientSecretName:
```
DanielZhangQD, line 62: Add an explanation, and if the DB of saving checkpoint is the same as the downstream DB, this can use the same secret as in L53, right? If yes, make it clear in the comments.
And, in this case, if users do not configure the `checkpoint`, we should use the configuration in L53-L56?

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+  # certAllowedCN is the Common Name that allowed
```
DanielZhangQD, line 63: Make it clear for the difference with L56.

charts/tidb-drainer/values.yaml
DanielZhangQD: Suggest commenting out the new configurations.

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
   certAllowedCN: []
   #  - TiDB
 
+# The tls config between drainer and the downstream database server (MySQL/TiDB)
+tlsSyncer:
```
DanielZhangQD, line 52:
```suggestion
tlsSyncer: {}
```

charts/tidb-drainer/values.yaml
DanielZhangQD:
```suggestion
  # tlsClientSecretName: ""
```

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+  tlsClientSecretName:
+
+  # certAllowedCN is the Common Name that allowed
+  certAllowedCN: []
```
DanielZhangQD, line 56:
```suggestion
  # certAllowedCN: []
```

charts/tidb-drainer/values.yaml
DanielZhangQD, line 62:
```suggestion
  # checkpoint:
    # tlsClientSecretName: ""
```

charts/tidb-drainer/values.yaml
```diff
@@ -48,6 +48,22 @@ tlsCluster:
...
+  #  - TiDB
+
+  # checkpoint is the tls config for the database we save binlog checkpoint
+  # Omit this part if you just want to save checkpoint in downstream database
+  checkpoint:
+    tlsClientSecretName:
+  # certAllowedCN is the Common Name that allowed
+    certAllowedCN: []
```
DanielZhangQD, line 64:
```suggestion
    # certAllowedCN: []
```

charts/tidb-drainer/values.yaml
lichunzhu: No, empty configuration is okay. binlog will fill the security options automatically if `[syncer.to.checkpoint.type]` if not specified.

lichunzhu:
@weekface PTAL again

PR #3043, opened 2020-07-30
fix  helm install question included in e2e shell under mac system

mikechengwei:
Under mac system,I cannot run e2e test  normally.

command：

`
./output/bin/helm init --service-account tiller --output yaml             | sed 's@apiVersion: extensions/v1beta1@apiVersion: apps/v1@'             | sed 's@  replicas: 1@  replicas: 1\n  selector: {"matchLabels": {"app": "helm", "name": "tiller"}}@'             | kubectl  apply -f -
`
error:
`
error: error parsing STDIN: error converting YAML to JSON: yaml: line 11: mapping values are not allowed in this context
`

I find `\n` nextline character not work in mac , so we can replace with `\'$'\n`.

### Check List 

Tests 

 - E2E test

Code changes

### DIFF
--- a/hack/run-e2e.sh
+++ b/hack/run-e2e.sh
@@ -174,7 +174,7 @@ function e2e::setup_helm_server() {
     if hack::version_ge $(e2e::get_kube_version) "v1.16.0"; then
         # workaround for https://github.com/helm/helm/issues/6374
         # TODO remove this when we can upgrade to helm 2.15+, see https://github.com/helm/helm/pull/6462
-        # \'$'\n  regex is used to be compatible with BSD sed (Darwin)
+        # \'$'\n is used to be compatible with BSD sed (Darwin)
         $HELM_BIN init --service-account tiller --output yaml \
             | sed 's@apiVersion: extensions/v1beta1@apiVersion: apps/v1@' \
             | sed 's@  replicas: 1@  replicas: 1\'$'\n  selector: {"matchLabels": {"app": "helm", "name": "tiller"}}@' \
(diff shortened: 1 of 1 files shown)


hack/run-e2e.sh
```diff
@@ -174,10 +174,18 @@ function e2e::setup_helm_server() {
...
         # workaround for https://github.com/helm/helm/issues/6374
         # TODO remove this when we can upgrade to helm 2.15+, see https://github.com/helm/helm/pull/6462
-        $HELM_BIN init --service-account tiller --output yaml \
+        if [[ "$OSTYPE" == "darwin"* ]]; then
+            # mac system
+            $HELM_BIN init --service-account tiller --output yaml \
+            | sed 's@apiVersion: extensions/v1beta1@apiVersion: apps/v1@' \
+            | sed 's@  replicas: 1@  replicas: 1\'$'\n  selector: {"matchLabels": {"app": "helm", "name": "tiller"}}@' \
+            | $KUBECTL_BIN --context $KUBECONTEXT apply -f -
+        else
```
cofyc, line 183: remove the Linux branch

GNU sed on Linux is compatible with BSD sed on Mac. So we can use the same method here.

mikechengwei: Under the mac system, \'$'\ n is correct, the original  \n is wrong. So do  i  replace origin regex directly ?

cofyc: yes

cofyc: also add a note that `\'$'\n` is used to be compatible with BSD sed (Darwin)

mikechengwei: ok.

hack/run-e2e.sh
```diff
@@ -174,9 +174,10 @@ function e2e::setup_helm_server() {
     if hack::version_ge $(e2e::get_kube_version) "v1.16.0"; then
         # workaround for https://github.com/helm/helm/issues/6374
         # TODO remove this when we can upgrade to helm 2.15+, see https://github.com/helm/helm/pull/6462
+        # \'$'\n  regex is used to be compatible with BSD sed (Darwin)
```
cofyc, line 177:
```suggestion
        # \'$'\n is used to be compatible with BSD sed (Darwin)
```

it's not regex actually.

mikechengwei: ok

Issue #3004, opened 2020-07-22
Patch PVC when storage request is increased


mightyguava:
## Feature Request

**Is your feature request related to a problem? Please describe:**
Increasing storage for tidb, pd, pump, monitor, require the user to manually patch PVCs. The `storage` setting in the CRDs only have effect when initially creating the cluster.

**Describe the feature you'd like:**
When the `storage` request for any component is increased, and the underlying storage-class supports expansion (`allowVolumeExpansion: true`), the operator should patch the PVC to increase the volume size automatically.

DanielZhangQD:
I think we can patch the PVC automatically, and users have to make sure that the underlying sc supports expansion.
The operator just patches the PVC according to the spec, if the sc does not support expansion, the patch just does not take effect. WDYT @cofyc

cofyc:
SGTM

cofyc:
related discussions:

- https://github.com/kubernetes/kubernetes/issues/68737
- https://github.com/kubernetes/enhancements/pull/660
- https://github.com/kubernetes/enhancements/issues/661
- https://github.com/kubernetes/enhancements/pull/1848

cofyc:
Current statefulset implementation does not allow `volumeClaimTemplates` to be changed, two possible workaround solutions:

#### recreate the statefulset, patch the PVCs

#### patch the PVCs only

mightyguava:
My intent with the feature request is to just patch the PVCs. I think that is sufficient.

cofyc:
hi, @mightyguava 

what volume plugin are you using? does it support online filesystem volume expansion? 
is it ok to patch all PVCs at the same time?

mightyguava:
We are using the in-tree aws EBS driver. I believe it require the pod to be restarted for the resize operation to run.

cofyc:
Thanks!

I didn't test with EKS. But according to [this blog](https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/#online-file-system-expansion), as for AWS-EBS, the pod does not need to be restarted. The feature `ExpandInUsePersistentVolumes` has been enabled by default since 1.15.

mightyguava:
We just upgraded to kubernetes 1.15 last week so I guess it’s on by default now! I will test it today.

mightyguava:
Just tried on our cluster, it didn't resize online.

<details>
  <summary>PVC</summary>

```
Name:          tikv-tidb-tikv-0
[...]

PR #3096, opened 2020-08-11
Patch PVCs when storage request is increased

cofyc:
fixes https://github.com/pingcap/tidb-operator/issues/3004
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Support patching PVCs when the storage request is increased. Due to the limitation of the current statefulset, new PVCs will be created with the old storage request and resized later
```

### DIFF
--- a/pkg/manager/member/pvc_resizer.go
+++ b/pkg/manager/member/pvc_resizer.go
@@ -58,7 +58,7 @@ import (
 //   https://github.com/kubernetes/enhancements/pull/1848) to be implemented.
 // - If the feature `ExpandInUsePersistentVolumes` is not enabled or the volume
 //   plugin does not support, the pod referencing the volume must be deleted and
-//   recreted after the `FileSystemResizePending` condition becomes true.
+//   recreated after the `FileSystemResizePending` condition becomes true.
 // - Shrinking volumes is not supported.
 //
 type PVCResizerInterface interface {
--- a/pkg/manager/member/pvc_resizer.go
+++ b/pkg/manager/member/pvc_resizer.go
@@ -1,0 +1,202 @@
+// Copyright 2020 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


cofyc:
cc @mightyguava

pkg/manager/member/pvc_resizer_test.go
```diff
@@ -0,0 +1,319 @@
...
+						},
+					},
+				},
+			},
+			sc: newStorageClass("sc", false),
+			pvcs: []*v1.PersistentVolumeClaim{
+				newPVCWithStorage("pd-0", label.PDLabelVal, "sc", "2Gi"),
+			},
+			wantPVCs: []*v1.PersistentVolumeClaim{
+				newPVCWithStorage("pd-0", label.PDLabelVal, "sc", "2Gi"),
```
mightyguava, line 277: nice test!

cofyc:
> can/should the operator also monitor that the volume expansion is eventually applied? I guess it can only do that if ExpandInUsePersistentVolumes is enabled, which it probably can't check

it can but the operator can only report some events, but users can watch the PVCs too. as for now, I think it's not necessary.

pkg/manager/member/pvc_resizer.go
```diff
@@ -0,0 +1,205 @@
...
+// Limitations:
+//
+// - Note that the current statfulset implementation does not allow
+//   `volumeClaimTemplates` to be changed, so new PVCs created by statefulset
+//   controller will use the old storage request.
+// - This is best effort, before statefulset volume resize feature (e.g.
+//   https://github.com/kubernetes/enhancements/pull/1848) to be implemented.
+// - If the feature `ExpandInUsePersistentVolumes` is not enabled or the volume
+//   plugin does not support, the pod referencing the volume must be deleted and
+//   recreted after the `FileSystemResizePending` condition becomes true.
```
DanielZhangQD, line 61:
```suggestion
//   recreated after the `FileSystemResizePending` condition becomes true.
```

Issue #3061, opened 2020-08-04
TLS support for heterogeneous components


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

mikechengwei:
I think heterogeneous cluster use tls feature just need to create the same secret as original cluster.

DanielZhangQD:
No need to create the secret anymore, we can just reuse the secret of the original cluster and update the code to get the secret name, for example, https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tidb_member_manager.go#L588 and https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tidb_member_manager.go#L594.

PR #3185, opened 2020-08-30
TLS support for heterogeneous components

mikechengwei:

#3061 
TLS support for heterogeneous components.
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test

Code changes

 - Has Go code change


AstroProfundis:
What would the pod and service names of heterogeneous components be? I'm not sure but suspects they might be not in the `DNSNames` list of the certs from original cluster.

mikechengwei:
Heterogeneous cluster container and service names depend on the cluster name, the same rules as origin cluster.
It is a problem that heterogeneous cluster dnsNames not exist in origin cluster certs .

mikechengwei:
@AstroProfundis Add a heterogeneous cluster to an existing cluster. Currently we choose the heterogeneous cluster to use the secret of the existing cluster, so we manually update the origin cluster.

AstroProfundis:
I see, so it's possible that pods in the original cluster can not connect to pods in the heterogeneous cluster, or pods from multiple heterogeneous clusters can not connect to each other.

We don't have PD in heterogeneous cluster so this might not be fatal, but it could be that some pods won't actually work at all. And the monitoring metrics of these pods in heterogeneous cluster may be unable to be collected.

Is it possible to renew the certs of original cluster when deploying a heterogeneous cluster and extend the DNS name list?

mikechengwei:
The current certs is manually configured by the user and create secrets. reference https://docs.pingcap.com/zh/tidb-in-kubernetes/stable/enable-tls-between-components.
If user config dnsName list by tcName field , they need to manually update the dnsName list and rolling update origin cluster before creating an heterogeneous cluster join the origin cluster.
I think it is okay for user.

DanielZhangQD:
@mikechengwei Did you test the following cases:
1. The heterogeneous cluster can join the cluster successfully with the TLS enabled. Please help add e2e for this case, you can refer to the case [here](https://github.com/pingcap/tidb-operator/blob/master/tests/e2e/tidbcluster/tidbcluster.go#L995).
2. The monitoring works as expected with TLS enabled.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -63,6 +64,7 @@ type tidbMemberManager struct {
 	podLister                    corelisters.PodLister
 	cmLister                     corelisters.ConfigMapLister
 	secretLister                 corelisters.SecretLister
+	tidbLister                   listers.TidbClusterLister
```
DanielZhangQD, line 67: Suggest using `tcLister` instead of `tidbLister`.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -551,16 +555,34 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
...
 			Name: "tidb-tls", ReadOnly: true, MountPath: clusterCertPath,
 		})
 	}
-	if tc.Spec.TiDB.IsTLSClientEnabled() {
+	if tc.Spec.TiDB.IsTLSClientEnabled() && !tc.IsHeterogeneous() {
 		volMounts = append(volMounts, corev1.VolumeMount{
 			Name: "tidb-server-tls", ReadOnly: true, MountPath: serverCertPath,
 		})
 	}
+	if tc.IsHeterogeneous() {
```
DanielZhangQD, line 568: I think we can combine this block with that from L627, no need to get the tc twice.

DanielZhangQD:
After further discussion with @mikechengwei, we think it's better to create separate TLS secret for the heterogenous clusters.
And we have to make sure that the following cases work well.  
1. The heterogeneous cluster can join the cluster successfully with the TLS enabled.
2. The monitoring works as expected with TLS enabled.

mikechengwei:
Reopen PR with #3253 .

Issue #3159, opened 2020-08-24
Sync tidbautoscaler with PD API


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Sync tidbautoscaler with PD API, please check detail in https://github.com/pingcap/tidb-operator/blob/f73551ea6c7cc979f729f03f7076b2ff7d181398/docs/design-proposals/2020-08-21-auto-scaling-with-pdapi.md#auto-scaling

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

howardlau1999:
Sync TidbMonitor clusters reference after clusters are created and deleted

howardlau1999:
Move resources definition in the `spec` to `tikv.spec` and `tidb.spec`

DanielZhangQD:
The sync logic would be like:
```
sync
  synctidb
    syncexternal
    syncpd
  synctikv
    syncexternal
    sycntikv
  syncMonitor
```

howardlau1999:
Check autoscaling interval according to status for each group

PR #3264, opened 2020-09-14
autoscaler: construct pd strategy request according to autoscaler spec

howardlau1999:
Signed-off-by: Howard Lau <howardlau1999@hotmail.com>

Construct pd strategy according to autoscaler spec

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

```release-note
NONE
```

### DIFF
--- a/pkg/autoscaler/autoscaler/autoscaler_manager.go
+++ b/pkg/autoscaler/autoscaler/autoscaler_manager.go
@@ -14,7 +14,6 @@
 package autoscaler
 
 import (
-	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
@@ -28,7 +27,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -86,13 +84,10 @@ func (am *autoScalerManager) Sync(tac *v1alpha1.TidbClusterAutoScaler) error {
 		}
 		return err
 	}
-	permitted, err := am.checkAutoScalerRef(tc, tac)
-	if err != nil {
-		klog.Error(err)
-		return err
-	}
-	if !permitted {
-		klog.Infof("tac[%s/%s]'s auto-scaling is no permitted", tac.Namespace, tac.Name)
+
+	defaultTAC(tac, tc)
+	if err := validateTAC(tac); err != nil {
+		klog.Errorf("invalid spec tac[%s/%s]: %s", tac.Namespace, tac.Name, err.Error())
 		return nil
 	}
 
@@ -107,7 +102,6 @@ func (am *autoScalerManager) Sync(tac *v1alpha1.TidbClusterAutoScaler) error {
 }
 
 func (am *autoScalerManager) syncAutoScaling(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler) error {
-	defaultTAC(tac)
 	oldTikvReplicas := tc.Spec.TiKV.Replicas
 	if err := am.syncTiKV(tc, tac); err != nil {
 		tc.Spec.TiKV.Replicas = oldTikvReplicas
@@ -199,46 +193,4 @@ func (am *autoScalerManager) updateTidbClusterAutoScaler(tac *v1alpha1.TidbClust
 		klog.Errorf("failed to update TidbClusterAutoScaler: [%s/%s], error: %v", ns, tacName, err)
 	}
 	return err
-}
-
-// checkAutoScalerRef will first check whether the target tidbcluster's auto-scaler reference have been occupied.
-// If it has been, and the reference scaler is the current auto-scaler itself, the auto-scaler would be permitted,
-// otherwise the auto-scaling would be forbidden.
-// If the target tidbcluster's auto-scaler reference is empty, then the auto-scaler will try to patch itself to the
-// references, and if the patching is success, the auto-scaling would discard the current syncing and wait for the next
-// syncing round.
-func (am *autoScalerManager) checkAutoScalerRef(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler) (bool, error) {
-	if tc.Status.AutoScaler != nil {
-		if tc.Status.AutoScaler.Name == tac.Name && tc.Status.AutoScaler.Namespace == tac.Namespace {
-			return true, nil
-		}
-		msg := fmt.Sprintf("tac[%s/%s]'s target tc[%s/%s] already controlled by another auto-scaler", tac.Namespace, tac.Name, tc.Namespace, tc.Name)
-		klog.Info(msg)
-		return false, nil
-	}
-	klog.Infof("tac[%s/%s]'s tc[%s/%s] start to occupy the auto-scaler ref", tac.Namespace, tac.Name, tc.Namespace, tc.Name)
-	err := am.patchAutoScalerRef(tc, tac)
-	return true, err
-}
-
-func (am *autoScalerManager) patchAutoScalerRef(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler) error {
-	mergePatch, err := json.Marshal(map[string]interface{}{
-		"status": map[string]interface{}{
-			"auto-scaler": map[string]interface{}{
-				"name":      tac.Name,
-				"namespace": tac.Namespace,
-			},
-		},
-	})
-	if err != nil {
-		klog.Error(err)
-		return err
-	}
-	_, err = am.cli.PingcapV1alpha1().TidbClusters(tc.Namespace).Patch(tc.Name, types.MergePatchType, mergePatch)
-	if err != nil {
-		klog.Error(err)
-		return err
-	}
-	msg := fmt.Sprintf("tac[%s/%s] patch itself to tc[%s/%s] auto-scaler ref success, do auto-scaling in next round", tac.Namespace, tac.Name, tc.Namespace, tc.Name)
-	return controller.RequeueErrorf(msg)
 }
--- a/pkg/autoscaler/autoscaler/util.go
+++ b/pkg/autoscaler/autoscaler/util.go
@@ -20,11 +20,16 @@ import (
 
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/label"
+	"github.com/pingcap/tidb-operator/pkg/pdapi"
 	operatorUtils "github.com/pingcap/tidb-operator/pkg/util"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog"
 	"k8s.io/utils/pointer"
 )
+
+var zeroQuantity = resource.MustParse("0")
 
 // checkStsAutoScalingPrerequisites would check the sts status to ensure wouldn't happen during
 // upgrading, scaling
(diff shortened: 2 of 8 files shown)


pkg/autoscaler/autoscaler/util.go
```diff
@@ -146,3 +147,51 @@ func genMetricsEndpoint(tac *v1alpha1.TidbClusterAutoScaler) (string, error) {
...
+func autoRulesToStrategyRule(component string, rules map[corev1.ResourceName]v1alpha1.AutoRule) *pdapi.Rule {
+	result := &pdapi.Rule{
+		Component: component,
+	}
+	for res, rule := range rules {
+		switch res {
+		case corev1.ResourceCPU:
+			result.CPURule = &pdapi.CPURule{
+				MaxThreshold:  rule.MaxThreshold,
+				MinThreshold:  *rule.MinThreshold,
```
howardlau1999, line 186: This should be defaulted so we will not dereference nil pointer

rleungx: How about adding a comment to explain why the min and max have different types?

pkg/autoscaler/autoscaler/util.go
```diff
@@ -146,3 +147,54 @@ func genMetricsEndpoint(tac *v1alpha1.TidbClusterAutoScaler) (string, error) {
 	}
 	return fmt.Sprintf("http://%s-prometheus.%s.svc:9090", tac.Spec.Monitor.Name, tac.Spec.Monitor.Namespace), nil
 }
+
+func autoscalerToStrategy(tac *v1alpha1.TidbClusterAutoScaler) *pdapi.Strategy {
```
DanielZhangQD, line 151: Suggest adding defaulting logic here as the code in this PR takes the assumption that all of the fields in the auto scaler CR have values.

Yisaer: This is unrelated to this issue. We can implement this feature later.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -146,3 +147,54 @@ func genMetricsEndpoint(tac *v1alpha1.TidbClusterAutoScaler) (string, error) {
...
+			result.CPURule = &pdapi.CPURule{
+				MaxThreshold:  rule.MaxThreshold,
+				MinThreshold:  *rule.MinThreshold,
+				ResourceTypes: rule.ResourceTypes,
+			}
+		case corev1.ResourceStorage:
+			result.StorageRule = &pdapi.StorageRule{
+				MinThreshold:  1.0 - rule.MaxThreshold,
+				ResourceTypes: rule.ResourceTypes,
+			}
```
Yisaer, line 196: We should directly deliver the pass `MinThreshold` to PD because user should directly define `MinThreshold`, so we can't get `MinThreshold` by calculating `1 - MaxThreshold` which also means `MaxThreshold` and `MinThreshold` should both be pointer and its existing should be verified by validating logic.

Yisaer (commented):
The rest LGTM. BTW, we don't need to handle the storage case, only `cpu` case is enough currently.

pkg/autoscaler/autoscaler/util.go
rleungx: IMO, we can keep it for now even if we don't take it into account.

pkg/autoscaler/autoscaler/util.go
DanielZhangQD: In this case, we have to review the codes in this PR later again as we're not sure if the values are defaulted correctly later.
Or we can have a separate PR to do the defaulting at the same time.

pkg/autoscaler/autoscaler/util.go
DanielZhangQD: 1. Agree with @rleungx to keep the storage code block, actually, there is very little extra support to support storage as it shares most of the logic with CPU.
2. For the storage, users should set MaxThreshold in the Operator API, as in this case, the MaxThreshold has the same meaning for both the CPU and storage rules, that is, when the usage exceeds the MaxThreshold, we will do auto-scaling, otherwise, we have to explain that, for CPU, we will scale when the usage exceeds the MaxThreshold and for storage, we will scale when the available space is less than the minThreshold, which is more complicated for users to understand than the current solution.

howardlau1999: I agree with the current design and we can keep this code

pkg/autoscaler/autoscaler/util.go
howardlau1999: Will implement in this PR

rleungx:
PTAL @Yisaer @DanielZhangQD

pkg/autoscaler/autoscaler/util_test.go
```diff
@@ -307,6 +339,46 @@ func TestGenMetricsEndpoint(t *testing.T) {
...
+			CPU:          resource.MustParse("1000m"),
+			Memory:       resource.MustParse("8Gi"),
+			Storage:      resource.MustParse("1000Gi"),
+			Count:        pointer.Int32Ptr(2),
+		},
+		{
+			ResourceType: "resource_b",
+			CPU:          resource.MustParse("2000m"),
+			Memory:       resource.MustParse("4Gi"),
+			Storage:      resource.MustParse("2000Gi"),
```
DanielZhangQD, line 357: storage may be empty.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -91,11 +93,54 @@ func limitTargetReplicas(targetReplicas int32, tac *v1alpha1.TidbClusterAutoScal
...
+
+		if tac.Spec.TiDB != nil {
+			defResources(&tac.Spec.Resources, v1alpha1.TiDBMemberType)
+		}
+	}
+
+	defResourceTypes := func(rule *v1alpha1.AutoRule) {
+		if len(rule.ResourceTypes) == 0 {
+			for _, res := range tac.Spec.Resources {
+				rule.ResourceTypes = append(rule.ResourceTypes, res.ResourceType)
```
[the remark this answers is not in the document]
DanielZhangQD: Especially for the `default_tidb`, whose storage would be 0 and should not be used as the resource type for TiKV.

pkg/autoscaler/autoscaler/util_test.go
DanielZhangQD: For the definition below, I think we should remove the `omitempty` for `ResourceType`, `CPU` and `Memory`.
For `Storage`, it's not necessary for TiDB but necessary for TiKV, WDYT @howardlau1999 @rleungx, how PD will handle this case now?
```
// AutoResource describes the resource type definitions
type AutoResource struct {
	// ResourceType identifies a specific resource type
	ResourceType string `json:"resource_type,omitempty"`
	// CPU defines the CPU of this resource type
	CPU resource.Quantity `json:"cpu,omitempty"`
	// Memory defines the memory of this resource type
	Memory resource.Quantity `json:"memory,omitempty"`
	// Storage defines the storage of this resource type
	Storage resource.Quantity `json:"storage,omitempty"`
	// Count defines the max availabel count of this resource type
	Count *int32 `json:"count,omitempty"`
}
```

pkg/autoscaler/autoscaler/util.go
```diff
@@ -91,11 +93,54 @@ func limitTargetReplicas(targetReplicas int32, tac *v1alpha1.TidbClusterAutoScal
 // If the minReplicas not set, the default value would be 1
 // If the Metrics not set, the default metric will be set to 80% average CPU utilization.
 // defaultTAC would default the omitted value
-func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
+func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler, tc *v1alpha1.TidbCluster) {
 	if tac.Annotations == nil {
 		tac.Annotations = map[string]string{}
 	}
 
+	defResources := func(resources *[]v1alpha1.AutoResource, component v1alpha1.MemberType) {
```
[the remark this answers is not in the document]
howardlau1999: it's ok to move out

pkg/autoscaler/autoscaler/util_test.go
howardlau1999: PD does not take `storage` into account for now

pkg/autoscaler/autoscaler/util.go
```diff
@@ -91,11 +93,54 @@ func limitTargetReplicas(targetReplicas int32, tac *v1alpha1.TidbClusterAutoScal
 // If the minReplicas not set, the default value would be 1
 // If the Metrics not set, the default metric will be set to 80% average CPU utilization.
 // defaultTAC would default the omitted value
-func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
+func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler, tc *v1alpha1.TidbCluster) {
```
DanielZhangQD, line 96: Suggest adding some basic validation, e.g. min<max<1

pkg/autoscaler/autoscaler/util_test.go
howardlau1999: Removing the `omitempty` for `ResourceType`, `CPU` and `Memory` is reasonable IMO

pkg/autoscaler/autoscaler/util.go
howardlau1999: what should we do if the validating failed, e.g. min > max

pkg/autoscaler/autoscaler/util.go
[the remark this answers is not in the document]
howardlau1999: what should we do if user has specified serveral resources with some having 0 storage and does not specify resource types for tikv?

pkg/autoscaler/autoscaler/util.go
DanielZhangQD: We can move the defaulting logic from https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L110 to the function https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L70, after the defaulting logic, call the validation function, if the validation fails, just print the detailed error and return nil, no need to retry the sync for this autoscaler CR.

DanielZhangQD: BTW, could you please also help remove the logic https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/autoscaler_manager.go#L89-L97, it's not necessary to update and check the TidbCluster status here.

pkg/autoscaler/autoscaler/util_test.go
rleungx: Also, PD doesn't handle the memory for now, I think we can remove it.

pkg/autoscaler/autoscaler/util.go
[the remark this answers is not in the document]
DanielZhangQD: 1. Should we default to `tikv.requests.storage` in TC?
2. Ignore them and use the `default_tikv`
WDYT? @howardlau1999 @rleungx 

BTW, I think we should define this field https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/tidbclusterautoscaler_types.go#L114 to be a map with the key as the resource_type, otherwise, users may configure two resources with the same resource_type.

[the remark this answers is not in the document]
howardlau1999: Default to `tikv.requests.storage` is unreasonable because user may not have such resources. And we don't have `default_tikv` when user has defined some resources.

How about:
1. If user does not provide a resource with storage and specifies TiKV autoscaling, the configuration is invalid and provides an error message
2. If user provides some resources with storage and some without, we can use only those with storage when defaulting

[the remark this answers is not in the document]
howardlau1999: Change resources to `map` is OK

Yisaer (dismissed):
LGTM

pkg/autoscaler/autoscaler/util.go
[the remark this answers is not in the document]
howardlau1999: @DanielZhangQD WDYT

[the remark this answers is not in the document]
DanielZhangQD: > Default to `tikv.requests.storage` is unreasonable because user may not have such resources. And we don't have `default_tikv` when user has defined some resources.
> 
> How about:
> 
> 1. If user provides custom resource types but does not provide a resource with storage and specifies TiKV autoscaling, the configuration is invalid and we should show an error message
> 2. If user provides some resources with storage and some without, we can use only those with storage when defaulting

OK. Then we also have to validate the storage if users have provided resource types in the rules.
cc @rleungx

pkg/autoscaler/autoscaler/util.go
```diff
@@ -137,6 +203,80 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+					}
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
+func validateTAC(tac *v1alpha1.TidbClusterAutoScaler) error {
```
rleungx, line 258: We need to add some tests for invalid cases. Maybe add TODO here?

howardlau1999: updated

rleungx (commented):
LGTM

pkg/autoscaler/autoscaler/autoscaler_manager.go
```diff
@@ -86,15 +84,11 @@ func (am *autoScalerManager) Sync(tac *v1alpha1.TidbClusterAutoScaler) error {
 		}
 		return err
 	}
-	permitted, err := am.checkAutoScalerRef(tc, tac)
-	if err != nil {
-		klog.Error(err)
+
+	defaultTAC(tac, tc)
+	if err := validateTAC(tac); err != nil {
 		return err
```
DanielZhangQD, line 90: Do not return error here, just print the error log and return nil, if we return err here, this CR will be synced again very soon.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -88,36 +93,97 @@ func limitTargetReplicas(targetReplicas int32, tac *v1alpha1.TidbClusterAutoScal
...
-			spec.MetricsTimeDuration = pointer.StringPtr("3m")
+
+		if tac.Spec.TiDB != nil {
+			defaultResources(tc, tac, v1alpha1.TiDBMemberType)
 		}
 	}
 
 	if tidb := tac.Spec.TiDB; tidb != nil {
-		def(&tidb.BasicAutoScalerSpec)
+		defaultBasicAutoScaler(tac, &tidb.BasicAutoScalerSpec, v1alpha1.TiDBMemberType)
```
DanielZhangQD, line 182: The `TiDBMemberType ` should be able to identify the `tidb.BasicAutoScalerSpec`.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -88,36 +93,97 @@ func limitTargetReplicas(targetReplicas int32, tac *v1alpha1.TidbClusterAutoScal
...
 	}
 
 	if tikv := tac.Spec.TiKV; tikv != nil {
-		def(&tikv.BasicAutoScalerSpec)
+		defaultBasicAutoScaler(tac, &tikv.BasicAutoScalerSpec, v1alpha1.TiKVMemberType)
```
DanielZhangQD, line 186: ditto

pkg/autoscaler/autoscaler/util.go
```diff
@@ -137,6 +203,84 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
 	}
 }
 
+func validateBasicAutoScalerSpec(tac *v1alpha1.TidbClusterAutoScaler, spec *v1alpha1.BasicAutoScalerSpec, component v1alpha1.MemberType) error {
```
DanielZhangQD, line 206: ditto

pkg/autoscaler/autoscaler/util.go
```diff
@@ -137,6 +203,84 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+			return err
+		}
+
+		switch res {
+		case corev1.ResourceCPU:
+			if *rule.MinThreshold > 1.0 || *rule.MinThreshold < 0.0 {
+				return fmt.Errorf("min threshold should be between 0 and 1 for rule %s of %s in %s/%s", res, component.String(), tac.Namespace, tac.Name)
+			}
+			if *rule.MinThreshold > rule.MaxThreshold {
+				return fmt.Errorf("min threshold > max threshold for cpu rule of %s in %s/%s", component.String(), tac.Namespace, tac.Name)
```
DanielZhangQD, line 245: Please also print the values of the thresholds.
Please also check the other logs.

pkg/autoscaler/autoscaler/util.go
```diff
@@ -137,6 +203,84 @@ func defaultTAC(tac *v1alpha1.TidbClusterAutoScaler) {
...
+		corev1.ResourceCPU:     {},
+		corev1.ResourceStorage: {},
+	}
+
+	checkCommon := func(res corev1.ResourceName, rule v1alpha1.AutoRule) error {
+		if _, ok := acceptableResources[res]; !ok {
+			return fmt.Errorf("unknown resource type %s of %s in %s/%s", res.String(), component.String(), tac.Namespace, tac.Name)
+		}
+		if rule.MaxThreshold > 1.0 || rule.MaxThreshold < 0.0 {
+			return fmt.Errorf("max threshold should be between 0 and 1 for rule %s of %s in %s/%s", res, component.String(), tac.Namespace, tac.Name)
```
DanielZhangQD, line 221: Print the threshold values.

pkg/autoscaler/autoscaler/util_test.go
```diff
@@ -307,6 +337,108 @@ func TestGenMetricsEndpoint(t *testing.T) {
...
+	tac.Spec.TiKV.Rules[corev1.ResourceStorage] = v1alpha1.AutoRule{
+		MaxThreshold:  0.8,
+		ResourceTypes: []string{"resource_a"},
+	}
+	strategy := autoscalerToStrategy(tac)
+	g.Expect(len(strategy.Resources)).Should(Equal(3))
+	g.Expect(len(strategy.Rules)).Should(Equal(2))
+}
+
+func TestValidateTidbClusterAutoScaler(t *testing.T) {
```
DanielZhangQD, line 382: Please make sure that the new functions are covered in the UT.

DanielZhangQD (approved):
LGTM

ti-srebot (approved):

PR #3265, opened 2020-09-14
autoscaler: setup framework for syncing pd plans

howardlau1999:

Based on https://github.com/pingcap/tidb-operator/pull/3264, added framework for syncing pd autoscaling plans and removed unused code.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

```release-note
NONE
```


howardlau1999:
Please review https://github.com/pingcap/tidb-operator/pull/3264 first

pkg/autoscaler/autoscaler/autoscaler_manager.go
```diff
@@ -55,14 +56,15 @@ func NewAutoScalerManager(
 	kubeInformerFactory kubeinformers.SharedInformerFactory,
 	recorder record.EventRecorder) *autoScalerManager {
 	tcLister := informerFactory.Pingcap().V1alpha1().TidbClusters().Lister()
-	stsLister := kubeInformerFactory.Apps().V1().StatefulSets().Lister()
 	return &autoScalerManager{
 		kubecli:   kubecli,
 		cli:       cli,
 		tcLister:  tcLister,
```
rleungx, line 62: How about
```suggestion
		tcLister:  informerFactory.Pingcap().V1alpha1().TidbClusters().Lister(),
```

howardlau1999: When creating `tcControl`, we also need `tcLister` and I think creating one is enough

Makefile
```diff
@@ -90,7 +90,6 @@ e2e-build:
 	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/webhook ./tests/cmd/webhook
 	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/blockwriter ./tests/cmd/blockwriter
 	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/apiserver ./tests/cmd/apiserver
-	$(GO_BUILD) -ldflags '$(LDFLAGS)' -o tests/images/e2e/bin/mock-prometheus ./tests/cmd/mock-monitor
```
DanielZhangQD, line 93: Why remove this?

pkg/autoscaler/autoscaler/autoscaler_manager.go
```diff
@@ -100,46 +102,30 @@ func (am *autoScalerManager) Sync(tac *v1alpha1.TidbClusterAutoScaler) error {
...
-func (am *autoScalerManager) syncTidbClusterReplicas(tac *v1alpha1.TidbClusterAutoScaler, tc *v1alpha1.TidbCluster, oldTc *v1alpha1.TidbCluster) error {
-	if tc.Spec.TiDB.Replicas == oldTc.Spec.TiDB.Replicas && tc.Spec.TiKV.Replicas == oldTc.Spec.TiKV.Replicas {
-		return nil
-	}
-	newTc := tc.DeepCopy()
-	_, err := am.tcControl.UpdateTidbCluster(newTc, &newTc.Status, &oldTc.Status)
+	// Request PD for auto-scaling plans
+	plans, err := controller.GetPDClient(am.pdControl, tc).GetAutoscalingPlans(*strategy)
 	if err != nil {
+		klog.Errorf("cannot get auto-scaling plans err:%v", err)
```
DanielZhangQD, line 118: Please add the tac namespace and name.

DanielZhangQD: Please also check the other logs.

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -0,0 +1,93 @@
...
+
+const groupLabelKey = "group"
+
+func (am *autoScalerManager) syncPlans(tc *v1alpha1.TidbCluster, tac *v1alpha1.TidbClusterAutoScaler, plans []pdapi.Plan) error {
+	planGroups := sets.String{}
+	groupPlanMap := make(map[string]pdapi.Plan)
+	for _, plan := range plans {
+		groupName := plan.Labels[groupLabelKey]
+		planGroups.Insert(groupName)
+		groupPlanMap[groupName] = plan
```
DanielZhangQD, line 33: There is once an issue that we have discussed which needs PD to support:
For the same metric and the same component, if only one resrouce_type is specified and then users updated its resource_type, in the following responses, it will include both of the groups.
So, in this case, the gourpName may need to including the resources info?

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	// Filter all autoscaled TidbClusters
+	requirement, err := labels.NewRequirement(label.AutoScalingGroupLabelKey, selection.Exists, nil)
+	if err != nil {
+		return err
+	}
+	selector := labels.NewSelector().Add(*requirement)
+	tcList, err := am.tcLister.List(selector)
```
DanielZhangQD, line 42: We need to add namespace here.

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return err
+	}
+
+	existedGroups := sets.String{}
+	groupTcMap := make(map[string]*v1alpha1.TidbCluster)
+	for _, tc := range tcList {
+		groupName := tc.Labels[label.AutoScalingGroupLabelKey]
```
DanielZhangQD, line 50: Consider the case that the label does not exist?

pkg/autoscaler/autoscaler/query/external.go
```diff
@@ -1,113 +0,0 @@
-// Copyright 2020 PingCAP, Inc.
```
DanielZhangQD, line 1: We still support external, keep this file or you want to do it all over again?

tests/cmd/mock-monitor/main.go
```diff
@@ -1,31 +0,0 @@
-// Copyright 2020 PingCAP, Inc.
```
DanielZhangQD, line 1: why remove this?

tests/pkg/mock/monitor.go
```diff
@@ -1,150 +0,0 @@
-// Copyright 2020 PingCAP, Inc.
```
DanielZhangQD, line 1: Can we keep these mock files in case we may use them later?

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
howardlau1999: The label selector makes sure it exist

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
Yisaer: Identifying the configuration for the same `resource_type` during the instances in groups is hard to maintain in PD and PD won't support this feature. Currently, I think we can `ignore situtation if resource_type configuraiton changed` because that is not our main target and it doesn't block our main process. 

To solve this problem, I think there are many ways, and we can discuss it later, but not now.

Yisaer (commented):
I think the mockPrometheus is necessary. Though operator fetch the information from PD instead of monitor, we still need to mock the data and register it into ETCD. So we can remain mockPrometheus and register it into PD.

howardlau1999:
Restored the mock prometheus

pkg/autoscaler/autoscaler/autoscaler_manager.go
howardlau1999: updated

pkg/autoscaler/autoscaler/query/external.go
howardlau1999: i will keep this file

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
howardlau1999: updated

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
DanielZhangQD: so what's the behavior for this issue now?

DanielZhangQD: And we ever had a deal during the design review @rleungx

DanielZhangQD: OK, discussed this with @rleungx, PD only supports one resource type now.

pkg/autoscaler/autoscaler/autoscaler_manager.go
```diff
@@ -55,14 +55,15 @@ func NewAutoScalerManager(
...
-	stsLister := kubeInformerFactory.Apps().V1().StatefulSets().Lister()
 	return &autoScalerManager{
 		kubecli:   kubecli,
 		cli:       cli,
 		tcLister:  tcLister,
 		tcControl: controller.NewRealTidbClusterControl(cli, tcLister, recorder),
+		pdControl: pdapi.NewDefaultPDControl(kubecli),
 		taLister:  informerFactory.Pingcap().V1alpha1().TidbClusterAutoScalers().Lister(),
-		stsLister: stsLister,
+		tmLister:  informerFactory.Pingcap().V1alpha1().TidbMonitors().Lister(),
```
Yisaer, line 65: Is this used?

Yisaer (commented):
Rest LGTM

pkg/autoscaler/autoscaler/autoscaler_manager.go
howardlau1999: Will be used when creating or deleting clusters to update tidbmonitors

Yisaer (approved):
LGTM

pkg/autoscaler/autoscaler/autoscaler_manager.go
DanielZhangQD: Why do we need to update tidbmonitors in the autoscaler controller?

howardlau1999: Because we need to add the created clusters to the monitor so that PD can fetch metrics

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -0,0 +1,95 @@
...
+		if len(groupName) == 0 {
+			klog.Warningf("tidbcluster [%s/%s] has empty value for label %s", tc.Namespace, tc.Name, label.AutoScalingGroupLabelKey)
```
rleungx, line 53: Is warning enough?

rleungx (commented):
The rest LGTM

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
howardlau1999: Updated to error

rleungx:
@DanielZhangQD PTAL

DanielZhangQD (approved):
LGTM

ti-srebot (approved):

PR #3267, opened 2020-09-14
autoscaler: update autoscaling clusters

howardlau1999:

Based on https://github.com/pingcap/tidb-operator/pull/3265 , implement update function for PD plan autoscaler.

### Check List 

Tests 

 - Manual test (refer to https://github.com/pingcap/tidb-operator/pull/3269)

Code changes

 - Has Go code change

Side effects

```release-note
NONE
```


howardlau1999:
Review https://github.com/pingcap/tidb-operator/pull/3265 first

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -80,11 +82,51 @@ func (am *autoScalerManager) deleteAutoscalingClusters(tc *v1alpha1.TidbCluster,
...
+		actual, oldTc, expected := groupTcMap[group].DeepCopy(), groupTcMap[group], groupPlanMap[group]
+		component := expected.Component
+
+		if !checkAutoscalingComponent(tac, component) {
+			continue
+		}
+
+		switch component {
+		case v1alpha1.TiKVMemberType.String():
+			sts, err := am.stsLister.StatefulSets(actual.Namespace).Get(operatorUtils.GetStatefulSetName(actual, v1alpha1.TiKVMemberType))
```
DanielZhangQD, line 96: why should care about the sts here, auto scaler controller should only make sure that the required TC is created.

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -80,11 +82,51 @@ func (am *autoScalerManager) deleteAutoscalingClusters(tc *v1alpha1.TidbCluster,
...
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			if !checkAutoScalingPrerequisites(actual, sts, v1alpha1.TiKVMemberType) {
```
DanielZhangQD, line 101: what's the point to call this function here?

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -80,11 +82,51 @@ func (am *autoScalerManager) deleteAutoscalingClusters(tc *v1alpha1.TidbCluster,
...
+				continue
+			}
+			if actual.Spec.TiKV.Replicas == int32(expected.Count) {
```
DanielZhangQD, line 104: So for the same group, the response from PD only differs in the count, right?

DanielZhangQD, line 104: What makes them different?

howardlau1999: Yes

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
howardlau1999: removed

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
howardlau1999: removed

pkg/autoscaler/autoscaler/util.go
```diff
@@ -344,3 +344,13 @@ func autoRulesToStrategyRule(component string, rules map[corev1.ResourceName]v1a
 	}
 	return result
 }
+
+func checkAutoscalingComponent(tas *v1alpha1.TidbClusterAutoScaler, component string) bool {
```
rleungx, line 348: How about putting this logic in line 99?

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -85,8 +86,35 @@ func (am *autoScalerManager) deleteAutoscalingClusters(tc *v1alpha1.TidbCluster,
...
+		actual, oldTc, plan := groupTcMap[group].DeepCopy(), groupTcMap[group], groupPlanMap[group]
+		component := plan.Component
+
+		if !checkAutoscalingComponent(tac, component) {
+			continue
+		}
+
+		switch component {
+		case v1alpha1.TiKVMemberType.String():
+			if actual.Spec.TiKV.Replicas == int32(plan.Count) {
```
rleungx, line 100: Maybe we don't need it?

pkg/autoscaler/autoscaler/util.go
howardlau1999: line 99 in which file?

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
howardlau1999: we can reduce the number of API calls to k8s if they are the same

pkg/autoscaler/autoscaler/util.go
rleungx: in `pdplan_autoscaler.go`

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
rleungx: make sense

pkg/autoscaler/autoscaler/util.go
howardlau1999: the function will be also used when creating autoscaling clusters. do you mean we put this function in that file?

pkg/autoscaler/autoscaler/autoscaler_manager.go
```diff
@@ -95,45 +97,80 @@ func (am *autoScalerManager) Sync(tac *v1alpha1.TidbClusterAutoScaler) error {
...
+			return err
+		}
+
+		if tc.Spec.TiDB.Replicas == targetReplicas {
+			return nil
+		}
+
+		updated := tc.DeepCopy()
+		updated.Spec.TiDB.Replicas = targetReplicas
+		if _, err = am.tcControl.UpdateTidbCluster(updated, &updated.Status, &tc.Status); err != nil {
```
DanielZhangQD, line 119: With the new design, we should also create new TidbCluster for the external query, so this will be done in the following PRs?

pkg/autoscaler/autoscaler/autoscaler_manager.go
```diff
@@ -100,8 +101,59 @@ func (am *autoScalerManager) Sync(tac *v1alpha1.TidbClusterAutoScaler) error {
...
+			return err
+		}
+	}
+
+	if tac.Spec.TiKV != nil && tac.Spec.TiKV.External != nil {
+		if err := am.syncExternal(tc, tac, v1alpha1.TiKVMemberType); err != nil {
+			return err
+		}
+	}
+
```
DanielZhangQD, line 156: If users configures `external` for TiDB/TiKV, should we still go through to the following code to call PD API?

DanielZhangQD: The https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/util.go#L160 and https://github.com/pingcap/tidb-operator/blob/master/pkg/autoscaler/autoscaler/util.go#L259 functions do not consider the external configuration.

pkg/autoscaler/autoscaler/autoscaler_manager.go
howardlau1999: yes, the hust team wants us to recover the external service asap

pkg/autoscaler/autoscaler/autoscaler_manager.go
howardlau1999: If user configures `external` for one component, it will be ignored when constructing PD API requests. If the constructed PD strategy has no rules, then we will skip making PD API requests.

howardlau1999: The defaulting function and the validating function have been updated

DanielZhangQD: > If user configures external for one component, it will be ignored when constructing PD API requests. If the constructed PD strategy has no rules, then we will skip making PD API requests.

Where is the PD API skipped?
What's the behavior of the following cases in PD (Y means exist, N means non-exist):
TiDB rules TiKV rules
Y Y
N Y
Y N
N N

DanielZhangQD (approved):
LGTM

pkg/autoscaler/autoscaler/pdplan_autoscaler.go
```diff
@@ -84,9 +87,34 @@ func (am *autoScalerManager) deleteAutoscalingClusters(tc *v1alpha1.TidbCluster,
...
+		case v1alpha1.TiDBMemberType.String():
+			if tac.Spec.TiDB == nil || actual.Spec.TiDB.Replicas == int32(plan.Count) {
+				continue
+			}
+			actual.Spec.TiDB.Replicas = int32(plan.Count)
+		default:
+			errs = append(errs, fmt.Errorf("unexpected component %s for group %s in autoscaling plan", plan.Component, group))
+			continue
+		}
+
```
DanielZhangQD, line 110: We should also check and update TidbMonitor here if necessary.

DanielZhangQD: We should add a function `syncMonitor` to sync the TidbMonitor out of the TC sync function.

Yisaer (approved):
LGTM

ti-srebot (approved):

DanielZhangQD (approved):
LGTM

Issue #2868, opened 2020-07-05
Support DM 2.0 in TiDB Operator


Ishiihara:
## Description

Integrates DM with TiDB Operator. Ideally, we would like TiDB Operator to manage DM as well.

### Integration Method

Add dm controller under the logic of the existing TiDB operator

### Deployment

1. No longer write a dm-operator alone, but add resource dmclusters in tidb-operator
2. `dm-master` and `dm-worker` are deployed through statefulsets
3. The dm-controller is only responsible for managing the cluster, and the subsequent dm task management requires the user to operate through the http/grpc api of the dm-master service, through the dm-ctl packaged in the image or run `dm-ctl` locally through `kubectl port-forward`.

### Configuration

Deployment configurations should be saved through k8s `configmaps` and updated by the mechanism of statefulsets.

### Service

`dm-master` should expose service `NodePort` method. 

#### dm-master

With the help of the partition of statefulset, rolling update in order from large to small numbers, check whether the POD is the latest state, and then check whether the POD is the leader. If it is the leader, deploy the leader migration, and then set it as the current POD sequence number partition.

#### dm-worker

Update directly.

### Scale

For scale-in operation, **delete member info first**, and then delete pod from cluster.

### High Availability

Achieved by DM's architecture.

#### dm-master

Assuming that the cluster has 3 master pods, if a dm-master pod goes down for more than 5 minutes (configurable), the operator will add a new dm-master pod. At this time, 4 pods will exist at the same time. After the failed dm-master node is restored, the operator will delete the newly started node. Then cluster will still have 3 master pods.

#### dm-worker

Almost the same as the above failover processing. The difference is that a new dm-worker may be started and the dm task has been assigned. If the worker goes offline at this time, the task will be re-assigned which will cause the task interrupted for a little while. Therefore, operator will not delete the newly started node, but keep it as 4 nodes. If users enables advanced-statefulsets, we can delete the intermediate node.

### Monitor

Add the container containing the new dm monitoring file to the existing tidbMonitor pod. When deploying the monitoring of the dm cluster, copy the promethus and grafana configs to the target volumns.

### Log

Reuse the current EFK system.

## Category

Feature

## Value

Support dm-controller in TiDB Operator so that users can easily deploy, scale, upgrade dm cluster through TiDB Operator.

## TODO lists

* [x] dmclusters CRD definition
  * [x] Define the dmclusters cluster spec, including the dm-master's service, configmap, statefulsets and other k8s resource definitions
[...]

PR #3271, opened 2020-09-15
dm-operator/: support tls for dm cluster and sources

lichunzhu:
https://github.com/pingcap/tidb-operator/issues/2868
1. Support cluster's tls between dm-masters and dm-workers.
2. Support MySQL client tls for upstream MySQL databases and downstream TiDB databases.

1. For cluster's tls, just read tls secret like `TidbCluster`.
2. For MySQL client tls, we need to mount tls to **every** dm-master and dm-worker pod. Because every master and worker is possible to use the added source. Every database source's TLS will be mounted to directory `/var/lib/source-tls/${secret_name}`. Customers need to configure TLS path to this format correctly in `source.yaml` and `dm-task.yaml`.

### Check List 

Tests 

 - Unit test
 - E2E test
  1. Test replicating data from upstream MySQL to downstream TiDB. Both two databases enable TLS. DM can replicate data correctly.
  2. Test start dm cluster with TLS enabled. dm-cluster can start with tls successfully.

Code changes

 - Has Go code change

```release-note
NONE
```


lonng (commented):
LGTM

lichunzhu:
@DanielZhangQD PTAL

pkg/util/crypto/certs.go
```diff
@@ -111,7 +111,9 @@ func LoadTlsConfigFromSecret(secret *corev1.Secret) (*tls.Config, error) {
 	rootCAs := x509.NewCertPool()
 	var tlsCert tls.Certificate
 
-	rootCAs.AppendCertsFromPEM(secret.Data[corev1.ServiceAccountRootCAKey])
+	if !rootCAs.AppendCertsFromPEM(secret.Data[corev1.ServiceAccountRootCAKey]) {
+		return nil, fmt.Errorf("failed to append ca certs")
```
july2993, line 115: do we need to use pingcap/errors for stack trace ?

july2993 (approved):
```
Every database source's TLS will be mounted to directory /var/lib/source-tls/${secret_name}. Customers need to configure TLS path to this format correctly in source.yaml and dm-task.yaml
```
need to document this rest LGTM

pkg/util/crypto/certs.go
lichunzhu: Maybe we can add more info like this issue? https://github.com/pingcap/tidb-operator/issues/1936

july2993 (commented):
LGTM

ti-srebot (approved):

Issue #3303, opened 2020-09-23
fail to generate the desired config for pump with float type config item


july2993:
## Bug Report

**What version of Kubernetes are you using?**

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T21:54:15Z", GoVersion:"go1.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of TiDB Operator are you using?**

2ec8dfbf05fbb3513aea878ce4c2c13bf64a97b2  at master

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

**What's the status of the TiDB cluster pods?**

**What did you do?**

`kubectl apply -f tidb-cluster.yaml`
<details><summary>CLICK ME for tidb-cluster.yaml</summary>

```yaml
apiVersion: pingcap.com/v1alpha1
kind: TidbCluster
metadata:
  name: basic
spec:
  version: v4.0.6
  timezone: UTC
  pvReclaimPolicy: Delete
  discovery: {}
  configUpdateStrategy: RollingUpdate
  pd:
    baseImage: pingcap/pd
    replicas: 1
    # if storageClassName is not set, the default Storage Class of the Kubernetes cluster will be used
    # storageClassName: local-storage
    requests:
      storage: "1Gi"
    config: {}
  tikv:
    baseImage: pingcap/tikv
    replicas: 1
    # if storageClassName is not set, the default Storage Class of the Kubernetes cluster will be used
    # storageClassName: local-storage
    requests:
      storage: "1Gi"
    config:
      storage:
        # In basic examples, we set this to avoid using too much storage.
        reserve-space: "2MB"
  tidb:
    baseImage: pingcap/tidb
    replicas: 1
    service:
      type: ClusterIP
    config:
      token-limit: 202
      max-server-connections: 51
  pump:
    baseImage: pingcap/tidb-binlog
    replicas: 1
    # storageClassName: local-storage
    requests:
[...]

PR #3312, opened 2020-09-24
Support pass raw toml config for pump as string

july2993:
Support pass raw toml config for pump as string

issue in #3303 still exist if config in the origin way.

- enhance the `GenericConfig` to support unmarshal from both string and object of JSON data.
- add some useful method to get&set value of the `GenericConfig`
- avoid code the json tag inside the `GenericConfig` map, so we can use it with a specified field name if need later.

### Check List 

Tests 

 
test with:
Screen Shot 2020-09-24 at 12 01 06 PM
check cm and pump works
Screen Shot 2020-09-24 at 12 00 40 PM

also, test with the origin yaml way.

Related changes

```release-note
Support passing raw toml config for the pump as a string
```


pkg/util/config/config.go
```diff
@@ -45,5 +114,166 @@ func (c *GenericConfig) DeepCopy() *GenericConfig {
...
+
+	default:
+		return 0, errors.Errorf("type is %v not integer", reflect.TypeOf(v.inner))
+	}
+}
+
+func (v *Value) MustFloat() float64 {
+	value, err := v.AsFloat()
+	if err != nil {
+		panic(err)
```
DanielZhangQD, line 201: Is it reasonable to panic here?
This will cause the operator process panic, right?

DanielZhangQD:
@july2993 Can the original config 
```
    config:
      addr: 0.0.0.0:8250
      gc: 7
      heartbeat-interval: 2
      security:
        cert-allowed-cn:
        - TiDB
        ssl-ca: /var/lib/pump-tls/ca.crt
        ssl-cert: /var/lib/pump-tls/tls.crt
        ssl-key: /var/lib/pump-tls/tls.key
```
be converted to the new format gracefully without triggering a rolling update?

july2993:
> @july2993 Can the original config
> 
> ```
>     config:
>       addr: 0.0.0.0:8250
>       gc: 7
>       heartbeat-interval: 2
>       security:
>         cert-allowed-cn:
>         - TiDB
>         ssl-ca: /var/lib/pump-tls/ca.crt
>         ssl-cert: /var/lib/pump-tls/tls.crt
>         ssl-key: /var/lib/pump-tls/tls.key
> ```
> 
> be converted to the new format gracefully without triggering a rolling update?

the generated toml will keep the same because we use a `map` to unmarshal, which sort the key in order.

pkg/util/config/config.go
july2993: no code except test code will call this directly now.
as the name `Must..` you should call this only if it should be the expected type or it's a program error.

DanielZhangQD: ok

DanielZhangQD (approved):
LGTM

DanielZhangQD:
@july2993 Please also help update the doc https://docs.pingcap.com/zh/tidb-in-kubernetes/dev/deploy-tidb-binlog in https://github.com/pingcap/docs-tidb-operator.

lichunzhu (approved):
LGTM

ti-srebot (approved):

PR #3322, opened 2020-09-24
Fix heterogeneous code cause tidb rolling update bug 

mikechengwei:
Fix heterogeneous code cause tidb rolling update bug .
 Non-heterogeneous cluster tidb startup use the old address format.
### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change


pkg/manager/member/template.go
```diff
@@ -77,6 +77,7 @@ exec /tidb-server ${ARGS}
 `))
 
 type TidbStartScriptModel struct {
+	ClusterName     string
```
DanielZhangQD, line 80: This is not used, right?

mikechengwei: Old logic have this field, i think it is unused. I can remove it.

DanielZhangQD (approved):
LGTM

pkg/manager/member/tidb_member_manager.go
```diff
@@ -402,7 +402,7 @@ func getTiDBConfigMap(tc *v1alpha1.TidbCluster) (*corev1.ConfigMap, error) {
 	if tc.IsHeterogeneous() {
 		tidbStartScriptModel.Path = controller.PDMemberName(tc.Spec.Cluster.Name) + ":2379"
```
lichunzhu, line 403: Do we need to update this one?

DanielZhangQD: No need, we just need to avoid trigger rolling update for the existing clusters.

lichunzhu (approved):
LGTM

ti-srebot (approved):

PR #3327, opened 2020-09-27
Support passing raw toml config for the tidb as a string

july2993:
### What problem does this PR solve? 

Support passing raw toml config for the tidb as a string

### What is changed and how does it work?
Add a TiDBConfigWraper to handle both type of objest or string when unmarshal json

After this pr we can write the config for tidb like this:
```
    tidb:
      baseImage: pingcap/tidb
      replicas: 1
      service:
        type: ClusterIP
      # string type to pass the toml config
      config: |
        token-limit = 102
        max-server-connections = 51
```
This pr also check whether the old toml config in configmap is equal to
the new generated one, if it's equal, we will still use the old one to
make sure will not trigger a rolling update.

### Check List 

Tests 

 - Manual test (add detailed scripts or steps below)
### will not trigger a rolling update
1. use the flowing config to apply with master controller
```
    tidb:
      baseImage: pingcap/tidb
      replicas: 1
      service:
        type: ClusterIP
      config:
        token-limit: 203
        max-server-connections: 51
```
2. update controller to this branch
3. check tidb will not rolling update.

### pass toml
1. use the flowing config to apply a cluster

```
    tidb:
      baseImage: pingcap/tidb
      replicas: 1
      service:
        type: ClusterIP
      config: |
        token-limit = 102
        max-server-connections = 51
```
2 check tidb and config work as expected.

### Does this PR introduce a user-facing change?:

```release-note
Support passing raw toml config for the tidb as a string
```

PR #3337, opened 2020-10-09
Support passing raw toml config for the tidb as a string (#3327)

ti-srebot:
cherry-pick #3327 to release-1.1

---

Support passing raw toml config for the tidb as a string

Add a TiDBConfigWraper to handle both type of objest or string when unmarshal json

After this pr we can write the config for tidb like this:
```
    tidb:
      baseImage: pingcap/tidb
      replicas: 1
      service:
        type: ClusterIP
      # string type to pass the toml config
      config: |
        token-limit = 102
        max-server-connections = 51
```
This pr also check whether the old toml config in configmap is equal to
the new generated one, if it's equal, we will still use the old one to
make sure will not trigger a rolling update.

### Check List 

Tests 

### will not trigger a rolling update
1. use the flowing config to apply with master controller
```
    tidb:
      baseImage: pingcap/tidb
      replicas: 1
      service:
        type: ClusterIP
      config:
        token-limit: 203
        max-server-connections: 51
```
2. update controller to this branch
3. check tidb will not rolling update.

### pass toml
1. use the flowing config to apply a cluster

```
    tidb:
      baseImage: pingcap/tidb
      replicas: 1
      service:
        type: ClusterIP
      config: |
        token-limit = 102
        max-server-connections = 51
```
2 check tidb and config work as expected.

```release-note
Support passing raw toml config for the tidb as a string
```


DanielZhangQD (approved):
LGTM

Issue #2592, opened 2020-05-29
TiDB Cluster deploy/upgrade failure when setting `pessimistic-txn` parameters for TiKV


DanielZhangQD:
## Bug Report

**What version of Kubernetes are you using?**

1.12.10
**What version of TiDB Operator are you using?**

v1.1.0
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

local-storage
**What's the status of the TiDB cluster pods?**

Running
**What did you do?**

1. Deploy TiDB Operator v1.1.0-rc.4 + TiDB v3.1.1
Configure below parameters for TiKV:
```
    config:
      log-level: info
      pessimistic-txn:
        wait-for-lock-timeout: 1000
        wake-up-delay-duration: 20
```
Upgrade TiDB Operator to v1.1.0, and then upgrade TiDB to v4.0, but upgrade failed with below logs:
```
E0529 09:24:40.252962       1 reflector.go:123] k8s.io/client-go@v0.0.0/tools/cache/reflector.go:96: Failed to list *v1alpha1.TidbCluster: v1alpha1.TidbClusterList.Items: []v1alpha1.TidbCluster: v1alpha1.TidbCluster.Spec: v1alpha1.TidbClusterSpec.TiKV: v1alpha1.TiKVSpec.Config: v1alpha1.TiKVConfig.TiKVPessimisticTxn: v1alpha1.TiKVPessimisticTxn.WaitForLockTimeout: ReadString: expects " or n, but found 1, error found in #10 byte of ...|timeout":1000,"wake-|..., bigger context ...|info","pessimistic-txn":{"wait-for-lock-timeout":1000,"wake-up-delay-duration":20}},"image":"pingcap|...
```
2. Deploy TiDB Operator v1.1.0 + TiDB v3.1.1
Configure below parameters for TiKV:
```
    config:
      log-level: info
      pessimistic-txn:
        wait-for-lock-timeout: 1000
        wake-up-delay-duration: 20
```
TiDB Cluster cannot start up with below logs:
```
[...]

PR #3342, opened 2020-10-09
Support passing raw toml config for tikv/pd as string type

july2993:

Support passing raw toml config for tikv/pd as string type
follow up #3327 

also fix a wrong tag name of tikv:
https://github.com/pingcap/tidb-operator/pull/3342/files#diff-35ce941397445e52d38814cd717d06d4R971

Add a TiKVConfigWraper/PDConfigWraper to handle both types of object or string when unmarshaling JSON

### Check List 

Tests 

basically like #3327 

Side effects

The `string` tag removed in pd_config.go is incompatible, but I think we should not use it.

Related changes

```release-note
Support passing raw toml config for tikv/pd as string type
```


pkg/manager/member/tikv_member_manager.go
```diff
@@ -575,26 +575,31 @@ func transformTiKVConfigMap(srcStr string, tc *v1alpha1.TidbCluster) string {
...
+				newString := fmt.Sprintf(`%s = %s`, waitForLockTimeOutKey, str)
 				srcStr = strings.ReplaceAll(srcStr, old, newString)
 			}
 		}
-		if config.TiKVPessimisticTxn.WakeUpDelayDuration != nil {
-			_, err := strconv.ParseInt(*config.TiKVPessimisticTxn.WakeUpDelayDuration, 10, 64)
+	}
+
+	if v := config.Get("pessimistic-txn.wake-up-delay-duration"); v != nil {
+		if str, err := v.AsString(); err == nil {
```
lonng, line 592: Do we need to output some warning messages if an error encountered.

lonng (commented):
The `PDConfigWrapper/TiDBConfigWrapper/TiKVConfigWrapper/...` have much duplicated code. Can we eliminate them?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -575,26 +575,31 @@ func transformTiKVConfigMap(srcStr string, tc *v1alpha1.TidbCluster) string {
...
+			_, err := strconv.ParseInt(str, 10, 64)
 			if err == nil {
 				wakeUpDelayDuration := "wake-up-delay-duration"
-				old := fmt.Sprintf(`%s = "%s"`, wakeUpDelayDuration, *config.TiKVPessimisticTxn.WakeUpDelayDuration)
-				newString := fmt.Sprintf(`%s = %s`, wakeUpDelayDuration, *config.TiKVPessimisticTxn.WakeUpDelayDuration)
+				old := fmt.Sprintf(`%s = "%s"`, wakeUpDelayDuration, str)
+				newString := fmt.Sprintf(`%s = %s`, wakeUpDelayDuration, str)
```
DanielZhangQD, line 597: The block here is to solve the incompatibility issue (the type is changed from `int32` to `string`) described in https://github.com/pingcap/tidb-operator/issues/2592, with this code change, is it OK to remove this block?

pkg/manager/member/tikv_member_manager.go
july2993: add in https://github.com/pingcap/tidb-operator/pull/3342/commits/6a23edae1b6a901bda60bb7418cc51e4c9d3e97f

july2993:
> The `PDConfigWrapper/TiDBConfigWrapper/TiKVConfigWrapper/...` have much duplicated code. Can we eliminate them?

avoid the core duplicate `UnmarshalJSON` others will keep it.
https://github.com/pingcap/tidb-operator/pull/3342/commits/6a23edae1b6a901bda60bb7418cc51e4c9d3e97f

pkg/manager/member/tikv_member_manager.go
july2993: for compatibility(user still using the origin way), we need to keep it.

DanielZhangQD (approved):
LGTM

july2993:
@lonng PTAL

lonng (approved):
LGTM

ti-srebot (approved):

PR #3350, opened 2020-10-11
Support passing raw toml config for tikv/pd as string type (#3342)

ti-srebot:
cherry-pick #3342 to release-1.1

---

Support passing raw toml config for tikv/pd as string type
follow up #3327 

also fix a wrong tag name of tikv:
https://github.com/pingcap/tidb-operator/pull/3342/files#diff-35ce941397445e52d38814cd717d06d4R971

Add a TiKVConfigWraper/PDConfigWraper to handle both types of object or string when unmarshaling JSON

### Check List 

Tests 

basically like #3327 

Side effects

The `string` tag removed in pd_config.go is incompatible, but I think we should not use it.

Related changes

```release-note
Support passing raw toml config for tikv/pd as string type
```


DanielZhangQD (approved):
LGTM

PR #3351, opened 2020-10-11
Fix pass *interface{} to json.Unmarshal

july2993:

https://github.com/pingcap/tidb-operator/pull/3342/commits/6a23edae1b6a901bda60bb7418cc51e4c9d3e97f in #3342 

introduce an issue.

we pass `&x` (type *interface{}) to `json.Unmarshal`
so type of x will be `map[string]interface{}` after it.
will fail the add test `TestUnmarshlJSON` in this pr.

pass origin `x` interface and improve test.
### Check List 

Tests 

 - Unit test

```release-note
NONE
```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/tidb_config_wraper.go
+++ b/pkg/apis/pingcap/v1alpha1/tidb_config_wraper.go
@@ -51,7 +51,7 @@ func (c *TiDBConfigWraper) MarshalJSON() ([]byte, error) {
 // for compatibility, if we use a map[string]interface{} to Unmarshal directly,
 // we can not distinct the type between integer and float for toml.
 func (c *TiDBConfigWraper) UnmarshalJSON(data []byte) error {
-	var deprecated *TiDBConfig
+	deprecated := new(TiDBConfig)
 	var err error
 	c.GenericConfig, err = unmarshalJSON(data, deprecated)
 	return err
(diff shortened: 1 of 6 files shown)


pkg/apis/pingcap/v1alpha1/tidb_config_wraper.go
```diff
@@ -78,7 +78,11 @@ func unmarshalJSON(data []byte, x interface{}) (g *config.GenericConfig, err err
 	case string:
 		tomlData = []byte(s)
 	case map[string]interface{}:
-		err = json.Unmarshal(data, &x)
+		if x == nil {
```
lonng, line 81: `json.Unmarshal` will panic if the `x == nil`. Do we need to check it here?

DanielZhangQD (approved):
LGTM

pkg/apis/pingcap/v1alpha1/tidb_config_wraper.go
july2993: https://github.com/pingcap/tidb-operator/pull/3351/commits/1a977112e0b07977cc83190e04c27ff0dd85d2a2 PTAL

lonng (approved):
LGTM

ti-srebot (approved):

Issue #3349, opened 2020-10-10
tidb operator failure self-recovery regenerate the new cluster id


WentaoJin:
1、When the pd cluster is unavailable, the regenerated pod from the tidb operator failure self-recovery will regenerate the new cluster id，then  tikv save old cluster id，so even if the pd cluster is alive, the tidb cluster cannot be served. solved by the according to tikv local cluster id pd recover(https://docs.pingcap.com/zh/tidb-in-kubernetes/stable/pd-recover)

tidb  operator failure self-recovery 8 pod
![image](https://user-images.githubusercontent.com/34805682/95658137-d725db80-0b4a-11eb-9bb1-db6a23f99bac.png)

tikv pd cluster id is not consistent，tikv tidb pod CrashLoopBackOff status
![image](https://user-images.githubusercontent.com/34805682/95657990-27506e00-0b4a-11eb-9a1c-28dbaee32ac5.png)

2、Spec：
When the pd cluster is unavailable, can the tidb operator control the generation of pods without self-recovery from failure

DanielZhangQD:
@WentaoJin What's the operator version?

DanielZhangQD:
[Uploading ods-cfp_discovery.log…]()

DanielZhangQD:
[ods-cfp_discovery.log](https://github.com/pingcap/tidb-operator/files/5362632/ods-cfp_discovery.log)

WentaoJin:
> @WentaoJin What's the operator version?

v1.0 version，rememeber v1.0.3

WentaoJin:
> @WentaoJin What's the operator version?

PR #3365, opened 2020-10-13
discovery: fix problem when restart with failover pods

lichunzhu:
https://github.com/pingcap/tidb-operator/issues/3349
Imagine the situation that we have 3 pd pods running normally. At one time pd-0's pvc is lost and failover starts another pod pd-3. pd-3 is blocked at discovering part because of some reason. Discovery will receive 2 requests now. If pd-1's pvc is also lost and restart again, the discovery service will receive 3 requests and give an `initial-cluster...` starting argument to either `pd-0`, `pd-1` or `pd-3`, which will cause two cluster-ids in one pd group.

Use failover desired replicas instead of replicas for discovery service.

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

Related changes

```release-note
Use failover desired replicas instead of replicas for discovery service
```


pkg/discovery/discovery.go
```diff
@@ -114,6 +113,12 @@ func (td *tidbDiscovery) Discover(advertisePeerUrl string) (string, error) {
 
 	membersArr := make([]string, 0)
 	for _, member := range membersInfo.Members {
+		// In some failure situations, for example, delete the dm-master's data directory, dm-master will try to restart
+		// and get join info from discovery service. But dm-master embed etcd may still have the registered member info,
+		// which will return the argument to join dm-master itself, which is not allowed in dm-master.
+		if member.Name == podName {
+			continue
+		}
```
DanielZhangQD, line 121: This should be updated in `DiscoveryDM`?

DanielZhangQD (approved):
LGTM

july2993 (approved):

ti-srebot (approved):

Issue #3182, opened 2020-08-28
Do not scale in TiKV if the available replicas are less than 3


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

With the current implementation, when there are only 3 available replicas, if users try to scale in TiKV, the Operator will delete the store, however, due to the constraint of raft group, the store cannot become tombstone and will remain `offline` forever, so it's better that Operator checks the available replicas before scaling in and do not delete store if the available replicas <= 3.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

cvvz:
If tidb-operator has modified `spec.replicas` of StatefulSet to 2, it will not be able to scaling out tidb-cluster by itself any more:
```shell
I0906 17:12:05.189967       1 tikv_scaler.go:68] scaling out tikv statefulset tidb-cluster/backdb-tikv, ordinal: 2 (replicas: 3, delete slots: [])
I0906 17:12:05.204073       1 tidb_cluster_controller.go:303] TidbCluster: tidb-cluster/backdb, still need sync: tikv.ScaleOut, cluster tidb-cluster/backdb ready to scale out, wait for next round, requeuing
``` 
Because it keeps waiting for the PVC of `tikv-2` to be deleted.

DanielZhangQD:
That's another issue, do you mean if scale in from 3 to 2 and then scale out from 2 to 3, the scale-out cannot be done?
If so, please help create another issue and attach the log of the tidb-controller-manager.

cvvz:
Yes, and the issue is here https://github.com/pingcap/tidb-operator/issues/3229. I think these two issues can be solved in the same way.

PR #3367, opened 2020-10-14
forbid to scale in tikv when number of stores is equal to or less than 3

cvvz:
https://github.com/pingcap/tidb-operator/issues/3182
forbid to scale in tikv when number of stores is equal to or less than 3.
### Check List 

Tests 

 - E2E test

Code changes

 - Has Go code change

Related changes

```release-note
NONE
```


DanielZhangQD:
@cvvz Please fix the CI.

pkg/manager/member/tikv_scaler.go
```diff
@@ -94,6 +94,14 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
 
 	switch meta.(type) {
 	case *v1alpha1.TidbCluster:
+		tc ,_ := meta.(*v1alpha1.TidbCluster)
+		storesInfo, err := controller.GetPDClient(tsd.pdControl, tc).GetStores()
```
DanielZhangQD, line 98: should be `tsd.deps.PDControl` ?

pkg/manager/member/tikv_scaler.go
```diff
@@ -94,6 +94,14 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+		if err != nil {
+			return fmt.Errorf("failed to get stores info when scaling in TiKV")
```
DanielZhangQD, line 100: Please add tc name and ns

pkg/manager/member/tikv_scaler.go
```diff
@@ -94,6 +94,14 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+		}
+		if storesInfo.Count <= 3 {
+			return fmt.Errorf("it is not allowed to scale in TiKV when stores number(%d) <= 3", storesInfo.Count)
```
DanielZhangQD, line 103: No need to return error because it will cause the TC to be re-synced, just print the error and return `nil`.
Please also add info of TC.

pkg/manager/member/tikv_scaler.go
```diff
@@ -94,6 +94,14 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
 
 	switch meta.(type) {
 	case *v1alpha1.TidbCluster:
+		tc ,_ := meta.(*v1alpha1.TidbCluster)
+		storesInfo, err := controller.GetPDClient(tsd.pdControl, tc).GetStores()
+		if err != nil {
+			return fmt.Errorf("failed to get stores info when scaling in TiKV")
+		}
+		if storesInfo.Count <= 3 {
```
DanielZhangQD, line 102: what about the following cases:
```
if storesInfo.Count > 3
   sum(UP) < 3
   sum(UP) = 3 && the state of the store for the Pod to be scaled in is UP
   sum(UP) > 3 (this should be good)
```

pkg/manager/member/tikv_scaler.go
DanielZhangQD: Can we add `status.tikv.storeCount` and sync it in the `syncTidbClusterStatus`, so here we can just retrieve the info from the `status.tikv` because the `GetStores` will be called in `syncTidbClusterStatus` just before the scaling function.

pkg/manager/member/tikv_scaler.go
cvvz: > No need to return error because it will cause the TC to be re-synced, just print the error and return `nil`.
> Please also add info of TC.

@DanielZhangQD Although the TC will be re-synced no matter whether error returns or not, it will be more logical and more friendly to write unit test ,

pkg/manager/member/tikv_scaler.go
```diff
@@ -103,16 +103,30 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+	for _, store := range tc.Status.TiKV.Stores {
+		if store.State == v1alpha1.TiKVStateUp {
+			upNumber++
+		}
+		if store.PodName == podName {
+			storeState = store.State
+		}
+	}
+	if upNumber < 3 {
+		return fmt.Errorf("the number of stores in Up state of TidbCluster [%s/%s] is %d, less than 3, can't scale in TiKV, podname %s ", meta.GetNamespace(), meta.GetName(), upNumber, podName)
```
DanielZhangQD, line 118: If we return an error here, the TC will be enqueued and synced immediately over and over again, however, the resync is useless, so we just print the log here and return nil, in this case, the TC will be resynced every 30s or until next update from the user.

cvvz:
@DanielZhangQD I've override the code and modified some of unit tests for adaptation. PTLA :)

pkg/manager/member/tikv_scaler.go
```diff
@@ -103,16 +103,30 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+	storeState := ""
+	for _, store := range tc.Status.TiKV.Stores {
+		if store.State == v1alpha1.TiKVStateUp {
+			upNumber++
+		}
+		if store.PodName == podName {
+			storeState = store.State
+		}
+	}
+	if upNumber < 3 {
```
DanielZhangQD, line 117: Suggest defining a constant for `3`.

pkg/manager/member/tikv_scaler.go
DanielZhangQD, line 118:
```suggestion
		return fmt.Errorf("the number of stores in Up state of TidbCluster [%s/%s] is %d, less than 3, can't scale in TiKV, podname %s", meta.GetNamespace(), meta.GetName(), upNumber, podName)
```

pkg/manager/member/tikv_scaler.go
```diff
@@ -103,16 +103,30 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+		return fmt.Errorf("the number of stores in Up state of TidbCluster [%s/%s] is %d, less than 3, can't scale in TiKV, podname %s ", meta.GetNamespace(), meta.GetName(), upNumber, podName)
+	} else if upNumber == 3 {
+		if storeState == v1alpha1.TiKVStateUp {
+			return fmt.Errorf("can't scale in TiKV of TidbCluster [%s/%s], cause the number of up stores is 3, and the store in Pod %s which is going to be deleted is up too", meta.GetNamespace(), meta.GetName(), podName)
```
DanielZhangQD, line 121: ditto

DanielZhangQD: BTW, did you test this case? The store cannot change to tombstone, right?

cvvz: > BTW, did you test this case? The store cannot change to tombstone, right?

@DanielZhangQD Yep. If the store state is tombstone, it will be moved to `Status.TiKV.TombstoneStores`.

pkg/manager/member/tikv_scaler.go
```diff
@@ -103,14 +108,30 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
 		return fmt.Errorf("tikvScaler.ScaleIn: failed to get pods %s for cluster %s/%s, error: %s", podName, ns, tcName, err)
 	}
 
-	if tsd.deps.CLIConfig.PodWebhookEnabled {
-		setReplicasAndDeleteSlots(newSet, replicas, deleteSlots)
+	tc, _ := meta.(*v1alpha1.TidbCluster)
```
DanielZhangQD, line 111: Please help add some UT cases to cover this code change, thanks!

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -1429,6 +1429,32 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+
+			pdClient, cancel, err := proxiedpdclient.NewProxiedPDClient(c, fw, ns, tc.Name, false)
+			framework.ExpectNoError(err, "create pdClient error")
+			defer cancel()
+			storesInfo, err := pdClient.GetStores()
+			framework.ExpectNoError(err, "get stores info error")
+			framework.ExpectEqual(storesInfo.Count, 3, "Expect number of stores is 3")
+			for _, store := range storesInfo.Stores {
+				framework.ExpectEqual(store.Store.StateName, "Up", "Expect state of stores are Up")
+			}
```
DanielZhangQD, line 1455: We should wait for several minutes (e.g. 2 or 3 min) to check the store state, you can refer to the code here https://github.com/pingcap/tidb-operator/blob/master/tests/e2e/tidbcluster/tidbcluster.go#L328.

DanielZhangQD (approved):
LGTM

pkg/manager/member/tikv_scaler.go
```diff
@@ -103,14 +108,30 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+	storeState := ""
+	for _, store := range tc.Status.TiKV.Stores {
+		if store.State == v1alpha1.TiKVStateUp {
+			upNumber++
+		}
+		if store.PodName == podName {
+			storeState = store.State
+		}
+	}
+	if upNumber < MinimalUpStateOfStores {
```
lonng, line 122: I think it's better to use the replica configuration instead of a hardcode constant `MinimalUpStateOfStores`. The replica configuration can be found via the PD client, see: https://docs.pingcap.com/tidb/stable/pd-configuration-file#replication.

pkg/manager/member/tikv_scaler.go
```diff
@@ -103,14 +108,30 @@ func (tsd *tikvScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+		klog.Errorf("the number of stores in Up state of TidbCluster [%s/%s] is %d, less than %d, can't scale in TiKV, podname %s ", meta.GetNamespace(), meta.GetName(), upNumber, MinimalUpStateOfStores, podName)
 		return nil
+	} else if upNumber == MinimalUpStateOfStores {
+		if storeState == v1alpha1.TiKVStateUp {
+			klog.Errorf("can't scale in TiKV of TidbCluster [%s/%s], cause the number of up stores is %d, and the store in Pod %s which is going to be deleted is up too", meta.GetNamespace(), meta.GetName(), MinimalUpStateOfStores, podName)
```
lonng, line 127: Do we need to record the failed event for debugging purposes? @DanielZhangQD

pkg/manager/member/tikv_scaler.go
DanielZhangQD: Good point.

pkg/manager/member/tikv_scaler.go
DanielZhangQD: Yes, it's better to record an event for TC here.

DanielZhangQD:
@cvvz Please help address the comments when you're available, thanks!

pkg/manager/member/tikv_scaler.go
DanielZhangQD: @cvvz You can use `GetConfig` function in the `pkg/pdapi/pdapi.go` to get the config from PD.

cvvz:
> @cvvz Please help address the comments when you're available, thanks!

@DanielZhangQD I've fix and test it now, PTAL.

DanielZhangQD:
@cvvz Please fix the CI.

DanielZhangQD (dismissed):
LGTM

pkg/manager/member/tikv_scaler.go
```diff
@@ -18,6 +18,8 @@ import (
 	"strconv"
 	"time"
 
+	v1 "k8s.io/api/core/v1"
+
```
DanielZhangQD, line 22: Please remove the empty line here.

cvvz: done

DanielZhangQD (approved):
LGTM

DanielZhangQD:
@lonng PTAL.

lonng (approved):
LGTM

ti-srebot (approved):

howardlau1999:
It seems this PR does not consider heterogeneous clusters. For example, when a user has created a TidbCluster with 3 TiKV and later created a hete tc with 3 TiKV to join the original tc, the user cannot scale in the hete tc although the replicas constraints can be satisfied.

DanielZhangQD:
> It seems this PR does not consider heterogeneous clusters. For example, when a user has created a TidbCluster with 3 TiKV and later created a hete tc with 3 TiKV to join the original tc, the user cannot scale in the hete tc although the replicas constraints can be satisfied.

@howardlau1999 You're right!
@cvvz Sorry that I did not consider this case during the review, could you please help update the code to get the stores info from PD API instead of `status`?
Sorry for the trouble!

PR #3370, opened 2020-10-14
discovery: fix problem when restart with failover pods (#3365)

ti-srebot:
cherry-pick #3365 to release-1.1

---

https://github.com/pingcap/tidb-operator/issues/3349
Imagine the situation that we have 3 pd pods running normally. At one time pd-0's pvc is lost and failover starts another pod pd-3. pd-3 is blocked at discovering part because of some reason. Discovery will receive 2 requests now. If pd-1's pvc is also lost and restart again, the discovery service will receive 3 requests and give an `initial-cluster...` starting argument to either `pd-0`, `pd-1` or `pd-3`, which will cause two cluster-ids in one pd group.

Use failover desired replicas instead of replicas for discovery service.

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

Related changes

```release-note
Use failover desired replicas instead of replicas for discovery service
```


DanielZhangQD (approved):
LGTM

lichunzhu (approved):
LGTM

PR #3376, opened 2020-10-15
fix: allow tidb init job to run a 2nd time

### LINKED ISSUE #3353: tidb-initializer job is not idempotent
## Bug Report

In our environment we always apply all yaml, using kustomize. THis means that we can't accept things that can only be applied once.

https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/templates/scripts/_initialize_tidb_users.py.tpl#L5

is an example of a violation of this principle. It assumes the db can be connected w/o a password, then changes the password. If run again, it will fail on the first assumption.

A possible fix for this might look like:
```
diff --git a/tidb-cluster-chart.yaml b/tidb-cluster-chart.yaml
index 6f5ca6c..d75d8e2 100644
--- a/tidb-cluster-chart.yaml
+++ b/tidb-cluster-chart.yaml
@@ -543,32 +543,35 @@ spec:
         - python
         - -c
         - |
-          import os, MySQLdb
+          import os, sys, MySQLdb
           host = 'db-tidb'
           permit_host = "%"
           port = 4000
-          conn = MySQLdb.connect(host=host, port=port, user='root', connect_timeout=5)
-          password_dir = '/etc/tidb/password'
-          for file in os.listdir(password_dir):
-              if file.startswith('.'):
-                  continue
-              user = file
-              with open(os.path.join(password_dir, file), 'r') as f:
-                  password = f.read()
-              if user == 'root':
-                  conn.cursor().execute("set password for 'root'@'%%' = %s;", (password,))
-              else:
-                  conn.cursor().execute("create user %s@%s identified by %s;", (user, permit_host, password,))
-          with open('/data/init.sql', 'r') as sql:
-              for line in sql.readlines():
-                  conn.cursor().execute(line)
-                  conn.commit()
-          if permit_host != '%%':
-              conn.cursor().execute("update mysql.user set Host=%s where User='root';", (permit_host,))
-          conn.cursor().execute("flush privileges;")
-          conn.commit()
-          conn.close()
-          
+          try:
+              conn = MySQLdb.connect(host=host, port=port, user='root', connect_timeout=5)
```

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T17:01:15Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.13-gke.1", GitCommit:"688c6543aa4b285355723f100302d80431e411cc", GitTreeState:"clean", BuildDate:"2020-07-21T02:37:26Z", GoVersion:"go1.13.9b4", Compiler:"gc", Platform:"linux/amd64"}

```

CHART_VER=v1.1.5

```
$ kubectl get sc
NAME                 PROVISIONER            AGE
managed-premium      kubernetes.io/gce-pd   461d
standard (default)   kubernetes.io/gce-pd   461d
```

**What did you do?**

Step 1, install tidb-cluster
Step 2, apply the same yaml again

i expected step 2 would be a no-op

**What did you see instead?**

job-initializer crashes.

donbowman:
Fixes #3353. If we can connect with the root password, assume
the job has previously been run and exit. Otherwise connect
with no password, set it, etc.

This will allow the tidb-initializer job to run a 2nd time
without crashing, allowing the helm chart to be upgraded.

Correct #3353 - if the init job is run a 2nd time it crashes since it assumes the password of the db is null

in _initialize_tidb_users.py.tpl, if we can connect w/ the root password, assume job has previously run and exit.

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/charts/tidb-cluster/templates/scripts/_initialize_tidb_users.py.tpl
+++ b/charts/tidb-cluster/templates/scripts/_initialize_tidb_users.py.tpl
@@ -1,7 +1,22 @@
-import os, MySQLdb
+import os, sys, MySQLdb
 host = '{{ template "cluster.name" . }}-tidb'
 permit_host = {{ .Values.tidb.permitHost | default "%" | quote }}
 port = 4000
+{{- if .Values.tidb.passwordSecretName }}
+# We make the assumption if we can connect as the root user w/ the password,
+# that we have been run already. If so, exit
+with open(os.path.join("/etc/tidb/password", "root"), 'r') as f:
+    setup_complete = False
+    password = f.read()
+    try:
+        conn = MySQLdb.connect(host=host, port=port, user='root', password=password, auth_plugin='mysql_native_password', connect_timeout=5)
+        setup_complete = True
+    except:
+        pass
+    if setup_complete:
+        print("Init job completed successfully on previous run, exiting")
+        sys.exit(0)
+{{- end }}
 conn = MySQLdb.connect(host=host, port=port, user='root', connect_timeout=5)
 {{- if .Values.tidb.passwordSecretName }}
 password_dir = '/etc/tidb/password'


lonng (approved):
LGTM

charts/tidb-cluster/templates/scripts/_initialize_tidb_users.py.tpl
```diff
@@ -1,7 +1,22 @@
-import os, MySQLdb
+import os, sys, MySQLdb
 host = '{{ template "cluster.name" . }}-tidb'
 permit_host = {{ .Values.tidb.permitHost | default "%" | quote }}
 port = 4000
+{{- if .Values.tidb.passwordSecretName }}
```
DanielZhangQD, line 5: The chart `tidb-cluster` is not used since TiDB Operator v1.1.
Please also help update the code here https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/template.go#L282.

DanielZhangQD (changes requested):

DanielZhangQD:
@donbowman You do have tested the code change, right? Could you please update the test result in the PR description?
Thanks!

charts/tidb-cluster/templates/scripts/_initialize_tidb_users.py.tpl
donbowman: ok will do.

DanielZhangQD:
@donbowman With the TidbInitializer CR, it can be applied multiple times because TiDB Operator will check if the job exists and if yes, it will not create the job again, did you test with the TidbInitializer CR?

DanielZhangQD:
> @donbowman With the TidbInitializer CR, it can be applied multiple times because TiDB Operator will check if the job exists and if yes, it will not create the job again, did you test with the TidbInitializer CR?

@donbowman Did you test with the TidbInitializer CR?

omaskery:
Hi, just wanted to get a sense of if this work is likely to be continued any time soon? @donbowman is there any way I can help?

donbowman:
i'll try to find a time, but since i have a solution w/ my re-written initialiser (that i posted earlier), i cannot apply as-is to my production type system since it would wreck the database.

I am really not clear on what you are asking that is different.
the problem i saw:

1. install. it runs a snippet of python which creates a db, changes the password
2. upgrade. it runs the same snippet of python. this fails, since the password was changed

so i think you are suggesting to edit the helm chart and remove that python snippet, and, use a new crd TidbInitializer and write that in some fashion so that it will do the same thing as my change?

Does everyone not have this problem? I am confused?

omaskery:
@donbowman to be honest I'm now in the same boat, it was easier for me to make our infrastructure overwrite the python script in the rendered helm chart with my own that essentially implements your fix 😅

DanielZhangQD:
The tidb-cluster chart is [deprecated](https://docs.pingcap.com/tidb-in-kubernetes/dev/notes-tidb-operator-v1.1) from v1.1.0 and TidbInitializer CR is used to initializer the database, in which case, the initializer job will not be created twice unless the original job is removed manually.

DanielZhangQD:
> The tidb-cluster chart is [deprecated](https://docs.pingcap.com/tidb-in-kubernetes/dev/notes-tidb-operator-v1.1) from v1.1.0 and TidbInitializer CR is used to initializer the database, in which case, the initializer job will not be created twice unless the original job is removed manually.

@donbowman Did you have a chance to try the TidbInitializer CR?

PR #3390, opened 2020-10-18
monitor: fix surprising prometheus commandOptions behavior

mightyguava:
# What problem does this PR solve? 

`prometheus.spec.config.commandOptions` has surprising behavior. Setting it overrides flags essential to correct operation, and other config options on the `spec`.

```
        - --web.enable-admin-api	
        - --web.enable-lifecycle	
        - --config.file=/etc/prometheus/prometheus.yml	
        - --storage.tsdb.path=/data/prometheus	
        - --storage.tsdb.retention=12d
```

while merging with with `--log-level`.

Now, `commandOptions` will only append to the existing set of flags. Prometheus will error on startup if a flag is duplicated.

Note that this is a breaking change. I'm assuming that `commandOptions` is a rarely used advanced option and anyone who uses it will be able to apply a fix pretty easily.

A non-breaking alternative is to add an `extraCommandOptions` flag, but that seems a bit overboard.

### Check List 

Tests

 - Unit test

Code changes

 - Has Go code change

Side effects

 - Breaking backward compatibility.

```release-note
The behavior of `prometheus.spec.config.commandOptions` has changed. Any duplicated flags must be removed or Prometheus will fail to start.
```

### DIFF
--- a/pkg/monitor/monitor/util.go
+++ b/pkg/monitor/monitor/util.go
@@ -396,6 +396,11 @@ func getMonitorPrometheusContainer(monitor *v1alpha1.TidbMonitor, tc *v1alpha1.T
 		Resources: controller.ContainerResource(monitor.Spec.Prometheus.ResourceRequirements),
 		Command: []string{
 			"/bin/prometheus",
+			"--web.enable-admin-api",
+			"--web.enable-lifecycle",
+			"--config.file=/etc/prometheus/prometheus.yml",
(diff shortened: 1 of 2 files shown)


DanielZhangQD (approved):
LGTM

mikechengwei:
LGTM

ti-srebot (approved):

Issue #3341, opened 2020-10-09
PD leader transfer does not consider the case where advanced statefulset controller enabled


DanielZhangQD:
## Bug Report

**What version of Kubernetes are you using?**

**What version of TiDB Operator are you using?**

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

**What's the status of the TiDB cluster pods?**

**What did you do?**

With advanced statefulset controller enabled, the pd-0 maybe deleted, however, the PD transfer logic in https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/pd_scaler.go#L135 and https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/pd_upgrader.go#L126 may still transfer leader to pd-0 without checking it.
We should use a similar logic with that in https://github.com/pingcap/tidb-operator/blob/master/pkg/webhook/pod/pd_deleter.go#L208-L213 for the min and max ordinals.

**What did you expect to see?**

**What did you see instead?**

Issue #1235, opened 2019-11-26
Optimize tranferring pd leader


Yisaer:
## Feature Request

**Is your feature request related to a problem? Please describe:**

https://github.com/pingcap/tidb-operator/issues/1015

Currently PD leader would be transferred between the first pd pod and the last pd pod without considering whether they were working or healthy. And each 2 upgrading would cause twice transferring pd leader during upgrading.

**Describe the feature you'd like:**

Optimizing the pd leader transferring logic. The target pd pod which is going to be transferred as pd leader should be checked whether it is avaiable. 

If the last pd pod was the leader before upgrading, I think there was at least and at most 2 times to transfer leader. I think we should decrease the frequenecy of 2 times pd leader transferring.

shonge:
Can you describe which scene will transfer leader more than 2 times? @Yisaer

PR #3395, opened 2020-10-19
fix: PD leader transfer does not consider the case where advanced statefulset controller enabled.

tangwz:
fixed: #3341
PD leader transfer does not consider the case where advanced statefulset controller enabled 

With advanced statefulset controller enabled, the pd-0 maybe deleted, We should use a similar logic with `pd_deleter.go` for the min and max ordinals.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

 - None

Related changes

```release-note
Support the case where advanced statefulset controller is enabled during PD leader transfer
```


pkg/manager/member/pd_upgrader.go
```diff
@@ -115,7 +115,7 @@ func (pu *pdUpgrader) upgradePDPod(tc *v1alpha1.TidbCluster, ordinal int32, newS
 		lastOrdinal := tc.PDStsActualReplicas() - 1
```
DanielZhangQD, line 115: The `lastOrdianl` here should also be calculated by `GetMaxPodOrdinal`, right?

pkg/manager/member/pd_scaler.go
```diff
@@ -127,7 +127,15 @@ func (psd *pdScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, newSe
...
-			err = pdClient.TransferPDLeader(fmt.Sprintf("%s-pd-%d", tc.GetName(), 0))
+			var targetName string
+			lastOrdinal := helper.GetMaxPodOrdinal(*newSet.Spec.Replicas, newSet)
+			if ordinal == lastOrdinal {
+				targetName = PdPodName(tcName, helper.GetMinPodOrdinal(*newSet.Spec.Replicas, newSet))
+			} else {
+				targetName = PdPodName(tcName, lastOrdinal)
+			}
+
+			err = pdClient.TransferPDLeader(targetName)
```
DanielZhangQD, line 138: The logic from L124 should be as following：
```
if ordinal > minOrdinal {
  transfer leader to Pod with MinOrdinal
}
```
WDYT?

pkg/manager/member/pd_upgrader.go
tangwz: I agree with you.Thanks.

pkg/manager/member/pd_scaler.go
tangwz: In this case no need to `lastOrdinal := helper.GetMaxPodOrdinal(*newSet.Spec.Replicas, newSet)` this logic ?

DanielZhangQD: Yes, I think so, because this is the scaling in operation, we can just transfer the leader to the min ordinal Pod if it's not the min ordinal Pod and if it is, no need to transfer the PD leader, which is the same logic as it is.

DanielZhangQD:
BTW, @tangwz would you like to take https://github.com/pingcap/tidb-operator/issues/1235 after this PR is merged?

pkg/manager/member/pd_scaler.go
```diff
@@ -121,13 +121,14 @@ func (s *pdScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, newSet
 	// If the pd pod was pd leader during scale-in, we would transfer pd leader to pd-0 directly
 	// If the pd statefulSet would be scale-in to zero and the pd-0 was going to be deleted,
 	// we would directly deleted the pd-0 without pd leader transferring
-	if ordinal > 0 {
+	minOrdinal := helper.GetMinPodOrdinal(*newSet.Spec.Replicas, newSet)
+	if ordinal > minOrdinal {
```
DanielZhangQD, line 125: Sorry that I missed one point that with advanced statefulset we can scale in any Pod, so the logic should be :
```
if *newSet.Spec.Replicas > 1
    if ordinal > minOrdinal, transfer to minOrdinal
    else, transfer to maxOrdinal
else, no need to transfer leader
```

pkg/manager/member/pd_scaler.go
```diff
@@ -121,13 +121,14 @@ func (s *pdScaler) ScaleIn(meta metav1.Object, oldSet *apps.StatefulSet, newSet
 	// If the pd pod was pd leader during scale-in, we would transfer pd leader to pd-0 directly
 	// If the pd statefulSet would be scale-in to zero and the pd-0 was going to be deleted,
 	// we would directly deleted the pd-0 without pd leader transferring
```
DanielZhangQD, line 123: Please also update the comments.

tangwz:
> BTW, @tangwz would you like to take #1235 after this PR is merged?

Sounds good.

DanielZhangQD:
@tangwz Please help check the new comments, thanks!

pkg/manager/member/pd_scaler.go
tangwz: I think we need check `leader.Name == memberName` first:
```
if leader.Name == memberName 
    if ordinal > minOrdinal, transfer to minOrdinal
    else, transfer to maxOrdinal
```

What do you think?

tangwz: > Sorry that I missed one point that with advanced statefulset we can scale in any Pod, so the logic should be :
> 
> ```
> if *newSet.Spec.Replicas > 1
>     if ordinal > minOrdinal, transfer to minOrdinal
>     else, transfer to maxOrdinal
> else, no need to transfer leader
> ```

I think we need check `leader.Name == memberName `first:
```
if leader.Name == memberName 
    if ordinal > minOrdinal, transfer to minOrdinal
    else, transfer to maxOrdinal
```
What do you think?

DanielZhangQD: What about:
```
if *newSet.Spec.Replicas > 1
    if leader.Name == memberName 
        if ordinal > minOrdinal, transfer to minOrdinal
        else, transfer to maxOrdinal
else, no need to transfer leader
```

tangwz: That's what I meant.

DanielZhangQD (approved):
LGTM

lichunzhu (approved):
LGTM

ti-srebot (approved):

Issue #3407, opened 2020-10-20
Tidb scheduler should support HA scheduling in failover situation


cvvz:
## Bug Report

**What did you do?**

When a Pod of PD or TiKV is down unexpectly, after a while, tidb-operator begins to do failover, that is, create a new Pod. 

**What did you expect to see?**
The new component with the new generated Pod should keep high availability.

**What did you see instead?**
The new pod may be scheduled to a topology that can ruin the high availability of the component.

cvvz:
I think tidb-scheduler should consider the failure members of PD and TiKV when scheduling. In other words, do not count the topology of the failure members. I'll fix this.

DanielZhangQD:
@cvvz #3388 fixed this issue, right?

cvvz:
@DanielZhangQD It's different. This is about to keep HA scheduling in failover situation.

PR #3419, opened 2020-10-21
support HA scheduling when failover happen

cvvz:
https://github.com/pingcap/tidb-operator/issues/3407
### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/scheduler/predicates/ha.go
+++ b/pkg/scheduler/predicates/ha.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/client/clientset/versioned"
 	"github.com/pingcap/tidb-operator/pkg/label"
+	"github.com/pingcap/tidb-operator/pkg/util"
 	apiv1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


pkg/scheduler/predicates/ha.go
```diff
@@ -469,3 +474,22 @@ func isPodDesired(tc *v1alpha1.TidbCluster, component, podName string) bool {
 	}
 	return ordinals.Has(ordinal)
 }
+
+func isFailureMember(tc *v1alpha1.TidbCluster, component, podName string) bool {
+	if component == v1alpha1.PDMemberType.String() {
+		for _, fm := range tc.Status.PD.FailureMembers {
+			if fm.PodName == podName {
+				return true
```
DanielZhangQD, line 482: We also need to check the `pd.members` to check the current status, because if one Pod is recovered, it may not be removed from the failureMembers.

pkg/scheduler/predicates/ha.go
```diff
@@ -469,3 +474,22 @@ func isPodDesired(tc *v1alpha1.TidbCluster, component, podName string) bool {
...
+			}
+		}
+		return false
+	}
+
+	for _, fs := range tc.Status.TiKV.FailureStores {
+		if fs.PodName == podName {
+			return true
```
DanielZhangQD, line 490: Also need to check the current store status, because the failurerStores will not be removed automatically.

pkg/scheduler/predicates/ha.go
cvvz: I think the Pod should be treated as a failure member before it is removed from the failureMembers. This won't cause any problem.

DanielZhangQD: What about a Pod failed, set in the failureMembers, deleted by the Operator, and then recovered, but the failureMember is not updated because we only clear the failureMembers when all of the failed Pods recovered, you can check the logic [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/pd_member_manager.go#L254).

DanielZhangQD (approved):
LGTM

lichunzhu (approved):
LGTM

ti-srebot (approved):

Issue #2183, opened 2020-04-15
TiKV Pod supports multiple PVCs 


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

TiKV Pod supports multiple PVCs so that Titan can be enabled with a separate disk.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

cofyc:
use case: https://github.com/pingcap/tidb-operator/issues/1357

mikechengwei:
mark

mikechengwei:
My design:

```
    config:
      storage:
        # In basic examples, we set this to avoid using too much storage.
        reserve-space: "0MB"
      rocksdb:
        wal-dir: "/var/lib/wal"
      titan:
        dirname: "/var/lib/titan"      
    volume:
      - name: wal
        storageClassName: local-storage
        storage: "2Gi"
        mountPath: "/val/lib/wal"
      - name: titan
        # if storageClassName is not set, the default Storage Class of the Kubernetes cluster will be used
        storage: "2Gi"
        mountPath: "/var/lib/titan"
```

DanielZhangQD:
@mikechengwei 
Just some minor suggestions:
`volume:` -> `storageVolumes:`
`storage: "2Gi"` -> `storageSize: "2Gi"`, to keep consistent with the other storage size definitions that are not defined in `requests/limits`.

PR #3425, opened 2020-10-24
Tikv multiple pvc feature

mikechengwei:
#2183 
Support tikv multiple volume mount.
### Check List 

Tests 

 - Unit test
 - No code

Code changes

 - Has Go code change


pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1832,3 +1837,11 @@ type WorkerFailureMember struct {
 	PodName   string      `json:"podName,omitempty"`
 	CreatedAt metav1.Time `json:"createdAt,omitempty"`
 }
+
+// StorageVolume is TiKV storage information
+type StorageVolume struct {
+	Name             string  `json:"name,omitempty"`
+	StorageClassName *string `json:"storageClassName,omitempty"`
+	StorageSize      string  `json:"storageSize,omitempty"`
+	MountPath        string  `json:"mountPath,omitempty"`
```
DanielZhangQD, line 1846:
```suggestion
	Name             string  `json:"name"`
	StorageClassName *string `json:"storageClassName,omitempty"`
	StorageSize      string  `json:"storageSize"`
	MountPath        string  `json:"mountPath"`
```

DanielZhangQD: The above updated three fields are required for each config item, right?

DanielZhangQD: We should also add validation for the `StorageSize ` field to ensure it's a valid storage config.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -372,6 +374,28 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
...
 	}
 
+	var additionalVolumeClaims []corev1.PersistentVolumeClaim
+	if len(tc.Spec.TiKV.StorageVolumes) > 0 {
+		for _, storageVolume := range tc.Spec.TiKV.StorageVolumes {
+			storageRequest, err := controller.ParseStorageRequest(corev1.ResourceList{
+				corev1.ResourceStorage: resource.MustParse(storageVolume.StorageSize),
+			})
+			if err != nil {
+				return nil, fmt.Errorf("cannot parse storage request for tikv, tidbcluster %s/%s, error: %v", tc.Namespace, tc.Name, err)
```
DanielZhangQD, line 384: We should validate this field first.
And if we still hit an error here, is it better to print the error log and continue to sync others instead of returning the error here?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -372,6 +374,28 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
...
+			}
+			var storageClassName *string
+			if storageVolume.StorageClassName != nil {
```
DanielZhangQD, line 387: && len(*storageVolume.StorageClassName) > 0

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -1458,6 +1458,31 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+					StorageSize: "2Gi",
+					MountPath:   "/var/lib/titan",
+				},
+			}
+			tc.Spec.TiKV.Config.Set("rocksdb.wal-dir", "/var/lib/wal")
+			tc.Spec.TiKV.Config.Set("titan.dirname", "/var/lib/titan")
+			err := genericCli.Create(context.TODO(), tc)
+			framework.ExpectNoError(err, "Expected TiDB cluster created")
+			err = oa.WaitForTidbClusterReady(tc, 30*time.Minute, 15*time.Second)
+			framework.ExpectNoError(err, "Expected TiDB cluster ready")
```
DanielZhangQD, line 1483: We also need to test scaling in and then scaling out work as expected.

pkg/apis/pingcap/v1alpha1/types.go
mikechengwei: ok

pkg/manager/member/tikv_member_manager.go
```diff
@@ -372,6 +374,29 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 		}
 	}
 
+	var additionalVolumeClaims []corev1.PersistentVolumeClaim
+	if len(tc.Spec.TiKV.StorageVolumes) > 0 {
+		for _, storageVolume := range tc.Spec.TiKV.StorageVolumes {
+			storageRequest, err := controller.ParseStorageRequest(corev1.ResourceList{
```
DanielZhangQD, line 380: No need to call `ParseStorageRequest` here, just construct the ResourceRequirements directly.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -372,6 +374,29 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
...
+				corev1.ResourceStorage: resource.MustParse(storageVolume.StorageSize),
```
DanielZhangQD, line 381: We should not use `MustParse` here which will cause tidb-controller-manager panic with invalid configuration.
Try to use `resource.ParseQuantity(storageSize)`, btw, I think we can call this function to validate the config in `validation.go`.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -1459,6 +1459,36 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+			},
+		}
+		cluster.Clustrer.Spec.TiKV.Config.Set("rocksdb.wal-dir", "/var/lib/wal")
+		cluster.Clustrer.Spec.TiKV.Config.Set("titan.dirname", "/var/lib/titan")
+		e2elog.Logf("deploying tidb cluster [%s/%s]", cluster.Namespace, cluster.ClusterName)
+		oa.DeployTidbClusterOrDie(&cluster)
+		oa.CheckTidbClusterStatusOrDie(&cluster)
+
+		ginkgo.By("scale multiple pvc tidb cluster")
+		cluster.ScaleTiKV(1)
```
DanielZhangQD: tikv cannot be scaled in with less than 3 replicas, we can test scaling in from 4 to 3 and then scale out from 3 to 4.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -1478,16 +1476,22 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
+
+		e2elog.Logf("deploying tidb cluster [%s/%s]", clusterConfig.Namespace, clusterConfig.ClusterName)
+		oa.DeployTidbClusterOrDie(&clusterConfig)
+		oa.CheckTidbClusterStatusOrDie(&clusterConfig)
 
 		ginkgo.By("scale multiple pvc tidb cluster")
-		cluster.ScaleTiKV(1)
-		oa.UpgradeTidbClusterOrDie(&cluster)
-		oa.CheckTidbClusterStatusOrDie(&cluster)
+		clusterConfig.ScaleTiKV(3)
```
DanielZhangQD, line 1492: Please also scale out again because scale-out involves PVC operation.

pkg/manager/member/tikv_member_manager.go
DanielZhangQD: This comment is not addressed.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -372,6 +374,32 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 		}
 	}
 
+	var additionalVolumeClaims []corev1.PersistentVolumeClaim
+	if len(tc.Spec.TiKV.StorageVolumes) > 0 {
+		for _, storageVolume := range tc.Spec.TiKV.StorageVolumes {
+			quantity, err := resource.ParseQuantity(storageVolume.StorageSize)
+			if err != nil {
+				klog.Errorf("cannot parse storage size for tikv, tidbcluster %s/%s, error: %v", tc.Namespace, tc.Name, err)
+				break
```
DanielZhangQD, line 383:
```suggestion
				klog.Errorf("Cannot parse storage size %v in Spec.TiKV.StorageVolumes, tidbcluster %s/%s, error: %v", storageVolume.StorageSize, tc.Namespace, tc.Name, err)
				continue
```

tests/e2e/tidbcluster/tidbcluster.go
DanielZhangQD: Can we scale-out again here?

DanielZhangQD (approved):
LGTM

lichunzhu (approved):
LGTM

ti-srebot (approved):

PR #3440, opened 2020-10-28
tidbmonitor deployment smooth migration to statefulset

mikechengwei:
#2968 

TidbMonitor smooth migration from deployment to statefulset directly and take over the old PV if found.
### Check List 

Tests 
 - Unit test
 - E2E test

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Support TidbMonitor smooth migration from deployment to statefulset directly and take over the old PV if found.
```


pkg/controller/equality.go
```diff
@@ -95,3 +95,27 @@ func IngressEqual(newIngress, oldIngres *extensionsv1beta1.Ingress) (bool, error
 	}
 	return false, nil
 }
+
+// DeploymentPodSpecChanged checks whether the new deployment differs with the old one's last-applied-config
```
DanielZhangQD, line 99: Please check if `templateEqual` [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/utils.go#L136) meets your requirement.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +435,64 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
 	_, err = m.deps.Clientset.PingcapV1alpha1().TidbClusters(tc.Namespace).Patch(tc.Name, types.MergePatchType, mergePatch)
 	return err
 }
+
+func (m *MonitorManager) smoothMigrationToStatefulSet(monitor *v1alpha1.TidbMonitor) error {
+	exist, err := m.deps.TypedControl.Exist(client.ObjectKey{
+		Namespace: monitor.Namespace,
+		Name:      monitor.Name,
+	}, &appsv1.Deployment{})
+	if err != nil {
```
DanielZhangQD, line 444: Is this reasonable for the new creation case?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +435,64 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+				Namespace: monitor.Namespace,
+			},
+		})
+		if err != nil {
+			klog.Errorf("tm[%s/%s]'s deployment failed to delete,err: %v", monitor.Namespace, monitor.Name, err)
+			return err
+		}
+		return nil
+	} else {
+		stsPvcName := fmt.Sprintf("%s-0", GetMonitorObjectName(monitor))
```
DanielZhangQD, line 461: Why creating PVC here if no deployment, the sts will create PVC automatically.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +435,64 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+		stsPvc, err := m.deps.TypedControl.CreateOrUpdatePVC(monitor, stsPvc, false)
+
+		if err != nil {
+			klog.Errorf("tm[%s/%s]'s pvc failed to sync,err: %v", monitor.Namespace, monitor.Name, err)
+			return err
+		}
+
+		deploymentPvcName := GetMonitorObjectName(monitor)
+		deploymentPvc, err := m.deps.PVCLister.PersistentVolumeClaims(monitor.Namespace).Get(deploymentPvcName)
+		if err != nil {
```
DanielZhangQD, line 472: If it's the not found error, I think we can just ignore it.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +435,64 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+			return err
+		}
+		if deploymentPvc != nil {
+			// update meta info for pv
+			deploymentPv, err := m.deps.PVLister.Get(deploymentPvc.Spec.VolumeName)
+			if err != nil {
+				return err
+			}
+			deploymentPv.Spec.ClaimRef.Name = stsPvc.Name
+			_, err = m.deps.PVControl.UpdateMetaInfo(monitor, deploymentPv)
```
DanielZhangQD, line 482: It's not enough to update the meta info, we have to bind the PV to the new PVC of sts, is it easy to do?

DanielZhangQD:
@mikechengwei Did you test your code change with the following cases?
1. New deploy with StatefulSet created.
2. Upgrade from the previous deployment and the old data can be retrieved.

pkg/monitor/monitor/monitor_manager.go
mikechengwei: I think it is reasonable that when monitor deployment exist, we can start migrate work.

pkg/monitor/monitor/monitor_manager.go
mikechengwei: I think we just need to update claimRef .

mikechengwei:
> @mikechengwei Did you test your code change with the following cases?
> 
> 1. New deploy with StatefulSet created.
> 2. Upgrade from the previous deployment and the old data can be retrieved.

First case, i have tested.
Second case, i will test.

mikechengwei:
Now, manual test for smooth migration is ok. Screenshot below:
Before upgrading:

image
image
image

After upgrading:
image
image
image

pkg/controller/generic_control.go
```diff
@@ -628,4 +630,47 @@ func (c *FakeGenericControl) Delete(controller, obj runtime.Object) error {
...
+		for k, v := range desiredSts.Annotations {
+			existingSts.Annotations[k] = v
+		}
+		// only override the default strategy if it is explicitly set in the desiredDep
+		if string(desiredSts.Spec.UpdateStrategy.Type) != "" {
+			existingSts.Spec.UpdateStrategy.Type = desiredSts.Spec.UpdateStrategy.Type
+			if existingSts.Spec.UpdateStrategy.RollingUpdate != nil {
+				existingSts.Spec.UpdateStrategy.RollingUpdate = desiredSts.Spec.UpdateStrategy.RollingUpdate
+			}
+		}
```
DanielZhangQD, line 653: Does `DeepCopy` work?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -66,72 +68,52 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
-		return rerr
-	}
-	if tc.Status.Monitor != nil {
-		if tc.Status.Monitor.Name != monitor.Name || tc.Status.Monitor.Namespace != monitor.Namespace {
-			err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor [%s/%s]", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name, tc.Status.Monitor.Namespace, tc.Status.Monitor.Name)
+	var firstTc *v1alpha1.TidbCluster
+	for index, tcRef := range monitor.Spec.Clusters {
+		if len(tcRef.Namespace) < 1 {
+			tcRef.Namespace = monitor.Namespace
+		}
```
DanielZhangQD, line 75: This has been done in `defaultTidbMonitor`.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -66,72 +68,52 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+		tc, err := m.deps.Clientset.PingcapV1alpha1().TidbClusters(tcRef.Namespace).Get(tcRef.Name, metav1.GetOptions{})
+		if index == 0 {
+			firstTc = tc
+		}
```
DanielZhangQD, line 79: Move this after L83.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -66,72 +68,52 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+			rerr := fmt.Errorf("get tm[%s/%s]'s target tc[%s/%s] failed, err: %v", monitor.Namespace, monitor.Name, tcRef.Namespace, tcRef.Name, err)
+			return rerr
+		}
+		if tc.Status.Monitor != nil {
+			if tc.Status.Monitor.Name != monitor.Name || tc.Status.Monitor.Namespace != monitor.Namespace {
+				err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor [%s/%s]", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name, tc.Status.Monitor.Namespace, tc.Status.Monitor.Name)
+				m.deps.Recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, err.Error())
+				return err
+			}
+		}
```
DanielZhangQD, line 90: Is this necessary?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +395,83 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
 	_, err = m.deps.Clientset.PingcapV1alpha1().TidbClusters(tc.Namespace).Patch(tc.Name, types.MergePatchType, mergePatch)
 	return err
 }
+
+func (m *MonitorManager) smoothMigrationToStatefulSet(monitor *v1alpha1.TidbMonitor) (bool, error) {
+	// determine whether there is an old deployment
+	deploymentName := GetMonitorObjectName(monitor)
+	exist, err := m.deps.TypedControl.Exist(client.ObjectKey{
```
DanielZhangQD, line 402: Just use the deploymentLister in `m.deps` to check.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +395,83 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+		Namespace: monitor.Namespace,
+		Name:      deploymentName,
+	}, &appsv1.Deployment{})
+	if err != nil {
+		klog.Errorf("smoothMigration tm[%s/%s]'s,old deployment failed to get,err: %v", monitor.Namespace, monitor.Name, err)
```
DanielZhangQD, line 407:
```suggestion
		klog.Errorf("Fail to get deployment for tm [%s/%s], err: %v", monitor.Namespace, monitor.Name, err)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +395,83 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+		return false, err
+	}
+	if exist {
+		klog.Infof("smoothMigration tm[%s/%s]'s,old deployment is exist", monitor.Namespace, monitor.Name)
```
DanielZhangQD, line 411:
```suggestion
		klog.Infof("The deployment exists, start smooth migration for tm [%s/%s]", monitor.Namespace, monitor.Name)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +395,83 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+		// if deployment exist, delete it and wait next reconcile.
+		err = m.deps.TypedControl.Delete(monitor, &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      deploymentName,
+				Namespace: monitor.Namespace,
+			},
+		})
+		if err != nil {
+			klog.Errorf("smoothMigration tm[%s/%s]'s,old deployment failed to delete,err: %v", monitor.Namespace, monitor.Name, err)
```
DanielZhangQD, line 420:
```suggestion
			klog.Errorf("Fail to delete the deployment for tm [%s/%s], err: %v", monitor.Namespace, monitor.Name, err)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +395,83 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+			return false, err
+		}
+
+		return !monitor.Spec.Persistent, nil
+	} else {
```
DanielZhangQD, line 425: The following block should also be done if the deployment is found in the above `if` branch.
Why separate it with the deployment deletion logic?

pkg/controller/generic_control.go
```diff
@@ -55,6 +56,8 @@ type TypedControlInterface interface {
 	CreateOrUpdateService(controller runtime.Object, svc *corev1.Service) (*corev1.Service, error)
 	// CreateOrUpdateDeployment create the desired deployment or update the current one to desired state if already existed
 	CreateOrUpdateDeployment(controller runtime.Object, deploy *appsv1.Deployment) (*appsv1.Deployment, error)
+	// CreateOrUpdateStatefulSet create the desired statefulset or update the current one to desired state if already existed
+	CreateOrUpdateStatefulSet(controller runtime.Object, deploy *appsv1.StatefulSet) (*appsv1.StatefulSet, error)
```
DanielZhangQD, line 60:
```suggestion
	CreateOrUpdateStatefulSet(controller runtime.Object, sts *appsv1.StatefulSet) (*appsv1.StatefulSet, error)
```

pkg/monitor/monitor/monitor_manager.go
DanielZhangQD: I mean move L79-L81 after L85.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -66,72 +73,49 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+			if tc.Status.Monitor.Name != monitor.Name || tc.Status.Monitor.Namespace != monitor.Namespace {
+				err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor [%s/%s]", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name, tc.Status.Monitor.Namespace, tc.Status.Monitor.Name)
+				m.deps.Recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, err.Error())
+				return err
+			}
+		}
+		// TODO: Support validating webhook that forbids the tidbmonitor to update the monitorRef for the tidbcluster whose monitorRef has already
+		// been set by another TidbMonitor.
+		// Patch tidbcluster status first to avoid multi tidbmonitor monitoring the same tidbcluster
+		if err := m.patchTidbClusterStatus(&tcRef, monitor); err != nil {
```
DanielZhangQD, line 96: Please ignore the heterogeneous cluster.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -157,31 +141,7 @@ func (m *MonitorManager) syncTidbMonitorService(monitor *v1alpha1.TidbMonitor) e
...
-	if err != nil {
-		return err
-	}
-	return nil
-}
-
-func (m *MonitorManager) syncTidbMonitorDeployment(tc *v1alpha1.TidbCluster, monitor *v1alpha1.TidbMonitor) error {
+func (m *MonitorManager) syncTidbMonitorStatefulset(tc *v1alpha1.TidbCluster, monitor *v1alpha1.TidbMonitor) error {
 
 	cm, err := m.syncTidbMonitorConfig(tc, monitor)
```
DanielZhangQD, line 146: The configmap name is fixed, so if we update the config in a new release, the  Pod will not be rolling updated, will this be addressed in another PR?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -200,12 +160,22 @@ func (m *MonitorManager) syncTidbMonitorDeployment(tc *v1alpha1.TidbCluster, mon
...
-	deployment, err := getMonitorDeployment(sa, cm, secret, monitor, tc)
+	result, err := m.smoothMigrationToStatefulSet(monitor)
 	if err != nil {
-		klog.Errorf("tm[%s/%s]'s deployment failed to generate,err: %v", monitor.Namespace, monitor.Name, err)
+		klog.Errorf("Fail to migrate from deployment to statefulset for tm [%s/%s], err: %v", monitor.Namespace, monitor.Name, err)
 		return err
 	}
-	_, err = m.deps.TypedControl.CreateOrUpdateDeployment(monitor, deployment)
+	if !result {
+		klog.Errorf("Wait for the smooth migration to be done successfully for tm [%s/%s], err: %v", monitor.Namespace, monitor.Name, err)
```
DanielZhangQD, line 169: `err` is `nil` here.

DanielZhangQD:
```suggestion
		klog.Infof("Wait for the smooth migration to be done successfully for tm [%s/%s]", monitor.Namespace, monitor.Name)
```

pkg/monitor/monitor/monitor_manager.go
DanielZhangQD: Why it's better? This will do API call to kube-apiserver each time which is not a good practice especially in a large-scale Kubernetes cluster where kube-apiserver may already be under pressure.

tests/e2e/tidbcluster/serial.go
```diff
@@ -507,6 +508,55 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+			oldVolumeName := deploymentPvc.Spec.VolumeName
+			ginkgo.By("Upgrade tidb-operator and CRDs to current version")
+			ocfg.Tag = cfg.OperatorTag
+			ocfg.Image = cfg.OperatorImage
+			oa.InstallCRDOrDie(ocfg)
+			oa.UpgradeOperatorOrDie(ocfg)
+
+			err = wait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
+				newStsPvcName := fmt.Sprintf("%s-monitor-0", monitorName)
+				stsPvc, err := c.CoreV1().PersistentVolumeClaims(tc.Namespace).Get(newStsPvcName, metav1.GetOptions{})
```
DanielZhangQD, line 543: Please also check the deployment and its PVC are removed.

tests/e2e/tidbcluster/tidbcluster.go
```diff
@@ -654,38 +654,21 @@ var _ = ginkgo.Describe("[tidb-operator] TiDBCluster", func() {
...
-		pvc, err := c.CoreV1().PersistentVolumeClaims(ns).Get("e2e-monitor-monitor", metav1.GetOptions{})
+		pvcName := "monitor-data-e2e-monitor-0"
+		pvc, err := c.CoreV1().PersistentVolumeClaims(ns).Get(pvcName, metav1.GetOptions{})
 		framework.ExpectNoError(err, "Expected fetch tidbmonitor pvc success")
 		pvName := pvc.Spec.VolumeName
 		pv, err := c.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
 		framework.ExpectNoError(err, "Expected fetch tidbmonitor pv success")
 
 		err = wait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
-			value, existed := pv.Labels[label.ComponentLabelKey]
```
DanielZhangQD, line 672: Why remove this? The PV for TidbMonitor should also sync the labels as we have done for TidbCluster.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -66,72 +73,49 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
 		m.deps.Recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, message)
 		return err
 	}
 
-	// After the pvc has consumer, we sync monitor pv's labels
-	if monitor.Spec.Persistent {
-		if err := m.syncTidbMonitorPV(monitor, pvc); err != nil {
-			return err
-		}
-	}
```
DanielZhangQD, line 134: Why remove the PV label sync?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -58,6 +60,11 @@ func NewMonitorManager(deps *controller.Dependencies) *MonitorManager {
 }
 
 func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
+	// syncing all PVs managed by operator's reclaim policy to Retain
+	if err := m.pvManager.SyncMonitor(monitor); err != nil {
+		return err
+	}
```
DanielZhangQD, line 66: We can move this block to the code where the PV labels are synced as they are both for PVs.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -66,72 +73,49 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
-		rerr := fmt.Errorf("get tm[%s/%s]'s target tc[%s/%s] failed, err: %v", monitor.Namespace, monitor.Name, tcRef.Namespace, tcRef.Name, err)
-		return rerr
-	}
-	if tc.Status.Monitor != nil {
-		if tc.Status.Monitor.Name != monitor.Name || tc.Status.Monitor.Namespace != monitor.Namespace {
-			err := fmt.Errorf("tm[%s/%s]'s target tc[%s/%s] already referenced by TidbMonitor [%s/%s]", monitor.Namespace, monitor.Name, tc.Namespace, tc.Name, tc.Status.Monitor.Namespace, tc.Status.Monitor.Name)
+	var firstTc *v1alpha1.TidbCluster
+	for index, tcRef := range monitor.Spec.Clusters {
+		tc, err := m.deps.Clientset.PingcapV1alpha1().TidbClusters(tcRef.Namespace).Get(tcRef.Name, metav1.GetOptions{})
+		if index == 0 {
```
DanielZhangQD, line 79: The first non-heterogeneous TC.

pkg/manager/meta/reclaim_policy_manager.go
```diff
@@ -42,7 +42,7 @@ func (m *reclaimPolicyManager) Sync(tc *v1alpha1.TidbCluster) error {
 }
 
 func (m *reclaimPolicyManager) SyncMonitor(tm *v1alpha1.TidbMonitor) error {
-	return m.sync(v1alpha1.TiDBMonitorKind, tm, false, *tm.Spec.PVReclaimPolicy)
+	return m.sync(v1alpha1.TiDBMonitorKind, tm, true, *tm.Spec.PVReclaimPolicy)
```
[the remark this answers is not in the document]
mikechengwei: I think we don’t need other field to enable pv policy sync.

pkg/monitor/monitor/monitor_manager.go
mikechengwei: Yes,another pr support it.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +397,83 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+				return false, err
+			}
+			stsPvcName := fmt.Sprintf("monitor-data-%s-0", GetMonitorObjectName(monitor))
+			deploymentPv, err := m.deps.PVLister.Get(deploymentPvc.Spec.VolumeName)
+			if err != nil && !errors.IsNotFound(err) {
+				klog.Errorf("smoothMigration tm[%s/%s]'s,old deployment pv failed to get,err: %v", monitor.Namespace, monitor.Name, err)
+				return false, err
+			}
+			deploymentPv.Spec.ClaimRef.Name = stsPvcName
+			err = m.deps.PVControl.PatchPVClaimRef(monitor, deploymentPv, stsPvcName)
```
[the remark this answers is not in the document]
mikechengwei: We must delete pvc before patch pv, because pv will roll back to bind old pvc immediately by apiserver when old pvc exist .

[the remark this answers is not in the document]
DanielZhangQD: @mikechengwei Then how can we handle the case that the PVC is deleted successfully but an error occurs in the following operations?

mikechengwei:
@codecov-io

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -180,8 +180,9 @@ type TidbClusterRef struct {
 	ClusterDomain string `json:"clusterDomain,omitempty"`
 }
 
-// TODO: sync status
 type TidbMonitorStatus struct {
+	// old deployment storage status
+	OldDeploymentStorageStatus *OldDeploymentStorageStatus `json:"oldDeploymentStatus,omitempty"`
```
DanielZhangQD, line 185:
```suggestion
	// Storage status for deployment 
	DeploymentStorageStatus *DeploymentStorageStatus `json:"deploymentStorageStatus,omitempty"`
```

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -195,3 +196,12 @@ type TidbMonitorList struct {
...
 	Items []TidbMonitor `json:"items"`
 }
+
+// GrafanaSpec is the desired state of grafana
+type OldDeploymentStorageStatus struct {
+	// old pvc name
+	PvcName string `json:"pvcName,omitempty"`
+
+	// old pv name
+	PvName string `json:"pvName,omitempty"`
```
DanielZhangQD, line 206:
```suggestion
// DeploymentStorageStatus is the storage information of the deployment
type DeploymentStorageStatus struct {
	// PVC name
	PVCName string `json:"pvcName,omitempty"`

	// PV name
	PVName string `json:"pvName,omitempty"`
```

pkg/apis/pingcap/v1alpha1/validation/validation.go
```diff
@@ -605,3 +609,21 @@ func validateAdditionalContainers(containers []corev1.Container, fldPath *field.
 
 	return allErrs
 }
+
+func validateStorageInfo(spec *v1alpha1.TidbMonitorSpec, fldPath *field.Path) field.ErrorList {
```
DanielZhangQD, line 613: Suggest passing the `spec.Storage` as a parameter instead of `spec` to make this function generic, which can be used by validating other similar fields.

pkg/apis/pingcap/v1alpha1/validation/validation.go
```diff
@@ -605,3 +609,21 @@ func validateAdditionalContainers(containers []corev1.Container, fldPath *field.
...
+	allErrs := field.ErrorList{}
+	if spec.Persistent {
```
DanielZhangQD, line 615: This is already checked in L73.

pkg/controller/pv_control.go
```diff
@@ -79,6 +82,41 @@ func (c *realPVControl) PatchPVReclaimPolicy(obj runtime.Object, pv *corev1.Pers
 	return err
 }
 
+func (c *realPVControl) GetPV(name string) (*corev1.PersistentVolume, error) {
+	return c.pvLister.Get(name)
+}
+
+func (c *realPVControl) CreatePV(obj runtime.Object, pv *corev1.PersistentVolume) error {
```
DanielZhangQD, line 89: Why do we need to create PV?

pkg/apis/pingcap/v1alpha1/validation/validation.go
```diff
@@ -61,14 +61,18 @@ func ValidateDMCluster(dc *v1alpha1.DMCluster) field.ErrorList {
 	return allErrs
 }
 
-func ValidateTidbMonitor(monitor *v1alpha1.TidbMonitor) field.ErrorList {
+func ValidateCreateTidbMonitor(monitor *v1alpha1.TidbMonitor) field.ErrorList {
```
DanielZhangQD, line 64: Please revert this.

pkg/controller/pv_control.go
```diff
@@ -255,4 +295,39 @@ func (c *FakePVControl) UpdateMetaInfo(obj runtime.Object, pv *corev1.Persistent
...
+	if c.createPVTracker.ErrorReady() {
+		defer c.createPVTracker.Reset()
+		return c.createPVTracker.GetError()
+	}
+
+	return c.PVIndexer.Add(pv)
+}
+
+func (c *FakePVControl) GetPV(name string) (*corev1.PersistentVolume, error) {
+	defer c.updatePVTracker.Inc()
```
DanielZhangQD, line 321: Why call `updatePVTracker` here?

pkg/controller/pvc_control.go
```diff
@@ -248,6 +269,17 @@ func (c *FakePVCControl) UpdatePVC(_ runtime.Object, pvc *corev1.PersistentVolum
 	return pvc, c.PVCIndexer.Update(pvc)
 }
 
+// AddPVC add new pvc
+func (c *FakePVCControl) CreatePVC(controller runtime.Object, pvc *corev1.PersistentVolumeClaim) error {
+	defer c.updatePVCTracker.Inc()
```
DanielZhangQD, line 274: Should add a `createPVCTracker `

pkg/controller/tidbmonitor/tidb_monitor_control.go
```diff
@@ -14,36 +14,70 @@
...
+	// Update tidbmonitor status
+	UpdateTidbMonitor(*v1alpha1.TidbMonitor) (*v1alpha1.TidbMonitor, error)
 }
 
 // NewDefaultTidbMonitorControl returns a new instance of the default TidbMonitor ControlInterface
-func NewDefaultTidbMonitorControl(monitorManager monitor.MonitorManager) ControlInterface {
-	return &defaultTidbMonitorControl{monitorManager: monitorManager}
+func NewDefaultTidbMonitorControl(cli versioned.Interface,
+	tmLister listers.TidbMonitorLister, monitorManager monitor.MonitorManager, recorder record.EventRecorder) ControlInterface {
+	return &defaultTidbMonitorControl{cli: cli, tmLister: tmLister, monitorManager: monitorManager, recorder: recorder}
```
DanielZhangQD, line 45: The `monitorManager.deps` includes all the `cli`s and `lister`s.

pkg/controller/tidbmonitor/tidb_monitor_control.go
```diff
@@ -14,36 +14,70 @@
...
 	if err := c.reconcileTidbMonitor(tm); err != nil {
 		errs = append(errs, err)
 	}
-	return errors.NewAggregate(errs)
+	return errorutils.NewAggregate(errs)
 }
 
 func (c *defaultTidbMonitorControl) reconcileTidbMonitor(tm *v1alpha1.TidbMonitor) error {
-	return c.monitorManager.SyncMonitor(tm)
+	if !c.validate(tm) {
```
DanielZhangQD, line 64: Usually, we should call the defaulting function before the validation.

pkg/controller/tidbmonitor/tidb_monitor_control.go
```diff
@@ -70,4 +104,53 @@ func (tmc *FakeTidbMonitorControl) ReconcileTidbMonitor(tm *v1alpha1.TidbMonitor
...
+}
+
 var _ ControlInterface = &FakeTidbMonitorControl{}
+
+func (c *defaultTidbMonitorControl) UpdateTidbMonitor(tm *v1alpha1.TidbMonitor) (*v1alpha1.TidbMonitor, error) {
+	ns := tm.GetNamespace()
+	tmName := tm.GetName()
+
+	status := tm.Status.DeepCopy()
+	var updateTC *v1alpha1.TidbMonitor
```
DanielZhangQD, line 119:
```suggestion
	var update *v1alpha1.TidbMonitor
```

pkg/controller/pv_control.go
mikechengwei: It used for unit test mock create pv.

pkg/controller/pv_control.go
mikechengwei: I refer the pvc code  https://github.com/pingcap/tidb-operator/blob/2ef71285d46e6b8ad962343d4b3c2ab8d4afa133/pkg/controller/pvc_control.go#L272.

pkg/controller/pvc_control.go
mikechengwei: It used for unit test mock create pv.

pkg/controller/tidbmonitor/tidb_monitor_control.go
mikechengwei: monitorManager is interface , we can't use monitorManager.deps

pkg/manager/meta/meta_manager.go
```diff
@@ -57,8 +57,7 @@ func (m *metaManager) Sync(tc *v1alpha1.TidbCluster) error {
 		if err != nil {
 			return err
 		}
-
-		if component := pod.Labels[label.ComponentLabelKey]; component != label.PDLabelVal && component != label.TiKVLabelVal && component != label.TiFlashLabelVal {
+		if component := pod.Labels[label.ComponentLabelKey]; component != label.PDLabelVal && component != label.TiKVLabelVal && component != label.TiFlashLabelVal && component != label.TiDBMonitorVal {
```
DanielZhangQD, line 60: This function is for TC, please check L41.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -100,38 +105,23 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
-	if err := m.syncTidbMonitorDeployment(tc, monitor); err != nil {
+	// Sync Statefulset
+	if err := m.syncTidbMonitorStatefulset(firstTc, monitor); err != nil {
 		message := fmt.Sprintf("Sync TidbMonitor[%s/%s] Deployment failed,err:%v", monitor.Namespace, monitor.Name, err)
 		m.deps.Recorder.Event(monitor, corev1.EventTypeWarning, FailedSync, message)
 		return err
 	}
 
-	// After the pvc has consumer, we sync monitor pv's labels
+	// Sync PVC
```
DanielZhangQD, line 125:
```suggestion
	// Sync PV
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -100,38 +105,23 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
 	if monitor.Spec.Persistent {
-		if err := m.syncTidbMonitorPV(monitor, pvc); err != nil {
+		// syncing all PVs managed by this tidbmonitor
+		if err := m.pvManager.SyncMonitor(monitor); err != nil {
 			return err
 		}
+		klog.V(4).Infof("tm[%s/%s]'s pv synced", monitor.Namespace, monitor.Name)
 	}
+
 	klog.V(4).Infof("tm[%s/%s]'s deployment synced", monitor.Namespace, monitor.Name)
```
DanielZhangQD, line 134:
```suggestion
	klog.V(4).Infof("tm[%s/%s]'s StatefulSet synced", monitor.Namespace, monitor.Name)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -147,71 +137,96 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
-
-	pvc := getMonitorPVC(monitor)
-	pvc, err := m.deps.TypedControl.CreateOrUpdatePVC(monitor, pvc, false)
-	if err != nil {
-		klog.Errorf("tm[%s/%s]'s pvc failed to sync,err: %v", monitor.Namespace, monitor.Name, err)
-		return nil, err
-	}
-	return pvc, nil
-}
+		equal, err := controller.ServiceEqual(newSvc, oldSvc)
```
DanielZhangQD, line 166: We have to preserve some fields in the service spec, e.g. clusterIP, NodePort, etc., please refer to the function for TiDB https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tidb_member_manager.go#L256.
Please also check the logs of `tidb-controller-manager` during your test to make sure that no unnecessary updates for the resources during the sync.

pkg/monitor/monitor/util.go
```diff
@@ -33,12 +33,17 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/klog"
 )
 
 func GetMonitorObjectName(monitor *v1alpha1.TidbMonitor) string {
 	return fmt.Sprintf("%s-monitor", monitor.Name)
 }
 
+func GetMonitorFirstPVCName(name string) string {
+	return fmt.Sprintf(v1alpha1.TidbMonitorMemberType.String()+"-%s-0", name)
```
DanielZhangQD, line 44: Is this the correct PVC name for sts?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +442,96 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
 	_, err = m.deps.Clientset.PingcapV1alpha1().TidbClusters(tc.Namespace).Patch(tc.Name, types.MergePatchType, mergePatch)
 	return err
 }
+
+func (m *MonitorManager) smoothMigrationToStatefulSet(monitor *v1alpha1.TidbMonitor) (bool, error) {
```
DanielZhangQD, line 455: What about the following logic:
Get deployment successfully, delete it and return
Get deployment error except for the non NotFound error, return
Get deployment NotFound error
    PVC found
        PV not found, return true, no need to migrate
        PV found
            patch the status of monitor to make sure the PV info is updated (return error for failure)
            delete PVC
            patch PV
    PVC not found
        no PV in status of monitor, return true, no need to migrate
        PV found in status of monitor, patch PV (only patch if necessary, check the claimRef first)

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +442,96 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+				}
+
+				err = m.deps.PVCControl.DeletePVC(monitor, &corev1.PersistentVolumeClaim{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      deploymentPvcName,
+						Namespace: monitor.Namespace,
+					},
+				})
+				if err != nil {
+					monitor.Status.DeploymentStorageStatus = &v1alpha1.DeploymentStorageStatus{
```
DanielZhangQD, line 496: This cannot be guaranteed to be updated successfully.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +442,96 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+					klog.Errorf("Smooth migration for tm[%s/%s], fail to get PV %s, err: %v", monitor.Namespace, monitor.Name, deploymentPvc.Spec.VolumeName, err)
+					return false, err
+				}
+				if deploymentPv.Spec.ClaimRef == nil {
+					deploymentPv.Spec.ClaimRef = &corev1.ObjectReference{}
+				}
+				deploymentPv.Spec.ClaimRef.Name = firstStsPvcName
+				err = m.deps.PVControl.PatchPVClaimRef(monitor, deploymentPv, firstStsPvcName)
+				if err != nil {
+					monitor.Status.DeploymentStorageStatus = &v1alpha1.DeploymentStorageStatus{
```
DanielZhangQD, line 517: ditto.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +442,96 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+						PvName: deploymentPvc.Spec.VolumeName,
+					}
+					klog.Errorf("Smooth migration for tm[%s/%s], fail to get PV %s, err: %v", monitor.Namespace, monitor.Name, deploymentPvc.Spec.VolumeName, err)
+					return false, err
+				}
+				if deploymentPv.Spec.ClaimRef == nil {
+					deploymentPv.Spec.ClaimRef = &corev1.ObjectReference{}
+				}
+				deploymentPv.Spec.ClaimRef.Name = firstStsPvcName
+				err = m.deps.PVControl.PatchPVClaimRef(monitor, deploymentPv, firstStsPvcName)
```
DanielZhangQD, line 515: We should check the ClaimRef before patch.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -54,6 +61,7 @@ func NewMonitorManager(deps *controller.Dependencies) *MonitorManager {
 		deps:               deps,
 		pvManager:          meta.NewReclaimPolicyManager(deps),
 		discoveryInterface: discoverycachedmemory.NewMemCacheClient(deps.KubeClientset.Discovery()),
+		recorder:           deps.Recorder,
```
DanielZhangQD, line 64: Why add another recorder here?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -147,71 +151,111 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
 
 func (m *MonitorManager) syncTidbMonitorService(monitor *v1alpha1.TidbMonitor) error {
 	services := getMonitorService(monitor)
-	for _, svc := range services {
-		_, err := m.deps.TypedControl.CreateOrUpdateService(monitor, svc)
+	for _, newSvc := range services {
+		oldSvcTmp, err := m.deps.ServiceLister.Services(newSvc.Namespace).Get(newSvc.Name)
```
DanielZhangQD, line 155: What about defining a function for this code block to create or update service?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +471,186 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+					klog.Errorf("Smooth migration for tm[%s/%s], the PVC of the deployment get error:%v", monitor.Namespace, monitor.Name, err)
+					return false, err
+				}
+
+				// firstly patch status
+				if monitor.Status.DeploymentStorageStatus == nil {
+					monitor.Status.DeploymentStorageStatus = &v1alpha1.DeploymentStorageStatus{
+						PvName: deploymentPvc.Spec.VolumeName,
+					}
+					err = m.deps.TypedControl.UpdateStatus(monitor)
```
DanielZhangQD, line 514: This will erase the other fields in the status, you can patch the status similar with what we're doing in https://github.com/pingcap/tidb-operator/blob/master/pkg/monitor/monitor/monitor_manager.go#L397

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +471,186 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+						return true, nil
+					}
+					klog.Errorf("Smooth migration for tm[%s/%s], the PVC of the deployment get error:%v", monitor.Namespace, monitor.Name, err)
+					return false, err
+				}
+
+				// firstly patch status
+				if monitor.Status.DeploymentStorageStatus == nil {
+					monitor.Status.DeploymentStorageStatus = &v1alpha1.DeploymentStorageStatus{
+						PvName: deploymentPvc.Spec.VolumeName,
```
DanielZhangQD, line 512: We should check `deploymentPvc.Spec.VolumeName`, if no value, we can skip the migration.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +471,186 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+				if deploymentPv.Spec.ClaimRef == nil {
+					deploymentPv.Spec.ClaimRef = &corev1.ObjectReference{}
+				}
+
+				deploymentPv.Spec.ClaimRef.Name = firstStsPvcName
+				err = m.deps.PVControl.PatchPVClaimRef(monitor, deploymentPv, firstStsPvcName)
+				if err != nil {
+					klog.Errorf("Smooth migration for tm[%s/%s], fail to patch PV %s, err: %v", monitor.Namespace, monitor.Name, deploymentPv.Name, err)
+					return false, err
+				}
```
DanielZhangQD, line 553: We can define a function for this and it can also be used around L495.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +471,186 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+				if err != nil {
+					if errors.IsNotFound(err) {
+						// If the PVC of the deployment does not exist and no old pvc status, we don't need to migrate.
+						if monitor.Status.DeploymentStorageStatus != nil && len(monitor.Status.DeploymentStorageStatus.PvName) > 0 {
+							deploymentPv, err := m.deps.PVLister.Get(monitor.Status.DeploymentStorageStatus.PvName)
+							if err != nil {
+								klog.Errorf("Smooth migration for tm[%s/%s], fail to get PV %s, err: %v", monitor.Namespace, monitor.Name, monitor.Status.DeploymentStorageStatus.PvName, err)
+								return false, err
+							}
+							deploymentPv.Spec.ClaimRef.Name = firstStsPvcName
```
DanielZhangQD, line 494: Check if it's already the value we want.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +471,186 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+						Namespace: monitor.Namespace,
+					},
+				})
+				if err != nil {
+					klog.Errorf("Fail to delete the PVC %s for tm [%s/%s], err: %v", deploymentPvcName, monitor.Namespace, monitor.Name, err)
+					return false, err
+				}
+
+				deploymentPv, err := m.deps.PVLister.Get(deploymentPvc.Spec.VolumeName)
+				if err != nil && !errors.IsNotFound(err) {
```
DanielZhangQD, line 533: What if `errors.IsNotFound(err)`?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +471,186 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func (m *MonitorManager) resolvePVCFromPod(pod *corev1.Pod) ([]*corev1.PersistentVolumeClaim, error) {
```
DanielZhangQD, line 635: Cannot use the function in `pkg/manager/meta/meta_manager.go`?

pkg/monitor/monitor/monitor_manager.go
DanielZhangQD: There are too many nested `if`s, please optimize the code to handle the code where it can return quickly first.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +471,186 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+	if len(errs) > 0 {
+		aggregatedErr := errs.ToAggregate()
+		klog.Errorf("tidbmonitor %s/%s is not valid and must be fixed first, aggregated error: %v", tidbmonitor.GetNamespace(), tidbmonitor.GetName(), aggregatedErr)
+		c.recorder.Event(tidbmonitor, corev1.EventTypeWarning, "FailedValidation", aggregatedErr.Error())
+		return false
+	}
+	return true
+}
+
+func (m *MonitorManager) syncTidbMonitorPV(tm *v1alpha1.TidbMonitor) error {
```
DanielZhangQD, line 592: This seems to duplicate with the `Sync` in `meta_manager.go`, I think we can change the parameter of `Sync` in `meta_manager.go` to handle different cases.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -60,6 +60,8 @@ const (
 	DMWorkerMemberType MemberType = "dm-worker"
 	// SlowLogTailerMemberType is tidb log tailer container type
 	SlowLogTailerMemberType MemberType = "slowlog"
+	// TidbMonitorMemberType is tidbmonitor type
+	TidbMonitorMemberType MemberType = "tidbmonitor"
```
DanielZhangQD, line 64:
```suggestion
	TidbMonitorMemberType MemberType = "monitor"
```

pkg/controller/tidbmonitor/tidb_monitor_control.go
mikechengwei: Do you mean deps as parameter directly?

pkg/monitor/monitor/monitor_manager.go
mikechengwei: record validate err event.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,243 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+	ns := tm.GetNamespace()
+	instanceName := tm.Name
+
+	l, err := label.New().Instance(instanceName).Selector()
+	if err != nil {
+		return err
+	}
+	pods, err := m.deps.PodLister.Pods(ns).List(l)
+	if err != nil {
+		return fmt.Errorf("metaManager.Sync: failed to list pods for tidbmonitor %s/%s, selector: %s, error: %v", ns, instanceName, l, err)
```
DanielZhangQD, line 539:
```suggestion
		return fmt.Errorf("fail to list pods for tidbmonitor %s/%s, selector: %s, error: %v", ns, instanceName, l, err)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,243 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+		return false
+	}
+	return true
+}
+
+func (m *MonitorManager) syncTidbMonitorPV(tm *v1alpha1.TidbMonitor) error {
+	ns := tm.GetNamespace()
+	instanceName := tm.Name
+
+	l, err := label.New().Instance(instanceName).Selector()
```
DanielZhangQD, line 533: We should also add component label here.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,243 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+	pods, err := m.deps.PodLister.Pods(ns).List(l)
+	if err != nil {
+		return fmt.Errorf("metaManager.Sync: failed to list pods for tidbmonitor %s/%s, selector: %s, error: %v", ns, instanceName, l, err)
+	}
+
+	for _, pod := range pods {
+		if component := pod.Labels[label.ComponentLabelKey]; component != label.TiDBMonitorVal {
+			// only update tidbmonitor pod
+			continue
+		}
```
DanielZhangQD, line 546: Add component label in L533.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,243 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+			return err
+		}
+		err = m.deps.ServiceControl.CreateService(monitor, newSvc)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("syncTidbMonitorService: failed to get svc %s for cluster %s/%s, error: %s", newSvc.Name, monitor.Namespace, monitor.Name, err)
```
DanielZhangQD, line 586:
```suggestion
		return fmt.Errorf("createOrUpdateService: fail to get svc %s for tm %s/%s, error: %s", newSvc.Name, monitor.Namespace, monitor.Name, err)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,243 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+	if err != nil {
+		klog.Errorf("Smooth migration for tm[%s/%s], fail to get PV %s, err: %v", monitor.Namespace, monitor.Name, pvName, err)
+		return err
+	}
+
+	if deploymentPv.Spec.ClaimRef != nil && deploymentPv.Spec.ClaimRef.Name == patchPvcName {
+		// smooth migration successfully and clean status
+		monitor.Status.DeploymentStorageStatus = nil
+		return nil
+	}
```
DanielZhangQD, line 635: This should be done outside of this function.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,253 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+		if err != nil {
+			klog.Errorf("Smooth migration for tm[%s/%s], fail to patch PV %s, err: %v", monitor.Namespace, monitor.Name, monitor.Status.DeploymentStorageStatus.PvName, err)
+			return false, err
+		}
+		// smooth migration successfully and clean status
+		monitor.Status.DeploymentStorageStatus = nil
+		return true, nil
+	}
+
+	if len(deploymentPvc.Spec.VolumeName) <= 0 {
```
DanielZhangQD, line 490: For this case, no need to migrate, we should print info log and `return true, nil`.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,253 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+		return false
+	}
+	return true
+}
+
+func (m *MonitorManager) syncTidbMonitorPV(tm *v1alpha1.TidbMonitor) error {
+	ns := tm.GetNamespace()
+	instanceName := tm.Name
+
+	l, err := label.New().Instance(instanceName).TiDBMonitor().Selector()
```
DanielZhangQD, line 555: Should use `buildTidbMonitorLabel` here?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,253 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+	if err != nil {
+		return err
+	}
+	pods, err := m.deps.PodLister.Pods(ns).List(l)
+	if err != nil {
+		return fmt.Errorf("fail to list pods for tidbmonitor %s/%s, selector: %s, error: %v", ns, instanceName, l, err)
+	}
+
+	for _, pod := range pods {
+		if component := pod.Labels[label.ComponentLabelKey]; component != label.TiDBMonitorVal {
```
DanielZhangQD, line 565: Still need to check here?

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,253 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func (m *MonitorManager) createOrUpdateService(newSvc *corev1.Service, monitor *v1alpha1.TidbMonitor) error {
```
DanielZhangQD, line 594: We can define this as a common function in util.go to share with other modules, right?

pkg/monitor/monitor/monitor_manager.go
mikechengwei: How to build labelSelector with buildTidbMonitorLabel return value?

pkg/monitor/monitor/monitor_manager.go
mikechengwei: No,i try to move this method to pkg/util package,it will cause package cycle call.

tests/e2e/util/image/image.go
```diff
@@ -31,6 +31,7 @@ const (
 	TiDBV3UpgradeVersion          = "v3.1.2"
 	TiDBV4Version                 = "v4.0.4"
 	TiDBV4UpgradeVersion          = "v4.0.5"
+	TiDBV47UpgradeVersion         = "v4.0.5"
```
DanielZhangQD, line 34: Please just update `TiDBV4Version` to v407 and `TiDBV4UpgradeVersion ` to v408.

pkg/monitor/monitor/monitor_manager.go
DanielZhangQD: What about make it a function in `pkg/member/manager` package just as you have done for `UpdateStatefulSet`.

pkg/manager/member/utils.go
```diff
@@ -447,3 +448,55 @@ func shouldRecoverDM(dc *v1alpha1.DMCluster, component string, podLister corelis
...
+	if errors.IsNotFound(err) {
+		err = controller.SetServiceLastAppliedConfigAnnotation(newSvc)
+		if err != nil {
+			return err
+		}
+		err = serviceControl.CreateService(monitor, newSvc)
+		if err != nil {
+			return err
+		}
+		return nil
```
DanielZhangQD, line 463:
```suggestion
		return serviceControl.CreateService(monitor, newSvc)
```

pkg/manager/member/utils.go
```diff
@@ -447,3 +448,55 @@ func shouldRecoverDM(dc *v1alpha1.DMCluster, component string, podLister corelis
...
+		}
+		// also override labels when adopt orphan
+		if isOrphan {
+			svc.OwnerReferences = newSvc.OwnerReferences
+			svc.Labels = newSvc.Labels
+		}
+		_, err = serviceControl.UpdateService(monitor, &svc)
+		if err != nil {
+			return err
+		}
```
DanielZhangQD, line 499:
```suggestion
		return err
```

pkg/manager/member/utils.go
```diff
@@ -447,3 +448,55 @@ func shouldRecoverDM(dc *v1alpha1.DMCluster, component string, podLister corelis
 	}
 	return true
 }
+
+func CreateOrUpdateService(serviceLister corelisters.ServiceLister, serviceControl controller.ServiceControlInterface, newSvc *corev1.Service, monitor *v1alpha1.TidbMonitor) error {
```
DanielZhangQD, line 452:
```suggestion
func CreateOrUpdateService(serviceLister corelisters.ServiceLister, serviceControl controller.ServiceControlInterface, newSvc *corev1.Service, obj runtime.Object) error {
```

DanielZhangQD:
```
	meta, ok := obj.(metav1.Object)
	if !ok {
		return fmt.Errorf("%T is not a metav1.Object", obj)
	}
```

DanielZhangQD: Then we can use the same function in the `syncTiDBService` function in `tidb_member_manager.go`.

tests/pkg/fixture/fixture.go
```diff
@@ -258,9 +258,12 @@ func NewTidbMonitor(name, namespace string, tc *v1alpha1.TidbCluster, grafanaEna
 				},
 				Envs: map[string]string{},
 			},
-			Persistent: persist,
```
DanielZhangQD, line 261: No need to set this anymore?

tests/e2e/tidbcluster/serial.go
```diff
@@ -507,6 +514,66 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+			framework.ExpectNoError(err, "Expected tidbmonitor deployment pvc success")
+			oldVolumeName := deploymentPvc.Spec.VolumeName
+			ginkgo.By("Upgrade tidb-operator and CRDs to current version")
+			ocfg.Tag = cfg.OperatorTag
+			ocfg.Image = cfg.OperatorImage
+			oa.InstallCRDOrDie(ocfg)
+			oa.UpgradeOperatorOrDie(ocfg)
+			err = tests.CheckTidbMonitor(tm, cli, c, fw)
+			framework.ExpectNoError(err, "Expected tidbmonitor checked success under migration")
+			err = wait.Poll(5*time.Second, 30*time.Minute, func() (done bool, err error) {
```
DanielZhangQD, line 546: We'd better decrease the timeout value when the case is stable.

tests/e2e/tidbcluster/serial.go
```diff
@@ -507,6 +514,66 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
+				tmSet, err := stsGetter.StatefulSets(ns).Get(monitor.GetMonitorObjectName(tm), metav1.GetOptions{})
+				if err != nil {
+					klog.Errorf("failed to get statefulset: %s/%s, %v", ns, tmSet, err)
+					return false, nil
+				}
+				return true, nil
+			})
+			framework.ExpectNoError(err, "Expected tidbmonitor sts success")
+			err = wait.Poll(5*time.Second, 30*time.Minute, func() (done bool, err error) {
```
DanielZhangQD, line 555: ditto.

tests/e2e/tidbcluster/serial.go
```diff
@@ -409,12 +415,13 @@ var _ = ginkgo.Describe("[tidb-operator][Serial]", func() {
...
 			ocfg = &tests.OperatorConfig{
-				Namespace:   ns,
-				ReleaseName: "operator",
-				Tag:         version,
-				Image:       fmt.Sprintf("pingcap/tidb-operator:%s", version),
+				Namespace:       ns,
+				ReleaseName:     "operator",
+				Tag:             version,
+				Image:           fmt.Sprintf("pingcap/tidb-operator:%s", version),
+				ImagePullPolicy: v1.PullAlways,
```
DanielZhangQD, line 424: Why need to add this since we're using officially released tags.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -427,3 +422,196 @@ func (m *MonitorManager) patchTidbClusterStatus(tcRef *v1alpha1.TidbClusterRef,
...
+
+	if deploymentPv.Spec.ClaimRef != nil && deploymentPv.Spec.ClaimRef.Name == firstStsPvcName {
+		// smooth migration successfully and clean status
+		monitor.Status.DeploymentStorageStatus = nil
+		return true, nil
+	}
+
+	// firstly patch status
+	if monitor.Status.DeploymentStorageStatus == nil {
+		err = m.patchTidbMonitorStatus(monitor, deploymentPvc.Spec.VolumeName)
```
DanielZhangQD, line 513: If the status patch succeeds here but the following operations fail, e.g. in L528, then the status will be updated in `UpdateTidbMonitor` in `pkg/controller/tidbmonitor/tidb_monitor_control.go`, in this case, the patched status will be overwritten.

tests/e2e/tidbcluster/serial.go
```diff
@@ -19,9 +19,12 @@ import (
 	_ "net/http/pprof"
 	"time"
 
+	"github.com/pingcap/tidb-operator/pkg/monitor/monitor"
+
```
DanielZhangQD, line 23: Please combine this with the following `tidb-operator` packages.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -510,10 +510,10 @@ func (m *MonitorManager) smoothMigrationToStatefulSet(monitor *v1alpha1.TidbMoni
 
 	// firstly patch status
 	if monitor.Status.DeploymentStorageStatus == nil {
-		err = m.patchTidbMonitorStatus(monitor, deploymentPvc.Spec.VolumeName)
-		if err != nil {
-			return false, err
+		monitor.Status.DeploymentStorageStatus = &v1alpha1.DeploymentStorageStatus{
+			PvName: deploymentPvc.Spec.VolumeName,
 		}
+		return false, nil
```
DanielZhangQD, line 516: What about return an error indicating requeue here?

DanielZhangQD (approved):
LGTM

AstroProfundis (approved):

PR #3443, opened 2020-10-29
Skip set gc life time

namco1992:
### What problem does this PR solve? 

Fix #3386.

### What is changed and how does it work?

1. Change CR structure in types.go as below:
"From TiDBAccessConfig `json:"from,omitempty"`" to "From *TiDBAccessConfig `json:"from,omitempty"`"
"To TiDBAccessConfig `json:"to,omitempty"`" to "To *TiDBAccessConfig `json:"to,omitempty"`"
2. Check `spec.from` and `spec.to` during the backup/restore, if not nil, go with the current logic to set GC; if nil, skip the GC setting logic

### Check List 

Tests 

 - Unit test
 - E2E test
 - Stability test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has Go code change

Side effects

N/A

Related changes

 - Need to update the documentation

### Does this PR introduce a user-facing change?:

```release-note
NONE
```

PR #3453, opened 2020-11-04
Automated cherry pick of #3443: skip setting GC life time if spec.from or spec.to is not

namco1992:
Cherry pick of #3443 on release-1.1.

#3443: skip setting GC life time if spec.from or spec.to is not


DanielZhangQD (approved):
LGTM

PR #3514, opened 2020-11-23
Filter out TiFlash stores when checking upStores before TiKV member scaling in

handlerww:
Filter out TiFlash stores when checking working TiKV Stores before TiKV member scaling in.
When controller get storesinfo from PD, it didn't tell apart the TiFlash stores or TiKV stores. It may cause error when TiFlash is deployed and TiKV is scaling in.
### Check List 

Tests 

image
image

Successfully set off the alert.

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Fix the issue that when TiFlash is deployed, TiKV can be scaled in even if there are only 3 UP stores in the cluster
```


pkg/manager/member/tikv_scaler.go
```diff
@@ -247,8 +248,11 @@ func (s *tikvScaler) preCheckUpStores(tc *v1alpha1.TidbCluster, podName string)
 	if err != nil {
 		return false, fmt.Errorf("failed to get stores info in TidbCluster %s/%s", tc.GetNamespace(), tc.GetName())
 	}
+	// filter out TiFlash
 	for _, store := range storesInfo.Stores {
-		if store.Store != nil && store.Store.StateName == v1alpha1.TiKVStateUp {
+		storePodName := strings.Split(store.Store.Address, ".")[0]
```
DanielZhangQD, line 253: Check `store.Store != nil` first.

pkg/manager/member/tikv_scaler.go
```diff
@@ -247,8 +248,11 @@ func (s *tikvScaler) preCheckUpStores(tc *v1alpha1.TidbCluster, podName string)
...
+		podKind := strings.Split(storePodName, "-")
+		if store.Store != nil && store.Store.StateName == v1alpha1.TiKVStateUp && podKind[len(podKind)-2] == "tikv" {
```
DanielZhangQD, line 255: Use `label.TiKVLabelVal` for `tikv`.

DanielZhangQD: Please also add UT to cover the code change.

pkg/manager/member/tikv_scaler_test.go
```diff
@@ -221,6 +222,9 @@ func TestTiKVScalerScaleIn(t *testing.T) {
 				store := &pdapi.StoreInfo{
 					Store: &pdapi.MetaStore{
 						StateName: v1alpha1.TiKVStateUp,
+						Store: &metapb.Store{
```
DanielZhangQD, line 225: Should also add tiflash stores to cover your code change?

handlerww: fixed

DanielZhangQD (approved):
LGTM

pkg/manager/member/tikv_scaler.go
```diff
@@ -247,9 +248,14 @@ func (s *tikvScaler) preCheckUpStores(tc *v1alpha1.TidbCluster, podName string)
 	if err != nil {
 		return false, fmt.Errorf("failed to get stores info in TidbCluster %s/%s", tc.GetNamespace(), tc.GetName())
 	}
+	// filter out TiFlash
 	for _, store := range storesInfo.Stores {
-		if store.Store != nil && store.Store.StateName == v1alpha1.TiKVStateUp {
-			upNumber++
+		if store.Store != nil {
+			storePodName := strings.Split(store.Store.Address, ".")[0]
+			podKind := strings.Split(storePodName, "-")
```
lonng, line 255: Do we need to check the `len(podKind)`?

handlerww: The pod name is generated in the format "xxx-xxx-tikv(component)-0(ordinals)", the last but one is fixed. Is there any corner case on it?

handlerww: Oh, I see. There might be some instances from outside K8S clusters.

handlerww: Fixed. I have updated UT to cover store named "outsidetikv".

pkg/util/util.go
```diff
@@ -355,6 +355,12 @@ func VolumeClaimTemplate(r corev1.ResourceRequirements, metaName string, storage
 	}
 }
 
+func MatchLabelFromStoreAddress(Address string, label string) bool {
+	storePodName := strings.Split(Address, ".")[0]
+	storeKind := strings.Split(storePodName, "-")
+	return len(storeKind) >= 2 && storeKind[len(storeKind)-2] == label
```
DanielZhangQD, line 361: If `len(storeKind) < 2`, we should also count them in.

pkg/util/util.go
```diff
@@ -355,6 +356,17 @@ func VolumeClaimTemplate(r corev1.ResourceRequirements, metaName string, storage
 	}
 }
 
+func MatchLabelFromStoreLabels(Labels []*metapb.StoreLabel, componentLabel string) bool {
```
DanielZhangQD, line 359:
```suggestion
func MatchLabelFromStoreLabels(labels []*metapb.StoreLabel, componentLabel string) bool {
```

DanielZhangQD (approved):
LGTM

PR #3518, opened 2020-11-24
Fix PD failover when multi-cluster enabled

handlerww:
When multi-cluster enabled, PD member name is in long format(FQDN), some member selections in pd_failover are using pod name.

Fix codes to be compatible with the long format.
### Check List 

Tests 

- Unit test

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/manager/member/pd_member_manager.go
```diff
@@ -267,9 +267,10 @@ func (m *pdMemberManager) shouldRecover(tc *v1alpha1.TidbCluster) bool {
 		if !podutil.IsPodReady(pod) {
 			return false
 		}
-		status, ok := tc.Status.PD.Members[pod.Name]
-		if !ok || !status.Health {
-			return false
+		for pdName, pdMember := range tc.Status.PD.Members {
```
DanielZhangQD, line 270: What if there is no member of this Pod?

handlerww: If there is no member of this pod, there is no need to set off the failover function. When a member is reported as unhealthy, the controller begins to do failover operations.

handlerww: It can be the launching stage.

pkg/manager/member/pd_member_manager.go
```diff
@@ -267,8 +267,16 @@ func (m *pdMemberManager) shouldRecover(tc *v1alpha1.TidbCluster) bool {
...
 		}
-		status, ok := tc.Status.PD.Members[pod.Name]
-		if !ok || !status.Health {
+		ok := false
+		for pdName, pdMember := range tc.Status.PD.Members {
+			if strings.Split(pdName, ".")[0] == pod.Name {
+				if !pdMember.Health {
+					return false
+				}
+				ok = true
```
DanielZhangQD, line 276: Should `break` here?

pkg/manager/member/pd_member_manager.go
```diff
@@ -267,8 +267,17 @@ func (m *pdMemberManager) shouldRecover(tc *v1alpha1.TidbCluster) bool {
 		if !podutil.IsPodReady(pod) {
 			return false
 		}
-		status, ok := tc.Status.PD.Members[pod.Name]
-		if !ok || !status.Health {
+		ok := false
+		for pdName, pdMember := range tc.Status.PD.Members {
```
DanielZhangQD, line 271: Please update UT to cover your code change.

DanielZhangQD (approved):
LGTM

lonng (approved):
LGTM

PR #3530, opened 2020-11-26
Add retry with PD PeerMembers on cross-region clusters when all PD members broken down

handlerww:
When cross-region feature enabled, if all PD is broken down, the reconcile will be blocked. 
We can use PeerMembers of PD to rescue the reconcile loop.
### Check List 

Tests 

Before this fix, the status of tidbcluster CR won't be updated. After the fix, tidbcluster have already updated the status and set off the failover logic.

The 2 clusters resources YAML are listed as below:
```yaml
kind: TidbCluster
metadata:
  name: cluster1
  namespace: pingcap
spec:
  version: v4.0.8
  timezone: UTC
  pvReclaimPolicy: Delete
  enableDynamicConfiguration: true
  configUpdateStrategy: RollingUpdate
  clusterDomain: "cluster1.com"
  discovery: {}
  pd:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/pd
    replicas: 1
    requests:
      storage: "1Gi"
    config: 
      schedule:
        max-store-down-time: 1m
  tikv:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/tikv
    replicas: 1
    requests:
      storage: "1Gi"
    config: {}
  tidb:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/tidb
    replicas: 1
    service:
      type: ClusterIP
    config: {}
```
```yaml
kind: TidbCluster
metadata:
  name: cluster2
  namespace: pingcap
spec:
  version: v4.0.8
  timezone: UTC
  pvReclaimPolicy: Delete
  enableDynamicConfiguration: true
  configUpdateStrategy: RollingUpdate
  clusterDomain: "cluster2.com"
  cluster:
    name: cluster1
    namespace: pingcap
    clusterDomain: "cluster1.com"
  discovery: {}
  pd:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/pd
    replicas: 2
    requests:
      storage: "1Gi"
    config: 
      schedule:
        max-store-down-time: 1m
  tikv:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/tikv
    replicas: 1
    requests:
      storage: "1Gi"
    config: {}
  tidb:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/tidb
    replicas: 0
    service:
      type: ClusterIP
    config: {}
```

When the clusters are operating normally, apply failure changes to PD on cluster1. 

```bash
kubectl patch pod cluster1-pd-0 -p '{"spec":{"containers":[{"name":"pd","image":"fake-registry.cn-beijing.aliyuncs.com/tidb/pd:v4.0.8"}]}}'
```

```output
# kubectl get pods
cluster1-pd-0                              0/1     ErrImagePull   0          4m7s
# kubectl describe tc |grep 'Message:'
Message:               PD(s) are not healthy
```

After a period of time(set in pdFailoverPeriod in helm chart values), You can see the PD get recovered. The events flow shows PD has gone completely failover routine, and the PD status shows the cluster goes right.

```output
Warning  PDMemberUnhealthy    27m (x50 over 32m)  tidb-controller-manager  cluster1-pd-0.cluster1-pd-peer.pingcap.svc.cluster1.com(16855841494087100559) is unhealthy
  Warning  Unhealthy            27m                 tidb-controller-manager  pd pod[cluster1-pd-0.cluster1-pd-peer.pingcap.svc.cluster1.com] is unhealthy, msg:pd member[16855841494087100559] is unhealthy
  Normal   SuccessfulDelete     27m                 tidb-controller-manager  delete PVC pd-cluster1-pd-0 in  cluster1 successful
  Warning  PDMemberDeleted      27m (x2 over 27m)   tidb-controller-manager  cluster1-pd-0(16855841494087100559) deleted from cluster
  Normal   SuccessfulDelete     27m (x2 over 27m)   tidb-controller-manager  delete Pod cluster1-pd-0 in  cluster1 successful
```

```output
Pd:
    Image:  registry.cn-beijing.aliyuncs.com/tidb/pd:v4.0.8
    Leader:
      Client URL:            http://cluster2-pd-0.cluster2-pd-peer.pingcap.svc.cluster2.com:2379
      Health:                true
      Id:                    7311293428786794222
      Last Transition Time:  2020-12-14T03:34:57Z
      Name:                  cluster2-pd-0.cluster2-pd-peer.pingcap.svc.cluster.local
    Members:
      Cluster 1 - Pd - 0 . Cluster 1 - Pd - Peer . Pingcap . Svc . Cluster . Local:
        Client URL:            http://cluster1-pd-0.cluster1-pd-peer.pingcap.svc.cluster1.com:2379
        Health:                true
        Id:                    13766930828778279053
        Last Transition Time:  2020-12-14T03:43:42Z
        Name:                  cluster1-pd-0.cluster1-pd-peer.pingcap.svc.cluster.local
    Peer Members:
      Cluster 2 - Pd - 0 . Cluster 2 - Pd - Peer . Pingcap . Svc . Cluster . Local:
        Client URL:            http://cluster2-pd-0.cluster2-pd-peer.pingcap.svc.cluster2.com:2379
        Health:                true
        Id:                    7311293428786794222
        Last Transition Time:  2020-12-14T03:34:57Z
        Name:                  cluster2-pd-0.cluster2-pd-peer.pingcap.svc.cluster.local
      Cluster 2 - Pd - 1 . Cluster 2 - Pd - Peer . Pingcap . Svc . Cluster . Local:
        Client URL:            http://cluster2-pd-1.cluster2-pd-peer.pingcap.svc.cluster2.com:2379
        Health:                true
        Id:                    13281533448990156986
        Last Transition Time:  2020-12-14T03:34:57Z
        Name:                  cluster2-pd-1.cluster2-pd-peer.pingcap.svc.cluster2.com
    Phase:                     Normal
    Stateful Set:
      Collision Count:      0
      Current Replicas:     1
      Current Revision:     cluster1-pd-57854957fb
      Observed Generation:  3
      Ready Replicas:       1
      Replicas:             1
      Update Revision:      cluster1-pd-57854957fb
      Updated Replicas:     1
    Synced:                 true
    Unjoined Members:
      Cluster 1 - Pd - 1:
        Created At:  2020-12-14T03:43:53Z
        Pod Name:    cluster1-pd-1
        Pvc UID:     59f29020-bb23-483b-a4f8-39dce36dcdf1
```

Deploy the clusters with 2 PDs on cluster-1 and 1 PD on cluster-2, the rest of the resources YAML as before.

Apply failure changes to PD on cluster2.
```bash
kubectl patch pod cluster2-pd-0 -p '{"spec":{"containers":[{"name":"pd","image":"fake-registry.cn-beijing.aliyuncs.com/tidb/pd:v4.0.8"}]}}'
```

After the failover period, both of the clusters recover.

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
When multi-cluster feature enabled, support tidb-operator keep working when all PD members in one Kubernetes cluster crashed down while the peer members in other clusters working and the PD cluster still have a quorum.
```


DanielZhangQD:
@handlerww Please also update the test result for PD failover in the first cluster.
And we also have to make sure that the TiDB service still can work in both clusters all the time during the test.

pkg/pdapi/pd_control.go
```diff
@@ -130,6 +132,28 @@ func (pdc *defaultPDControl) GetClusterRefPDClient(namespace Namespace, tcName s
...
+	pdc.mutex.Lock()
+	defer pdc.mutex.Unlock()
+
+	var tlsConfig *tls.Config
+	var err error
+
+	if tlsEnabled {
+		tlsConfig, err = GetTLSConfig(pdc.kubeCli, namespace, tcName, util.ClusterClientTLSSecretName(tcName))
+		if err != nil {
+			klog.Errorf("Unable to get tls config for tidb cluster %q, pd client may not work: %v", tcName, err)
```
DanielZhangQD, line 145: Add ns info of the TC?

pkg/controller/pd_control.go
```diff
@@ -32,6 +32,34 @@ func GetPDClient(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) p
 	return pdControl.GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.IsTLSClusterEnabled())
 }
 
+// Retry to GetPDClient for multi-cluster
```
DanielZhangQD, line 35:
```suggestion
// GetPDClient tries to return an available PDClient
```

pkg/controller/pd_control.go
```diff
@@ -32,6 +32,34 @@ func GetPDClient(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) p
...
+
+	for _, pdMember := range tc.Status.PD.PeerMembers {
+		pdClient = pdControl.GetPeerPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.IsTLSClusterEnabled(), pdMember.ClientURL, pdMember.Name)
+		_, err := pdClient.GetHealth()
+		if err == nil {
+			return pdClient
+		}
+	}
+
+	return pdClient
```
DanielZhangQD, line 60: If you plan to return the pdClient built in L41, then this has been changed in L53.

pkg/discovery/discovery.go
```diff
@@ -110,15 +110,23 @@ func (d *tidbDiscovery) Discover(advertisePeerUrl string) (string, error) {
...
 		namespace := tc.Spec.Cluster.Namespace
 		if len(namespace) == 0 {
 			namespace = tc.GetNamespace()
 		}
 		pdClients = append(pdClients, d.pdControl.GetClusterRefPDClient(pdapi.Namespace(namespace), tc.Spec.Cluster.Name, tc.Spec.Cluster.ClusterDomain, tc.IsTLSClusterEnabled()))
 	}
-	if tc.Spec.PD != nil {
-		pdClients = append(pdClients, d.pdControl.GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.IsTLSClusterEnabled()))
+
+	if len(tc.Status.PD.PeerMembers) > 0 {
```
DanielZhangQD, line 126: No need to check len here, right?

pkg/pdapi/pd_control.go
```diff
@@ -33,6 +33,8 @@ type PDControlInterface interface {
 	GetPDClient(namespace Namespace, tcName string, tlsEnabled bool) PDClient
 	// GetClusterRefPDClient provides PDClient of the tidb cluster.
 	GetClusterRefPDClient(namespace Namespace, tcName string, clusterDomain string, tlsEnabled bool) PDClient
+	// GetPeerPDClient provides PD etcd Client of the tidb cluster.
```
DanielZhangQD, line 36:
```suggestion
	// GetPeerPDClient provides PD Client of the tidb cluster from peerURL.
```

DanielZhangQD (approved):
LGTM

dragonly:
IMHO, the release note should start with "Fix the issue that", because it's a user facing message. We should explain to the user what issue we have fixed/feature we have added, but not how. 
@handlerww

pkg/controller/pd_control.go
```diff
@@ -18,8 +18,8 @@ import (
 	"github.com/pingcap/tidb-operator/pkg/pdapi"
 )
 
-// GetPDClient gets the pd client from the TidbCluster
-func GetPDClient(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) pdapi.PDClient {
+// GetPDClientBasic gets the pd client from the TidbCluster
+func GetPDClientBasic(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) pdapi.PDClient {
```
dragonly, line 22:
```suggestion
// getPDClientInCurrentK8s gets the pd client from the TidbCluster
func getPDClientInCurrentK8s(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) pdapi.PDClient {
```
should this be a private function?

pkg/controller/pd_control.go
```diff
@@ -32,6 +32,34 @@ func GetPDClient(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) p
 	return pdControl.GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.IsTLSClusterEnabled())
 }
 
+// GetPDClient tries to return an available PDClient
+// If the pdClient built from the PD service name is unavailable, try to
+// build another one with the ClientURL in the PeerMembers.
+// ClientURL example:
+// ClientURL: https://cluster2-pd-0.cluster2-pd-peer.pingcap.svc.cluster2.local
```
dragonly, line 39:
```suggestion
// ClientURL example: https://cluster2-pd-0.cluster2-pd-peer.pingcap.svc.cluster2.local
```

pkg/controller/pd_control.go
```diff
@@ -32,6 +32,34 @@ func GetPDClient(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) p
...
+func GetPDClient(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) pdapi.PDClient {
+	pdClient := GetPDClientBasic(pdControl, tc)
```
dragonly, line 41:
```suggestion
	pdClient := getPDClientInCurrentK8s(pdControl, tc)
```

pkg/pdapi/pd_control.go
```diff
@@ -33,6 +33,8 @@ type PDControlInterface interface {
 	GetPDClient(namespace Namespace, tcName string, tlsEnabled bool) PDClient
 	// GetClusterRefPDClient provides PDClient of the tidb cluster.
 	GetClusterRefPDClient(namespace Namespace, tcName string, clusterDomain string, tlsEnabled bool) PDClient
+	// GetPeerPDClient provides PD Client of the tidb cluster from peerURL.
+	GetPeerPDClient(namespace Namespace, tcName string, tlsEnabled bool, peerURL string, peerKey string) PDClient
```
dragonly, line 37: `peerKey` should be `peerName`, because it's actually `pdMember.Name`

pkg/controller/pd_control_test.go
```diff
@@ -0,0 +1,91 @@
...
+				}
+			},
+			expectFn: func(g *GomegaWithT, b bool) {
+				pdClientCluster1 := NewFakePDClient(pdControl, tc)
+
+				pdClientCluster1.AddReaction(pdapi.GetHealthActionType, func(action *pdapi.Action) (interface{}, error) {
+					return nil, fmt.Errorf("Fake cluster 1 PD crashed")
+				})
+				pdClientCLUSTER2 := NewFakePDClientWithAddress(pdControl, "pd-0")
+				pdClientCLUSTER2.AddReaction(pdapi.GetHealthActionType, func(action *pdapi.Action) (interface{}, error) {
```
dragonly, line 76:
```suggestion
				pdClientCluster2 := NewFakePDClientWithAddress(pdControl, "pd-0")
				pdClientCluster2.AddReaction(pdapi.GetHealthActionType, func(action *pdapi.Action) (interface{}, error) {
```

pkg/controller/pd_control_test.go
```diff
@@ -0,0 +1,91 @@
...
+					return &pdapi.HealthInfo{Healths: []pdapi.MemberHealth{
+						{Name: "pd-0", MemberID: uint64(1), ClientUrls: []string{"http://pd-0.pd.pingcap.cluster2.com:2379"}, Health: true},
+					}}, nil
+				})
+				pdClient := GetPDClient(pdControl, tc)
+				_, err := pdClient.GetHealth()
+				g.Expect(err).To(BeNil())
+			},
+		},
+	}
```
dragonly, line 86: should also test the case when cluster1/cluster2 pd members are all unhealthy.
btw, can we check the pd client url to match the expected value here?

pkg/controller/pd_control.go
```diff
@@ -32,6 +32,34 @@ func GetPDClient(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) p
...
+
+	if len(tc.Status.PD.PeerMembers) == 0 {
+		return pdClient
+	}
+
+	_, err := pdClient.GetHealth()
```
csuzhangxc, line 47: Should we need to check the returned `HealthInfo` here?

handlerww: It just checks the connectivity here. If the in-cluster PD endpoint works, there's no need to switch to the PD in other clusters or somewhere else.

csuzhangxc: If all of the in-cluster PDs are unhealthy, it still not switches to the PD in other clusters?

handlerww: The HTTP request will fail, the PD svc in this cluster will not respond to the requests. Then this function will return an error, we begin to find the available PD client in the PeerMembers. The connectivity is important here. If GetHealth works, return an "Unhealthy" result, the failover loop in tidb-operator, and the health check of the service in Kubernetes can handle these issues.

pkg/pdapi/pd_control.go
handlerww: It's compatible with the code above, we may consider abstract these code in the future. For now, it's better to use similar logic and variable rule.

pkg/controller/pd_control.go
handlerww: We can provide more examples in the future if needed. There's no need to merge these lines.

pkg/controller/pd_control.go
handlerww: Yes, I will change it.

pkg/controller/pd_control_test.go
handlerww: 1. In a practical scenario, the expected value is not important, or not exists. It just tries every PeerMembers until the tidb-operator finds one client works.
2. If all PD members are unhealthy, the tidb-operator will act in the usual way. I will add this case, thanks.

pkg/discovery/discovery.go
```diff
@@ -110,15 +110,21 @@ func (d *tidbDiscovery) Discover(advertisePeerUrl string) (string, error) {
...
 		if len(namespace) == 0 {
 			namespace = tc.GetNamespace()
 		}
 		pdClients = append(pdClients, d.pdControl.GetClusterRefPDClient(pdapi.Namespace(namespace), tc.Spec.Cluster.Name, tc.Spec.Cluster.ClusterDomain, tc.IsTLSClusterEnabled()))
 	}
-	if tc.Spec.PD != nil {
-		pdClients = append(pdClients, d.pdControl.GetPDClient(pdapi.Namespace(tc.GetNamespace()), tc.GetName(), tc.IsTLSClusterEnabled()))
+
+	for _, pdMember := range tc.Status.PD.PeerMembers {
+		pdClients = append(pdClients, d.pdControl.GetPeerPDClient(pdapi.Namespace(ns), tc.Name, tc.IsTLSClusterEnabled(), pdMember.ClientURL, pdMember.Name))
```
csuzhangxc, line 127: Is there any potential problem for the `if member.Name == podName` bellow if multiple members from different clusters have the same pod name (may also from `tc.Spec.Cluster`?).

pkg/pdapi/pd_control.go
dragonly: I don't see it's necessary to use `key` in the variable name just because it's a map key. BTW there's a `tcName` parameter here, so I suppose we should keep the variable name as close to the current meaning as possible, which is also convenient for future refactoring.

csuzhangxc, line 37: `peerURL` often refers to `xxx:2380` (not `xxx:2379`) in PD/etcd, but it's peer-member's client URL here, how about giving it a better name?

handlerww: It's a little tricky in multi-cluster. If the URL contains the pod name, we can ignore the port spec in the peer service.
It's the opposite concept to ClusterIP in the concept of the VPC-native Kubernetes cluster. We need the FQDN and the pod IP available in the VPC scope rather than the ClusterIP. We can refer to https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips for more details about the VPC-native cluster. In TiDB-Operator, the pod IP is provided by the peer service.

pkg/discovery/discovery.go
handlerww: Oh, it's a missing change for multi-cluster, I will submit another PR for this. Thanks a lot. 

For now, this check won't work. The PD member name in multi-cluster will use the FQDN(cluster2-pd-0.cluster2-pd-peer.pingcap.svc.cluster2.local), rather than pod name(cluster2-pd-0).

DanielZhangQD: @handlerww @csuzhangxc The change is done in https://github.com/pingcap/tidb-operator/pull/3365/, I had discussed that with @lichunzhu, it seems that it's not necessary for PD, please confirm with him and if no need for PD, we can remove the check in another PR.

csuzhangxc (approved):
LGTM after you confirmed https://github.com/pingcap/tidb-operator/pull/3530#discussion_r544741762

pkg/discovery/discovery.go
lichunzhu: I suggest keeping this for safety. PD will return an error if `cfg.Join == cfg.AdvertiseClientUrls`. https://github.com/tikv/pd/blob/bb5abcf0b9b38c845b6afc7d55d7317ef23363d4/server/join/join.go#L87
However, PD will use its own cached data when restarting.

PR #3532, opened 2020-11-27
Monitor: support monitoring lightning with TLS enabled

shuijing198799:
When the monitor tls enable, the scheme of lightning in prom rules should be https instead of http 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change


pkg/monitor/monitor/template.go
```diff
@@ -381,7 +381,7 @@ func addTlsConfig(pc *config.Config) {
 		// TODO support tiflash tls when it gets ready
 		if sconfig.JobName == "pd" || sconfig.JobName == "tidb" || sconfig.JobName == "tikv" ||
 			sconfig.JobName == "pump" || sconfig.JobName == "drainer" ||
-			sconfig.JobName == "tiflash" || sconfig.JobName == "tiflash-proxy" {
+			sconfig.JobName == "tiflash" || sconfig.JobName == "tiflash-proxy" || sconfig.JobName == "lightning" {
```
DanielZhangQD, line 384: The lightning chart does not support TLS yet.

shuijing198799: DBaaS will not use lightning chart, Just set scheme is all right.

LinuxGit (approved):
LGTM

pkg/monitor/monitor/template.go
```diff
@@ -390,6 +390,11 @@ func addTlsConfig(pc *config.Config) {
 			pc.ScrapeConfigs[id] = sconfig
 			sconfig.Scheme = "https"
 		}
+		// lightning does not need to monitor the access of other components,
```
DanielZhangQD, line 393:
```suggestion
		// lightning does not need to verify the access of other components,
```

DanielZhangQD (approved):
LGTM

PR #3572, opened 2020-12-07
add options for local and gcs provider in backup

### LINKED ISSUE #3561: import: add -v=4 in rclone command.
## Feature Request

When import lot of data. it first use `rclone` to copy data from remote storage. `rclone` would pre alloc some disk space, and if data didn't reach this size. it will looks like `rclone` is stuck.

append v=4 in rclone command. which will print status once in a minute. so the user will be able to see the real problem.

BinChenn:
Fix #3561 , user can configure logging options for rclone  in yaml file, example as below.
- Code chages
    Add options in LocalStorageProvider and GcsStorageProvider
    
- How does it work
    User can configure rclone options in yaml file when starting backup/restore jobs like below. We set `-v` flag as default for rclone command to output [logging info](https://rclone.org/docs/#logging). If user set logging flags such as `-q` , `-v=4` , user defined flags will take precedence.

### Check List 

Tests 

- Unit test

- E2E test

- Stability test

  - Backup&Restore using aws s3 AWS-KMS enabled

    1. Create a kind cluster and generate test data with go-tpc

        ```shell
        kind create cluster
        ./hack/local-up-operator.sh
        kubectl apply -f example/basic/tidb-cluster.yaml
        go-tpc -H 127.0.0.1 -P 4000 -D tpcc tpcc --warehouses 8 --parts 4 prepare
        ```

    2. [Backup data to aws s3 with dumpling](https://docs.pingcap.com/tidb-in-kubernetes/dev/backup-to-s3)

        ```
        kubectl apply -f https://github.com/pingcap/tidb-operator/blob/master/manifests/backup/backup-rbac.yaml -n test1
        kubectl create secret generic s3-secret --from-literal=access_key=xxx --from-literal=secret_key=yyy
        kubectl create secret generic backup-demo1-tidb-secret --from-literal=password=${password}
        ```

        backup-s3.yaml

        ```yaml
        ---
        kind: Backup
        metadata:
          name: demo1-backup-s3
        spec:
          from:
            host: basic-tidb.default
            port: 4000
            user: root
            secretName: backup-demo1-tidb-secret
          s3:
            provider: aws
            secretName: s3-secret
            region: ${region}
            bucket: ${bucket}
            prefix: ${prefix}
            options:
            - --ignore-checksum
          storageSize: 10Gi
        ```

        ```shell
        kubectl apply -f backup-s3.yaml
        ```

        If backup success, rclone logging will be as below.

        ```text
        I1209 07:26:38.010409       1 export.go:96] rclone copy data from /backup/chenbinbackup/tidbbackup/backup-2020-12-09T07:22:30Z.tgz data to s3:chenbinbackup/tidbbackup/backup-2020-12-09T07:22:30Z.tgz.tmp. rclone log: 
        Transferred:   	      110M / 319.592 MBytes, 34%, 1.845 MBytes/s, ETA 1m53s
        Transferred:            0 / 1, 0%
        Elapsed time:        59.6s
        Transferring:
        *               backup-2020-12-09T07:22:30Z.tgz: 34% /319.592M, 2.004M/s, 1m44s
        2020/12/09 07:25:35 INFO  :
        Transferred:   	      230M / 319.592 MBytes, 72%, 1.923 MBytes/s, ETA 46s
        Transferred:            0 / 1, 0%
        Elapsed time:      1m59.6s
        Transferring:
        *               backup-2020-12-09T07:22:30Z.tgz: 71% /319.592M, 1.743M/s, 51s
        2020/12/09 07:26:35 INFO  :
        Transferred:   	  319.592M / 319.592 MBytes, 100%, 1.779 MBytes/s, ETA 0s
        Transferred:            0 / 1, 0%
        Elapsed time:      2m59.6s
        Transferring:
        *               backup-2020-12-09T07:22:30Z.tgz:100% /319.592M, 1.121M/s, 0s
        2020/12/09 07:26:37 INFO  : backup-2020-12-09T07:22:30Z.tgz: Copied (new)
        2020/12/09 07:26:37 INFO  :
        Transferred:   	  319.592M / 319.592 MBytes, 100%, 1.755 MBytes/s, ETA 0s
        Transferred:            1 / 1, 100%
        Elapsed time:       3m2.1s
        ```

  - Backup&Restore using aws s3

      Config is like previous subsection, refer to [doc](https://docs.pingcap.com/tidb-in-kubernetes/dev/backup-to-gcs), rclone logging  outputs like previous subsection.

Code changes

- Has Go code change

Side effects
None

Related changes

```release-note
Append verbose flag '-v' to rclone command as default in lightning and dumpling for troubleshooting.
```

### DIFF
--- a/cmd/backup-manager/app/export/export.go
+++ b/cmd/backup-manager/app/export/export.go
@@ -90,24 +90,24 @@ func (bo *Options) dumpTidbClusterData(backup *v1alpha1.Backup) (string, error) 
 func (bo *Options) backupDataToRemote(source, bucketURI string, opts []string) error {
 	destBucket := backupUtil.NormalizeBucketURI(bucketURI)
 	tmpDestBucket := fmt.Sprintf("%s.tmp", destBucket)
-	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket, true)
+	args := backupUtil.ConstructRcloneArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket, true)
 	// TODO: We may need to use exec.CommandContext to control timeouts.
 	output, err := exec.Command("rclone", args...).CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone copyto command for upload backup data %s failed, output: %s, err: %v", bo, bucketURI, string(output), err)
 	}
 
-	klog.Infof("cluster %s rclone copy data from %s data to %s, log: %s", bo, source, tmpDestBucket, output)
+	klog.Infof("cluster %s, rclone copy data from %s to %s, log: %s", bo, source, tmpDestBucket, output)
 	klog.Infof("upload cluster %s backup data to %s successfully, now move it to permanent URL %s", bo, tmpDestBucket, destBucket)
 
 	// the backup was a success
 	// remove .tmp extension
-	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket, true)
+	args = backupUtil.ConstructRcloneArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket, true)
 	output, err = exec.Command("rclone", args...).CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone moveto command failed, output: %s, err: %v", bo, string(output), err)
 	}
-	klog.Infof("cluster %s rclone move data from %s data to %s, log: %s", bo, tmpDestBucket, destBucket, output)
+	klog.Infof("cluster %s, rclone move data from %s to %s, log: %s", bo, tmpDestBucket, destBucket, output)
 	return nil
 }
 
@@ -117,7 +117,7 @@ func getBackupSize(backupPath string, opts []string) (int64, error) {
 	if exist := backupUtil.IsFileExist(backupPath); !exist {
 		return size, fmt.Errorf("file %s does not exist or is not regular file", backupPath)
 	}
-	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, nil, "ls", backupPath, "", false)
+	args := backupUtil.ConstructRcloneArgs(constants.RcloneConfigArg, nil, "ls", backupPath, "", false)
 	out, err := exec.Command("rclone", args...).CombinedOutput()
 	if err != nil {
 		return size, fmt.Errorf("failed to get backup %s size, err: %v", backupPath, err)
--- a/cmd/backup-manager/app/export/export.go
+++ b/cmd/backup-manager/app/export/export.go
@@ -93,21 +93,21 @@ func (bo *Options) backupDataToRemote(source, bucketURI string, opts []string) e
 	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket, true)
 	// TODO: We may need to use exec.CommandContext to control timeouts.
 	output, err := exec.Command("rclone", args...).CombinedOutput()
-	klog.Infof("rclone copy data from %s data to %s. rclone log: %s", source, tmpDestBucket, output)
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone copyto command for upload backup data %s failed, output: %s, err: %v", bo, bucketURI, string(output), err)
 	}
 
+	klog.Infof("cluster %s rclone copy data from %s data to %s, log: %s", bo, source, tmpDestBucket, output)
 	klog.Infof("upload cluster %s backup data to %s successfully, now move it to permanent URL %s", bo, tmpDestBucket, destBucket)
 
 	// the backup was a success
 	// remove .tmp extension
 	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket, true)
 	output, err = exec.Command("rclone", args...).CombinedOutput()
-	klog.Infof("rclone move data from %s data to %s. rclone log: %s", tmpDestBucket, destBucket, output)
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone moveto command failed, output: %s, err: %v", bo, string(output), err)
 	}
+	klog.Infof("cluster %s rclone move data from %s data to %s, log: %s", bo, tmpDestBucket, destBucket, output)
 	return nil
 }
 
--- a/cmd/backup-manager/app/export/export.go
+++ b/cmd/backup-manager/app/export/export.go
@@ -90,9 +90,10 @@ func (bo *Options) dumpTidbClusterData(backup *v1alpha1.Backup) (string, error) 
 func (bo *Options) backupDataToRemote(source, bucketURI string, opts []string) error {
 	destBucket := backupUtil.NormalizeBucketURI(bucketURI)
 	tmpDestBucket := fmt.Sprintf("%s.tmp", destBucket)
-	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket)
+	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket, true)
 	// TODO: We may need to use exec.CommandContext to control timeouts.
 	output, err := exec.Command("rclone", args...).CombinedOutput()
+	klog.Infof("rclone copy data from %s data to %s. rclone log: %s", source, tmpDestBucket, output)
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone copyto command for upload backup data %s failed, output: %s, err: %v", bo, bucketURI, string(output), err)
 	}
@@ -101,8 +102,9 @@ func (bo *Options) backupDataToRemote(source, bucketURI string, opts []string) e
 
 	// the backup was a success
 	// remove .tmp extension
-	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket)
+	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket, true)
 	output, err = exec.Command("rclone", args...).CombinedOutput()
+	klog.Infof("rclone move data from %s data to %s. rclone log: %s", tmpDestBucket, destBucket, output)
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone moveto command failed, output: %s, err: %v", bo, string(output), err)
 	}
@@ -115,7 +117,7 @@ func getBackupSize(backupPath string, opts []string) (int64, error) {
 	if exist := backupUtil.IsFileExist(backupPath); !exist {
 		return size, fmt.Errorf("file %s does not exist or is not regular file", backupPath)
 	}
-	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "ls", backupPath, "")
+	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, nil, "ls", backupPath, "", false)
 	out, err := exec.Command("rclone", args...).CombinedOutput()
 	if err != nil {
 		return size, fmt.Errorf("failed to get backup %s size, err: %v", backupPath, err)
--- a/cmd/backup-manager/app/util/util.go
+++ b/cmd/backup-manager/app/util/util.go
@@ -391,16 +391,21 @@ func GetCommitTsFromBRMetaData(provider v1alpha1.StorageProvider) (uint64, error
 // ConstructArgs constructs the rclone args
 func ConstructArgs(conf string, opts []string, command, source, dest string, rcloneLog bool) []string {
 	var args []string
+	defaultLog := true
 	if conf != "" {
 		args = append(args, conf)
 	}
 	if len(opts) > 0 {
 		for _, opt := range opts {
-			if opt == "-q" || strings.HasPrefix(opt, "-v") {
-				rcloneLog = false
+			if rcloneLog == false && strings.HasPrefix(opt, "-v") {
+				defaultLog = false
+				continue
 			}
+			if opt == "-q" || opt == "--quiet" {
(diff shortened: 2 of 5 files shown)


DanielZhangQD:
@BinChenn Did you test with `-v=4` for both S3 and GCS?

cmd/backup-manager/app/util/util.go
```diff
@@ -298,6 +298,10 @@ func GetOptions(provider v1alpha1.StorageProvider) []string {
 	switch st {
 	case v1alpha1.BackupStorageTypeS3:
 		return provider.S3.Options
+	case v1alpha1.BackupStorageTypeGcs:
+		return provider.Gcs.Options
+	case v1alpha1.BackupStorageTypeLocal:
+		return provider.Local.Options
```
DanielZhangQD, line 304: Currently, the `options` is used for rclone, so no need to update for local storage as it does not use rclone.

BinChenn: yes, options of local provider has been removed.

BinChenn:
> @BinChenn Did you test with `-v=4` for both S3 and GCS?

Test done. "-v=4" is equivalent to “-vv”.

cmd/backup-manager/app/util/util.go
```diff
@@ -389,12 +389,17 @@ func GetCommitTsFromBRMetaData(provider v1alpha1.StorageProvider) (uint64, error
...
 // ConstructArgs constructs the rclone args
-func ConstructArgs(conf string, opts []string, command, source, dest string) []string {
+func ConstructArgs(conf string, opts []string, command, source, dest string, rcloneLog bool) []string {
 	var args []string
 	if conf != "" {
 		args = append(args, conf)
 	}
 	if len(opts) > 0 {
+		for _, opt := range opts {
+			if opt == "-q" || strings.HasPrefix(opt, "-v") {
```
DanielZhangQD, line 399: Also need to consider `--quiet`.

DanielZhangQD: No need to check `-v`, just to append them is OK, right?

handlerww (commented):
/lgtm

handlerww (approved):

cmd/backup-manager/app/export/export.go
```diff
@@ -115,7 +117,7 @@ func getBackupSize(backupPath string, opts []string) (int64, error) {
 	if exist := backupUtil.IsFileExist(backupPath); !exist {
 		return size, fmt.Errorf("file %s does not exist or is not regular file", backupPath)
 	}
-	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "ls", backupPath, "")
+	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, nil, "ls", backupPath, "", false)
```
DanielZhangQD, line 120: Are you sure that no need to add `opts` for `ls`?
Did you test the case where `AWS-KMS` is enabled for S3?

cmd/backup-manager/app/export/export.go
```diff
@@ -90,9 +90,10 @@ func (bo *Options) dumpTidbClusterData(backup *v1alpha1.Backup) (string, error)
 func (bo *Options) backupDataToRemote(source, bucketURI string, opts []string) error {
 	destBucket := backupUtil.NormalizeBucketURI(bucketURI)
 	tmpDestBucket := fmt.Sprintf("%s.tmp", destBucket)
-	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket)
+	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket, true)
 	// TODO: We may need to use exec.CommandContext to control timeouts.
 	output, err := exec.Command("rclone", args...).CombinedOutput()
+	klog.Infof("rclone copy data from %s data to %s. rclone log: %s", source, tmpDestBucket, output)
```
DanielZhangQD, line 96: Please move this to L100.

DanielZhangQD, line 96:
```suggestion
	klog.Infof("cluster %s rclone copy data from %s to %s, log: %s", bo, source, tmpDestBucket, output)
```

cmd/backup-manager/app/export/export.go
```diff
@@ -101,8 +102,9 @@ func (bo *Options) backupDataToRemote(source, bucketURI string, opts []string) e
 
 	// the backup was a success
 	// remove .tmp extension
-	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket)
+	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket, true)
 	output, err = exec.Command("rclone", args...).CombinedOutput()
+	klog.Infof("rclone move data from %s data to %s. rclone log: %s", tmpDestBucket, destBucket, output)
```
DanielZhangQD, line 107: Move this after L110 and add cluster info.

cmd/backup-manager/app/export/export.go
BinChenn: In here, `rclone ls` is just used to get the size of local backup archive, just like:
```
$ rclone ls ${archive}
1256 ${archive}
```
So `rclone ls` do not need options about aws, and do not need any options about logging, otherwise, the first word of output may be logging info rather than size number, strToInt will throw a error.

cmd/backup-manager/app/util/util.go
BinChenn: It's ok!

BinChenn: Not ok. If user define `-v`, we should filter `-v` for `rclone ls` cmd in fun `getBackupSize` in `cmd/backup-manager/app/export/export.go`, or cmd would like `rclone ls -v` and fun `getBackupSize` will fail. I wil fix this logic in next commit.

cmd/backup-manager/app/export/export.go
BinChenn: fixed

cmd/backup-manager/app/export/export.go
BinChenn: fixed

BinChenn: fixed

dragonly:
@BinChenn Please run `make lint` to check for linting problems

cmd/backup-manager/app/export/export.go
```diff
@@ -90,22 +90,24 @@ func (bo *Options) dumpTidbClusterData(backup *v1alpha1.Backup) (string, error)
...
 	tmpDestBucket := fmt.Sprintf("%s.tmp", destBucket)
-	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket)
+	args := backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "copyto", source, tmpDestBucket, true)
 	// TODO: We may need to use exec.CommandContext to control timeouts.
 	output, err := exec.Command("rclone", args...).CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone copyto command for upload backup data %s failed, output: %s, err: %v", bo, bucketURI, string(output), err)
 	}
 
+	klog.Infof("cluster %s rclone copy data from %s data to %s, log: %s", bo, source, tmpDestBucket, output)
```
DanielZhangQD, line 100:
```suggestion
	klog.Infof("cluster %s, rclone copy data from %s to %s, log: %s", bo, source, tmpDestBucket, output)
```

cmd/backup-manager/app/export/export.go
```diff
@@ -90,22 +90,24 @@ func (bo *Options) dumpTidbClusterData(backup *v1alpha1.Backup) (string, error)
...
 
 	// the backup was a success
 	// remove .tmp extension
-	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket)
+	args = backupUtil.ConstructArgs(constants.RcloneConfigArg, opts, "moveto", tmpDestBucket, destBucket, true)
 	output, err = exec.Command("rclone", args...).CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("cluster %s, execute rclone moveto command failed, output: %s, err: %v", bo, string(output), err)
 	}
+	klog.Infof("cluster %s rclone move data from %s data to %s, log: %s", bo, tmpDestBucket, destBucket, output)
```
DanielZhangQD, line 110:
```suggestion
	klog.Infof("cluster %s, rclone move data from %s to %s, log: %s", bo, tmpDestBucket, destBucket, output)
```

cmd/backup-manager/app/export/export.go
DanielZhangQD: Did you test the case where AWS-KMS is enabled for S3?

cmd/backup-manager/app/util/util.go
```diff
@@ -389,13 +389,23 @@ func GetCommitTsFromBRMetaData(provider v1alpha1.StorageProvider) (uint64, error
...
+func ConstructArgs(conf string, opts []string, command, source, dest string, rcloneLog bool) []string {
 	var args []string
+	defaultLog := true
 	if conf != "" {
 		args = append(args, conf)
 	}
 	if len(opts) > 0 {
-		args = append(args, opts...)
+		for _, opt := range opts {
+			if rcloneLog == false && strings.HasPrefix(opt, "-v") {
```
DanielZhangQD, line 400: Also have to consider `--verbose`.

DanielZhangQD: Could you please add UT for this function?

dragonly, line 400:
```suggestion
			if !rcloneLog && strings.HasPrefix(opt, "-v") {
```

dragonly: if seems here if `rcloneLog == false` and `opt == "-v"`, `defaultLog` will be `false`, but actually we should enable verbose log in this situation, which means `defaultLog` should be `true`?

and why use `strings.HasPrefix(opt, "-v")` but not `opt == "-v"`?

BinChenn: 1. `rcloneLog == false` is set only when command is `rclone ls` in fun `getBackupSize` in `cmd/backup-manager/app/export/export.go`. In this situation, we should filter any flag that could output logging, otherwise the output of `rclone ls` will not be size of data, and fun `getBackupSize` will fail.
2. `rclone` [logging](https://rclone.org/docs/#logging) could be setted such as `-v`, `-vv`, `-v 4`, `-v=4`, etc.

cmd/backup-manager/app/export/export.go
BinChenn: fixed

BinChenn: fixed

cmd/backup-manager/app/util/util.go
BinChenn: UT has added, fix `--verbose`.

DanielZhangQD (approved):
LGTM

cmd/backup-manager/app/export/export.go
BinChenn: Test done!

cmd/backup-manager/app/util/util.go
dragonly: > rcloneLog == false is setted only when command is rclone ls in fun getBackupSize in cmd/backup-manager/app/export/export.go. In this situation, we should filter any flag that could output logging, otherwise the output of rclone ls will not be size of data, and fun getBackupSize will fail.

If so, `rcloneLog` is not a proper parameter name, which actually represents such a complicated combination case. And note that `ConstructArgs()` has nothing todo with `rclone`, this is also weird, because I only see this function be called with "rclone" command.
If this function should be kept generic (not specific to "rclone"), then I think a more proper parameter would be `allowVerboseOutput`.
I also suggest a comment here, like `// if command don't allow verbose output, and user proveded args contains items starting with -v or --verbose, we should not add them to the final arguments`.

dragonly: I suggest rethinking about the design and purpose of `ConstructArgs()`, and answer the following questions:
1. is this specific to "rclone"?
2. if yes, we should rename this function.
3. if not, why is the verbose logic so specific to "rclone"? like `-v / -vv / -v=4` are all captured by `strings.HasPrefix()`?

dragonly (changes requested):
PTAL at the recent comments on `ConstructArgs()`

cmd/backup-manager/app/util/util.go
DanielZhangQD: @dragonly @BinChenn Currently, `ConstructArgs()` is used for rclone only, I think it's OK to rename this function and rename the parameter name from `rcloneLog` to `verboseLog`, WDYT?

BinChenn: @dragonly  @DanielZhangQD Yes, `ConstructArgs()` is used in rclone only, i will rename fun and logging parameters.

BinChenn:
> PTAL at the recent comments on `ConstructArgs()`

fixed

cmd/backup-manager/app/util/util.go
```diff
@@ -388,14 +388,27 @@ func GetCommitTsFromBRMetaData(provider v1alpha1.StorageProvider) (uint64, error
...
+func ConstructRcloneArgs(conf string, opts []string, command, source, dest string, verboseLog bool) []string {
 	var args []string
+	defaultLog := true
 	if conf != "" {
 		args = append(args, conf)
 	}
 	if len(opts) > 0 {
-		args = append(args, opts...)
+		for _, opt := range opts {
+			// If forbid logging with verboseLog==false, user proveded args contains items starting with -v or --verbose should be filtered.
```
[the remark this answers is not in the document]
BinChenn: fixed

BinChenn:
> @BinChenn Please update the release note, because this feature introduces new logs for user to troubleshoot problems.
> I also updated some comments, PTAL

Release notes added.

cmd/backup-manager/app/util/util.go
dragonly: Then LGTM

dragonly:
LGTM

dragonly (approved):

DanielZhangQD (approved):
LGTM

Issue #3585, opened 2020-12-08
TidbMonitor controller should watch for the event of StatefulSet


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Since we have migrated the deployment resource of TidbMonitor from deployment to StatefulSet in https://github.com/pingcap/tidb-operator/pull/3440, the TidbMonitor controller should watch for the event of StatefulSet here https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/tidbmonitor/tidb_monitor_controller.go#L51.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

DanielZhangQD:
cc @mikechengwei

PR #3587, opened 2020-12-08
Fix tidbmonitor watch from deployment to statefulset

mikechengwei:
#3585 

Change TiDBMonitor controller watch from deployment to statefulset.


DanielZhangQD (approved):
LGTM

lonng (approved):

PR #3631, opened 2020-12-21
Fix monitor for TiKV-Importer

### LINKED ISSUE #3618: TiKV-Importer supports to be pulled metrics directly without Pushgateway
## Feature Request

TiKV-Importer has `status-server-address` config item for metrics pull, then the TiKV-Importer Helm chart can remove Pushgateway to make it simpler.

TiKV-Importer supports to be pulled metrics directly without Pushgateway

### LINKED ISSUE #3628: Prometheus can't scrape TiDB-Lightning's metrics with the default values.yaml
## Bug Report

v1.1.7

**What did you do?**

query TiDB-Lighting's metrics in Prometheus with TiDB-Lighting installed with the default `values.yaml`.

metrics can be seen in Prometheus.

**What did you see instead?**

scrape metrics timeout.

this is caused by:

- default `values.yaml` set `service.type: NodePort` in https://github.com/pingcap/tidb-operator/blob/ba2f64ce6e26a151f7583fc955429390fd2ad51d/charts/tidb-lightning/values.yaml#L13
- but Prometheus try to scrape metrics via `{service-name}.{namespace}:8289/metrics`

csuzhangxc:

- fix #3618

- remove Prometheus Pushgateway for TiKV-Importer and enable status server
- fix labels for TiKV-Importer so that Prometheus can re-label `__address__` correctly

some useful links:

- [`kubernetes_sd_config`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config)

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
  1. `helm install` TiKV-Importer and TiDB-Lightning
  2. import some data via TiDB-Lighting
  3. query metrics in Prometheus for TiDB-Lighting and TiKV-Importer
    
image

lightning-metrics

importer-metrics
image

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fix the issue that Prometheus can't pull metrics for TiKV-Importer
```

### DIFF
--- a/charts/tidb-lightning/values.yaml
+++ b/charts/tidb-lightning/values.yaml
@@ -9,8 +9,8 @@ image: pingcap/tidb-lightning:v4.0.9
 imagePullPolicy: IfNotPresent
 # imagePullSecrets: []
 
-service: {}
-  # type: NodePort
+service:
+   type: NodePort
 
 # failFast causes the lightning pod fails when any error happens.
 # when disabled, the lightning pod will keep running when error happens to allow manual intervention, users have to check logs to see the job status.
--- a/charts/tidb-lightning/values.yaml
+++ b/charts/tidb-lightning/values.yaml
@@ -10,7 +10,7 @@ imagePullPolicy: IfNotPresent
 # imagePullSecrets: []
 
 service:
-   type: NodePort
+  type: NodePort
 
 # failFast causes the lightning pod fails when any error happens.
 # when disabled, the lightning pod will keep running when error happens to allow manual intervention, users have to check logs to see the job status.
--- a/charts/tidb-lightning/values.yaml
+++ b/charts/tidb-lightning/values.yaml
@@ -5,7 +5,7 @@
 # timezone is the default system timzone
 timezone: UTC
 
-image: pingcap/tidb-lightning:v4.0.8
+image: pingcap/tidb-lightning:v4.0.9
 imagePullPolicy: IfNotPresent
 # imagePullSecrets: []
 
--- a/charts/tidb-lightning/values.yaml
+++ b/charts/tidb-lightning/values.yaml
@@ -9,8 +9,8 @@ image: pingcap/tidb-lightning:v4.0.8
 imagePullPolicy: IfNotPresent
 # imagePullSecrets: []
 
-service:
-  type: NodePort
+service: {}
+  # type: NodePort
 
 # failFast causes the lightning pod fails when any error happens.
 # when disabled, the lightning pod will keep running when error happens to allow manual intervention, users have to check logs to see the job status.
--- a/charts/tikv-importer/templates/tikv-importer-configmap.yaml
+++ b/charts/tikv-importer/templates/tikv-importer-configmap.yaml
@@ -3,7 +3,7 @@ kind: ConfigMap
 metadata:
   name: {{ .Values.clusterName }}-importer-{{ template "importer-configmap.data-digest" . }}
   labels:
-    app.kubernetes.io/name: {{ .Values.clusterName }}-importer
+    app.kubernetes.io/name: {{ template "chart.name" . }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Values.clusterName }}
     app.kubernetes.io/component: importer
--- a/charts/tikv-importer/templates/tikv-importer-configmap.yaml
+++ b/charts/tikv-importer/templates/tikv-importer-configmap.yaml
@@ -3,9 +3,9 @@ kind: ConfigMap
 metadata:
   name: {{ .Values.clusterName }}-importer-{{ template "importer-configmap.data-digest" . }}
   labels:
-    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/name: {{ .Values.clusterName }}-importer
     app.kubernetes.io/managed-by: {{ .Release.Service }}
-    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/instance: {{ .Values.clusterName }}
     app.kubernetes.io/component: importer
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
 data:
--- a/charts/tikv-importer/templates/tikv-importer-service.yaml
+++ b/charts/tikv-importer/templates/tikv-importer-service.yaml
@@ -9,6 +9,7 @@ metadata:
     app.kubernetes.io/component: importer
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
 spec:
+  clusterIP: None
   ports:
   - name: importer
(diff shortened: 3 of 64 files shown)


charts/tikv-importer/templates/tikv-importer-service.yaml
```diff
@@ -3,17 +3,18 @@ kind: Service
 metadata:
   name: {{ .Values.clusterName }}-importer
```
DanielZhangQD, line 4: I think we only need to change the service name from `{{ .Values.clusterName }}-importer` to `{{ .Release.Name }}-importer`, no need to change the other labels or selectors, please help confirm.

csuzhangxc: we use `__meta_kubernetes_pod_label_app_kubernetes_io_name` (and some other labels and annotations) in the Prometheus config. so we need to change the labels.

charts/tidb-lightning/values.yaml
```diff
@@ -9,8 +9,8 @@ image: pingcap/tidb-lightning:v4.0.9
 imagePullPolicy: IfNotPresent
 # imagePullSecrets: []
 
-service:
-  type: NodePort
+service: {}
```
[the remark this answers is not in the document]
csuzhangxc: we use `__meta_kubernetes_pod_annotation_%s_prometheus_io_port` in the Prometheus config. If we set `type: NodePort`, then this automatically created node port can not be re-label with any Prometheus _meta labels_.

charts/tikv-importer/templates/tikv-importer-service.yaml
mikechengwei: I don’t think we need to change the existing label of helm, we can add a cluster label. `app.kubernetes.io/name` and `app.kubernetes.io/instance` represents the container iteself.

dragonly:
LGTM

dragonly (approved):

charts/tidb-lightning/values.yaml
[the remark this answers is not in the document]
DanielZhangQD: Prometheus uses the service name to access the lightning service, the service name + service port can be accessed no matter the service type is nodePort or service.

charts/tikv-importer/templates/tikv-importer-service.yaml
DanielZhangQD: I think the issue here is that the value of instance label is `app.kubernetes.io/instance: {{ .Release.Name }}`, which is used to construct the service name in TidbMonitor controller, however, the service name deployed by the chart is `{{ .Values.clusterName }}-importer`, and they do not match.

charts/tidb-lightning/values.yaml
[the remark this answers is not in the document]
csuzhangxc: sorry, my mistake, have reverted it back.

charts/tikv-importer/templates/tikv-importer-service.yaml
mikechengwei: If we not spec clustername fied , it will use release name , right? How to deal with it?

csuzhangxc: > If we don't spec clustername field , it will use release name , right?

NO, and I have removed `if not specified, the chart release name will be used` in this PR.

if don't specify this filed, an error will be reported, like:

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in ConfigMap.metadata.labels.app.kubernetes.io/instance
```

pkg/monitor/monitor/template.go
```diff
@@ -240,11 +240,11 @@ func buildAddressRelabelConfigByComponent(kind string) *config.RelabelConfig {
 		return &config.RelabelConfig{
 			Action:      config.RelabelReplace,
 			Regex:       addressPattern,
-			Replacement: "$1.$2-importer.$3:$4",
+			Replacement: "$2.$3:$4",
```
DanielZhangQD, line 243: Please revert the change to this file, after changing the instance label, this should work.

charts/tikv-importer/templates/tikv-importer-configmap.yaml
```diff
@@ -3,9 +3,9 @@ kind: ConfigMap
 metadata:
   name: {{ .Values.clusterName }}-importer-{{ template "importer-configmap.data-digest" . }}
   labels:
-    app.kubernetes.io/name: {{ template "chart.name" . }}
+    app.kubernetes.io/name: {{ .Values.clusterName }}-importer
```
DanielZhangQD, line 6: I think we do not need to change this, just change the instance label is OK.

pkg/monitor/monitor/template.go
csuzhangxc: reverted

charts/tikv-importer/templates/tikv-importer-configmap.yaml
csuzhangxc: reverted

DanielZhangQD (approved):
LGTM

DanielZhangQD:
@csuzhangxc If users deploy tikv-importer of v1.1.8, can they upgrade to the chart in this PR?

csuzhangxc:
> @csuzhangxc If users deploy tikv-importer of v1.1.8, can they upgrade to the chart in this PR?

```
Error: UPGRADE FAILED: cannot patch "basic-importer" with kind Service: Service "basic-importer" is invalid: spec.clusterIP: Invalid value: "": field is immutable && cannot patch "basic-importer" with kind StatefulSet: StatefulSet.apps "basic-importer" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```

charts/tikv-importer/templates/tikv-importer-service.yaml
```diff
@@ -5,15 +5,16 @@ metadata:
   labels:
     app.kubernetes.io/name: {{ template "chart.name" . }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
-    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/instance: {{ .Values.clusterName }}
     app.kubernetes.io/component: importer
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
 spec:
-  clusterIP: None
```
DanielZhangQD, line 12: We should keep this, why remove it here?
This service is a headless service through which Prometheus can access the Pod via Pod name.

DanielZhangQD:
LGTM cancel

charts/tikv-importer/templates/tikv-importer-service.yaml
csuzhangxc: reverted.

DanielZhangQD (approved):
LGTM

Issue #3620, opened 2020-12-17
TiDB-Lighting supports to persist checkpoint during restart


csuzhangxc:
## Feature Request

**Is your feature request related to a problem? Please describe:**

now, TiDB-Lighting in the Helm chart save checkpoint in `/tmp/xxx`, if we restart the job, the checkpoint will be lost.

**Describe the feature you'd like:**

persist checkpoint under `dataSource.local.hostPath`/`dataSource.adhoc.pvcName`/PVC with `dataSource.remote.storageClassName`

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #3644, opened 2020-12-24
charts: add local backend support for TiDB-Lightning

### LINKED ISSUE #3162: tidb-lightning supports local backend on k8s
tidb-lightning supports local backend on k8s.
The following two ways of configuration are supported:
1. Support configuration of s3 data source;
2. Support configuration of local directory as data source, that is, import from pod local disk.

---------------------------------------------------------------------------------

支持以下两种方式配置：
1. 支持配置s3的数据源，从s3上导入;
2.支持配置本地目录作为数据源，从pod本地盘导入。

csuzhangxc:

add `local` backend support for TiDB-Lightning Helm chart, may close #3162.

- create a PVC for `local` backend
- update some config items

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
  1. set `backend: local`
  2. `helm install` the chart
  3. verify the data has been imported into the TiDB cluster

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Add local backend support for the TiDB-Lightning chart
```


charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
```diff
@@ -29,7 +29,11 @@ fi
 {{- if eq .Values.backend "importer" }}
     --importer={{ .Values.targetTidbCluster.name }}-importer.{{ .Values.targetTidbCluster.namespace | default .Release.Namespace }}:8287 \
 {{- else if eq .Values.backend "tidb" }}
-    --backend tidb \
+    --backend=tidb \
+    --tidb-port=4000 \
```
csuzhangxc, line 33: for `tidb` backend, we should set `tidb-port` as well because TiDB-Lightning wouldn't auto setting this based on `/settings`.

if we do not need `tidb` backend in the chart anymore, maybe we can just remove it?

charts/tidb-lightning/templates/persistentvolumeclaim.yaml
```diff
@@ -18,3 +18,23 @@ spec:
...
+metadata:
+  name: {{ include "tidb-lightning.name" . }}-sorted-kv
+  labels:
+    app.kubernetes.io/name: {{ include "tidb-lightning.name" . }}-sorted-kv
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Values.targetTidbCluster.name }}
+    app.kubernetes.io/component: tidb-lightning
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+spec:
+  storageClassName: {{ .Values.sortedKV.storageClassName }}
```
csuzhangxc, line 34: If we specify `.Values.dataSource.local`, then we will set `.spec.nodeName`. but this can't work if the storage-class has `waitForFirstConsumer`. ref https://github.com/openebs/openebs/issues/2915 and https://github.com/kubernetes/kubernetes/issues/89953.

Should we need to:
- just document this behavior
- or update the chart to use `nodeSelector` or `nodeAffinity` rather than `nodeName`

WDYT @DanielZhangQD

DanielZhangQD: @csuzhangxc If `.Values.dataSource.local` is used, we can also use that hostPath to save the sst files instead of using PV.

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
DanielZhangQD: Sorry that I did not notice that we support tidb backend in the chart, is `tidb-port` the only parameter missed? Can we just let the user to set the `config` filed in the values.yaml and we have to make it clear in the doc.

charts/tidb-lightning/values.yaml
```diff
@@ -90,7 +90,14 @@ annotations: {}
...
 affinity: {}
 
-backend: importer # importer | tidb
+# The delivery backend used to import data.
+# If set to `local`, then the following `sortedKV` should be set.
+backend: importer # importer | local | tidb
+
+# For `local` backend, an extra PV is needed for local KV sorting.
+sortedKV: {}
+#  storageClassName: local-path
```
DanielZhangQD, line 99:
```suggestion
#  storageClassName: local-storage
```

DanielZhangQD: To keep consistent with the existing one.

csuzhangxc: fixed in https://github.com/pingcap/tidb-operator/pull/3644/commits/b6718ed24a28dd24f770cbed84dbda7b865dd2f2.

charts/tidb-lightning/templates/scripts/_start_lightning.sh.tpl
csuzhangxc: YES, currently only `tidb-port` missed and I have tested manually (https://github.com/pingcap/tidb-lightning/blob/463e7289fcbc412a49e50b307e3b33d30da7f218/lightning/config/config.go#L591-L614). I think we can add it in this script as we did for `2379` and `8287`. BTW, we also hardcode `tidb-host` in this script.

charts/tidb-lightning/templates/persistentvolumeclaim.yaml
csuzhangxc: but for Lightning, It may be I/O bound. If we use the source hostPath to save sst files. It may become slower.

BTW, I'm also thinking about where can we store checkpoint (#3620) for the same reason. for Lightning, we often suggest the user to store the checkpoint neither in the source disk nor the target TiDB cluster.

DanielZhangQD: `or update the chart to use nodeSelector or nodeAffinity rather than nodeName` this is good.

charts/tidb-lightning/values.yaml
```diff
@@ -90,7 +90,14 @@ annotations: {}
 tolerations: []
 affinity: {}
 
-backend: importer # importer | tidb
+# The delivery backend used to import data.
+# If set to `local`, then the following `sortedKV` should be set.
+backend: importer # importer | local | tidb
```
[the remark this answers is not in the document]
csuzhangxc: fixed in https://github.com/pingcap/tidb-operator/pull/3644/commits/2f33eb47006dd5be820eabf1ad4260322252ad21

charts/tidb-lightning/templates/persistentvolumeclaim.yaml
csuzhangxc: updated to `nodeSelector` in https://github.com/pingcap/tidb-operator/pull/3644/commits/2f33eb47006dd5be820eabf1ad4260322252ad21.

DanielZhangQD (approved):
LGTM

handlerww (approved):
LGTM

dragonly:
LGTM

PR #3684, opened 2021-01-06
change alertManagerRulesVersion from TiDB version to initializer image version

### LINKED ISSUE #3676: Get the value of TIDB_VERSION from the version of initializer instead of tidb in TidbMonitor
## Feature Request

Currently, in the TidbMonitor controller, an environment variable named TIDB_VERSION is defined for the initializer container which is used to create a directory to save the alert rules that customized by the users, however, its value is retrieved from the TiDB version in the TidbCluster that this TidbMonitor refers to.
With this implementation, the TidbMonitor deployment/sts will be rolling updated each time the TiDB version is updated which does not make sense.
In addition, the original alert rules are provided by the initializer container, so the value of TIDB_VERSION should be the version of the initializer image.
To keep backward compatibility, we need to keep the name `TIDB_VERSION` although it does not make sense anymore after this change.

Test Cases:
We have to run the following cases with this change:
1. Monitor Pod can be created successfully
2. Alert rules can be customized and saved in the PV in the expected dir
3. Upgrade the TiDB version will not cause the rolling update of Monitor Pod
4. 

BinChenn:
To avoid that the TidbMonitor deployment/sts will be rolling updated when the TiDB version is updated, **change** the value of TIDB_VERSION and DM_VERSION from the version of initializer instead of TiDB in TidbMonitor, so that the `alertManagerRulesVersion` will keep constant when the TiDB version is updated.

Close #3676 

Get the value of TIDB_VERSION and DM_VERSION from the version of initializer instead of TiDB in TidbMonitor.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 

- [ ] E2E test 

- [x] Manual test

    1. Create cluster and deploy TiDB Cluster

        ```shell
        kind create cluster 
        ./hack/local-up-operator.sh 
        kc apply -f examples/basic/tidb-cluster.yaml
        kc apply -f examples/basic/tidb-monitor.yaml
        ```

        Note: I set the version to v4.0.8, like:

        tidb-cluster.yaml

        ```yaml
        kind: TidbCluster
        metadata:
          name: basic
        spec:
          version: v4.0.8
          ...
        ```

        tidb-monitor.yaml

        ```yaml
        kind: TidbMonitor
        metadata:
          name: basic
        spec:
          clusters:
          - name: basic
          prometheus:
            baseImage: prom/prometheus
            version: v2.18.1
          grafana:
            baseImage: grafana/grafana
            version: 6.1.6
          initializer:
            baseImage: pingcap/tidb-monitor-initializer
            version: v4.0.8
          reloader:
            baseImage: pingcap/tidb-monitor-reloader
            version: v1.0.1
          imagePullPolicy: IfNotPresent
        ```

    2. Alert rules reloading

        ```shell
        kc port-forward svc/basic-monitor-reloader 9089  > /dev/null &
        ```

        Access `http://127.0.0.1:9089/#`, and update the alert rules of tidb, then we can see that rules file will be stored into PV.

        ```shell
        kc exec -it basic-monitor-0 -c reloader -- /bin/sh
        ls /data/latest-rules/tidb\:v4.0.8/
        output:===========
        tidb.rules.yml 
        ```

    3. Upgrade Tidb Cluster

        Following [upgrade-a-tidb-cluster](https://docs.pingcap.com/zh/tidb-in-kubernetes/stable/upgrade-a-tidb-cluster) , upgrade tidb cluster to `v4.0.9`, the TidbMonitor deployment/sts will  notbe rolling updated.

    4. Alert rules reloading

        After upgrade tidb cluster to `v4.0.9`， update the alertrules of tikv like step 2. We can see that the directory is same as the directory before upgrading.

        ```
        kc exec -it basic-monitor-0 -c reloader -- /bin/sh
        ls /data/latest-rules/tidb\:v4.0.8/
        output:===========
        tidb.rules.yml  tikv.rules.yml
        ```

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


Yisaer (approved):
LGTM

DanielZhangQD (approved):
LGTM

PR #3694, opened 2021-01-11
Sync TiKV Statefulset even when TiKV is not bootstrapped

### LINKED ISSUE #3612: Sync TiKV Statefulset no matter whether the TiKV cluster is bootstrapped
## Bug Report

**What did you do?**

1. Create a tidb cluster with wrong value of `TC.spec.tikv.affinity.nodeAffinity` field, which caused the TiKV cluster not able to be scheduled. And all TiKV Pods are in status of `Pending`.
2. Correcte the value, but all TiKV Pods are still `Pending`, and the Statefulsets of TiKV are not updated.  

Tidb-operator sync loop should not be blocked no matter what status of TiKV cluster is.

**What did you see instead?**
When TiKV cluster is not bootstrapped successfully(**no Pod(store) created**), Tidb operator return error  [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_member_manager.go#L647-L651), which block the sync loop.

**Root Cause**
PD client return error when tidb operator tried to get stores information but no store created(error infomation: `TiKV cluster not bootstrapped, please start TiKV first`), which blocks the sync loop of TiKV in [syncTidbClusterStatus](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_member_manager.go#L647-L651).

cvvz:

fix #3612 
### Code changes

- [x] Has Go code change

### Tests

- [x] Manual test 
* Upgrade
1. Create `TidbCluster` with wrong image of tikv
2. TiKV Pod will be in `ImagePullBackOff` status
3. Correct the image of TiKV, TiKV Pod will be upgraded to `Running` one by one
* Scale in/out
1. Create `TidbCluster` with wrong image of tikv
2. TiKV Pod will be in `ImagePullBackOff` status
3. Scale in/out TiKV will be successfully implemented

### Release Notes

```release-note
Fix the issue that when TiKV cluster is not bootstrapped, the following updates of TiKV component are not able to be synced.
```


dragonly:
@cvvz please add a release note in the description like "Fix the issue that ..."

pkg/pdapi/pdapi.go
```diff
@@ -698,3 +706,23 @@ func getLeaderEvictSchedulerInfo(storeID uint64) *schedulerInfo {
 func getLeaderEvictSchedulerStr(storeID uint64) string {
 	return fmt.Sprintf("%s-%d", "evict-leader-scheduler", storeID)
 }
+
+// TiKVNotBootstrappedError represents that TiKV cluster is not bootstrapped yet
+type TiKVNotBootstrappedError struct {
```
dragonly, line 711: Should this and related types go into `pkg/manager/member/utils.go`?

pkg/pdapi/pdapi.go
```diff
@@ -315,7 +316,14 @@ func (c *pdClient) getStores(apiURL string) (*StoresInfo, error) {
 }
 
 func (c *pdClient) GetStores() (*StoresInfo, error) {
-	return c.getStores(fmt.Sprintf("%s/%s", c.url, storesPrefix))
+	storesInfo, err := c.getStores(fmt.Sprintf("%s/%s", c.url, storesPrefix))
+	if err != nil {
+		if strings.HasSuffix(err.Error(), tiKVNotBootstrapped+"\n") {
```
dragonly, line 321: Maybe use `pdapi.IsTiKVNotBootstrappedError`?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -646,6 +646,11 @@ func (m *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, set
 	// This only returns Up/Down/Offline stores
 	storesInfo, err := pdCli.GetStores()
 	if err != nil {
+		if pdapi.IsTiKVNotBootstrappedError(err) {
+			klog.V(4).Infof("TiKV of Cluster %s/%s is not bootstrapped yet, get no store", tc.Namespace, tc.Name)
```
dragonly, line 650:
```suggestion
			klog.V(4).Infof("TiKV of Cluster %s/%s is not bootstrapped yet, got no store", tc.Namespace, tc.Name)
```

pkg/manager/member/tikv_member_manager.go
```diff
@@ -646,6 +646,11 @@ func (m *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, set
...
+			tc.Status.TiKV.Synced = true
+			return nil
```
dragonly, line 652: Should we return an error and stop syncing tikv here?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -742,6 +747,10 @@ func (m *tikvMemberManager) setStoreLabelsForTiKV(tc *v1alpha1.TidbCluster) (int
 	pdCli := controller.GetPDClient(m.deps.PDControl, tc)
 	storesInfo, err := pdCli.GetStores()
 	if err != nil {
+		if pdapi.IsTiKVNotBootstrappedError(err) {
+			klog.V(4).Infof("TiKV of Cluster %s/%s is not bootstrapped yet, get no store", tc.Namespace, tc.Name)
```
dragonly, line 751:
```suggestion
			klog.V(4).Infof("TiKV of Cluster %s/%s is not bootstrapped yet, got no store", tc.Namespace, tc.Name)
```

pkg/manager/member/tikv_member_manager.go
```diff
@@ -742,6 +747,10 @@ func (m *tikvMemberManager) setStoreLabelsForTiKV(tc *v1alpha1.TidbCluster) (int
...
+			return setCount, nil
```
dragonly, line 752: ditto

pkg/manager/member/tikv_member_manager.go
cvvz: If we return an error, the sync loop will be blocked here.

pkg/manager/member/tikv_member_manager.go
cvvz: ditto

pkg/pdapi/pdapi.go
cvvz: I think this is the error that belongs to `pd api`, maybe suitable to place it here.

pkg/pdapi/pdapi.go
cvvz: `pdapi.IsTiKVNotBootstrappedError` is used to judge whether the type of error is `TiKVNotBootstrappedError`. Here is to return `TiKVNotBootstrappedError`.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -742,6 +747,10 @@ func (m *tikvMemberManager) setStoreLabelsForTiKV(tc *v1alpha1.TidbCluster) (int
 	pdCli := controller.GetPDClient(m.deps.PDControl, tc)
 	storesInfo, err := pdCli.GetStores()
 	if err != nil {
+		if pdapi.IsTiKVNotBootstrappedError(err) {
+			klog.V(4).Infof("TiKV of Cluster %s/%s is not bootstrapped yet, got no store", tc.Namespace, tc.Name)
```
DanielZhangQD, line 751:
```suggestion
			klog.Infof("TiKV of Cluster %s/%s is not bootstrapped yet, got no store", tc.Namespace, tc.Name)
```

pkg/pdapi/pdapi.go
```diff
@@ -315,7 +316,14 @@ func (c *pdClient) getStores(apiURL string) (*StoresInfo, error) {
...
+			err = TiKVNotBootstrappedErrorf("TiKV cluster not bootstrapped, please start TiKV first")
```
DanielZhangQD, line 322:
```suggestion
			err = TiKVNotBootstrappedErrorf(tiKVNotBootstrapped)
```

pkg/manager/member/tikv_upgrader.go
```diff
@@ -99,7 +99,8 @@ func (u *tikvUpgrader) Upgrade(meta metav1.Object, oldSet *apps.StatefulSet, new
 		i := podOrdinals[_i]
 		store := u.getStoreByOrdinal(meta.GetName(), *status, i)
 		if store == nil {
-			continue
+			setUpgradePartition(newSet, i)
```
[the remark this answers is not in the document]
cvvz: @DanielZhangQD If not do so, the partition of Statefulset will always be the same as the replicas of component, which will make Statefulset unable to rolling upgrade all the Pods.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -646,6 +646,11 @@ func (m *tikvMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, set
 	// This only returns Up/Down/Offline stores
 	storesInfo, err := pdCli.GetStores()
```
DanielZhangQD, line 647: We may also have to update the code in the following files:
```
pkg/manager/member/tikv_scaler.go:	storesInfo, err := pdClient.GetStores()
pkg/manager/member/tiflash_member_manager.go:	storesInfo, err := pdCli.GetStores()
pkg/manager/member/tiflash_member_manager.go:	storesInfo, err := pdCli.GetStores()
```

pkg/manager/member/tikv_upgrader.go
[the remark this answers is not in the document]
DanielZhangQD: @cvvz In this case, since no error returned with the change in this PR, so the next Pod will be upgraded in the next sync (30s by default, however, there is something that we need to investigate that even if no errors returned, the tc will be synced in about 10s), right? That's not expected.
What about adding a filed `bootStrapped` in the tikv status and check this field in the different logics? For the code here, if not bootstrapped, just set partition to 0.

pkg/pdapi/pdapi.go
dragonly, line 321:
```suggestion
		if pdapi.IsTiKVNotBootstrappedError(err) {
```

pkg/manager/member/tikv_scaler.go
```diff
@@ -18,6 +18,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/pingcap/tidb-operator/pkg/pdapi"
+
```
DanielZhangQD, line 22: Remove the empty line.

DanielZhangQD:
@cvvz Can scale-in be done successfully if TiKV cluster is not bootstrapped yet?
If yes, could you please update the test result in the PR description?

pkg/manager/member/tikv_upgrader.go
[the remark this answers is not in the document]
cvvz: Ok, that is more efficient

pkg/manager/member/tiflash_member_manager.go
```diff
@@ -720,6 +727,11 @@ func (m *tiflashMemberManager) setStoreLabelsForTiFlash(tc *v1alpha1.TidbCluster
 	// for unit test
 	setCount := 0
 
+	if !tc.TiFlashBootStrapped() {
```
DanielZhangQD, line 730: What about using TiKVBootStrapped here?
Because `GetStores` will return normally only if TiKV is bootstrapped.

pkg/manager/member/tikv_upgrader.go
```diff
@@ -93,13 +93,20 @@ func (u *tikvUpgrader) Upgrade(meta metav1.Object, oldSet *apps.StatefulSet, new
...
+
 	setUpgradePartition(newSet, *oldSet.Spec.UpdateStrategy.RollingUpdate.Partition)
 	podOrdinals := helper.GetPodOrdinals(*oldSet.Spec.Replicas, oldSet).List()
 	for _i := len(podOrdinals) - 1; _i >= 0; _i-- {
 		i := podOrdinals[_i]
 		store := u.getStoreByOrdinal(meta.GetName(), *status, i)
 		if store == nil {
-			continue
+			setUpgradePartition(newSet, i)
+			return nil
```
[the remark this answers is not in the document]
cvvz: if no store in the Pod, it would be more appropriate to upgrade this Pod directly.

pkg/manager/member/tiflash_member_manager.go
cvvz: OK, that's reasonable

pkg/manager/member/tikv_upgrader.go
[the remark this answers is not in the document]
cvvz: I think `continue` would be ok

pkg/manager/member/tikv_upgrader.go
```diff
@@ -105,8 +105,7 @@ func (u *tikvUpgrader) Upgrade(meta metav1.Object, oldSet *apps.StatefulSet, new
 		i := podOrdinals[_i]
 		store := u.getStoreByOrdinal(meta.GetName(), *status, i)
 		if store == nil {
-			setUpgradePartition(newSet, i)
```
DanielZhangQD: Why remove this? If there are 3 TiKVs, tikv-2 is up but tikv-0 and tikv-1 are Pending, in this case, the tikv-0 and tikv-1 cannot be upgraded because there is no store for them.

pkg/manager/member/tikv_scaler.go
```diff
@@ -243,6 +245,12 @@ func (s *tikvScaler) preCheckUpStores(tc *v1alpha1.TidbCluster, podName string)
 	pdClient := controller.GetPDClient(s.deps.PDControl, tc)
 	// get the number of stores whose state is up
 	upNumber := 0
+
+	if !tc.TiKVBootStrapped() {
+		klog.Infof("TiKV of Cluster %s/%s is not bootstrapped yet, skip pre check when scale in TiKV", tc.Namespace, tc.Name)
+		return true, nil
+	}
+
```
DanielZhangQD, line 253: Move to the beginning of this function or checking this in the caller function?

pkg/manager/member/tikv_upgrader.go
```diff
@@ -93,12 +93,19 @@ func (u *tikvUpgrader) Upgrade(meta metav1.Object, oldSet *apps.StatefulSet, new
 		return nil
 	}
 
+	// if TiKV not bootstrapped, we set partition to 0 to upgrade quickly
+	if !status.BootStrapped {
+		setUpgradePartition(newSet, 0)
+		return nil
+	}
+
```
DanielZhangQD, line 101: Is this necessary after the change of L108?

cvvz: Good suggestion, I will delelte this code.

pkg/manager/member/tikv_scaler.go
cvvz: ok

DanielZhangQD (approved):
LGTM

DanielZhangQD:
@dragonly @csuzhangxc PTAL.

pkg/pdapi/pdapi.go
```diff
@@ -315,7 +316,14 @@ func (c *pdClient) getStores(apiURL string) (*StoresInfo, error) {
 }
 
 func (c *pdClient) GetStores() (*StoresInfo, error) {
-	return c.getStores(fmt.Sprintf("%s/%s", c.url, storesPrefix))
+	storesInfo, err := c.getStores(fmt.Sprintf("%s/%s", c.url, storesPrefix))
+	if err != nil {
+		if strings.HasSuffix(err.Error(), tiKVNotBootstrapped+"\n") {
+			err = TiKVNotBootstrappedErrorf(tiKVNotBootstrapped)
```
csuzhangxc, line 322: how about using `err.Error()` instead of `tiKVNotBootstrapped` here?

pkg/manager/member/tiflash_member_manager.go
```diff
@@ -629,6 +629,11 @@ func (m *tiflashMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, s
 	// This only returns Up/Down/Offline stores
 	storesInfo, err := pdCli.GetStores()
 	if err != nil {
+		if pdapi.IsTiKVNotBootstrappedError(err) {
```
csuzhangxc, line 632: Can we still keep the original logic for TiFlash?

pkg/pdapi/pdapi.go
cvvz: But how can we know whether the error type is `NotBootstrapped`?

pkg/manager/member/tiflash_member_manager.go
cvvz: OK

pkg/pdapi/pdapi.go
csuzhangxc: `_, ok := err.(*TiKVNotBootstrappedError)` this should still work? I mean, only replace the error message with the original `err.Error()`.

cvvz: I see.

csuzhangxc (approved):
LGTM

PR #3702, opened 2021-01-13
The default value of statefulset partition should take `delete-slots` into consideration

cvvz:

Fix the bug that when upgrade TidbCluster with `delete-slots` annotations, the Pods whose ordinal bigger than `replicaton numbers - 1` will be terminated directly without any pre-delete operations such as evicting leaders. 

The default value of Statefulset's `partition` is added with the number of `delete-slots`.

### Code changes

- [x] Has Go code change

### Tests

- [x] Manual test 
1. Create a cluster with `delete-slots` annotation
2. The `partition` value of Statefulset is `replicas + number of delete-slots`
3. Upgade the cluster, all Pods will be upgraded normally

### Related changes

### Release Notes

```release-note
Fix the bug that when upgrade TidbCluster with `delete-slots` annotations, the Pods whose ordinal bigger than `replicaton numbers - 1` will be terminated directly without any pre-delete operations such as evicting leaders. 
```


pkg/manager/member/tidb_member_manager.go
```diff
@@ -729,13 +733,18 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
...
+	}
+
 	updateStrategy := apps.StatefulSetUpdateStrategy{}
 	if baseTiDBSpec.StatefulSetUpdateStrategy() == apps.OnDeleteStatefulSetStrategyType {
 		updateStrategy.Type = apps.OnDeleteStatefulSetStrategyType
 	} else {
 		updateStrategy.Type = apps.RollingUpdateStatefulSetStrategyType
 		updateStrategy.RollingUpdate = &apps.RollingUpdateStatefulSetStrategy{
-			Partition: pointer.Int32Ptr(tc.TiDBStsDesiredReplicas()),
+			Partition: pointer.Int32Ptr(tc.TiDBStsDesiredReplicas() + deleteSlotsNumber),
```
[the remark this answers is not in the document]
cvvz: Can we get the `deleteSlots` parameter of this function conveniently? If so, using this function would be more graceful.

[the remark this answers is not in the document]
DanielZhangQD: You can get the deleteSlots from the annotation as you're doing, with the current change, it seems that it's a combination of non-asts and asts style, but with that function, it's totally asts style and should be more readable, WDYT?

pkg/manager/member/pd_member_manager.go
```diff
@@ -681,6 +681,11 @@ func getNewPDSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap) (
 	podAnnotations := CombineAnnotations(controller.AnnProm(2379), basePDSpec.Annotations())
 	stsAnnotations := getStsAnnotations(tc.Annotations, label.PDLabelVal)
 
+	deleteSlotsNumber, err := util.GetDeleteSlotsNumber(stsAnnotations)
+	if err != nil {
+		return nil, fmt.Errorf("get delete slots number failed, err:%v", err)
```
DanielZhangQD, line 686: Add sts info.

pkg/manager/member/tidb_member_manager.go
```diff
@@ -729,13 +733,18 @@ func getNewTiDBSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 	podAnnotations := CombineAnnotations(controller.AnnProm(10080), baseTiDBSpec.Annotations())
 	stsAnnotations := getStsAnnotations(tc.Annotations, label.TiDBLabelVal)
 
+	deleteSlotsNumber, err := util.GetDeleteSlotsNumber(stsAnnotations)
+	if err != nil {
+		return nil, fmt.Errorf("get delete slots number failed, err:%v", err)
```
DanielZhangQD, line 738: ditto

pkg/manager/member/tikv_member_manager.go
```diff
@@ -427,6 +427,11 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 	capacity := controller.TiKVCapacity(tc.Spec.TiKV.Limits)
 	headlessSvcName := controller.TiKVPeerMemberName(tcName)
 
+	deleteSlotsNumber, err := util.GetDeleteSlotsNumber(stsAnnotations)
+	if err != nil {
+		return nil, fmt.Errorf("get delete slots number failed, err:%v", err)
```
DanielZhangQD, line 432: ditto

pkg/util/util.go
```diff
@@ -125,6 +125,22 @@ func GetPodOrdinals(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType) (s
 	return podOrdinals, nil
 }
 
+func GetDeleteSlotsNumber(annotations map[string]string) (int32, error) {
```
DanielZhangQD, line 128: Please add UT for this function.

DanielZhangQD (approved):
LGTM

pkg/util/util.go
```diff
@@ -125,6 +125,22 @@ func GetPodOrdinals(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType) (s
...
+	if len(annotations) == 0 {
+		return 0, nil
+	}
```
dragonly, line 131:
```suggestion
```
I think these lines are unnecessary, because if len is 0, then in L132 `ok` will always be false.

pkg/util/util.go
```diff
@@ -125,6 +125,22 @@ func GetPodOrdinals(tc *v1alpha1.TidbCluster, memberType v1alpha1.MemberType) (s
...
+	value, ok := annotations[helper.DeleteSlotsAnn]
+	if !ok {
+		return 0, nil
+	}
+	var slice []int32
+	err := json.Unmarshal([]byte(value), &slice)
+	if err != nil {
+		return 0, err
+	}
```
dragonly, line 140:
```suggestion
	if err := json.Unmarshal([]byte(value), &slice); err != nil {
		return 0, err
	}
```

dragonly:
please also fix the CI

pkg/util/util.go
cvvz: That's right

DanielZhangQD:
@cvvz Please fix the CI.

dragonly (approved):
LGTM

PR #3554, opened 2020-12-03
supoport canary deployment

cvvz:
### What problem does this PR solve? 
https://github.com/pingcap/tidb-operator/pull/3243
https://github.com/pingcap/tidb-operator/pull/3166
https://github.com/pingcap/tidb-operator/pull/1866

### What is changed and how does it work?
This PR is merged with https://github.com/pingcap/tidb-operator/pull/3548 ,  @shonge and I did this together. **Please review the code of `helm` part in his PR https://github.com/pingcap/tidb-operator/pull/3548.**
Users now can deploy mutiple operators(tidb-controller-manager and tidb-scheduler) in k8s by helm. And tidb-controller-manager will work independently by filtering the CR by label.

### Check List 

Tests 

- Manual test 
  - Install two operator releases by helm with values of `scheduler.serviceAccount`, `scheduler.schedulerName`, `controllerManager.serviceAccount` different and only install `controllerManager` and `scheduler`. (`advancedStatefulset` and `admissionWebhook` is not allowed to install mutiple) And one controller is set the `controllerManager.selector` as `k1=v1`, another is set as `k1!=v1`.
  - Create two `TidbCluster`, one is labeled with `k1=v1` and set the scheduler name as `tidb-scheduler-1`, another is not labeled and set the scheduler name as `tidb-scheduler-2`.
  - Notice that each `TidbCluster` will **only** be watched by specific controller.

Code changes

- Has Go code change

Related changes

- Need to update the documentation

### Does this PR introduce a user-facing change?:

```release-note
Support configuring `--selector` for the tidb-controller-manager to manage specific CRs 
```

PR #3548, opened 2020-12-03
[WIP] Support tidb-operator to canary deploy in helm chart

shonge:
### What problem does this PR solve? 
#3243 #3166 #1866

Dependent: #3611

### What is changed and how does it work?
Support tidb-controller-manager and scheduler to canary deploy.
Add `controllerManager.selector` in values.yaml
When user set seletcor, it will be appended to tidb-controller-manager command args `-selector=`

### Check List 

Tests 
- Manual test (add detailed scripts or steps below)

<details>
  <summary>Install with the new chart without specifying values.yaml</summary>

```
helm-v2 install --name tidb-operator-v1 --namespace tidb-admin .
```
```
NAME:   tidb-operator-v1
LAST DEPLOYED: Thu Dec 10 16:10:49 2020
NAMESPACE: tidb-admin
STATUS: DEPLOYED

RESOURCES:
==> v1/ConfigMap
NAME                   DATA  AGE
tidb-scheduler-policy  1     0s

==> v1/Deployment
NAME                                      READY  UP-TO-DATE  AVAILABLE  AGE
tidb-controller-manager-tidb-operator-v1  0/1    1           0          0s
tidb-scheduler                            0/1    1           0          0s

==> v1/Pod(related)
NAME                                                      READY  STATUS             RESTARTS  AGE
tidb-controller-manager-tidb-operator-v1-cd9bd8464-5j6rz  0/1    ContainerCreating  0         0s
tidb-scheduler-585c86b68d-dl4vj                           0/2    ContainerCreating  0         0s

==> v1/ServiceAccount
NAME                     SECRETS  AGE
tidb-controller-manager  1        0s
tidb-scheduler           1        0s

==> v1beta1/ClusterRole
NAME                                      AGE
tidb-operator-v1:tidb-controller-manager  0s
tidb-operator-v1:tidb-scheduler           0s

==> v1beta1/ClusterRoleBinding
NAME                                      AGE
tidb-operator-v1:kube-scheduler           0s
tidb-operator-v1:tidb-controller-manager  0s
tidb-operator-v1:tidb-scheduler           0s
tidb-operator-v1:volume-scheduler         0s
```
</details>

<details>
  <summary>Install with the new chart with the new values.yaml explicitly</summary>

```
helm-v2 install --name tidb-operator-v1 --namespace tidb-admin . -f values.yaml
```
```
NAME:   tidb-operator-v1
LAST DEPLOYED: Thu Dec 10 16:13:44 2020
NAMESPACE: tidb-admin
[...]

PR #3772, opened 2021-02-02
Don't create ClusterRole unless clusterScoped

tirsen:

Without this you can't install multiple helm charts (even in separate namespaces) in a Kubernetes cluster. You get the following error when you try to install the second one:
```
Error: Failed to render chart: exit status 1: Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: , name: tidb-operator:tidb-controller-manager, existing_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole, new_kind: rbac.authorization.k8s.io/v1, Kind=ClusterRole
```

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

### DIFF
--- a/charts/tidb-operator/templates/controller-manager-rbac.yaml
+++ b/charts/tidb-operator/templates/controller-manager-rbac.yaml
@@ -16,6 +16,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/component: controller-manager
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+"  "_" }}
+{{- if .Values.clusterScoped }}
 ---
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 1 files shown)


DanielZhangQD:
@tirsen Thanks for submitting the fix!
However, even for non cluster-scoped TiDB Operator, we still have to create RBAC rules for ClusterRoleBinding and ClusterRole listed [here](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-operator/templates/controller-manager-rbac.yaml#L78-L96).

> Without this you can't install multiple helm charts (even in separate namespaces) in a Kubernetes cluster

For this concern, after #3548 and #3554, you should be able to do that, could you please try [this](https://docs.pingcap.com/tidb-in-kubernetes/stable/deploy-multiple-tidb-operator) with v1.1.10.

PR #3789, opened 2021-02-18
Add TiFlash rolling upgrade logic

### LINKED ISSUE #3786: TiFlash does not support rolling upgrade
## Bug Report

TiFlash can't be ready to serve request immediately when restarted, if there are lots of table schemas in the cluster it would take a while to be ready. However, currently [tiflash_upgrader.go](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tiflash_upgrader.go#L33-L45) does not query tiflash store status during upgrade.

So the next tiflash pod upgrade may happen when the previous tiflash hasn't been ready to serve request thus causing service outage for TiFlash.

Note: even if we wait until the tiflash store ready, users still need to set the tiflash tables replica to at least 2 in order to achieve online rolling upgrade.

handlerww:

Fixes #3786

Add rolling upgrade logic for TiFlash component.

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

1. Create a cluster with TiFlash

> Note: The replicas of TiKV should be more than TiFlash.
```yaml
tikv:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/tikv
    version: v4.0.8
    replicas: 3
    requests:
      storage: "5Gi"
    config: {}
tiflash:
    baseImage: registry.cn-beijing.aliyuncs.com/tidb/tiflash
    version: v4.0.8
    maxFailoverCount: 3
    annotations:
      tidb.pingcap.com/sysctl-init: "true"
    replicas: 2
    storageClaims:
      - resources:
          requests:
            storage: 1Gi
```

3. Use go-tpc tpch to generate data into the TiFlash

```bash
go-tpc tpch -H 10.96.138.6 --tiflash --sf=50 prepare 
```

Modify the TiFlash replicas of the tables in SQL.

```sql
ALTER TABLE `test`.`nation` SET TIFLASH REPLICA 2;
ALTER TABLE `test`.`region` SET TIFLASH REPLICA 2;
ALTER TABLE `test`.`lineitem` SET TIFLASH REPLICA 2;
ALTER TABLE `test`.`supplier` SET TIFLASH REPLICA 2;
ALTER TABLE `test`.`customer` SET TIFLASH REPLICA 2;
ALTER TABLE `test`.`orders` SET TIFLASH REPLICA 2;
ALTER TABLE `test`.`partsupp` SET TIFLASH REPLICA 2;
ALTER TABLE `test`.`part` SET TIFLASH REPLICA 2;
```

4. Use command to check if TiFlash is working as expected.

```sql
SELECT * FROM information_schema.tiflash_replica;
```

If `available` is 1, it shows that TiFlash is working. You can see the TiFlash working status from TiDBMonitor.

5. Change tidbcluster CR to upgrade TiFlash, and monitor TiFlash with `select` query and the TiFlash metrics panel in Grafana.

```bash
kubectl patch tc cluster1 --type merge -p '{"spec":{"tiflash":{"version":"v4.0.9"}}}'
```

Continuely counting from `test.lineitem` to check if TiFlash working normally.

```bash
watch "mysql -h10.96.138.6 -P4000 -e'select count(*) from test.lineitem';"
```

Watch the Grafana panel to verify the fixes are working as expected. Before we add rolling upgrade logic, the pod cluster1-tiflash-0 will immediately terminate after the pod cluster1-tiflash-1 is launched.

image

We can verify this in the `Memory` panel. The memory usage of cluster1-tiflash-0 and cluster1-tiflash-1 cross at the bottom of the panel.

After we add rolling upgrade logic, the pod cluster1-tiflash-1 will first restart, and after the store state is up, the pod cluster1-tiflash-0 will restart later, which shows in Grafana that the memory usage of the pod cluster1-tiflash-1 will rise from the bottom to the peak for a while, then the pod cluster1-tiflash-1 will restart.

image

After the fix, we can see that there are two stores in Up status while there are only 1 TiFlash pod upgraded, and the `select` query is still working, There is an upgrading gap between 2 TiFlash pods.

image

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Add TiFlash rolling upgrade logic to avoid all TiFlash stores unavailable.
```

### DIFF
--- a/pkg/manager/member/tiflash_upgrader.go
+++ b/pkg/manager/member/tiflash_upgrader.go
@@ -39,7 +39,6 @@ func (u *tiflashUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Statefu
 	ns := tc.GetNamespace()
 	tcName := tc.GetName()
 
-	var status *v1alpha1.TiFlashStatus
 	if tc.Status.PD.Phase == v1alpha1.UpgradePhase || tc.Status.TiKV.Phase == v1alpha1.UpgradePhase ||
 		tc.Status.TiDB.Phase == v1alpha1.UpgradePhase {
 		_, podSpec, err := GetLastAppliedConfig(oldSet)
@@ -50,16 +49,16 @@ func (u *tiflashUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Statefu
 		return nil
 	}
 
-	if !status.Synced {
+	if !tc.Status.TiFlash.Synced {
 		return fmt.Errorf("cluster: [%s/%s]'s TiFlash status sync failed, can not to be upgraded", ns, tcName)
 	}
 
-	status.Phase = v1alpha1.UpgradePhase
+	tc.Status.TiFlash.Phase = v1alpha1.UpgradePhase
 	if !templateEqual(newSet, oldSet) {
 		return nil
 	}
 
-	if status.StatefulSet.UpdateRevision == status.StatefulSet.CurrentRevision {
+	if tc.Status.TiFlash.StatefulSet.UpdateRevision == tc.Status.TiFlash.StatefulSet.CurrentRevision {
 		return nil
 	}
 
@@ -77,7 +76,7 @@ func (u *tiflashUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.Statefu
 	podOrdinals := helper.GetPodOrdinals(*oldSet.Spec.Replicas, oldSet).List()
 	for _i := len(podOrdinals) - 1; _i >= 0; _i-- {
 		i := podOrdinals[_i]
-		store := getTiFlashStoreByOrdinal(tc.GetName(), *status, i)
+		store := getTiFlashStoreByOrdinal(tc.GetName(), tc.Status.TiFlash, i)
 		if store == nil {
 			setUpgradePartition(newSet, i)
 			continue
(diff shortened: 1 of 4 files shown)


pkg/manager/member/tiflash_upgrader.go
```diff
@@ -41,6 +48,78 @@ func (tku *tiflashUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.State
...
+			if store.State != v1alpha1.TiKVStateUp {
+				return controller.RequeueErrorf("tidbcluster: [%s/%s]'s upgraded TiFlash pod: [%s] is not all ready", ns, tcName, podName)
+			}
+			continue
+		}
+
+		if u.deps.CLIConfig.PodWebhookEnabled {
+			setUpgradePartition(newSet, i)
+			return nil
+		}
```
handlerww, line 107:
```suggestion
```

DanielZhangQD (approved):
LGTM

pkg/manager/member/tiflash_upgrader.go
```diff
@@ -41,6 +48,73 @@ func (tku *tiflashUpgrader) Upgrade(tc *v1alpha1.TidbCluster, oldSet *apps.State
 		newSet.Spec.Template.Spec = *podSpec
 		return nil
 	}
+
+	if !tc.Status.TiFlash.Synced {
```
DanielZhangQD, line 52: Please add UT cases for this file.

pkg/manager/member/tiflash_upgrader.go
```diff
@@ -14,11 +14,15 @@
...
+	"fmt"
+
+	"github.com/pingcap/advanced-statefulset/client/apis/apps/v1/helper"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/controller"
 	apps "k8s.io/api/apps/v1"
+	"k8s.io/klog"
+	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 )
-
```
DanielZhangQD, line 21: Please revert this.

PR #3790, opened 2021-02-19
move WaitForTidbClusterReady out of operatorActions

dragonly:
there are still a lot of operator unrelated functions needs to be
moved out from operatorActions. At the last, we will have a clean
struct and no redundant interface.

make operatorActions cleaner.

- move unrelated code out
- remove almost-identical functions
- more usage of controller-runtime/client (genericCli), which makes less dependencies such as stsGetter

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
None
```


dragonly:
this PR is complicated and not quite necessary for now, we can stick with OperatorActions for a while and simplify code on the go.

Issue #3292, opened 2020-09-20
Refactor Tidbmonitor design.


mikechengwei:
## Description

Currently, there are some problems with TiDBMonitor. We organize the requirements and list the requirements that need improvement.

1. TiDBMonitor is a stateful application but is created as a deployment. We should switch to StatefulSet. #2968
2. TiDBMonitor monitors multiple TiDB clusters for both TLS and non-TLS environment.
    After discussion, we will not support monitoring multiple clusters with one TidbMonitor when TLS enabled because this would require the following restrictions:
* All of the monitored clusters have to enable TLS
* Users must ensure that all the monitored clusters use the same CA certificate
* A dedicated cluster client secret needs to be created for TidbMonitor

The above restrictions are not reasonable so we will just support monitoring multiple clusters with TLS disabled with one TidbMonitor.
3. Monitor multiple TiDB Clusters with users own Prometheus #3184
    We can provide an example with Prometheus Operator.
4. Aggregate the monitoring data of multiple TiDB Clusters, we can refer to the Prometheus operator and support the Thanos spec.  #3118 #3117
5. Combination of Kubernetes monitoring and TiDB cluster monitoring. #3099
6. add external fields for `kubectl get tm` command. #2169 #2982
7. Grafana dashboards for multiple clusters.
8. The PD Dashboard monitor address is currently TidbCluster controller updated, it should be updated by tidbmonitor controller. And PD dashboard doesn't support multiple clusters.
9. Prometheus need to reload configuration  when spec filed cause configMap update .

   -  We can regenerate configMap name cause deployment rolling update.
   - Hot update configuration through reloader and call prometheus reload api (More user-friendly).

## Category
Feature

## Value
Refactor tidbMonitor.

## TODO lists
- P0 features
	- [x] Tidbmonitor monitors multiple clusters across multiple ns in tls and non-tls envrionments. #3307 
	- [x] Support generating the Grafana dashboard of multiple clusters.
- P1 features	
	- [x] Smooth upgrade deployment to StatefulSet. #2968
	- [x] Support Thanos spec and optimize Thanos example. #3553
	- [ ] Optimize TiDB service, more friendly support with Prometheus operator `ServiceMonitor`.
	- [x] Refactor for PD dashboard address writing.  #3924
	- [x] TidbMonitor multiple replica. #2968 #3815 sloved.

- P2 feature
	- [ ] Fix the combination of Kubernetes monitoring and TiDB cluster monitoring.
[...]

Issue #3659, opened 2020-12-27
Sample yaml file can automatically generated by template


mikechengwei:
## Feature Request

**Is your feature request related to a problem? 
Now we provide a lot of example yaml files, but due to the upgrade of the component version, the verison field or other fields are often updated.
We can provide developers with a mechanism to automatically generate example yaml files. Some of these fields in example yaml file  are configurable.

**Describe the feature you'd like:**
1. We can use jsonnet and gojsontoyaml to generate example yaml file.
2. Write a `update-example.sh` to auto generate example file.

DanielZhangQD:
@mikechengwei Good idea!
Would you like to help improve this?

PR #3867, opened 2021-03-21
multiple cluster tls monitoring

### LINKED ISSUE #3883: [Design] TiDBMonitor monitor multiple tls clusters accross multiple ns.
## Feature Request

Now , tidbmonitor can't monitor multiple tls clusters across multiple ns , we need to support it.

In order to allow users to cooperate with tls as little as possible and user can only need to spec tidbcluster name and ns fields, tidbmonitor can automatically decide whether to set up tls configurations according to the cluster itself.

We can take tidbmonitor get tidb tls configuration by k8s  apiserver and create a public secret which contain all cluster secrets under his own namespace.

We can create a secret assets cache and sync loop update.

The cache use map class store , key and value like below:
```
secret_<ns>_<name>_<key> : value
```
Then we create secret with it  and mount in tidbmonitor , when store value change , operator will update it.

mikechengwei:
Support multiple clusters tls monitoring.
Refer #3292  Fix #3883

Tidbmonitor create a asset secret which combine all monitoring clusters tls configuartion across different namespace.
### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

```release-note
Support monitoring multiple  tls  tidb clusters.
```


DanielZhangQD:
@mikechengwei Please fix the CI.

examples/basic/tidb-monitor.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
 spec:
   clusters:
   - name: basic
+  - name: heterogeneous
```
DanielZhangQD, line 8: Why change this file?

examples/monitor-multiple-cluster-tls/cert-manager/ns1/cluster-client-secret.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: cert-manager.io/v1alpha2
```
DanielZhangQD, line 1: Can we just add some templates in `examples/cert-manager` with the TidbCluster as a special name like TIDB_CLUSTER_NAME in the templates, write a simple shell to generate the desired yamls from the templates? Then we no need to add so many yamls in this PR and also the ones in `examples/heterogeneous-tls/cert-manager`?

examples/monitor-multiple-cluster-tls/ns1-cluster.yaml
```diff
@@ -0,0 +1,46 @@
+# IT IS NOT SUITABLE FOR PRODUCTION USE.
+# This YAML describes a basic TiDB cluster with minimum resource requirements,
+# which should be able to run in any Kubernetes cluster with storage support.
+apiVersion: pingcap.com/v1alpha1
+kind: TidbCluster
+metadata:
+  name: ns1
+  namespace: ns1
```
DanielZhangQD, line 8: The secrets are generated in the default namespace.

examples/monitor-multiple-cluster-tls/cert-manager/ns1/cluster-client-secret.yaml
DanielZhangQD: And also consider the namespace, which will be more common to use.

examples/monitor-multiple-cluster-tls/ns2-cluster.yaml
```diff
@@ -0,0 +1,46 @@
+# IT IS NOT SUITABLE FOR PRODUCTION USE.
+# This YAML describes a basic TiDB cluster with minimum resource requirements,
+# which should be able to run in any Kubernetes cluster with storage support.
+apiVersion: pingcap.com/v1alpha1
+kind: TidbCluster
+metadata:
+  name: ns2
+  namespace: ns2
```
DanielZhangQD, line 8: ditto.

examples/monitor-multiple-cluster-tls/tidb-monitor.yaml
```diff
@@ -0,0 +1,40 @@
+apiVersion: pingcap.com/v1alpha1
+kind: TidbMonitor
+metadata:
+  name: monitor
+spec:
+  clusterScoped: true
+  clusters:
+    - name: ns1
+      namespace: ns1
+      enableTls: true
```
DanielZhangQD, line 10: TiDB Operator should be able to get this value automatically.

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -246,6 +246,14 @@ type TidbClusterRef struct {
 	// ClusterDomain is the domain of TidbCluster object
 	// +optional
 	ClusterDomain string `json:"clusterDomain,omitempty"`
+
+	// TlsFileDir is the path of TidbCluster tls files.
+	// +optional
+	TLSFileDir string `json:"tlsFileDir,omitempty"`
```
DanielZhangQD, line 252:
```suggestion
	TLSCertsDir string `json:"tlsCertsDir,omitempty"`
```

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -246,6 +246,14 @@ type TidbClusterRef struct {
...
+
+	// whether enable tls request.
+	// +optional
+	EnableTLS *bool `json:"enableTls,omitempty"`
```
DanielZhangQD, line 256: This is not necessary.

pkg/monitor/monitor/template.go
```diff
@@ -149,8 +149,10 @@ type MonitorConfigModel struct {
 
 // ClusterRegexInfo is the monitor cluster info
 type ClusterRegexInfo struct {
-	Name      string
-	Namespace string
+	Name       string
+	Namespace  string
+	EnableTLS  bool
+	TLSFileDir string
```
DanielZhangQD, line 155:
```suggestion
	TLSCertsDir string
```

pkg/monitor/monitor/util.go
```diff
@@ -118,10 +118,23 @@ func getMonitorConfigMap(tc *v1alpha1.TidbCluster, dc *v1alpha1.DMCluster, monit
 
 	var releaseClusterInfos []ClusterRegexInfo
 	for _, cluster := range monitor.Spec.Clusters {
-		releaseClusterInfos = append(releaseClusterInfos, ClusterRegexInfo{
+		clusterRegex := ClusterRegexInfo{
 			Name:      cluster.Name,
 			Namespace: cluster.Namespace,
-		})
+		}
+		//  Old logical , firstTc enable tls and use default tls path
```
DanielZhangQD, line 125: We can check if the TLS enabled for each TC, why do we need to use the original logic here?

examples/monitor-multiple-cluster-tls/cert-manager/ns1/cluster-client-secret.yaml
mikechengwei: You're right. Ref issue: https://github.com/pingcap/tidb-operator/issues/3659

mikechengwei:
Manual test:
ns1 cluster in ns1 namespace and ns2 cluster in ns2 namespace , tidbmonitor  monitor ns1 and ns2 in default namespace:
image

image

pkg/controller/generic_control.go
```diff
@@ -43,7 +43,7 @@ type TypedControlInterface interface {
 	CreateOrUpdateConfigMap(controller runtime.Object, cm *corev1.ConfigMap) (*corev1.ConfigMap, error)
 	// CreateOrUpdateClusterRole the desired clusterRole or update the current one to desired state if already existed
 	CreateOrUpdateClusterRole(controller runtime.Object, clusterRole *rbacv1.ClusterRole) (*rbacv1.ClusterRole, error)
-	// CreateOrUpdateClusterRoleBinding create the desired clusterRoleBinding or update the current one to desired state if already existed
+	// CreateOrUpdateClusterRoleBindinag create the desired clusterRoleBinding or update the current one to desired state if already existed
```
csuzhangxc, line 46: revert this?

mikechengwei: Sorry and i have done.

csuzhangxc (approved):
LGTM

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -75,12 +75,24 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
 	}
 
 	var firstTc *v1alpha1.TidbCluster
+	assetStore := NewStore(m.deps.ConfigMapLister, m.deps.SecretLister)
```
DanielZhangQD, line 78: The `ConfigMapLister` only watches the configmaps managed by TiDB Operator, and it cannot get all the configmaps in the cluster.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -75,12 +75,24 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
 			rerr := fmt.Errorf("get tm[%s/%s]'s target tc[%s/%s] failed, err: %v", monitor.Namespace, monitor.Name, tcRef.Namespace, tcRef.Name, err)
 			return rerr
 		}
+
+		// If cluster enable tls
+		if tc.IsTLSClusterEnabled() {
+			tcTlsSecretName := util.ClusterClientTLSSecretName(tc.Name)
+			err := assetStore.addTLSAssets(tc.Namespace, tcTlsSecretName)
+			if err != nil {
+				return err
```
DanielZhangQD, line 92: Can we just continue to sync for other clusters? For the error here, we can choose one of the solutions:
1. print error log and continue, wait for the next resync period (30s)
2. aggrate and return the error in the end of this function

pkg/monitor/monitor/monitor_manager.go
DanielZhangQD: https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/dependences.go#L294 Otherwise, it may take too many memories for a large Kubernetes clusters.

pkg/monitor/monitor/util.go
```diff
@@ -44,6 +44,14 @@ const (
 	defaultReplicaExternalLabelName = "prometheus_replica"
 )
 
+func GetTlsAssetsSecretName(name string) string {
```
DanielZhangQD, line 47:
```suggestion
func GetTLSAssetsSecretName(name string) string {
```

pkg/monitor/monitor/util.go
```diff
@@ -478,6 +478,11 @@ func getMonitorPrometheusContainer(monitor *v1alpha1.TidbMonitor, tc *v1alpha1.T
 				MountPath: "/prometheus-rules",
 				ReadOnly:  false,
 			},
+			{
```
DanielZhangQD, line 481: What if TLS is disabled for all TiDB clusters?

pkg/monitor/monitor/util.go
```diff
@@ -742,6 +728,18 @@ func getMonitorVolumes(config *core.ConfigMap, monitor *v1alpha1.TidbMonitor, tc
 		volumes = append(volumes, monitor.Spec.AdditionalVolumes...)
 	}
 
+	// add asset tls
```
DanielZhangQD, line 731: ditto

examples/basic/tidb-monitor.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     version: 6.1.6
   initializer:
     baseImage: pingcap/tidb-monitor-initializer
-    version: v4.0.10
```
DanielZhangQD, line 16: Just revert this, we will have dedicated PR to update the version to v5.0 later.

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -246,6 +246,9 @@ type TidbClusterRef struct {
 	// ClusterDomain is the domain of TidbCluster object
 	// +optional
 	ClusterDomain string `json:"clusterDomain,omitempty"`
+
+	// TLS configuration to use when scraping the cluster mertric.
```
DanielZhangQD, line 250:
```suggestion
	// TLS configuration to use when scraping the cluster metrics.
```

DanielZhangQD: Is this still used now?

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -398,3 +421,9 @@ type QueueConfig struct {
 	MinBackoff time.Duration `json:"minBackoff,omitempty"`
 	MaxBackoff time.Duration `json:"maxBackoff,omitempty"`
 }
+
+// ClusterTLSConfig specifies TLS configuration parameters.
```
DanielZhangQD, line 425: ditto

pkg/monitor/monitor/monitor_manager.go
DanielZhangQD: It seems that the configmap lister is not used.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -75,12 +75,24 @@ func (m *MonitorManager) SyncMonitor(monitor *v1alpha1.TidbMonitor) error {
...
+			}
+		}
+
 		if firstTc == nil && !tc.IsHeterogeneous() {
 			firstTc = tc
```
DanielZhangQD, line 97: Is this still necessary?

DanielZhangQD: How are the TidbCluster related envs defined in this function https://github.com/pingcap/tidb-operator/blob/master/pkg/monitor/monitor/util.go#L274 used now?

pkg/monitor/monitor/template.go
```diff
@@ -151,6 +151,7 @@ type MonitorConfigModel struct {
 type ClusterRegexInfo struct {
 	Name      string
 	Namespace string
+	enableTls bool
```
DanielZhangQD, line 154:
```suggestion
	enableTLS bool
```

pkg/monitor/monitor/util.go
```diff
@@ -136,7 +136,7 @@ func getMonitorConfigMap(tc *v1alpha1.TidbCluster, dc *v1alpha1.DMCluster, monit
 
 	model := &MonitorConfigModel{
 		AlertmanagerURL:    "",
-		ClusterInfos:       releaseClusterInfos,
+		ClusterInfos:       monitorClusterInfos,
 		DMClusterInfos:     releaseDMClusterInfos,
 		EnableTLSCluster:   tc.IsTLSClusterEnabled(),
```
DanielZhangQD, line 141: Is this still used? It seems the TLS info is included in the ClusterInfos?

pkg/monitor/monitor/template.go
DanielZhangQD: Where is this used?

pkg/monitor/monitor/util.go
```diff
@@ -44,6 +44,14 @@ const (
 	defaultReplicaExternalLabelName = "prometheus_replica"
 )
 
+func GetTlsAssetsSecretName(name string) string {
+	return fmt.Sprintf("%s-tls-assets", prefixedName(name))
+}
+
+func prefixedName(name string) string {
```
DanielZhangQD, line 51: Where is this used?

pkg/monitor/monitor/util.go
```diff
@@ -114,15 +122,7 @@ func getAlertManagerRulesVersion(tc *v1alpha1.TidbCluster, monitor *v1alpha1.Tid
...
-func getMonitorConfigMap(tc *v1alpha1.TidbCluster, dc *v1alpha1.DMCluster, monitor *v1alpha1.TidbMonitor) (*core.ConfigMap, error) {
-
-	var releaseClusterInfos []ClusterRegexInfo
-	for _, cluster := range monitor.Spec.Clusters {
-		releaseClusterInfos = append(releaseClusterInfos, ClusterRegexInfo{
-			Name:      cluster.Name,
-			Namespace: cluster.Namespace,
-		})
-	}
+func getMonitorConfigMap(tc *v1alpha1.TidbCluster, dc *v1alpha1.DMCluster, monitor *v1alpha1.TidbMonitor, monitorClusterInfos []ClusterRegexInfo) (*core.ConfigMap, error) {
```
DanielZhangQD, line 125: Is `tc` still necessary?

pkg/monitor/monitor/util.go
mikechengwei: GetTlsAssetsSecretName method and prefixedName can be  merged.

pkg/monitor/monitor/monitor_manager.go
mikechengwei: I can print log,but this for loop code just returns now when it encounters an error ,you can look at the context around the code.

pkg/monitor/monitor/util.go
mikechengwei: I think we can take the `tls-assets` secret as default secret.When user update monitoring cluster targets, tls-assets can hot-update.

pkg/monitor/monitor/util.go
mikechengwei: I think we can take the `tls-assets` secret as default volume.Not only prometheus use it save secret,other container will  
also need ,for example thanos etc.

pkg/monitor/monitor/monitor_manager.go
mikechengwei: After discussion, i remove tc from configuraiton generate process firstly.  

syncTidbMonitorConfig(tc *v1alpha1.TCCluster, dc *v1alpha1.DMCluster, monitor *v1alpha1.TidbMonitor)
simple to below method:
syncTidbMonitorConfig(dc *v1alpha1.DMCluster, monitor *v1alpha1.TidbMonitor)

pkg/monitor/monitor/monitor_manager.go
mikechengwei: Currently, only secrets are used, but there will definitely be a need for configmap in the future. I have removed configmapLister.

mikechengwei: Do you mean user must add tidb-operator label to secret ,right?

DanielZhangQD: Add related label to the configmaps.

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -613,3 +649,27 @@ func (m *MonitorManager) patchPVClaimRef(pv *corev1.PersistentVolume, patchPvcNa
...
+		},
+		Data: make(map[string][]byte, len(store.TLSAssets)),
+	}
+	for key, asset := range store.TLSAssets {
+		tlsAssetsSecret.Data[key.String()] = []byte(asset)
+	}
+
+	_, err := m.deps.TypedControl.CreateOrUpdateSecret(monitor, tlsAssetsSecret)
+	if err != nil {
+		klog.Errorf("tm[%s/%s]'s secret failed to sync,err: %v", ns, name, err)
```
DanielZhangQD, line 671:
```suggestion
		klog.Errorf("Fail to sync tm[%s/%s]'s secret assets, err: %v", ns, name, err)
```

pkg/monitor/monitor/monitor_manager.go
```diff
@@ -613,3 +649,27 @@ func (m *MonitorManager) patchPVClaimRef(pv *corev1.PersistentVolume, patchPvcNa
...
+	tlsAssetsSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            GetTlsAssetsSecretName(monitor.Name),
+			Namespace:       monitor.Namespace,
+			Labels:          buildTidbMonitorLabel(monitor.Name),
+			OwnerReferences: []metav1.OwnerReference{controller.GetTiDBMonitorOwnerRef(monitor)},
+		},
+		Data: make(map[string][]byte, len(store.TLSAssets)),
+	}
+	for key, asset := range store.TLSAssets {
```
DanielZhangQD, line 665: We should return early if `len(store.TLSAssets)==0`, the function `CreateOrUpdateSecret` will call API to kube-apiserver directly during each sync, which is not necessary if no cluster has TLS enabled.

pkg/monitor/monitor/util.go
DanielZhangQD: If the user adds one TLS tc to some non-TLS tcs, then the monitor Pod will be rolling updated automatically, I think it's OK.

pkg/apis/pingcap/v1alpha1/tidbmonitor_types.go
```diff
@@ -307,6 +307,26 @@ type SafeTLSConfig struct {
 	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
 }
 
+// Validate semantically validates the given TLSConfig.
```
DanielZhangQD, line 310: The code changes in this file are not used.

pkg/monitor/monitor/util.go
```diff
@@ -478,6 +473,11 @@ func getMonitorPrometheusContainer(monitor *v1alpha1.TidbMonitor, tc *v1alpha1.T
 				MountPath: "/prometheus-rules",
 				ReadOnly:  false,
 			},
+			{
+				Name:      "tls-assets",
+				MountPath: util.ClusterAssetsTLSPath,
+				ReadOnly:  false,
```
DanielZhangQD, line 479: Why `false` here?

pkg/monitor/monitor/util.go
```diff
@@ -742,6 +723,18 @@ func getMonitorVolumes(config *core.ConfigMap, monitor *v1alpha1.TidbMonitor, tc
 		volumes = append(volumes, monitor.Spec.AdditionalVolumes...)
 	}
 
+	// add asset tls
```
DanielZhangQD, line 726:
```suggestion
	// Add TLS assets secret volume by default, so that no need to rolling update the monitor
	// Pod if users add new TLS clusters even if non of the original clusters has TLS enabled
```

DanielZhangQD (approved):
LGTM

PR #3869, opened 2021-03-22
Add kv default open state port 20180 mapping

hhy5861:
Add kv default open state port 20180 mapping

Convenient to use prometheus operator podmonitor to collect and discover


DanielZhangQD:
@hhy5861 You must have commited some commits with `Mike`, you can just pull the latest master, checkout a new branch, update the code in `types.go`, `pkg/manager/member/tidb_discovery_manager.go`, `pkg/manager/member/tikv_member_manager.go` and run `hack/update-all.sh` and then submit a new PR.

Issue #3080, opened 2020-08-06
TidbClusters are synced too frequently even if there is no update to it


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

By default, the resync duration for the CRs are 30s, however, currently, the TidbClusters are synced very frequently even if no update to them.
**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

DanielZhangQD:
```
I0806 01:01:28.468298       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:01:40.670287       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:01:54.069548       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:02:02.666732       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:02:13.867462       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:02:25.866712       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:02:39.467229       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:02:50.068625       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:02:59.467018       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:03:10.266458       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:03:23.869315       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
I0806 01:03:35.267130       1 tidbcluster_control.go:66] TidbCluster: [ns1/ns1] updated successfully
```

dragonly:
I also observed this issue by apply https://github.com/pingcap/tidb-operator/blob/master/examples/advanced/tidb-cluster.yaml
It seems there's another problem: the operator starts to sync tikv after pd is ready for about 20s.
the requeuing log is as follows:
```
[...]

PR #3886, opened 2021-03-31
remove LastHeartbeatTime(pingcap#3080)

songjiansuper:
in operator,this value does not seem to be necessary.
At the same time, every sync will cause this value to change,
which seems to trigger more update events

Every sync will cause LastHeartbeatTime value to change,which seems to trigger more update events.
But there is not much need for this value for the operator, so we can remove it
Fix https://github.com/pingcap/tidb-operator/issues/3080
### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Remove `lastHeartbeatTime` field from `status.tikv.stores` because it will cause the TidbCluster CR to be updated during each sync and actually is useless
```


DanielZhangQD:
@songjiansuper Thanks for the contribution!
Please sign the CLA.
BTW, what's the sync interval now with a TidbCluster that includes PD, TiDB, TiKV, and TiFlash?

pkg/controller/tidbcluster_control.go
```diff
@@ -114,7 +113,6 @@ func deepEqualExceptHeartbeatTime(newStatus *v1alpha1.TidbClusterStatus, oldStat
 
 func sweepHeartbeatTime(stores map[string]v1alpha1.TiKVStore) {
```
DanielZhangQD, line 114: This function and `deepEqualExceptHeartbeatTime` can be removed.

july2993 (approved):

songjiansuper:
@DanielZhangQD seems like 30 seconds,but it seems that this 30s is not the same as the 30s of the informer. My own informer resync time is set to 60s, and I will still get an update event in 30s.

DanielZhangQD:
> @DanielZhangQD seems like 30 seconds,but it seems that this 30s is not the same as the 30s of the informer. My own informer resync time is set to 60s, and I will still get an update event in 30s.

It should be controlled by the resync period similar to the code [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/dependences.go#L308).
BTW, did you test the case that deploys the old CRD and a TidbCluster, after all the Pods ready, apply the new CRD in this PR, upgrade to the operator built from this PR and everything works well?

DanielZhangQD:
@songjiansuper Please also help check the comment when you're available, thanks!

DanielZhangQD:
@songjiansuper Just a friendly ping.

songjiansuper:
@DanielZhangQD hello,I’ve been a bit busy with the company’s work recently, I may have to verify the above comment for a while, thx

DanielZhangQD:
> @DanielZhangQD hello,I’ve been a bit busy with the company’s work recently, I may have to verify the above comment for a while, thx

OK, thanks a lot!

DanielZhangQD:
> @DanielZhangQD hello,I’ve been a bit busy with the company’s work recently, I may have to verify the above comment for a while, thx

OK, thanks a lot!

pkg/manager/member/tidb_init_manager.go
```diff
@@ -64,18 +64,7 @@ func NewTiDBInitManager(deps *controller.Dependencies) InitManager {
...
-	tc, err := m.deps.TiDBClusterLister.TidbClusters(ns).Get(tcName)
-	if err != nil {
-		return fmt.Errorf("TidbInitManager.Sync: failed to get tidbcluster %s for TidbInitializer %s/%s, error: %s", tcName, ns, ti.Name, err)
-	}
-	if tc.Spec.TiDB == nil {
-		klog.Infof("TidbInitManager.Sync: Spec.TiDB is nil in tidbcluster %s, skip syncing TidbInitializer %s/%s", tcName, ns, ti.Name)
-		return nil
-	}
-
-	err = m.syncTiDBInitConfigMap(ti)
```
DanielZhangQD, line 78: Why remove this block?

songjiansuper: Shocked, I have no impression of it...this doesn't look like it should be deleted.Modify it together with the content to be removed above

pkg/manager/member/tidb_init_manager.go
```diff
@@ -74,7 +74,6 @@ func (m *tidbInitManager) Sync(ti *v1alpha1.TidbInitializer) error {
 		klog.Infof("TidbInitManager.Sync: Spec.TiDB is nil in tidbcluster %s, skip syncing TidbInitializer %s/%s", tcName, ns, ti.Name)
 		return nil
 	}
```
DanielZhangQD, line 76:
```suggestion
	}

```

DanielZhangQD (approved):
LGTM

PR #3896, opened 2021-04-06
Support installing tidb-operator with namespace scoped permissions

### LINKED ISSUE #3873: Support installing tidb-operator with namespace scoped permissions
## Feature Request

Currently, we support setting `clusterScoped: false` in the [values.yaml](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-operator/values.yaml#L5) during installation of TiDB Operator, however, [clusterrole and clusterrolebinding](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-operator/templates/controller-manager-rbac.yaml#L78-L96) will still be created in this case due to the following features in TiDB Operator：

- Sync some cluster related labels to PV
- `spec.EnablePVReclaim: true` to reclaim PV
- Resizing the storage request of PVCs
- Retrieve values from node labels and set them to the TiKV stores
- Create clusterrole and clusterrolebinding for TidbMonitor

However, some users do not have the cluster-admin permission for the Kubernetes cluster, they can only have the permission to manage one namespace, in which case, they have to install both the TiDB Operator and TiDB cluster in that namespace, so we have to support this.

- tidb-controller-manager
  - Add the following parameters in the values.yaml of tidb-operator chart

    ```
    clusterPermissions: {}
    # nodes: true
    # persistentvolumes: true
    # storageclasses: true
    ```

    - Each of them controls whether to add related permissions to the service account of tidb-controller-manager
    - tidb-controller-manager process also has to be aware of the configurations and skip the related functionalities if necessary
  - For the clusterroles & clusterrolebindings, should only be set with `clusterScoped: true`
- tidb-scheduler
  - The `kube-scheduler` container needs the `system:kube-scheduler` and `system:volume-scheduler` permissions, so if users do not want to grant the permissions, they can set `scheduler.create: false` to skip the creation of tidb-scheduler and use the default scheduler

Refer to the [blog](https://jeremievallee.com/2018/05/28/kubernetes-rbac-namespace-user.html) to create a namespace scoped kubeconfig.
To add the permissions for the custom resources:
```
- apiGroups: ["apps.pingcap.com"]
  resources: ["statefulsets", "statefulsets/status"]
  verbs: ["*"]
- apiGroups: ["pingcap.com"]
  resources: ["*"]
  verbs: ["*"]
```

csuzhangxc:

close #3873

- define some permission parameters
- skip operations without permissions

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
   - see `examples/namespaced/README.md`

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Support installing tidb-operator with namespace scoped permissions
```


charts/tidb-operator/templates/controller-manager-deployment.yaml
```diff
@@ -53,6 +53,11 @@ spec:
           {{- end }}
           - -tidb-discovery-image={{ .Values.operatorImage }}
           - -cluster-scoped={{ .Values.clusterScoped }}
+          {{- if .Values.clusterPermissions }}
+          - -cluster-permission-node={{ .Values.clusterPermissions.nodes | default false }}
+          - -cluster-permission-pv={{ .Values.clusterPermissions.persistentvolumes | default false }}
+          - -cluster-permission-sc={{ .Values.clusterPermissions.storageclasses | default false }}
```
DanielZhangQD, line 59: We'd better set the default value to true to keep backward compatibility.

csuzhangxc: default set these args to `true` when `clusterScoped: false`?

charts/tidb-operator/templates/controller-manager-rbac.yaml
```diff
@@ -117,7 +116,59 @@ roleRef:
   kind: ClusterRole
   name: {{ .Release.Name }}:tidb-controller-manager
   apiGroup: rbac.authorization.k8s.io
-{{- if (not .Values.clusterScoped) }}
+{{- else }}
+{{- if and .Values.clusterPermissions (or .Values.clusterPermissions.nodes .Values.clusterPermissions.persistentvolumes .Values.clusterPermissions.storageclasses) }}
```
DanielZhangQD, line 120: We have to cover the following cases:
1. No such configurations in the values.yaml, e.g. with values.yaml of the previous version
    We have to set the missed configurations to `true`, refer to the logic for  `.Values.controllerManager` in L4.
2. With those configurations and set to `true`
3. With those configurations and set to `false`

charts/tidb-operator/values.yaml
```diff
@@ -1,9 +1,15 @@
...
 # clusterScoped is whether tidb-operator should manage kubernetes cluster wide tidb clusters
-# Also see rbac.create and controllerManager.serviceAccount
+# Also see rbac.create, controllerManager.serviceAccount, scheduler.create and clusterPermissions.
 clusterScoped: true
 
+# clusterPermissions are some cluster scoped permissions that will be used even if `clusterScoped: false`.
+clusterPermissions: {}
+#  nodes: true
+#  persistentvolumes: true
+#  storageclasses: true
```
DanielZhangQD, line 11:
```suggestion
clusterPermissions:
  nodes: true
  persistentvolumes: true
  storageclasses: true
```

examples/namespaced/rbac.yaml
```diff
@@ -0,0 +1,83 @@
...
+#metadata:
+#  name: namespaced
+#rules:
+#  # this namespaced user will be used to crate other users when installing tidb-operator with cluster scoped permissions.
+#  - apiGroups: ["rbac.authorization.k8s.io"]
+#    resources: ["clusterroles"]
+#    verbs: ["escalate","create","get","update", "delete"]
+#  - apiGroups: ["rbac.authorization.k8s.io"]
+#    resources: ["clusterrolebindings"]
+#    verbs: ["create","get","update", "delete"]
```
DanielZhangQD, line 61: These should be removed.

pkg/controller/dependences.go
```diff
@@ -210,9 +234,13 @@ func newRealControls(
 		statefulSetLister = kubeInformerFactory.Apps().V1().StatefulSets().Lister()
 		serviceLister     = kubeInformerFactory.Core().V1().Services().Lister()
 		pvcLister         = kubeInformerFactory.Core().V1().PersistentVolumeClaims().Lister()
-		pvLister          = kubeInformerFactory.Core().V1().PersistentVolumes().Lister()
 		podLister         = kubeInformerFactory.Core().V1().Pods().Lister()
+		pvLister          corelisterv1.PersistentVolumeLister
```
DanielZhangQD, line 238: This should be removed?

examples/namespaced/README.md
```diff
@@ -0,0 +1,61 @@
...
+    namespace: namespaced
+    user: namespaced
+  name: namespaced
+
+current-context: namespaced
+```
+
+3. install a namespaced scope tidb-operator.
+
+```bash
```
DanielZhangQD, line 51: We need to update this to use the official chart and values.yaml after this fix is released.

pkg/manager/member/pvc_cleaner.go
```diff
@@ -149,21 +149,25 @@ func (c *realPVCCleaner) reclaimPV(meta metav1.Object) (map[string]string, error
...
+			if pv.Spec.PersistentVolumeReclaimPolicy != corev1.PersistentVolumeReclaimDelete {
+				err := c.deps.PVControl.PatchPVReclaimPolicy(runtimeMeta, pv, corev1.PersistentVolumeReclaimDelete)
+				if err != nil {
+					return skipReason, fmt.Errorf("%s %s/%s patch pv %s to %s failed, err: %v", clusterType, ns, metaName, pvName, corev1.PersistentVolumeReclaimDelete, err)
+				}
+				klog.Infof("%s %s/%s patch pv %s to policy %s success", clusterType, ns, metaName, pvName, corev1.PersistentVolumeReclaimDelete)
 			}
-			klog.Infof("%s %s/%s patch pv %s to policy %s success", clusterType, ns, metaName, pvName, corev1.PersistentVolumeReclaimDelete)
+		} else {
+			klog.Warningf("persistent volumes lister is unavailable, skip reclaiming policy for %s. this may be caused by no relevant permissions", pvName)
```
DanielZhangQD, line 170: If we print the warning log here, the logs will be flooded with the warning logs for the namespace scoped users, what about change to `klog.V(4).Infof`?

examples/namespaced/rbac.yaml
csuzhangxc: but it seems `this namespaced user will be used to create other users when installing tidb-operator with cluster scoped permissions.`

pkg/manager/meta/meta_manager.go
```diff
@@ -73,6 +73,11 @@ func (m *metaManager) Sync(tc *v1alpha1.TidbCluster) error {
 			if pvc.Spec.VolumeName == "" {
 				continue
 			}
+
+			if m.deps.PVLister == nil {
+				klog.Warningf("persistent volumes lister is unavailable, skip updating meta info for %s. this may be caused by no relevant permissions", pvc.Spec.VolumeName)
+				continue
+			}
```
DanielZhangQD, line 80: Move this block after L72?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -807,6 +807,11 @@ func getTiKVStore(store *pdapi.StoreInfo) *v1alpha1.TiKVStore {
 }
 
 func (m *tikvMemberManager) setStoreLabelsForTiKV(tc *v1alpha1.TidbCluster) (int, error) {
+	if m.deps.NodeLister == nil {
```
DanielZhangQD, line 810: What about checking the lister in the caller function https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_member_manager.go#L213?

pkg/manager/member/pvc_resizer.go
```diff
@@ -226,6 +226,11 @@ func (p *pvcResizer) isVolumeExpansionSupported(storageClassName string) (bool,
 
 // patchPVCs patches PVCs filtered by selector and prefix.
 func (p *pvcResizer) patchPVCs(ns string, selector labels.Selector, pvcQuantityInSpec map[string]resource.Quantity) error {
+	if p.deps.StorageClassLister == nil {
```
DanielZhangQD, line 229: What about we just skip the expansion check here https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/pvc_resizer.go#L260-L267, so that if the sc supports expansion, they can still resize the PVC and just return an error after the patch as we do now if not support.

charts/tidb-operator/templates/controller-manager-deployment.yaml
DanielZhangQD: Sorry, I should go through all the codes before submitting this comment, so the logic is:

- For `clusterscoped: true`, no need to set these three fields
- For `clusterscoped: true`, users have to set the fields explicitly, otherwise, take them as false

In this case, the code here is good.

DanielZhangQD: We have to cover the case that if users ever created a tidb-operator with `clusterscoped: false` with cluster-admin privilege, we have to make sure that they can upgrade TiDB Operator successfully and it works as expected (the configurations listed here need to be configured explicitly).
We also have to make it clear in the release notes.

examples/namespaced/rbac.yaml
DanielZhangQD: ok.

pkg/manager/member/pvc_resizer.go
csuzhangxc: 👍 fix in https://github.com/pingcap/tidb-operator/pull/3896/commits/956bf6118808e4ed6e4c559b3aa6d10d0c4e1634

charts/tidb-operator/templates/controller-manager-deployment.yaml
csuzhangxc: added some comments and updated the release notes in PR description.

csuzhangxc:
@DanielZhangQD PTAL again

DanielZhangQD:
The following action is only necessary if the user ever created a non-cluster scoped TiDB Operator, right?
```
ACTION REQUIRED: when upgrading from a previous version of TiDB Operator, you need to comment `clusterPermissions` out or set all fields (`clusterPermissions.nodes`, `clusterPermissions.persistentvolumes` and `clusterPermissions.storageclasses`) to `true` under `clusterPermissions` explicitly.
```

charts/tidb-operator/templates/controller-manager-deployment.yaml
DanielZhangQD: After further thought, I think we still have to set the default value (even if not set) to true to keep backward compatibility because, for all the TiDB Operators installed till this PR, the values for the three fields are true.

charts/tidb-operator/values.yaml
```diff
@@ -1,9 +1,19 @@
...
-# Also see rbac.create and controllerManager.serviceAccount
+# Also see rbac.create, controllerManager.serviceAccount, scheduler.create and clusterPermissions.
 clusterScoped: true
 
+# clusterPermissions are some cluster scoped permissions that will be used even if `clusterScoped: false`.
+# NOTE: `clusterPermissions: {}` means `false` for all following permissions,
+# but not defined `clusterPermissions` (commented out) means `true` for all following permissions for back compatibility.
+# when upgrading from a version of TiDB Operator without `clusterPermissions` config, you need to
+# comment `clusterPermissions` out or set all fields to `true` under `clusterPermissions` explicitly.
+clusterPermissions: {}
```
DanielZhangQD, line 12: Should we move this under the `controllerManager` section?

charts/tidb-operator/values.yaml
```diff
@@ -1,9 +1,19 @@
 # Default values for tidb-operator
 
 # clusterScoped is whether tidb-operator should manage kubernetes cluster wide tidb clusters
-# Also see rbac.create and controllerManager.serviceAccount
+# Also see rbac.create, controllerManager.serviceAccount, scheduler.create and clusterPermissions.
 clusterScoped: true
 
+# clusterPermissions are some cluster scoped permissions that will be used even if `clusterScoped: false`.
+# NOTE: `clusterPermissions: {}` means `false` for all following permissions,
+# but not defined `clusterPermissions` (commented out) means `true` for all following permissions for back compatibility.
```
DanielZhangQD, line 9: I think we can have the same logic for these two cases, no need to differentiate them.

csuzhangxc: both of them mean `true` for all?

csuzhangxc:
> The following action is only necessary if the user ever created a non-cluster scoped TiDB Operator, right?
> 
> ```
> ACTION REQUIRED: when upgrading from a previous version of TiDB Operator, you need to comment `clusterPermissions` out or set all fields (`clusterPermissions.nodes`, `clusterPermissions.persistentvolumes` and `clusterPermissions.storageclasses`) to `true` under `clusterPermissions` explicitly.
> ```

removed from release notes.

PTAL again

pkg/manager/member/pvc_cleaner.go
```diff
@@ -149,21 +149,25 @@ func (c *realPVCCleaner) reclaimPV(meta metav1.Object) (map[string]string, error
...
+			if pv.Spec.PersistentVolumeReclaimPolicy != corev1.PersistentVolumeReclaimDelete {
+				err := c.deps.PVControl.PatchPVReclaimPolicy(runtimeMeta, pv, corev1.PersistentVolumeReclaimDelete)
+				if err != nil {
+					return skipReason, fmt.Errorf("%s %s/%s patch pv %s to %s failed, err: %v", clusterType, ns, metaName, pvName, corev1.PersistentVolumeReclaimDelete, err)
+				}
+				klog.Infof("%s %s/%s patch pv %s to policy %s success", clusterType, ns, metaName, pvName, corev1.PersistentVolumeReclaimDelete)
 			}
-			klog.Infof("%s %s/%s patch pv %s to policy %s success", clusterType, ns, metaName, pvName, corev1.PersistentVolumeReclaimDelete)
+		} else {
+			klog.V(4).Infof("persistent volumes lister is unavailable, skip reclaiming policy for %s. this may be caused by no relevant permissions", pvName)
```
DanielZhangQD, line 170:
```suggestion
			klog.V(4).Infof("Persistent volumes lister is unavailable, skip updating the reclaim policy for %s. This may be caused by no relevant permissions", pvName)
```

pkg/manager/member/pvc_resizer.go
```diff
@@ -257,13 +257,18 @@ func (p *pvcResizer) patchPVCs(ns string, selector labels.Selector, pvcQuantityI
...
+				if err != nil {
+					return err
+				}
+				if !volumeExpansionSupported {
+					klog.Warningf("Storage Class %q used by PVC %s/%s does not support volume expansion, skipped", *pvc.Spec.StorageClassName, pvc.Namespace, pvc.Name)
+					continue
+				}
+			} else {
+				klog.V(4).Infof("storage classes lister is unavailable, skip checking volume expansion support for PVC %s with storage class %s. this may be caused by no relevant permissions",
+					pvc.Name, *pvc.Spec.StorageClassName)
```
DanielZhangQD, line 271:
```suggestion
				klog.V(4).Infof("Storage classes lister is unavailable, skip checking volume expansion support for PVC %s/%s with storage class %s. This may be caused by no relevant permissions",
					pvc.Namespace, pvc.Name, *pvc.Spec.StorageClassName)
```

pkg/manager/member/tiflash_member_manager.go
```diff
@@ -728,6 +728,11 @@ func (m *tiflashMemberManager) getTiFlashStore(store *pdapi.StoreInfo) *v1alpha1
 }
 
 func (m *tiflashMemberManager) setStoreLabelsForTiFlash(tc *v1alpha1.TidbCluster) (int, error) {
+	if m.deps.NodeLister == nil {
+		klog.V(4).Infof("node lister is unavailable, skip setting store labels for tiflash. this may be caused by no relevant permissions")
```
DanielZhangQD, line 732:
```suggestion
		klog.V(4).Infof("Node lister is unavailable, skip setting store labels for TiFlash of TiDB cluster %s/%s. This may be caused by no relevant permissions", tc.Namespace, tc.Name)
```

pkg/manager/member/tikv_member_manager.go
```diff
@@ -807,6 +807,11 @@ func getTiKVStore(store *pdapi.StoreInfo) *v1alpha1.TiKVStore {
...
+		klog.V(4).Infof("node lister is unavailable, skip setting store labels for tikv. this may be caused by no relevant permissions")
```
DanielZhangQD, line 811: ditto.

pkg/manager/meta/meta_manager.go
```diff
@@ -73,6 +73,11 @@ func (m *metaManager) Sync(tc *v1alpha1.TidbCluster) error {
 			if pvc.Spec.VolumeName == "" {
 				continue
 			}
+
+			if m.deps.PVLister == nil {
+				klog.V(4).Infof("persistent volumes lister is unavailable, skip updating meta info for %s. this may be caused by no relevant permissions", pvc.Spec.VolumeName)
```
DanielZhangQD, line 78:
```suggestion
				klog.V(4).Infof("Persistent volumes lister is unavailable, skip updating meta info for %s. This may be caused by no relevant permissions", pvc.Spec.VolumeName)
```

pkg/manager/meta/reclaim_policy_manager.go
```diff
@@ -50,6 +51,11 @@ func (m *reclaimPolicyManager) SyncDM(dc *v1alpha1.DMCluster) error {
 }
 
 func (m *reclaimPolicyManager) sync(kind string, obj runtime.Object, isPVReclaimEnabled bool, policy corev1.PersistentVolumeReclaimPolicy) error {
+	if m.deps.PVLister == nil {
+		klog.V(4).Infof("persistent volumes lister is unavailable, skip syncing reclaim policy for %s. this may be caused by no relevant permissions", kind)
```
DanielZhangQD, line 55:
```suggestion
		klog.V(4).Infof("Persistent volumes lister is unavailable, skip syncing reclaim policy for %s. This may be caused by no relevant permissions", kind)
```

csuzhangxc:
@DanielZhangQD addressed PTAL again

DanielZhangQD (approved):
LGTM

handlerww (approved):
LGTM

PR #3918, opened 2021-04-14
[WIP] Fast upgrade tikv pod

### LINKED ISSUE #3902: No need to wait tikv transfer leader timeout while upgrading tidb cluster with single tikv node
## Feature Request

It will take 10 minutes to wait tikv leader to transfer. Although there is only one tikv node. It's not necessary.

Upgrade the tikv directly if there is only one tikv

shonge:

Close #3902

When upgrade TiKV and stores number are less than 2, tidb-operator will update the tikv pod immediately.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [x] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fast upgrade TiKV when tikv replicas number are less than 2
```


shonge:
@DanielZhangQD  PTAL

pkg/manager/member/tikv_upgrader.go
```diff
@@ -209,6 +209,10 @@ func (u *tikvUpgrader) readyToUpgrade(upgradePod *corev1.Pod, tc *v1alpha1.TidbC
 			klog.Errorf("parse annotation:[%s] to time failed.", EvictLeaderBeginTime)
 			return false
 		}
+		if tc.Status.TiKV.StatefulSet.CurrentReplicas < 2 {
+			klog.Infof("TiKV stores are less than 2, skip waiting to evict region leader for Pod %s/%s", upgradePod.Namespace, upgradePod.Name)
+			return true
+		}
```
csuzhangxc, line 215: Can we move this block before L206?

csuzhangxc:
BTW, if we have more TiKV pods but with only `1` replica in PD config, can we also fast upgrade TiKV too?

pkg/manager/member/tikv_upgrader.go
DanielZhangQD: Do not use `CurrentReplicas` which is the number of Pods created by the StatefulSet controller from the StatefulSet version indicated by currentRevision. Here is an example during upgrade:
```
status:
  collisionCount: 0
  currentReplicas: 1
  currentRevision: t-tikv-5b4cd9d4df
  observedGeneration: 42
  readyReplicas: 2
  replicas: 3
  updateRevision: t-tikv-8444b66dc8
  updatedReplicas: 2
```

DanielZhangQD: BTW, should we move this block to the beginning of this function?

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -205,7 +205,7 @@ func (pc *PodAdmissionControl) admitDeleteUpTiKVPodDuringUpgrading(payload *admi
 		}
 	}
 
-	if !isTiKVReadyToUpgrade(payload.pod, store, tc.TiKVEvictLeaderTimeout()) {
+	if !isTiKVReadyToUpgrade(payload.pod, store, stores, tc.TiKVEvictLeaderTimeout()) {
```
DanielZhangQD, line 208: What about checking the replicas as you're doing in the tikv_upgrader.go and only execute L193-L212 if replicas >= 2?

pkg/manager/member/tikv_upgrader.go
DanielZhangQD: Refer to `tc.TiKVStsActualReplicas()`

shonge:
> BTW, if we have more TiKV pods but with only `1` replica in PD config, can we also fast upgrade TiKV too?

I think we can fast upgrade TiKV when tc.spec.tikv.replicas=1 and tc.status.tikv.statefulSet.replicas more than 1.
We can transfer leaders to the smallest ordinal of tikv pods and waiting to evict leaders.

pkg/manager/member/tikv_upgrader.go
shonge: update

pkg/webhook/pod/tikv_deleter.go
shonge: update

csuzhangxc (approved):
LGTM

pkg/manager/member/tikv_upgrader.go
```diff
@@ -190,6 +190,11 @@ func (u *tikvUpgrader) upgradeTiKVPod(tc *v1alpha1.TidbCluster, ordinal int32, n
 func (u *tikvUpgrader) readyToUpgrade(upgradePod *corev1.Pod, tc *v1alpha1.TidbCluster) bool {
 	evictLeaderTimeout := tc.TiKVEvictLeaderTimeout()
 	tlsEnabled := tc.IsTLSClusterEnabled()
+	if tc.TiKVStsActualReplicas() < 2 {
```
DanielZhangQD, line 193: What about we checking the `len(podOrdinals)` in https://github.com/pingcap/tidb-operator/blob/86ea694ef0/pkg/manager/member/tikv_upgrader.go#L95 and just update the partition and return if the length < 2?

pkg/webhook/pod/tikv_deleter.go
```diff
@@ -176,20 +176,26 @@ func (pc *PodAdmissionControl) admitDeleteUpTiKVPod(payload *admitPayload, store
 			klog.Infof("%s[%s/%s]'s tikv pod[%s/%s] failed to delete,%v", controllerKind, namespace, controllerName, namespace, name, err)
 			return util.ARFail(err)
 		}
-		return pc.admitDeleteUpTiKVPodDuringUpgrading(payload, store)
+		return pc.admitDeleteUpTiKVPodDuringUpgrading(payload, store, storesInfo)
```
DanielZhangQD, line 179: Where the `storesInfo` is used now?

DanielZhangQD:
Close via #4091.

Issue #3697, opened 2021-01-12
Operator should remove members from the Dashboard topology when scaling in


kolbe:
## Feature Request

**Is your feature request related to a problem? Please describe:**

At present, when a component is scaled in, it still shows up in TiDB Dashboard. Dashboard expects that it is the deployment tool's responsibility to remove entries for scaled-in instances.

**Describe the feature you'd like:**

When nodes are removed from the cluster, TiDB Operator should ensure that they're gracefully removed from TiDB Dashboard.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

DanielZhangQD:
Only handle the TiDB scaling in is good now, for the other components, e.g. PD, TiKV, TiFlash, please help verify that the members/stores scaled in will disappear from the TiDB Dashboard.

Issue #3058, opened 2020-08-03
TiDB Operator OOM


DanielZhangQD:
## Bug Report

**What version of Kubernetes are you using?**

**What version of TiDB Operator are you using?**

v1.1.3
**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

**What's the status of the TiDB cluster pods?**

**What did you do?**

**What did you expect to see?**
TiDB Operator works normally
**What did you see instead?**
The memory usage of TiDB Operator increases slowly to several GB and eventually, gets OOM killed.

DanielZhangQD:
![image](https://user-images.githubusercontent.com/36026334/89185802-66ec5e00-d5cd-11ea-9abf-8032d56e6945.png)

DanielZhangQD:
Heap profile:
```
File: tidb-controller-manager
Type: inuse_space
Time: Aug 3, 2020 at 2:05pm (CEST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 531.32MB, 81.41% of 652.66MB total
Dropped 245 nodes (cum <= 3.26MB)
Showing top 10 nodes out of 146
      flat  flat%   sum%        cum   cum%
  301.60MB 46.21% 46.21%   301.60MB 46.21%  go.uber.org/zap/zapcore.newCounters
   45.21MB  6.93% 53.14%    45.21MB  6.93%  google.golang.org/grpc/internal/transport.newBufWriter
   39.87MB  6.11% 59.25%    39.87MB  6.11%  bufio.NewReaderSize
   39.19MB  6.00% 65.25%    39.19MB  6.00%  bytes.makeSlice
   26.02MB  3.99% 69.24%    55.03MB  8.43%  crypto/x509.parseCertificate
   20.03MB  3.07% 72.31%    20.03MB  3.07%  encoding/pem.Decode
   17.29MB  2.65% 74.96%    17.29MB  2.65%  reflect.unsafe_NewArray
   15.55MB  2.38% 77.34%    15.55MB  2.38%  crypto/tls.(*halfConn).encrypt
      14MB  2.15% 79.49%       14MB  2.15%  math/big.nat.make
   12.55MB  1.92% 81.41%    12.55MB  1.92%  bufio.NewWriterSize
```

DanielZhangQD:
After several hours:
```
(pprof) top
Showing nodes accounting for 1793.98MB, 84.91% of 2112.83MB total
[...]

PR #3929, opened 2021-04-20
delete orphan tidb info key to make dashboard happy

july2993:

fix https://github.com/pingcap/tidb-operator/issues/3697

for every orphan  `/topology/tidb/{addr}/info` we delete it if the instance is not expected to be exist

for PD, TiKV, TiFlash, no need to do this, it will get the logical instances from pd

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- deploy a cluster with 3 tidb
- update replica to be 1 and check we can see only one in dashboard
- update tc to use a wrong image and delete the tidb pod(just to make tidb failed to start), and check tidb instance still exist at dashboard but are not connected

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Delete tidb info key to make the instance disappear at dashboard after scale in
```


pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,83 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
 	return nil
 }
 
+// ref https://github.com/pingcap/tidb/blob/36b04d1aa01db722b3f07af759168c6b8da33801/domain/infosync/info.go#L72
```
csuzhangxc, line 137: How about putting this ref above L34?

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,83 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+
+	for _, kv := range kvs {
+		key := kv.Key
+		klog.V(4).Infof("get tidb info key %s", key)
+
+		if strings.HasSuffix(key, "ttl") {
+			ttls[key] = kv
+		}
+
+		if strings.HasSuffix(key, "info") {
```
csuzhangxc, line 155: use `else if` to combine with the previous `if`?

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,83 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+			infos[key] = kv
+		}
+	}
+
+	for key, kv := range infos {
+		if _, ok := ttls[strings.Replace(key, "/info", "/ttl", -1)]; ok {
```
csuzhangxc, line 161: how about only replacing the suffix?

pkg/manager/member/tidb_failover.go
```diff
@@ -57,23 +57,30 @@ func (f *tidbFailover) Failover(tc *v1alpha1.TidbCluster) error {
 	maxFailoverCount := *tc.Spec.TiDB.MaxFailoverCount
 	for _, tidbMember := range tc.Status.TiDB.Members {
 		_, exist := tc.Status.TiDB.FailureMembers[tidbMember.Name]
+		if exist {
+			continue
+		}
+
 		deadline := tidbMember.LastTransitionTime.Add(f.deps.CLIConfig.TiDBFailoverPeriod)
-		if !tidbMember.Health && time.Now().After(deadline) && !exist {
+		if !tidbMember.Health && time.Now().After(deadline) {
```
DanielZhangQD, line 65: Move `tidbMember.Health` to the block in L60?

pkg/manager/member/tidbcluster_status_manager.go
july2993: There are no false-positive case by using `ReplaceAll` here and only replacing the suffix need more tedious code, i prefer to just keep `ReplaceAll`

pkg/manager/member/tidbcluster_status_manager.go
july2993: keep here and add more comments.
I like to tell the reader how the two key is write but not just the definition of the prefix

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -135,6 +135,7 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
 }
 
 // ref https://github.com/pingcap/tidb/blob/36b04d1aa01db722b3f07af759168c6b8da33801/domain/infosync/info.go#L72
+// search `TopologyInformationPath` about how theey with 'ttl' and 'info' key is write in that file.
```
csuzhangxc, line 138: typo for `theey`?

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -94,13 +106,10 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
 		return nil
 	}
 	pdEtcdClient, err := m.deps.PDControl.GetPDEtcdClient(pdapi.Namespace(tc.Namespace), tc.Name, tc.IsTLSClusterEnabled())
-
 	if err != nil {
 		return err
 	}
-	if tc.IsTLSClusterEnabled() {
```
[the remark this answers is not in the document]
july2993: I have make it use the some logic in pkg/pdapi/pd_control.go

both tls enable and not enable will just return a reuse client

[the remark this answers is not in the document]
DanielZhangQD: Yes, but with TLS enabled, this will cause memory leak, please refer to https://github.com/pingcap/tidb-operator/issues/3058.

pkg/manager/member/tidbcluster_status_manager.go
july2993: fix in https://github.com/pingcap/tidb-operator/pull/3929/commits/08a6ee483dade1d8947c187fa8f9ec1db921a17b

pkg/manager/member/tidbcluster_status_manager.go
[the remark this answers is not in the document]
july2993: we *will not* create a new client every time when calling `GetPDEtcdClient` now.
is there any reason for creating a new client every time?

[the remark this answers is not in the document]
DanielZhangQD: OK, sorry that I did not review the code change in that file yet. For the current behavior to create a new client each time, it's to cover the case that the secret containing the certs may be rotated to avoid expiration, in which case, the client may fail and we have to restart the tidb-controller-manager container.

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,82 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
 	return nil
 }
 
+// ref https://github.com/pingcap/tidb/blob/36b04d1aa01db722b3f07af759168c6b8da33801/domain/infosync/info.go#L72
+// search `TopologyInformationPath` about how the key with 'ttl' and 'info' suffix is write in that file.
```
DanielZhangQD, line 138:
```suggestion
// search `TopologyInformationPath` about how the key with 'ttl' and 'info' suffix is updated in that file.
```

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,82 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+func getNoLeastTidbInfoKey(ctx context.Context, client pdapi.PDEtcdClient) (noleast []*pdapi.KeyValue, err error) {
```
DanielZhangQD, line 139:
```suggestion
func getStaleTidbInfoKey(ctx context.Context, client pdapi.PDEtcdClient) (staleKeys []*pdapi.KeyValue, err error) {
```

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,82 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+		return err
+	}
+
+	kvs, err := getNoLeastTidbInfoKey(context.TODO(), pdEtcdClient)
+	if err != nil {
+		return err
+	}
+
+	expectNames := make(map[string]struct{})
+	for ordinal := range tc.TiDBStsDesiredOrdinals(true) {
```
DanielZhangQD, line 192:
```suggestion
	for ordinal := range tc.TiDBStsDesiredOrdinals(false) {
```

DanielZhangQD: The Pods created due to failover should also be valid ones.

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,82 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+	if err != nil {
+		return nil, perrors.AddStack(err)
+	}
+
+	infos := make(map[string]*pdapi.KeyValue)
+	ttls := make(map[string]*pdapi.KeyValue)
+
+	for _, kv := range kvs {
+		key := kv.Key
+		klog.V(4).Infof("get tidb info key %s", key)
```
DanielZhangQD, line 150:
```suggestion
		klog.V(4).Infof("Get TiDB info key %s", key)
```

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +134,82 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+	for ordinal := range tc.TiDBStsDesiredOrdinals(true) {
+		name := fmt.Sprintf("%s-%d", controller.TiDBMemberName(tc.GetName()), ordinal)
+		expectNames[name] = struct{}{}
+	}
+
+	for _, kv := range kvs {
+		name := getTidbName(kv.Key)
+
+		if _, ok := expectNames[name]; !ok {
+			klog.V(2).Infof("delete tidb info key %s", kv.Key)
```
DanielZhangQD, line 201:
```suggestion
			klog.V(2).Infof("Delete TiDB info key %s", kv.Key)
```

pkg/manager/member/tidbcluster_status_manager.go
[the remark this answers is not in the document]
july2993: https://github.com/pingcap/tidb-operator/pull/3929/commits/9fd4f25de573f15f57385c67c192562f7e6fcb4f

csuzhangxc (approved):
LGTM

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +136,84 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
 	return nil
 }
 
+// ref https://github.com/pingcap/tidb/blob/36b04d1aa01db722b3f07af759168c6b8da33801/domain/infosync/info.go#L72
+// search `TopologyInformationPath` about how the key with 'ttl' and 'info' suffix is updated in that file.
+func getStaleTidbInfoKey(ctx context.Context, client pdapi.PDEtcdClient) (staleKeys []*pdapi.KeyValue, err error) {
+	kvs, err := client.Get(tidbPrefix, true /*prefix*/)
```
DanielZhangQD, line 142: The key we get here should be for this cluster only:

- /topology/tidb/basic-tidb-0.basic-tidb-peer.tidb-cluster.svc:4000/info (to match the TiDB sts name and namespace)
- /topology/tidb/basic-tidb-0.basic-tidb-peer.tidb-cluster.svc.cluster.local:4000/info (for across Kubernetes deployment, to match the TiDB sts name, namespace, and cluster domain)

pkg/pdapi/pd_control.go
```diff
@@ -41,13 +41,23 @@ type PDControlInterface interface {
...
+	mutex     sync.Mutex
+	pdClients map[string]PDClient
+
 	etcdmutex     sync.Mutex
-	kubeCli       kubernetes.Interface
-	pdClients     map[string]PDClient
 	pdEtcdClients map[string]PDEtcdClient
 }
 
+type noOpClose struct {
```
DanielZhangQD, line 53:
```suggestion
type nonTLSClient struct {
```

pkg/manager/member/tidbcluster_status_manager.go
DanielZhangQD: Refer to https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_member_manager.go#L724.

july2993: https://github.com/pingcap/tidb-operator/pull/3929/commits/42ea6a6bb8206d76a796ccdfbfb28a9734811747

pkg/pdapi/pd_control.go
july2993: prefer to keep `noOpClose`

pkg/manager/member/tidbcluster_status_manager.go
DanielZhangQD: Suggest using `/topology/tidb/basic-tidb` as the prefix here so that we can get as least number of keys as we can.

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +136,92 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+
+func (m *TidbClusterStatusManager) syncTiDBInfoKey(tc *v1alpha1.TidbCluster) error {
+	pdEtcdClient, err := m.deps.PDControl.GetPDEtcdClient(pdapi.Namespace(tc.Namespace), tc.Name, tc.IsTLSClusterEnabled())
+	if err != nil {
+		return err
+	}
+
+	defer pdEtcdClient.Close()
+
+	kvs, err := getStaleTidbInfoKey(context.TODO(), pdEtcdClient)
```
DanielZhangQD, line 190: The kvs here should only include the keys for this TiDB cluster, otherwise, the following keys may be deleted incorrectly in L214:

- The TiDB address of the TiDB cluster with the same name in another namespace
- The TiDB address of the TiDB cluster with the same name and namespace but with a different cluster domain when deployed across Kubernetes.

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +136,92 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+		staleKeys = append(staleKeys, kv)
+	}
+
+	return
+}
+
+// /topology/tidb/basic-tidb-0.basic-tidb-peer.tidb-cluster.svc:4000/info -> basic-tidb-0.basic-tidb-peer.tidb-cluster.svc
+func getTidbAddr(key string) (addr string) {
+	key = strings.TrimPrefix(key, tidbPrefix)
+	key = strings.TrimPrefix(key, "/")
```
DanielZhangQD, line 175: What about changing `tidbPrefix = "/topology/tidb/"`?

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -125,6 +136,92 @@ func (m *TidbClusterStatusManager) syncDashboardMetricStorage(tc *v1alpha1.TidbC
...
+	tmp := strings.Split(key, ":")
+
+	addr = tmp[0]
+	return
+}
+
+func (m *TidbClusterStatusManager) syncTiDBInfoKey(tc *v1alpha1.TidbCluster) error {
+	pdEtcdClient, err := m.deps.PDControl.GetPDEtcdClient(pdapi.Namespace(tc.Namespace), tc.Name, tc.IsTLSClusterEnabled())
```
DanielZhangQD, line 183: We have to consider the heterogeneous cluster case, refer to https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tikv_upgrader.go#L267。

july2993: https://github.com/pingcap/tidb-operator/pull/3929/commits/e2671c566ef69b095de88f4b9c2b3d7bbb01845f
tested

pkg/manager/member/tidbcluster_status_manager.go
july2993: https://github.com/pingcap/tidb-operator/pull/3929/commits/0c142d5247143fac34492c7b1925413e42370282

july2993: > Suggest using `/topology/tidb/basic-tidb` as the prefix here so that we can get as least number of keys as we can.

keep fetch all now since it will not come up with too much overhead and we can know all instance if need

pkg/manager/member/tidbcluster_status_manager.go
july2993: prefer to keep now

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -28,6 +32,9 @@ const (
 	prometheusComponent = "prometheus"
 	grafanaComponent    = "grafana"
 	componentPrefix     = "/topology"
+	tidbPrefix          = "/topology/tidb"
+
+	tidbAddrPattern = `%s-tidb-\d+\.%s-tikv-peer\.%s\.svc%s\:\d+`
```
DanielZhangQD, line 37:
```suggestion
	tidbAddrPattern = `%s-tidb-\d+\.%s-tidb-peer\.%s\.svc%s\:\d+`
```

DanielZhangQD (approved):
LGTM

csuzhangxc (approved):
LGTM

pkg/manager/member/tidbcluster_status_manager_test.go
```diff
@@ -24,6 +26,18 @@ import (
...
 
+func TestTidbPattern(t *testing.T) {
+	name := "basic"
+	ns := "tidb-cluster"
+	reg := fmt.Sprintf(tidbAddrPattern, name, name, ns, controller.FormatClusterDomainForRegex(""))
+	pattern, err := regexp.Compile(reg)
+	g := NewGomegaWithT(t)
+	g.Expect(err).Should(BeNil())
+
+	m := pattern.Match([]byte("basic-tidb-0.basic-tidb-peer.tidb-cluster.svc"))
```
DanielZhangQD, line 37: Please add more cases, e.g. with cluster domain, heterogeneous cluster, and across Kubernetes deployment, etc.

pkg/manager/member/tidbcluster_status_manager.go
```diff
@@ -34,7 +34,7 @@ const (
 	componentPrefix     = "/topology"
 	tidbPrefix          = "/topology/tidb"
 
-	tidbAddrPattern = `%s-tidb-\d+\.%s-tidb-peer\.%s\.svc%s\:\d+`
+	tidbAddrPattern = `%s-tidb-\d+\.%s-tidb-peer\.%s\.svc%s`
```
DanielZhangQD, line 37: Why remove `\d+` in the end? In this case, the addresses in the other Kubernetes clusters with the same name and namespace will also be matched and deleted, right?

july2993: https://github.com/pingcap/tidb-operator/pull/3929/commits/d570f0f7f4e0f804fd66c08049aaeb91ca939856

pkg/manager/member/tidbcluster_status_manager_test.go
july2993: https://github.com/pingcap/tidb-operator/pull/3929/commits/d570f0f7f4e0f804fd66c08049aaeb91ca939856

PR #3940, opened 2021-04-23
fix the method counting pd members when scaling out pd 

cvvz:
fix the method counting pd members when scaling out pd.

The actual count of pd members should come from `pd api`, or `tc.status.pd.members`, not the replicas of pod, because pod may not run normally

### Code changes

- [x] Has Go code change

### Tests

- [x] Manual test 

1. Create a tc with 3 PDs, 2 runs normally, 1 is Error
2. Scale out pd, the pd can scale out sccessfully without log: 
`E0423 19:41:41.260220       1 tidb_cluster_controller.go:126] TidbCluster: xxx, sync failed TidbCluster: xxx's pd 2/3 is ready, can't scale out now, requeuing`.

### Release Notes

```release-note
fix the method counting pd members when scaling out pd.
```


july2993 (approved):

DanielZhangQD:
@cvvz Please fix the CI.

pkg/manager/member/pd_scaler.go
```diff
@@ -74,15 +74,15 @@ func (s *pdScaler) ScaleOut(meta metav1.Object, oldSet *apps.StatefulSet, newSet
...
+	totalCount := len(tc.Status.PD.Members)
 	podOrdinals := helper.GetPodOrdinals(*oldSet.Spec.Replicas, oldSet).List()
 	for _, i := range podOrdinals {
 		targetPdName := PdName(tcName, i, tc.Namespace, tc.Spec.ClusterDomain)
 		if member, ok := tc.Status.PD.Members[targetPdName]; ok && member.Health {
 			healthCount++
 		}
 	}
-	if healthCount < int(totalCount) {
+	if healthCount < totalCount {
```
DanielZhangQD, line 85: Suggest removing the check here.
Why do we need to check the cluster status during scale out?

DanielZhangQD: This is introduced in https://github.com/pingcap/tidb-operator/pull/142 before the discovery service is introduced in https://github.com/pingcap/tidb-operator/pull/262.

handlerww: What's the problem in scaling when there are some unhealthy members for PD?
I think we can keep the check here and make a survey on this situation.

PR #3962, opened 2021-05-06
[WIP] ensure crds exist before the controller starts

### LINKED ISSUE #3939: Is it needed to make sure  CRDs exist before starting the controller ?
https://github.com/pingcap/tidb-operator/blob/c76d162b88fa02047193c19480cf5e1044d50f34/cmd/controller-manager/main.go#L183

zhucan:
Signed-off-by: zhucan <zhucan.k8s@gmail.com>

feat: ensure crds exists before controller start
Closes #3939 

Only crd installed that the controller can start

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/cmd/controller-manager/main.go
+++ b/cmd/controller-manager/main.go
@@ -167,9 +167,11 @@ func main() {
 			deps.KubeInformerFactory,
 			deps.LabelFilterKubeInformerFactory,
 		}
+		waitCtx, cancel := context.WithTimeout(context.Background(), cliCfg.CacheSyncDuration)
+		defer cancel()
 		for _, f := range informerFactories {
 			f.Start(ctx.Done())
-			for v, synced := range f.WaitForCacheSync(wait.NeverStop) {
+			for v, synced := range f.WaitForCacheSync(waitCtx.Done()) {
 				if !synced {
 					klog.Fatalf("error syncing informer for %v", v)
 				}
--- a/pkg/controller/dependences.go
+++ b/pkg/controller/dependences.go
@@ -64,6 +64,7 @@ type CLIConfig struct {
 	RenewDeadline         time.Duration
 	RetryPeriod           time.Duration
 	WaitDuration          time.Duration
+	CacheSyncDuration     time.Duration
 	// ResyncDuration is the resync time of informer
 	ResyncDuration time.Duration
 	// Defines whether tidb operator run in test mode, test mode is
@@ -96,6 +97,7 @@ func DefaultCLIConfig() *CLIConfig {
 		RetryPeriod:            2 * time.Second,
 		WaitDuration:           5 * time.Second,
 		ResyncDuration:         30 * time.Second,
+		CacheSyncDuration:      30 * time.Second,
 		TiDBBackupManagerImage: "pingcap/tidb-backup-manager:latest",
 		TiDBDiscoveryImage:     "pingcap/tidb-operator:latest",
 		Selector:               "",
@@ -116,6 +118,7 @@ func (c *CLIConfig) AddFlag(_ *flag.FlagSet) {
 	flag.DurationVar(&c.MasterFailoverPeriod, "dm-master-failover-period", c.MasterFailoverPeriod, "dm-master failover period")
 	flag.DurationVar(&c.WorkerFailoverPeriod, "dm-worker-failover-period", c.WorkerFailoverPeriod, "dm-worker failover period")
 	flag.DurationVar(&c.ResyncDuration, "resync-duration", c.ResyncDuration, "Resync time of informer")
+	flag.DurationVar(&c.CacheSyncDuration, "cache-sync-duration", c.CacheSyncDuration, "Timeout of the cache sync")
 	flag.BoolVar(&c.TestMode, "test-mode", false, "whether tidb-operator run in test mode")
 	flag.StringVar(&c.TiDBBackupManagerImage, "tidb-backup-manager-image", c.TiDBBackupManagerImage, "The image of backup manager tool")
 	// TODO: actually we just want to use the same image with tidb-controller-manager, but DownwardAPI cannot get image ID, see if there is any better solution


cmd/controller-manager/main.go
```diff
@@ -54,6 +55,73 @@ import (
...
 )
 
+const defaultNamespace = "kube-system"
+
+// ensureCRDExists to ensure CRD exists before start
+func ensureCRDExists(kubeClient kubernetes.Interface, clientSet *versioned.Clientset) error {
+	condition := func() (bool, error) {
+		var err error
+		_, err = kubeClient.CoreV1().Namespaces().Get(defaultNamespace, metav1.GetOptions{})
+		if err == nil {
```
csuzhangxc, line 65: How about `return` as early as we can (`err != nil`)?

cmd/controller-manager/main.go
```diff
@@ -54,6 +55,73 @@ import (
...
+			// Only namespace "kube-system" exists, to list resources
+			_, err = clientSet.PingcapV1alpha1().Backups(defaultNamespace).List(metav1.ListOptions{})
```
csuzhangxc, line 67: Can we check CRD scheme instead of listing resources here?

zhucan: I was also thinking of checking the CRD directly but that will require new rbacs.

csuzhangxc:
BTW, if the user deploy tidb-operator with namespace-scoped permissions (see #3896), this feature may block controller-manager to startup, right? if so, do we need to skip this check sometime?

cmd/controller-manager/main.go
csuzhangxc: OK. you may add some comments to explain this. 😂

cmd/controller-manager/main.go
zhucan: other branch that I had done

zhucan:
> BTW, if the user deploy tidb-operator with namespace-scoped permissions (see #3896), this feature may block controller-manager to startup, right? if so, do we need to skip this check sometime or only check with current installed namespace not `kube-system`)?

no, the purpose of the check is that the endpoint exists, it doesn't matter if there's nothing in the namespace.

cmd/controller-manager/main.go
zhucan: good suggestion.

csuzhangxc:
> > BTW, if the user deploy tidb-operator with namespace-scoped permissions (see #3896), this feature may block controller-manager to startup, right? if so, do we need to skip this check sometime or only check with current installed namespace not `kube-system`)?
> 
> no, the purpose of the check is that the endpoint exists, it doesn't matter if there's nothing in the namespace.

but check with `kube-system` may return errors if no permissions granted, right?

cmd/controller-manager/main.go
csuzhangxc: I know. I mean, fewer indentations and handling errors close to their returned may better. but this is just my preference and you can keep this.

zhucan:
> > > BTW, if the user deploy tidb-operator with namespace-scoped permissions (see #3896), this feature may block controller-manager to startup, right? if so, do we need to skip this check sometime or only check with current installed namespace not `kube-system`)?
> > 
> > 
> > no, the purpose of the check is that the endpoint exists, it doesn't matter if there's nothing in the namespace.
> 
> but check with `kube-system` may return errors if no permissions granted, right?

Maybe list the resources under the namespace that set to the env is ok?

csuzhangxc:
> > > > BTW, if the user deploy tidb-operator with namespace-scoped permissions (see #3896), this feature may block controller-manager to startup, right? if so, do we need to skip this check sometime or only check with current installed namespace not `kube-system`)?
> > > 
> > > 
> > > no, the purpose of the check is that the endpoint exists, it doesn't matter if there's nothing in the namespace.
> > 
> > 
> > but check with `kube-system` may return errors if no permissions granted, right?
> 
> Maybe list the resources under the namespace that set to the env is ok?

I think it's OK.

zhucan:
> > > > > BTW, if the user deploy tidb-operator with namespace-scoped permissions (see #3896), this feature may block controller-manager to startup, right? if so, do we need to skip this check sometime or only check with current installed namespace not `kube-system`)?
> > > > 
> > > > 
> > > > no, the purpose of the check is that the endpoint exists, it doesn't matter if there's nothing in the namespace.
> > > 
> > > 
> > > but check with `kube-system` may return errors if no permissions granted, right?
> > 
> > 
> > Maybe list the resources under the namespace that set to the env is ok?
> 
> I think it's OK.

done

cmd/controller-manager/main.go
```diff
@@ -54,6 +55,72 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// ensureCRDExists to ensure CRD exists before start
+func ensureCRDExists(ns string, kubeClient kubernetes.Interface, clientSet *versioned.Clientset) error {
+	condition := func() (bool, error) {
+		var err error
+		_, err = kubeClient.CoreV1().Namespaces().Get(ns, metav1.GetOptions{})
+		if err == nil {
+			// Only namespace "kube-system" exists, to list resources
```
csuzhangxc, line 64: please fix this comment for `kube-system`

zhucan: ok

cmd/controller-manager/main.go
zhucan: I know, It's only to debug with bin file not with container. so when running with bin file, we need to check the namespace whether it exists.

cmd/controller-manager/main.go
```diff
@@ -54,6 +55,72 @@ import (
...
+	condition := func() (bool, error) {
+		var err error
+		_, err = kubeClient.CoreV1().Namespaces().Get(ns, metav1.GetOptions{})
+		if err == nil {
+			// Only namespace(the controller deploy under it) exists, to list resources
+			// checking the CRD directly but that will require new rbacs, the purpose of the check is that the endpoint exists, it doesn't matter if there's nothing in the namespace
+			_, err = clientSet.PingcapV1alpha1().Backups(ns).List(metav1.ListOptions{})
+			if err != nil {
+				klog.Errorf("Failed to list backups with error=%+v", err)
+				return false, err
```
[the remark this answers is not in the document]
zhucan: yes

cmd/controller-manager/main.go
```diff
@@ -54,6 +55,72 @@ import (
...
+			}
+			return true, nil
+		}
+		klog.Errorf("Failed to get namespace %s, error=%+v", ns, err)
+		return false, err
+	}
+	// with a Factor of 1.5 we wait up to 7.5 seconds (the 10th attempt)
+	backoff := wait.Backoff{
+		Duration: 100 * time.Millisecond,
```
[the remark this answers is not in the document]
zhucan: I think no need to wait longer, we had tested on csi snapshottor with same way.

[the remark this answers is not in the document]
DanielZhangQD: Then what's the point of waiting here? We can just exit here with `klog.Fatalf` if some CRDs are missed.

cmd/controller-manager/main.go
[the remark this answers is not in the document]
DanielZhangQD: Please check the comment https://github.com/pingcap/tidb-operator/pull/3962#discussion_r628012339.

cmd/controller-manager/main.go
DanielZhangQD: OK, then please return here is `err!=nil`.

cmd/controller-manager/main.go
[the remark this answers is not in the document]
zhucan: we don't want to the container crash Immediately，if there is bad network,  so we retry some times.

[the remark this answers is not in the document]
DanielZhangQD: If there is a network issue, it will return an error in the check function above, and it will abort immediately, there is no chance to wait.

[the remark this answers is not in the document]
zhucan: If list the resource with errors that return (false, nil), is it ok?  retry sometimes.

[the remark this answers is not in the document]
zhucan: Is it needed to wait a moment when crds not deployed?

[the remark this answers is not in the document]
DanielZhangQD: I think it's OK not to wait here, WDYT? cc @liubog2008 @csuzhangxc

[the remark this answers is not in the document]
csuzhangxc: wait or not wait are both OK for me.

cmd/controller-manager/main.go
zhucan: done

cmd/controller-manager/main.go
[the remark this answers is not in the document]
zhucan: I think no need to crash immediately.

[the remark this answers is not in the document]
liubog2008: I think "just let it crash" is more reasonable. @DanielZhangQD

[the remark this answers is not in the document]
liubog2008: @zhucan Informer cache will wait after starting, no need to wait again.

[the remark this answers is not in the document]
liubog2008: Use `ctx.Done` of timeout context to replace `wait.NeverStop` channel.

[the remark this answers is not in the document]
zhucan: @liubog2008 List the resources throught API server, not informer,  we retry it only wait crd deployed. In general, we prefer retrying with exponential backoff so that we don't potentially overload the API server with too many requests in a short time.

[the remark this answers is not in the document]
zhucan: however we'd like to have a faster startup time, if the container exits quickly and goes into CrashLoopBackOff it'll be restarted every 10s, 20s, 40s and so on, during the first run instead of failing quickly it might be possible that the CRDs are ready after a few seconds and in one of these 10 runs the wait condition might succeed @DanielZhangQD @liubog2008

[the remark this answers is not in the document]
liubog2008: @zhucan I mean informer also will wait before CRD is deployed. It's no need to wait twice at begin.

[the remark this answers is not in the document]
zhucan: @liubog2008 yes, where is the better place to wait? do you have better suggestion?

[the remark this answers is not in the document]
liubog2008: @zhucan See https://github.com/pingcap/tidb-operator/blob/master/cmd/controller-manager/main.go#L181
you can change `wait.NeverStop` to a context created by `context.WithTimeout()`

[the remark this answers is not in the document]
liubog2008: I don't know whether it meets your requirement.

DanielZhangQD:
@zhucan 
Could you please check the comment https://github.com/pingcap/tidb-operator/pull/3962#discussion_r629211134?
It's an elegant solution.

zhucan:
> @zhucan How about changing https://github.com/pingcap/tidb-operator/blob/master/cmd/controller-manager/main.go#L179 like below.
> 
> ```
> waitCtx, cancel := context.WithTimeout(context.Background(), time.Second*30)
> defer cancel()
> for _, f := range informerFactories {
> 	f.Start(ctx.Done())
> 	for v, synced := range f.WaitForCacheSync(waitCtx.Done()) {
> 		if !synced {
> 			klog.Fatalf("error syncing informer for %v", v)
> 		}
> 	}
> }
> ```
> 
> Use context with timeout to wait an arbitrary period.

ok, I will do it.

cmd/controller-manager/main.go
```diff
@@ -114,6 +175,11 @@ func main() {
 		klog.Fatalf("failed to get the generic kube-apiserver client: %v", err)
 	}
 
+	err = ensureCRDExists(ns, kubeCli, cli)
```
liubog2008, line 178: This is not needed anymore

cmd/controller-manager/main.go
```diff
@@ -54,6 +55,66 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+//ensureCRDExists to ensure CRD exists before start
+func ensureCRDExists(ns string, kubeClient kubernetes.Interface, clientSet *versioned.Clientset) error {
```
liubog2008, line 59: not needed

cmd/controller-manager/main.go
zhucan: yes, I overlooked this place

cmd/controller-manager/main.go
```diff
@@ -176,9 +177,11 @@ func main() {
 			deps.KubeInformerFactory,
 			deps.LabelFilterKubeInformerFactory,
 		}
+		waitCtx, cancel := context.WithTimeout(context.Background(), time.Second*30)
```
DanielZhangQD, line 180: @zhucan As you can see that in L176-L178, there are three informer factories including the ones list-watch the Kubernetes native resources, however, for some users, their Kubernetes cluster may have too many resources, in which case, we cannot sync all the resources in 30s, so we'd better add a startup parameter for tidb-controller-manager and make the default value as 30s.
You can refer to the code here https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/dependences.go#L111.
Please also add a parameter in the helm chart https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-operator/values.yaml#L42 in case that users want to customize this value.

zhucan: copy that, you want to make it configurable and default value as 30s?

DanielZhangQD: Yes, thanks!

liubog2008: @zhucan It seems not changed as above comments

zhucan: the"resync-duration" can solve the issue as above comments. you mean I have no modify the yaml ?

liubog2008: Resync duration is not prepared for this case. It is defined for interval of informer cache resync. I think a new option is needed.

liubog2008: If use "resync-duration" in here, other codes using this flag will be affected

Issue #3969, opened 2021-05-09
Rolling update happen after changing configUpdateStrategy from  InPlace to RollingUpdate


cvvz:
## Bug Report

**What version of Kubernetes are you using?**

v1.16.9

**What version of TiDB Operator are you using?**

TiDB Operator Version: version.Info{GitVersion:"v1.1.10-46+c715442c7ad0e1", GitCommit:"c715442c7ad0e115bacae3b5cc27946e13a24ce9", GitTreeState:"clean", BuildDate:"2021-04-26T12:13:47Z", GoVersion:"go1.15.7", Compiler:"gc", Platform:"linux/amd64"}

**What did you do?**

1. Modify configuration when  the `configUpdateStrategy` of cluster is `InPlace`
2. Remove `configUpdateStrategy` , whose default value is `RollingUpdate`

**What did you expect to see?**
The cluster stays normal

**What did you see instead?**
Components whose configuration was modified rolling upgrade because new confimap is generated.

cvvz:
This is because the suffix generated [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/config.go#L79) is not equal to the old one. However, the data of the old one is actually the same as the new one.

PR #3970, opened 2021-05-09
Fix the bug that rolling update may happen after changing `configUpdateStrategy` from `InPlace` to `RollingUpdate`.

cvvz:

Fix the bug that rolling update may happen after changing `configUpdateStrategy` from `InPlace` to `RollingUpdate`.(https://github.com/pingcap/tidb-operator/issues/3969)

When data of configmap is equal but name is different, use the existing one to avoid rolling update.
When data of configmap is different but name is equal, add a special suffix to the name of cm to trigger rolling update. 

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [x] Manual test  
  1. Modify the config while  `configUpdateStrategy` is `InPlace`, then change `configUpdateStrategy` to `RollingUpdate`, the Pod still use the old cm and do not rolling update.
  2. After that, modify the config back to the original value, the component rolling update. 

### Release Notes

```release-note
Fix the bug that rolling update may happen after changing `configUpdateStrategy` from `InPlace` to `RollingUpdate`.
```

### DIFF
--- a/pkg/manager/member/config.go
+++ b/pkg/manager/member/config.go
@@ -14,6 +14,8 @@
 package member
 
 import (
+	"fmt"
+
 	perrors "github.com/pingcap/errors"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/util/toml"
@@ -22,12 +24,17 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 )
 
-func updateConfigMap(old, new *corev1.ConfigMap) error {
+func updateConfigMap(old, new *corev1.ConfigMap) (bool, error) {
 	tomlField := []string{"config-file" /*pd,tikv,tidb */, "pump-config", "config_templ.toml" /*tiflash*/, "proxy_templ.toml" /*tiflash*/}
+	dataEqual := true
 
 	for _, k := range tomlField {
 		oldData, oldOK := old.Data[k]
 		newData, newOK := new.Data[k]
+
+		if oldOK != newOK {
+			dataEqual = false
+		}
 
 		if !oldOK || !newOK {
 			continue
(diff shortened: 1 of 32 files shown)


pkg/manager/member/config.go
```diff
@@ -22,28 +24,35 @@ import (
...
+		}
+
 		if !oldOK || !newOK {
 			continue
 		}
 
 		equal, err := toml.Equal([]byte(oldData), []byte(newData))
 		if err != nil {
-			return perrors.Annotatef(err, "compare %s/%s %s and %s failed", old.Namespace, old.Name, oldData, newData)
+			return dataEqual, perrors.Annotatef(err, "compare %s/%s %s and %s failed", old.Namespace, old.Name, oldData, newData)
```
july2993, line 45:
```suggestion
			return false, perrors.Annotatef(err, "compare %s/%s %s and %s failed", old.Namespace, old.Name, oldData, newData)
```

july2993 (approved):

pkg/manager/member/config.go
```diff
@@ -71,15 +80,31 @@ func updateConfigMapIfNeed(
...
+// confirmNameByData is used to fix the problem that
+// when configUpdateStrategy is changed from InPlace to RollingUpdate for the first time,
+// the name of desired configmap maybe different from the existing one while
+// the data of them are the same, which will cause rolling update.
+func confirmNameByData(existing, desired *corev1.ConfigMap, dataEqual bool) {
+	if dataEqual && existing.Name != desired.Name {
+		desired.Name = existing.Name
+	}
+	if !dataEqual && existing.Name == desired.Name {
+		desired.Name = fmt.Sprintf("%s-new", desired.Name)
```
DanielZhangQD, line 108: This will create a new CM if `InPlace` is set?
What's the issue to fix with this block?

pkg/manager/member/config_test.go
```diff
@@ -67,9 +70,10 @@ func TestUpdateConfigMap(t *testing.T) {
 
 	tests := []testcase{
 		{
-			name: "two empty config map",
-			old:  &corev1.ConfigMap{},
-			new:  &corev1.ConfigMap{},
+			name:  "two empty config map",
+			old:   &corev1.ConfigMap{},
```
DanielZhangQD, line 74: Please also add a case where both CMs have the same data.

pkg/manager/member/config.go
cvvz: If `InPlace` is set, it will just update current CM, the code is [here](https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/config.go#L58-L62).

pkg/manager/member/config_test.go
cvvz: ok

DanielZhangQD (approved):
LGTM

PR #3974, opened 2021-05-14
examples: Update AWS ELB config

dveeden:

- Remove incorrect `aws-load-balancer-internal`.
- Change the loadbalancer for Grafana from "CLB" (Classic) to "NLB"
  (Network)

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

Deploy on EKS with https://docs.pingcap.com/tidb-in-kubernetes/stable/deploy-on-aws-eks

On the AWS Console check the loadbalancers that were created. This should show three NLB type loadbalancers with the 'internet facing' scheme. Previously this would show 1 NLB and 1 CLB and the NLB had the 'internal' scheme.

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


dveeden:
Should we also add a LoadBalancer for PD as that is hosting the dashboard on port 2379?

july2993 (approved):

examples/aws/tidb-cluster.yaml
```diff
@@ -12,6 +12,14 @@ spec:
...
     baseImage: pingcap/pd
     replicas: 3
+    service:
+      annotations:
+        service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
+        service.beta.kubernetes.io/aws-load-balancer-type: nlb
+        service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
+      exposeStatus: true
+      externalTrafficPolicy: Local
+      type: LoadBalancer
```
DanielZhangQD, line 22: Please do not expose the PD service as LB since PD does not support authentication/authorization yet, it's dangerous to do this.

examples/aws/tidb-cluster.yaml
```diff
@@ -47,8 +55,8 @@ spec:
     service:
       annotations:
         service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
-        service.beta.kubernetes.io/aws-load-balancer-internal: '0.0.0.0/0'
-        service.beta.kubernetes.io/aws-load-balancer-type: nlb 
+        service.beta.kubernetes.io/aws-load-balancer-type: nlb
+        service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
```
DanielZhangQD, line 59: Please do not set this to `internet-facing`, we use internal LB by default, if some users want to use internal-facing LB, they can just update this.
BTW, to keep backward compatibility, what about also adding:
`service.beta.kubernetes.io/aws-load-balancer-internal: "true"`?

examples/aws/tidb-monitor.yaml
```diff
@@ -22,7 +22,12 @@ spec:
     logLevel: info
     password: admin
     service:
-      portName: http-grafana
+      annotations:
+        service.beta.kubernetes.io/aws-load-balancer-cross-zone-load-balancing-enabled: 'true'
```
DanielZhangQD, line 26: This is not necessary as there will be only one monitoring Pod.

examples/aws/tidb-cluster.yaml
dveeden: I'm afraid many users are not sufficiently aware of this risk as well. This might need a follow-up in the docs.

Instead of not exposing this (and maybe having users doing this themselves) can we expose this securely with an Ingress controller?

I haven't tested this, but something along the lines of this:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  namespace: default
  name: ingress
  annotations:
    kubernetes.io/ingress.class: alb
    alb.ingress.kubernetes.io/scheme: internet-facing
spec:
  rules:
    - http:
        - path: /dashboard
          backend:
            serviceName: pd
            servicePort: use-annotation
```
This is intended to only forward calls to /dashboard, which should only allow access to the dashboard and not to other paths of PD.

dveeden: This is actually already documented on https://docs.pingcap.com/tidb-in-kubernetes/stable/access-dashboard but I wasn't aware of this. I added a link in the docs in https://github.com/pingcap/docs-tidb-operator/pull/1212

examples/aws/tidb-monitor.yaml
```diff
@@ -22,7 +22,11 @@ spec:
     logLevel: info
     password: admin
     service:
-      portName: http-grafana
+      annotations:
+        service.beta.kubernetes.io/aws-load-balancer-type: nlb
+        service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
+      exposeStatus: true
+      externalTrafficPolicy: Local
```
DanielZhangQD, line 29: These two fields are not supported for the Grafana service.

DanielZhangQD (approved):
LGTM

PR #4014, opened 2021-06-10
The comment of StorageVolume is confusing.

cvvz:
The comment of `StorageVolume` makes me confuesed, I can't tell what the difference between it and [AdditionalVolumes and AdditionalVolumes](https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/types.go#L892-L895) at the beginning.


csuzhangxc (approved):
LGTM

PR #4015, opened 2021-06-10
Do not generate volumeMount if not neccessary

cvvz:

Sometimes users may not want to generate `volumeMount` by tidb-operator automatically, for example, users may inject `volumeMoumt` in their Mutating Admission Webhook.
### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [x] Manual test 
1. Create `TidbCluster` with 
```yaml
ticdc:
  storageVolumes:
     - name: storage
       storageClassName: local-storage
       storageSize: 10Gi
```
2.  Statefulset with `volumeClaimTemplates` and without `volumeMount` in `template` generate normally.
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
```


DanielZhangQD:
@cvvz What are you trying to do with this PR?

> Support subPath for additional volume.

This is supported by default with `additonalVolumeMount`.

> Sometimes users may not want to generate volumeMount by tidb-operator.

What does this mean? What's the user scenario? Just configure a volume but do not mount to the container?

cvvz:
1. We inject `volumeMount` by Our mutatingwebhook. What we need is just an additional PV for sts such as TiCDC, it will be unnecessary for us to create a `volumeMount` without use. PS, once I specify `volumeMount` through this field, it may cause rollingupdate next time if I need to add item in `volumeMount`, so we inject it by webhook.
2. About `subPath`, sure you can use `additonalVolumeMount` to specify additional volumeMount, but it will be more handy to create `volumeMount` with `subPath` by `storageVolume`.

DanielZhangQD:
> 1. We inject `volumeMount` by Our mutatingwebhook. What we need is just an additional PV for sts such as TiCDC, it will be unnecessary for us to create a `volumeMount` without use. PS, once I specify `volumeMount` through this field, it may cause rollingupdate next time if I need to add item in `volumeMount`, so we inject it by webhook.

For this case, you can just configure `additionalVolume`.

> 2. About `subPath`, sure you can use `additonalVolumeMount` to specify additional volumeMount, but it will be more handy to create `volumeMount` with `subPath` by `storageVolume`.

What's the value added by `subPath` when you mount a PV to a container?

cvvz:
> For this case, you can just configure additionalVolume.
But for now, `storageVolume ` will generate a `volumeMount` which we don't need.
> What's the value added by subPath when you mount a PV to a container?
This is prepare for the future.

DanielZhangQD:
> > For this case, you can just configure additionalVolume.
> 
> But for now, `storageVolume ` will generate a `volumeMount` which we don't need because it force user to specify `volumeMount`.

For `storageVolume`, TiDB Operator will generate the `volumeMount` automatically, it does not requre user to specify manually.

> > What's the value added by subPath when you mount a PV to a container?
> 
> This is prepare for the future.

For which user scenario?

cvvz:
> For storageVolume, TiDB Operator will generate the volumeMount automatically, it does not requre user to specify manually.

That's the problem, I don't want it generate volumeMount for me, because I inject volumeMount by webhook already. There is no need for me to specify `volumeMount` in `storageVolume`. 

> For which user scenario?
For example, sidecar and app container may both want to mount log file to path `/var/log/app.log`, mount to `/var/log/` directly may cause override.

cvvz:
PTAL @DanielZhangQD

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -2083,13 +2083,16 @@ type WorkerFailureMember struct {
 	CreatedAt metav1.Time `json:"createdAt,omitempty"`
 }
 
-// StorageVolume configures additional storage for PD/TiDB/TiKV pods.
-// If `StorageClassName` not set, default to the `spec.[pd|tidb|tikv].storageClassName`
+// StorageVolume configures additional PVC template for statefulsets and volumeMount for pods which mount this pvc .
```
DanielZhangQD, line 2086:
```suggestion
// StorageVolume configures additional PVC template for StatefulSets and volumeMount for pods that mount this PVC.
```

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -2083,13 +2083,16 @@ type WorkerFailureMember struct {
...
+// Note:
+// If `MountPath` is not set, no volumeMount will not be generated. (You may not want to set this field when you inject volumeMount
```
DanielZhangQD, line 2088:
```suggestion
// If `MountPath` is not set, volumeMount will not be generated. (You may not want to set this field when you inject volumeMount
```

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -2083,13 +2083,16 @@ type WorkerFailureMember struct {
...
+// in somewhere else such as Mutating Admission Webhook)
+// If `StorageClassName` is not set, default to the `spec.${component}.storageClassName`
 type StorageVolume struct {
 	Name             string  `json:"name"`
 	StorageClassName *string `json:"storageClassName,omitempty"`
 	StorageSize      string  `json:"storageSize"`
-	MountPath        string  `json:"mountPath"`
+	MountPath        *string `json:"mountPath,omitempty"`
```
DanielZhangQD, line 2095:
```suggestion
	MountPath        string `json:"mountPath,omitempty"`
```

DanielZhangQD: No need to be a pointer, right?

cvvz: Yes

DanielZhangQD (approved):
LGTM

handlerww (approved):

PR #4023, opened 2021-06-15
add design of tidb monitor 'path'

songjiansuper:
This is a design about adding the variable path to the ingressspec of tidb monitor

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

NONE

```release-note

```

### DIFF
--- a/docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
+++ b/docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
@@ -1,0 +1,41 @@
+# Pause TiDB Cluster
+
+This document presents a design to add path of tidb monitor in
+operator.
+
+## Motivation
+
+At present, when we want to expose monitoring services (including grafana and prothumes) through Ingress, our design is to use different hosts for different tidb-monitors to distinguish which specific monitoring service is being accessed.This brings some inflexible experiences. If we want to distinguish between different monitoring services through one host but different paths, it is impossible to do it now, but in fact this is also a feasible solution
+
+## Proposal
+
+We can add path to the spec and apply it to the ingress spec application, which can bring us more flexible path parameters [spec code](https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/types.go#L1709)
+
+### Spec
+
+Add a new field to `Path` to Ingress Spec:
+
+```
+    // Indicates ingress path variable
+    // +optional
+    Path string `json:"path,omitempty"`
+```
+
+### Implementation
+
+If not specified the value is '/'
+
+If value specified,it will take effect on [operator code](https://github.com/pingcap/tidb-operator/blob/master/pkg/monitor/monitor/util.go#L942)
+
+- tidb monitor ingress
+
+## Testing plan
+
+### TiDB cluster can be paused and unpaused
+
+- Deploy a tidb cluster
+- Deploy a have 'path' tidb monitor
+- Verify ingress spec
+- Verify monitor service connectivity
+
+## Open Questions


docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
+# Pause TiDB Cluster
```
csuzhangxc, line 1: please fix the title

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+
+This document presents a design to add path of tidb monitor in
+operator.
```
csuzhangxc, line 4: unnecessary newline?

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Motivation
+
+At present, when we want to expose monitoring services (including grafana and prothumes) through Ingress, our design is to use different hosts for different tidb-monitors to distinguish which specific monitoring service is being accessed.This brings some inflexible experiences. If we want to distinguish between different monitoring services through one host but different paths, it is impossible to do it now, but in fact this is also a feasible solution
```
csuzhangxc, line 8:
```suggestion
At present, when we want to expose monitoring services (including grafana and prothumes) through Ingress, our design is to use different hosts for different tidb-monitors to distinguish which specific monitoring service is being accessed. This brings some inflexible experiences. If we want to distinguish between different monitoring services through one host but different paths, it is impossible to do it now, but in fact, this is also a feasible solution.
```

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Proposal
+
+We can add path to the spec and apply it to the ingress spec application, which can bring us more flexible path parameters [spec code](https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/types.go#L1709)
```
csuzhangxc, line 12: I suggest link to a permalink (to a specified git commit, like https://github.com/pingcap/tidb-operator/blob/2c7a0cae64a6ec8ec5e93033203a8087fc22d819/pkg/apis/pingcap/v1alpha1/types.go#L1709) rather than `master`.

if so, ditto for L28.

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+    Path string `json:"path,omitempty"`
+```
+
+### Implementation
+
+If not specified the value is '/'
+
+If value specified,it will take effect on [operator code](https://github.com/pingcap/tidb-operator/blob/master/pkg/monitor/monitor/util.go#L942)
+
+- tidb monitor ingress
```
csuzhangxc, line 30: what does this line mean?

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Testing plan
+
+### TiDB cluster can be paused and unpaused
```
csuzhangxc, line 34: please fix this title.

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+We can add path to the spec and apply it to the ingress spec application, which can bring us more flexible path parameters [spec code](https://github.com/pingcap/tidb-operator/blob/master/pkg/apis/pingcap/v1alpha1/types.go#L1709)
+
+### Spec
+
+Add a new field to `Path` to Ingress Spec:
+
+```
+    // Indicates ingress path variable
+    // +optional
+    Path string `json:"path,omitempty"`
```
DanielZhangQD, line 21: How does the user access one specific TidbMonitor service with this `Path`?
Any special configuration should be done for Grafana?
Did you do a PoC with this change?

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+### Implementation
+
+If not specified the value is '/'
```
DanielZhangQD, line 26: Should it default to `""`? Because this is the behavior now.

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+
+If value specified,it will take effect on [operator code](https://github.com/pingcap/tidb-operator/blob/master/pkg/monitor/monitor/util.go#L942)
```
DanielZhangQD, line 28: Please add more detailed description here, how the `Path` is handled exactly？

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,41 @@
...
+- tidb monitor ingress
+
+## Testing plan
+
+### TiDB cluster can be paused and unpaused
+
+- Deploy a tidb cluster
+- Deploy a have 'path' tidb monitor
+- Verify ingress spec
+- Verify monitor service connectivity
```
DanielZhangQD, line 39: Can we describe the case in more detail? e.g. show the important configurations, how to access the service, etc.

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
songjiansuper: In fact, when setting this "" will be set to'/', you can refer to here,https://github.com/pingcap/tidb-operator/blob/master/pkg/monitor/monitor/util.go#L942

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
songjiansuper: Now it is a simple experiment. I tried to manually create ingress for different tidb monitors. Specify the host as one and the path as different. It seems to be accessible. I refer to this document, https:// docs.ksyun.com/documents/39303.Yes, we need to change the configuration of grafana, GF_SERVER_ROOT_URL, this needs to be changed

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
songjiansuper: This is used to indicate where the code actually takes effect, that is, where tm spec is transformed into ingress spec

DanielZhangQD: OK, please describe it in more detail, thanks!

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
DanielZhangQD: OK, thanks for the explanation.
We have to make the design as much detail as possible so that others can know the following info:

- Why we should do this?
- How to do this exactly?

docs/design-proposals/2021-06-15-add-path-to-tidb-monitor.md
```diff
@@ -0,0 +1,45 @@
...
+
+- tidb monitor ingress
+
+## Testing plan
+
+### TiDB monitor can use path parameter
+
+- Deploy a tidb cluster
+- Deploy a have 'path' tidb monitor
+  - Change grafana spec value 'GF_SERVER_ROOT_URL',make it match the path in ingress
```
mikechengwei, line 37: The env logic of prometheus is missing. Prometheus have support shard, can you add prometheus multiple shards and multiple ingress design? Ref: #4198

mikechengwei: If we have two replicas and two shards , it will have four prometheus instances. So how will these four prometheus ingress be designed?

PR #4116, opened 2021-08-04
Issues

hihihuhu:
this pr addresses a few issues:

1. add namespace to the components in operator chart, so that the manifest from `helm template` will be self contained
2. add the ability to customize tidb service selector label, which our internal service discovery system relies on
3. add the ability to customize scheduler config map annotation, so that spinnaker won't version the config map when it applied the generated manifest from `helm template`
4. ignore the new lines when set the password in tidb initializer
5. use internal service name for `discovery` in multi k8s setup, alternatively, could customize the service definition and set `ClusterIp` to `none` for discovery service as well, however, the components don't have to talk to it cross k8s cluster, thus only use internal dns record should be enough.

see section above

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


KanShiori:
Thanks for submitting the PR.

Should we split it into multi PRs ?

DanielZhangQD:
> Thanks for submitting the PR.
> 
> Should we split it into multi PRs ?

Agree, we should focus on one issue in one PR.
@hihihuhu Could you please help split to multiple PRs?

pkg/manager/member/ticdc_member_manager.go
```diff
@@ -395,19 +395,18 @@ func getNewTiCDCStatefulSet(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap) (*ap
 		} else {
 			pdAddr = fmt.Sprintf("http://%s-pd:2379", tcName)
 		}
-		formatClusterDomain := controller.FormatClusterDomain(tc.Spec.ClusterDomain)
 
 		str := `set -uo pipefail
 pd_url="%s"
 encoded_domain_url=$(echo $pd_url | base64 | tr "\n" " " | sed "s/ //g")
-discovery_url="%s-discovery.${NAMESPACE}.svc%s:10261"
+discovery_url="%s-discovery.${NAMESPACE}.svc:10261"
```
DanielZhangQD, line 402:
```suggestion
discovery_url="%s-discovery.${NAMESPACE}:10261"
```

pkg/manager/member/tidb_member_manager.go
```diff
@@ -448,6 +448,7 @@ func getNewTiDBServiceOrNil(tc *v1alpha1.TidbCluster) *corev1.Service {
 	tidbSelector := label.New().Instance(instanceName).TiDB()
 	svcName := controller.TiDBMemberName(tcName)
 	tidbLabels := util.CombineStringMap(tidbSelector.Copy().UsedByEndUser().Labels(), svcSpec.Labels)
+	tidbSelectorLabels := util.CombineStringMap(tidbSelector.Copy().Labels(), svcSpec.SelectorLabels)
```
DanielZhangQD, line 451: Why do you need to update the selector? Or you actually need to update the Pod labels?

pkg/manager/member/tiflash_member_manager.go
```diff
@@ -407,11 +407,10 @@ func getNewStatefulSet(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap) (*apps.St
 		} else {
 			pdAddr = fmt.Sprintf("http://%s-pd:2379", tcName)
 		}
-		formatClusterDomain := controller.FormatClusterDomain(tc.Spec.ClusterDomain)
 		str := `pd_url="%s"
 set +e
 encoded_domain_url=$(echo $pd_url | base64 | tr "\n" " " | sed "s/ //g")
-discovery_url="%s-discovery.%s.svc%s:10261"
+discovery_url="%s-discovery.%s.svc:10261"
```
DanielZhangQD, line 413:
```suggestion
discovery_url="%s-discovery.%s:10261"
```

pkg/manager/member/template.go
```diff
@@ -145,7 +145,7 @@ POD_NAME=${POD_NAME:-$HOSTNAME}
 cluster_name=` + "`" + `echo ${PEER_SERVICE_NAME} | sed 's/-pd-peer//'` + "`" +
 	`
 domain="${POD_NAME}.${PEER_SERVICE_NAME}.${NAMESPACE}.svc{{ .FormatClusterDomain }}"
-discovery_url="${cluster_name}-discovery.${NAMESPACE}.svc{{ .FormatClusterDomain }}:10261"
+discovery_url="${cluster_name}-discovery.${NAMESPACE}.svc:10261"
```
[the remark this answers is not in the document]
DanielZhangQD: Please make sure that your changes will not trigger a rolling update of the current TiDB clusters deployed in one Kubernetes cluster.

PR #4122, opened 2021-08-04
customize annotation for scheduler configmap

hihihuhu:
using spinnaker to deploy the manifest generated by `helm template` requires additional annotation for the configmap, so that it won't and add version to the name

pass through the annotation to the config map

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

`helm template` with the value set, unset and comment out

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


charts/tidb-operator/values.yaml
```diff
@@ -163,6 +163,10 @@ scheduler:
   # Refer to https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
   podAnnotations: {}
 
+  # additional annotations for the configmap, mainly to prevent spinnaker versioning the cm
+  configmap:
+    annotations: {}
```
csuzhangxc, line 168: how about renaming it to something like `configmapAnnotations` just as `podAnnotations` did?

csuzhangxc (approved):

charts/tidb-operator/values.yaml
hihihuhu: thanks for pointing out! fixed

DanielZhangQD (approved):
LGTM

PR #4124, opened 2021-08-04
add additional selector labels for tidb service

hihihuhu:
our internal service discovery works in the following way.

we use webhookapi to inject a sidecar to tidb pod, which monitors the health state of the pod and registers on our service discovery system.

however, the implement details of the sidecar is,
1. it adds a label to the pod to say it is ready
2. it gets the endpoint for k8s endpoint
3. it only marks the pod ready when 2 is success

in order to work with the service discovery system we have, the service object for tidb needs to tolerate unready pod but use label to determine which pod could be add to the service. that is the reason we need to customize the selector label of the service

passthrough additional selector label for tidb service

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


csuzhangxc (approved):

DanielZhangQD:
@hihihuhu Sorry for the late review.
I have some questions for the description in the PR description:
```
it adds a label to the pod to say it is ready
it gets the endpoint for k8s endpoint
it only marks the pod ready when 2 is success
```

Since only when a Pod is ready then it will be included in the endpoint, why do we need another label to indicate that the Pod is ready?
Is it OK that your webhook gets the endpoint and updates your discovery system?

hihihuhu:
Thanks for the review. I guess this requirement is a bit weird, we can keep this on our fork and not upstream it for now. Closing this one

Issue #3497, opened 2020-11-13
Deprecate pod validating and mutating webhook


DanielZhangQD:
## Feature Request

**Is your feature request related to a problem? Please describe:**

In TiDB Operator v1.1.0, we introduced Pod validating webhook to handle the upgrade and scale operation of PD and TiKV, which has the similar logic with that in PD and TiKV manager in the tidb-controller-manager, however, as we integrate more and more components in TiDB Operator, I propose to deprecate the Pod validating webhook for the following reasons:

1. There is not much benefit to move the upgrade and scale logic from the manager in the tidb-controller-manager to the webhook, check the logic in the webhook now, the code logic is complicated too.
2. We have to maintain two logics for the same component as we cannot totally discard the logic in tidb-controller-manager now, which is a big burden for us
3. The webhook can only handle the upgrade and scale of TiKV and PD now, for all the other components, e.g. TiFlash, DM, etc., it does not support yet.

The Pod mutating webhook is introduced in v1.1.0 to add a specific label for the auto-scaled TiKV Pods, however, in v1.2.0, the auto-scaling logic has been refactored ([proposal](https://github.com/pingcap/tidb-operator/blob/master/docs/design-proposals/2020-08-21-auto-scaling-with-pdapi.md)) and there is no need to mutate the auto-scaled TiKV Pods anymore, so that we can remove it.

 We will keep the following functionalities in the webhook:

1. Canary upgrade of StatefulSet
2. defaulting and validating of CRs

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

PR #4214, opened 2021-10-08
implement charts for deploying admission-webhook

spxcmss:
add nodeselector and torelation in admission-webhook's charts


DanielZhangQD:
@spxcmss Do you use the Pod webhook during your work?
We will deprecate the Pod webhook in v1.3.0, please check detail in https://github.com/pingcap/tidb-operator/issues/3497.

PR #4123, opened 2021-08-04
use interanl svc name for discovery instead of external dns

hihihuhu:
### What problem does this PR solve?
the discovery url doesn't work in a cross k8s cluster setup, the discovery service is defined without set `ClusterIp` to `none`, thus normally external-dns won't create record for it. even it does create, it will map to a virtual ip in k8s and cannot be accessed across k8s.

### What is changed and how does it work?
the discovery service doesn't have to be accessed across k8s, thus use internal dns instead of external

### Code changes

- [x] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

deployed the pr will stage the change in configmap, but it doesn't make the components rolling restart

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [ ] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note
NONE
```

PR #4248, opened 2021-11-04
Check startup script when check if configmap has beed updated

KanShiori:

Check startup script when check if configmap has beed updated

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
Upgrade operator v1.1.7 to v1.2.4:
* If set `clusterDomain` in TidbCluster, `PD`,`TiKV` and `TiDB` roll update. Because of PR #4123
* If don't set `clusterDomain` in TidbCluster, `PD`,`TiKV` and `TiDB` don't roll update.
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Roll update `PD`, `TiKV`, `TiDB` if set `clusterDomain`.
```

### DIFF
--- a/pkg/manager/member/config.go
+++ b/pkg/manager/member/config.go
@@ -24,10 +24,12 @@ import (
 	corelisters "k8s.io/client-go/listers/core/v1"
 )
 
+// updateConfigMap check if ConfigMap is has actually been updated
 func updateConfigMap(old, new *corev1.ConfigMap) (bool, error) {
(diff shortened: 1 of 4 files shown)


pkg/manager/member/config.go
```diff
@@ -52,6 +53,18 @@ func updateConfigMap(old, new *corev1.ConfigMap) (bool, error) {
 		}
 	}
```
july2993, line 55: how about just
```
if !reflect.DeepEqual(old.Data, new.Data) {
    return false nil
}

return true, nil

```

july2993 (approved):

DanielZhangQD (approved):
LGTM

PR #4260, opened 2021-11-16
Update image version of busybox helper

dveeden:

The `busybox:1.26.2` image is getting old and only available for `amd64` and not for `arm64`.

- https://hub.docker.com/_/busybox?tab=tags&page=1&name=1.26.2
- https://hub.docker.com/_/busybox?tab=tags&page=1&name=1.34.1

This updates the version of the helper container

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [x] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
The helper image was updated to a newer version of busybox
```


KanShiori:
It will cause rolling update after upgrading operator, right?

dveeden:
> It will cause rolling update after upgrading operator, right?

I think it will, but I'm not 100% sure

dveeden:
Another PR that is related to amd64 and tidb-operator is https://github.com/pingcap/docs-tidb-operator/pull/1301

KanShiori:
> > It will cause rolling update after upgrading operator, right?
> 
> I think it will (and should), but I'm not 100% sure

clients don't expect to roll update after upgrade operator, so maybe we need to implement a feature to control rolling update firstly 🤔

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -27,7 +27,7 @@ import (
 
 const (
 	// defaultHelperImage is default image of helper
-	defaultHelperImage        = "busybox:1.26.2"
+	defaultHelperImage        = "busybox:1.34.1"
```
DanielZhangQD, line 30: We cannot update the default image here, which will cause the rolling update of TiDB after upgrading the TiDB Operator, and is not a good practice.
This is configurable in the CR so if the user wants to use arm64 image, they can just configure it.

dveeden: I understand that we want to be careful with existing installations. However for new installations (especially on ARM64, but also on any other platform) having and old and possibly insecure busybox versions isn't great.

Can we modify this to use 1.34.1 for new installations while not updating older installations?

Would it be ok to set it to 1.34.1 in all example files (including those where currently no version is specified)?

Is this something that can be done in a new major version of the operator?

DanielZhangQD: @dveeden I understand your concern.
Even if with the major version, we still prefer not to trigger a rolling update, as you know that TiDB connections are not handled well during rolling update.

I think it's OK `to set it to 1.34.1 in all example files (including those where currently no version is specified)`, could you please help to improve the examples?

dveeden: @DanielZhangQD done

Not sure what's needed for `charts/tidb-cluster/values.yaml` etc.

dveeden: I've updated `charts/tidb-cluster/templates/tidb-cluster.yaml` and `charts/tidb-cluster/values.yaml`.

docs/api-references/docs.md
```diff
@@ -6606,7 +6606,7 @@ string
 <td>
 <em>(Optional)</em>
 <p>Image used to tail slow log and set kernel parameters if necessary, must have <code>tail</code> and <code>sysctl</code> installed
-Optional: Defaults to busybox:1.26.2</p>
+Optional: Defaults to busybox:1.26.2. Recommended to set to 1.34.1 for new installations.</p>
```
DanielZhangQD, line 6609: This file is generated automatically, please update the description in types.go, thanks!

pkg/apis/pingcap/v1alpha1/tidbcluster.go
DanielZhangQD: The `tidb-cluster` chart has been deprecated since v1.1.0 but your change is OK, thanks!

dveeden:
@DanielZhangQD PTAL

docs/api-references/docs.md
DanielZhangQD: @dveeden Please check this comment since it will cause CI failure, thanks!

DanielZhangQD (approved):
LGTM

docs/api-references/docs.md
dveeden: I've ran `hack/update-all.sh`, the result is in c20a731daaa543dc1b90965cf4d59f6fe15a58a5

For this I had to modify my `go.sum` like this:
```patch
diff --git a/go.sum b/go.sum
index 5d4d2fdd..6308383b 100644
--- a/go.sum
+++ b/go.sum
@@ -80,7 +80,7 @@ github.com/PuerkitoBio/purell v1.1.1/go.mod h1:c11w/QuzBsJSee3cPx9rAFu61PvFxuPbt
 github.com/PuerkitoBio/urlesc v0.0.0-20160726150825-5bd2802263f2/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 h1:d+Bc7a5rLufV/sSk/8dngufqelfh6jnri85riMAaF/M=
 github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578/go.mod h1:uGdkoq3SwY9Y+13GIhn11/XLaGBb4BfwItxLd5jeuXE=
-github.com/agiledragon/gomonkey/v2 v2.2.0 h1:QJWqpdEhGV/JJy70sZ/LDnhbSlMrqHAWHcNOjz1kyuI=
+github.com/agiledragon/gomonkey/v2 v2.2.0 h1:SXUF5UN7ZH67ORjFK+e1FoFRZczCSVUvwTywNY2+fcg=
 github.com/agiledragon/gomonkey/v2 v2.2.0/go.mod h1:ap1AmDzcVOAz1YpeJ3TCzIgstoaWLA6jbbgxfB4w2iY=
 github.com/agnivade/levenshtein v1.0.1/go.mod h1:CURSv5d9Uaml+FovSIICkLbAUZ9S4RqaHDIsdSBg7lM=
 github.com/ajg/form v0.0.0-20160822230020-523a5da1a92f/go.mod h1:uL1WgH+h2mgNtvBq0339dVnzXdBETtL2LeUXaIv25UY=
```
I haven't checked in the `go.sum` changes.

Without this it complained like this:
```
[dvaneeden@dve-carbon tidb-operator]$ ./hack/update-all.sh 
warning: ignoring symlink /home/dvaneeden/dev/pingcap/tidb-operator/deploy/gcp/charts/tidb-cluster
warning: ignoring symlink /home/dvaneeden/dev/pingcap/tidb-operator/deploy/gcp/charts/tidb-operator
warning: ignoring symlink /home/dvaneeden/dev/pingcap/tidb-operator/deploy/gcp/manifests/gke
go: downloading cloud.google.com/go v0.51.0
go: downloading github.com/robfig/cron v1.1.0
go: downloading go.etcd.io/etcd v0.5.0-alpha.5.0.20200819165624-17cef6e3e9d5
go: downloading github.com/agiledragon/gomonkey/v2 v2.2.0
go: downloading golang.org/x/sync v0.0.0-20201207232520-09787c993a3a
go: downloading github.com/pingcap/tidb v2.1.0-beta+incompatible
verifying github.com/agiledragon/gomonkey/v2@v2.2.0: checksum mismatch
	downloaded: h1:SXUF5UN7ZH67ORjFK+e1FoFRZczCSVUvwTywNY2+fcg=
	go.sum:     h1:QJWqpdEhGV/JJy70sZ/LDnhbSlMrqHAWHcNOjz1kyuI=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```

This could be related to goproxy etc.

DanielZhangQD (approved):
LGTM

july2993 (approved):

PR #4277, opened 2021-11-24
RFC: Add 2021-11-24-graceful-restart-tikv-pod.md

july2993:
RFC: Add 2021-11-24-graceful-restart-tikv-pod.md

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,107 @@
...
+Support user to add an annotation to CRD `TidbCluster` to trigger an gracefule reschedule.
+
+annotation key: `tikv.tidb.pingcap.com/reschedule`
+
+annotation value is JSON in the flowing struct:
+
+```go
+{
+  // the flowing field set by user.
+  NodeName *string
```
csuzhangxc, line 51: Why not support specify more than one node?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,107 @@
...
+
+- Provider a way to make tikv graceful stop when just recive a `SIGTERM` signal.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+suppost `tikv-0` is running at node `node0`,  there are many free resource at other nodes, I would like to make tikv-0 run at other node and shutdown `node0` to reduce the cost. I may apply the flowing operations:
```
liubog2008, line 30:
```suggestion
suppose `tikv-0` is running at node `node0`,  there are many free resource at other nodes, I would like to make tikv-0 run at other node and shutdown `node0` to reduce the cost. I may apply the flowing operations:
```

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,107 @@
...
+   }
+} else {
+  // Find a slot to hanle by comaring `FinishedSlots` and `Slots` or slots belong to the specify NodeName.
+  // set RunningSlot or mark Status as Finished
+}
+
+
+```
+
+An example for Story 1 step 2, user might add annotation with the flowing value for key `tikv.tidb.pingcap.com/reschedule`:
```
csuzhangxc, line 94: did the user need to do step 1 (`kubectl cordon node0`) manually? this will also affect the schedule of other pods? can we modify our sts to achieve this goal?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,107 @@
...
+2. delete pod `tikv-0` to let it re-create and reschedule to other nodes.
+3. draine `node-0` in the nomal way as [safely drain a node](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/) and shotdown `node0`
+
+### Risks and Mitigations
+
+- PD/TiKV fails to evict all leader and we may just keep checking it.
+
+## Design Details
+
+Support user to add an annotation to CRD `TidbCluster` to trigger an gracefule reschedule.
```
liubog2008, line 42: Why not update pod annotation to reschedule it?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
liubog2008: Why nodeName is needed?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
july2993: > did the user need to do step 1 (kubectl cordon node0) manually?

yes

I prefer operator does not care anything about the schedule. that means a user can do not do `kubectl cordon node0`
and just let it re-create and schedule in the normal way(can still run at `node0` or other nodes) if the user wants this(like graceful restart in tiup).

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
july2993: generally, we just want to take one node per time?
you can do it one by one.

csuzhangxc: > generally, we just want to take one node per time?

but if the user set more than one node, operator can still do it one by one, right?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
liubog2008: > then we may lose the update(leave an evict-leader scheduler at pd)
While the pod has the annotation, add evict-leader scheduler at pd. If the pod is deleted or updated(annotation is removed), just remove scheduler at pd. It's level triggered and cancelable so scheduler will not be lefted at pd.

july2993: You can not always catch the `delete`(like at the time pod deleted operator not running normal), you need to check time from time.

another possible issue is we may also delete evict-leader scheduler not added by us.

liubog2008: > You can not always catch the delete(like at the time pod deleted operator not running normal), you need to check time from time.

No need to catch the event. We just check the current tikv pod. While pod is deleted and operator is not running, delete event will miss but evict-leader scheduler can be removed after operator is restarted.

> another possible issue is we may also delete evict-leader scheduler not added by us.

It seems a common issue for both of two ways?

july2993: > No need to catch the event. We just check the current tikv pod. While pod is deleted and operator is not running, delete event will miss but evict-leader scheduler can be removed after operator is restarted.
--

yes, you need to check all clusters as mentioned before. (if not annotation but having an evict-leader scheduler, we delete the scheduler)


> It seems a common issue for both of two ways?

no, we have the annotation at tc, we will not delete the scheduler added by the user.

liubog2008: Fine, I understand your concern. Let me rethink it because I prefer to act it in pod level.

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,107 @@
...
+
+- PD/TiKV fails to evict all leader and we may just keep checking it.
+
+## Design Details
+
+Support user to add an annotation to CRD `TidbCluster` to trigger an graceful reschedule.
+
+annotation key: `tikv.tidb.pingcap.com/reschedule`
+
+annotation value is JSON in the flowing struct:
```
DanielZhangQD, line 46: This json is too complicated for users to use.

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,107 @@
...
+   switch RunningSlotStatus {
+     case "Evicting"
+       // 1. make sure evict-scheduler is added
+       // 2. update RunningSlot.Status as ReCreating if leader-count is 0
+     case "ReCreating"
+     if the creationTimestamp of pod is less than RunningSlot.startTimestamp {
+        delete the pod.
+     }
+     if the pod is already running normal {
+        append RunningSlot.Slot in RunningSlot and set RunningSlot as nil
```
DanielZhangQD, line 83: `append RunningSlot.Slot in RunningSlot` to `FinishedSlots`?

DanielZhangQD: When to delete the evict leader scheduler?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
DanielZhangQD: Yes, the original requirement is just to restart TiKV gracefully.

july2993:
@DanielZhangQD PTAL

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,137 @@
...
+- `delete-pod`: delete pod and remove the evict-leader scheduler from PD.
+
+
+
+An EvictLeader status willbe added to the `TiKVStatus`:
+
+```go
++type EvictLeaderStatus struct {
++       PodCreateTime metav1.Time `json:"podCreateTime,omitempty"`
++       Value         string      `json:"value,omitempty"`
```
csuzhangxc, line 61: do we need to add pod name into this status?

july2993: https://github.com/pingcap/tidb-operator/pull/4277/commits/edece78c434d7035b303b470ac8e8f3c38f88862
s/*EvictLeader/map[string]*EvictLeader
in tikv status to support multi pod annotated in the same time.

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,138 @@
...
+func sync(pod *corev1.Pod, tc *v1alpha1.TidbCluster) (ctrl.Result, error) {
+        value, ok := pod.Annotations[v1alpha1.EvictLeaderAnnKey]
+
+        if ok {
+                evictStatus := &v1alpha1.EvictLeaderStatus{
+                        PodCreateTime: pod.CreationTimestamp,
+                        Value:         value,
+                }
+                nowStatus := tc.Status.TiKV.EvictLeader[pod.Name]
+                if nowStatus == nil || *nowStatus != evictStatus {
```
july2993, line 89:
```suggestion
                if nowStatus == nil || *nowStatus != *evictStatus {
```

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,138 @@
...
+        TombstoneStores map[string]TiKVStore            `json:"tombstoneStores,omitempty"`
+        FailureStores   map[string]TiKVFailureStore     `json:"failureStores,omitempty"`
+        Image           string                          `json:"image,omitempty"`
++       EvictLeader     map[string]*EvictLeaderStatus   `json:"evictLeaderStatus,omitempty"`
+ }
+```
+
+
+
+At controller side, the reconciler to handle pod will be like:
```
DanielZhangQD, line 77: A new controller?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Risks and Mitigations
+
+- PD/TiKV fails to evict all leader and we may just keep checking it.
+
+## Design Details
+
+Support user to add an annotation to tikv pod to trigger an graceful reschedule.
+
+Annotation key `EvictLeaderAnnKey`: `tikv.tidb.pingcap.com/evict-leader`
```
DanielZhangQD, line 47: What about using two separate keys `tidb.pingcap.com/evict-leader` and `tidb.pingcap.com/restart`?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
july2993: yes

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,138 @@
...
+
+the value controler the behavier when the leander count drop to zero, valid value is one of:
+
+- `none`: doing nothing
+- `delete-pod`: delete pod and remove the evict-leader scheduler from PD.
+
+
+
+An EvictLeader status willbe added to the `TiKVStatus`:
```
liubog2008, line 56:
```suggestion
An EvictLeader status will be added to the `TiKVStatus`:
```

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
july2993: so the behavior of `tidb.pingcap.com/evict-leader` with any value is the same as a `none` value here.
and `tidb.pingcap.com/restart` is same as a `delete-pod` value here?

DanielZhangQD: Yes.

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,138 @@
...
+- `none`: doing nothing
+- `delete-pod`: delete pod and remove the evict-leader scheduler from PD.
+
+
+
+An EvictLeader status will be added to the `TiKVStatus`:
+
+```go
++type EvictLeaderStatus struct {
++       PodCreateTime metav1.Time `json:"podCreateTime,omitempty"`
```
DanielZhangQD, line 60: Is this field used anywhere?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,164 @@
...
++}
+
+ // TiKVStatus is TiKV status
+ type TiKVStatus struct {
+        Synced          bool                            `json:"synced,omitempty"`
+@@ -1139,6 +1151,7 @@ type TiKVStatus struct {
+        TombstoneStores map[string]TiKVStore            `json:"tombstoneStores,omitempty"`
+        FailureStores   map[string]TiKVFailureStore     `json:"failureStores,omitempty"`
+        Image           string                          `json:"image,omitempty"`
++       EvictLeader     map[string]*EvictLeaderStatus   `json:"evictLeaderStatus,omitempty"`
```
DanielZhangQD, line 84: Is `EvictLeaderStatus` useful here? If not, can we just use `EvictLeaderPods map[string]EmptyStruct `json:"evictLeaderPods,omitempty"``?

docs/design-proposals/2021-11-24-graceful-reschedule-tikv-pod.md
```diff
@@ -0,0 +1,164 @@
...
+ }
+```
+
+The `Value` of `EvictLeaderStatus` control the behavior when the leader count drops to zero, the valid value is one of:
+
+- `none`: doing nothing (for `tidb.pingcap.com/evict-leader`)
+
+- `delete-pod`: delete pod and remove the evict-leader scheduler from PD. (for `tidb.pingcap.com/restart`)
```
DanielZhangQD, line 92: Needs to update here.

DanielZhangQD (approved):
LGTM

csuzhangxc (approved):

PR #4278, opened 2021-11-25
fix backupschedule.status.lastBackup empty problem when it is the fir…

### LINKED ISSUE #4220: backupschedule two backup running at the same time
## Bug Report

v1.1.12
i think there exsits the same problem on the newest code.

**What did you do?**

1. create backupschedule, let it have some backups
2. pause backupschedule by set backupschedule.Spec.Pause = true
3. wait enough time when all the backup should be cleaned by config.backupschedule.Spec.MaxReservedTime
4. restart backupschedule by set backupschedule.Spec.Pause = false

as i do that, what i see should be like this:
1.a new backup is created because it should be schedule again
2.all old backups should be cleaned because it execeed the MaxReservedTime
3. backupschedule.Status.Last-Backup should be the new backup

**What did you see instead?**

* backupschedule.Status.Last-Backup is nil
* backupschedule.Status.All-Backup-Clean-Time is set 

because the last-backup is nil, although there is a backup is running, after next sched time, it can not check the running backup, and create a new backup again.

then two backup running together.

jiyfhust:
…st backup

Issue Number: close #4220

### Code changes

- [ ] Has Go code change

if last Backup not found in backupsList from cache, re-check from apiserver directly to make sure the Backup really not exist

### Tests

i add a log in func needResetLastBackup: 
````
func (bm *backupScheduleManager) needResetLastBackup(bs *v1alpha1.BackupSchedule, backupsList []*v1alpha1.Backup) bool {
       ...
        if errors.IsNotFound(err) {
                return true
        }
        klog.Infof("found last backup %s/%s from apiserver, not reset last backup", bs.GetNamespace(), bs.Status.LastBackup)
        return false
}
````

then create a backupschdule：
1. the tidb-controller-manager logs like this:
2. backupschedule.status.lastBackup is right set

````

I1125 10:23:31.519293       1 backup_schedule_status_updater.go:61] BackupSchedule: [tidb-test/auto-backup-test] updated successfully
I1125 10:24:01.520776       1 backup_schedule_manager.go:360] found last backup tidb-test/auto-backup-test-2021-11-25t10-24-00 from apiserver, not reset last backup
I1125 10:24:01.524844       1 backup_schedule_status_updater.go:61] BackupSchedule: [tidb-test/auto-backup-test] updated successfully



````

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

v1.2.0-rc.2.89+bbf4644cbd449d

```release-note

```


csuzhangxc:
@jiyfhust please fix the CI

jiyfhust:
> @jiyfhust please fix the CI

@csuzhangxc hi, done.

csuzhangxc (approved):

pkg/backup/backupschedule/backup_schedule_manager.go
```diff
@@ -332,12 +333,33 @@ func (bm *backupScheduleManager) backupGCByMaxBackups(bs *v1alpha1.BackupSchedul
...
+		}
+	}
+	// if last Backup not found in backupsList from cache, re-check from apiserver directly to make sure the Backup really not exist
+	_, err := bm.deps.Clientset.PingcapV1alpha1().Backups(bs.GetNamespace()).Get(context.TODO(), bs.Status.LastBackup, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		klog.Errorf("check last backup exsit from apiserver directly, err: %s", err)
+		return false
+	}
+	if errors.IsNotFound(err) {
+		return true
```
DanielZhangQD, line 358: If we find the backup in cache in L348, we return true, but here we do not find the backup in kube-apiserver here and we return false, why?

pkg/backup/backupschedule/backup_schedule_manager.go
```diff
@@ -332,12 +333,33 @@ func (bm *backupScheduleManager) backupGCByMaxBackups(bs *v1alpha1.BackupSchedul
 		klog.Infof("backup schedule %s/%s gc backup %s success", ns, bsName, backup.GetName())
 	}
 
-	if deleteCount == len(backupsList) {
+	if deleteCount == len(backupsList) && bm.needResetLastBackup(bs, backupsList) {
```
DanielZhangQD, line 336: Why we should check the last backup here?
If all the backups have been cleaned, we should reset the backupSchedule here.
The fix in this PR does not align with the issue described in https://github.com/pingcap/tidb-operator/issues/4220, in #4220, the cleanup behavior is good, so the GC behavior is good, right?
The issue occurs when we unpause the backupSchedule, right?
If yes, we have to see that if it is caused by the stale cache in the controller (not synced from kube-apiserver in time), if yes, what we can do to avoid or fix this.

pkg/backup/backupschedule/backup_schedule_manager.go
jiyfhust: needResetLastBackup func is called when all listed backup from cache is need to be gc:
1. if the cache in L348 list out the lastBackup，then the last backup is really cleaned, so return true to reset status.lastBackup
2. if step 1 false, then we check the last backup  in kube-apiserver.   and when we do not find the backup in kube-apiserver here what we return is true, not return false. L357、L358 return true.

pkg/backup/backupschedule/backup_schedule_manager.go
jiyfhust: the problem is when deleteCount == len(backupsList) condition true，then status.lastBackup is nil，there are two case when deleteCount == len(backupsList) condition true:
1. as issue described
2. the first backup of the backupscheduler. 
the problem can easily be reoccurred on a normal k8s enviarenment as the test log described.

jiyfhust: the problem is caused by the stale cache in the controller as you  describe above，and i find the re-check method to make sure it in other part like this:
https://github.com/pingcap/tidb-operator/blob/b953252754b0ecd4d6041708f8c3f0574eb4484a/pkg/manager/member/orphan_pods_cleaner.go#L132

pkg/backup/backupschedule/backup_schedule_manager.go
DanielZhangQD: The `bs.Status.LastBackup` is the last backup created before the backupSchedule is paused, and it needs to be deleted, finally, all the previous backups will be deleted, and the status will be as below with or without this PR:
```
status:
  allBackupCleanTime: "2021-12-03T13:09:56Z"
  lastBackup: ""
  lastBackupTime: null
```

So how can this PR fix the issue in #4220?
Could you please explain this in more detail?

jiyfhust: I think it occurred because when backupschedule is restart, a new backup is created, but gc does not list it out from cache，then lastBackup is set nil. 

next turn,  canPerformNextBackup func get lashBackup nil，so it skip check the  backup just created. and go on to create next scheduled backup.
https://github.com/pingcap/tidb-operator/blob/b953252754b0ecd4d6041708f8c3f0574eb4484a/pkg/backup/backupschedule/backup_schedule_manager.go#L107

i think it depend on the lastBackup is not right set.  In the issue in #4220，i think it can re-occurr lastBackup set nil，so i described a scene like that: backupschedule pause then restart.

in fact, it occurred in a not normal k8s env. I observer many times, some consecutive backup is created，but it is hard to have such a env now.

 In the issue in #4220，may be it is more appropriate to describe like this: lastbackup is not right set.

DanielZhangQD: > I think it occurred because when backupschedule is restart, a new backup is created, but gc does not list it out from cache，then lastBackup is set nil.

If this is the issue, we should check the `deleteCount>=1` here https://github.com/pingcap/tidb-operator/blob/release-1.2/pkg/backup/backupschedule/backup_schedule_manager.go#L303?

> in fact, it occurred in a not normal k8s env

I have tried this in my private k8s and I cannot reproduce this issue.
Could you please help add debug log and verify this in the k8s that you mentioned?
We need to confirm the root cause of the issue and then we can also verify if this PR really fixes this issue.
Thanks!

jiyfhust: Because it is too long after the issue occurr, there is no debug log any more to help us now. It is hard to re-occurr the k8s env, it is a problem described here: https://pingcap.com/zh/blog/fix-two-linux-kernel-bugs-while-testing-tidb-operator-in-k8s

I can re-occurr the lastBackup is nil, by setting:
Max Reserved Time:  2m 
Schedule:           */2 * * * *

tidb-operator version: v1.1.14

the backupschedule status info like this:

Status:
  All Backup Clean Time:  2021-12-06T08:30:22Z
  Last Backup:            
  Last Backup Time:       <nil>

while there exist a backup crd :

$ kubectl get backup -n tidb-test
NAME                                   STATUS     BACKUPPATH                                                             BACKUPSIZE   COMMITTS             AGE
auto-backup-test-2021-12-06t08-30-00   Complete   s3://backup/auto/tidb-for-test-pd.tidb-test-2379-2021-12-06t08-30-00   160 kB       429596193482080257   40s

In the issue in #4220, i find it is hard to re-occurr "lastBackup is nil"，because i make a mistake, ignore that the gc will always go on even if backupschedule if paused.

In version v1.1.14，we can also conclusion that a new created backup maybe can not be listed out from cache by gc。

if we only check the deleteCount>=1，there is also case like the example above，then lastBackup is set nil, but it is not commonly occurr.

jiyfhust: I1206 08:50:22.648484       1 event.go:255] Event(v1.ObjectReference{Kind:"Backup", Namespace:"tidb-test", Name:"auto-backup-test-2021-12-06t08-50-00", UID:"72e5f54d-d727-4ba7-aa7d-f05f03f73efd", APIVersion:"pingcap.com/v1alpha1", ResourceVersion:"5155046", FieldPath:""}): type: 'Normal' reason: 'SuccessfulCreate' create Backup tidb-test/auto-backup-test-2021-12-06t08-50-00 for backupSchedule/auto-backup-test successful
I1206 08:50:22.654106       1 backup_schedule_manager.go:300] backup schedule tidb-test/auto-backup-test gc backup auto-backup-test-2021-12-06t08-48-00 success
I1206 08:50:22.654213       1 event.go:255] Event(v1.ObjectReference{Kind:"Backup", Namespace:"tidb-test", Name:"auto-backup-test-2021-12-06t08-48-00", UID:"3a20ddef-78bf-4c40-ba3a-c938562be028", APIVersion:"pingcap.com/v1alpha1", ResourceVersion:"5154692", FieldPath:""}): type: 'Normal' reason: 'SuccessfulDelete' delete Backup tidb-test/auto-backup-test-2021-12-06t08-48-00 for backupSchedule/auto-backup-test successful
I1206 08:50:22.654464       1 backup_controller.go:146] backup tidb-test/auto-backup-test-2021-12-06t08-48-00 is being deleted


the tidb-controller-manager log likes above.

DanielZhangQD: @jiyfhust So we should take the following actions:
1. In https://github.com/pingcap/tidb-operator/blob/release-1.2/pkg/backup/backupschedule/backup_schedule_manager.go#L107 this function, if `bs.Status.LastBackup` is empty, we need to access the kube-apiserver to check if any backups exist for this backupSchedule, if yes, check the status as in https://github.com/pingcap/tidb-operator/blob/release-1.2/pkg/backup/backupschedule/backup_schedule_manager.go#L119
2. We should check the deleteCount>=1 here https://github.com/pingcap/tidb-operator/blob/release-1.2/pkg/backup/backupschedule/backup_schedule_manager.go#L303 because with the current implementation, if the backupSchedule is paused, the status will be reset during each sync and `allBackupCleanTime` will be updated each time, which causes useless sync.

WDYT?

jiyfhust: hi, how about we only check the deleteCount>=1 as you mentioned above:
1. it will not reset status.lastBackup even the new created backup (also the first backup of the backupschedule)  is not list out from cache. 
2. At the same time it will not causes useless sync when there is no backup, like a case: the backupSchedule is paused.

It only has a case when all backup from cache is expired and the new created backup is not list out from cache, just like the scenary i tested above, it is rarely not common.

if we  access the kube-apiserver to check if any backups exist for this backupSchedule when bs.Status.LastBackup is empty as you described: it will always access the kube-apiserver to check if there is no backup exist during each sync.

or if we can add a line like this:
func (bm *backupScheduleManager) needResetLastBackup(bs *v1alpha1.BackupSchedule, backupsList []*v1alpha1.Backup) bool {
	if len(backupsList)  == 0 {
		return false
	}
        ...
}

I think it can solve problems above this way.


In fact, i am not very good at k8s programming, do not know what is the best practice.

DanielZhangQD: > how about we only check the deleteCount>=1

OK.

DanielZhangQD (approved):
LGTM

PR #4377, opened 2022-01-19
Fixed the problem where pump would get stuck when local pds are down

### LINKED ISSUE #4361: Sync for Pump is blocked when the PDs in one Kubernetes cluster are all down in across Kubernetes deployment
## Feature Request

Sync for Pump is blocked when the PDs in one Kubernetes cluster are all down in across Kubernetes deployment, which hangs the tidbcluster reconcile process up. Possible causes are listed below:
1. `buildBinlogClient` are using the pd only in current kubernetes cluster.
https://github.com/pingcap/tidb-operator/blob/4629d6354d97a41bba9cbc8811825ce86265c0a8/pkg/manager/member/pump_member_manager.go#L140-L152
2. context won't timeout while requesting etcd.
https://github.com/pingcap/tidb-operator/blob/4629d6354d97a41bba9cbc8811825ce86265c0a8/pkg/manager/member/pump_member_manager.go#L185

just1900:
Closes #4361 

For the 2 problems mentioned in #4361 :

1. add all peermembers to endpoints when initializing etcd client. 

2. Instead of timedout every context in pumpclient, I just making the clientv3.New() to return error when underlying endpoints are not available(see https://github.com/etcd-io/etcd/issues/9877), thus to avoid the following client call get stucked indefinitely.

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [x] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fixed the problem where sync pump would get stuck when the PDs of one Kubernetes cluster are all down in across Kubernetes deployment.
```

### DIFF
--- a/pkg/binlog/binlog.go
+++ b/pkg/binlog/binlog.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pingcap/errors"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
 	"go.etcd.io/etcd/clientv3"
+	"google.golang.org/grpc"
 )
 
 // Client is the client of binlog.
@@ -46,11 +47,12 @@ func (c *Client) hookAddr(addr string) string {
 	return addr
 }
 
-// NewBinlogClient create a Client.
+// NewBinlogClient create a Client and return an error if the underlying conn is not up within 5 seconds.
 func NewBinlogClient(pdEndpoint []string, tlsConfig *tls.Config) (*Client, error) {
 	etcdClient, err := clientv3.New(clientv3.Config{
 		Endpoints:   pdEndpoint,
 		DialTimeout: time.Second * 5,
+		DialOptions: []grpc.DialOption{grpc.WithBlock()},
 		TLS:         tlsConfig,
 	})
 	if err != nil {
--- a/pkg/binlog/binlog.go
+++ b/pkg/binlog/binlog.go
@@ -33,6 +33,7 @@ type Client struct {
 	tls        *tls.Config
 	httpClient *http.Client
 	etcdClient *clientv3.Client
+	timeout    time.Duration
 
 	// if setted, will call HookAddr to change the address of pump/drainer
 	// before accessing pump/drainer.
@@ -47,11 +48,11 @@ func (c *Client) hookAddr(addr string) string {
 	return addr
 }
 
-// NewBinlogClient create a Client and return an error if the underlying conn is not up within 5 seconds.
-func NewBinlogClient(pdEndpoint []string, tlsConfig *tls.Config) (*Client, error) {
+// NewBinlogClient create a Client and return an error if the underlying conn is not up within timeout duration.
+func NewBinlogClient(pdEndpoint []string, tlsConfig *tls.Config, timeout time.Duration) (*Client, error) {
 	etcdClient, err := clientv3.New(clientv3.Config{
 		Endpoints:   pdEndpoint,
-		DialTimeout: time.Second * 5,
+		DialTimeout: timeout,
 		DialOptions: []grpc.DialOption{grpc.WithBlock()},
 		TLS:         tlsConfig,
 	})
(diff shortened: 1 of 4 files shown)


pkg/manager/member/pump_member_manager.go
```diff
@@ -149,9 +149,18 @@ func buildBinlogClient(tc *v1alpha1.TidbCluster, control pdapi.PDControlInterfac
...
 	client, err = binlog.NewBinlogClient(endpoints, tlsConfig)
-	if err != nil {
-		return nil, err
+	if err == nil {
+		return client, nil
+	}
+
+	for _, pdMember := range tc.Status.PD.PeerMembers {
+		endpoints = []string{pdMember.ClientURL}
+		client, err = binlog.NewBinlogClient(endpoints, tlsConfig)
```
csuzhangxc, line 160: Can we use all peer members here?

just1900: Nice catch! It should work, will test it out.

csuzhangxc (approved):

july2993 (approved):

pkg/binlog/binlog.go
```diff
@@ -46,11 +47,12 @@ func (c *Client) hookAddr(addr string) string {
 	return addr
 }
 
-// NewBinlogClient create a Client.
+// NewBinlogClient create a Client and return an error if the underlying conn is not up within 5 seconds.
 func NewBinlogClient(pdEndpoint []string, tlsConfig *tls.Config) (*Client, error) {
 	etcdClient, err := clientv3.New(clientv3.Config{
 		Endpoints:   pdEndpoint,
 		DialTimeout: time.Second * 5,
+		DialOptions: []grpc.DialOption{grpc.WithBlock()},
```
DanielZhangQD, line 55: Even if we succeed here, we still have to add timeout for the operations, e.g. https://github.com/pingcap/tidb-operator/blob/master/pkg/binlog/binlog.go#L181.
You can refer to the logic here https://github.com/pingcap/tidb-operator/blob/master/pkg/pdapi/pdetcd.go#L69.

DanielZhangQD: BTW, please also add timout to the HTTP client in case someone else use it later.

tests/e2e/tidbcluster/across-kubernetes.go
```diff
@@ -401,8 +401,6 @@ var _ = ginkgo.Describe("[Across Kubernetes]", func() {
 			tc1 := GetTCForAcrossKubernetes(ns1, tcName1, version, clusterDomain, nil)
 			tc2 := GetTCForAcrossKubernetes(ns2, tcName2, version, clusterDomain, tc1)
 			tc3 := GetTCForAcrossKubernetes(ns3, tcName3, version, clusterDomain, tc1)
-			// FIXME(jsut1900): remove this after #4361 get fixed.
```
DanielZhangQD, line 404: Why skip TiKV in L526?

pkg/binlog/binlog.go
DanielZhangQD: The `DialTimeout` should be OK here, right? So no need to add the `DialOptions` here, please check the example here https://pkg.go.dev/go.etcd.io/etcd/clientv3#pkg-overview and the description for `DialTimeout`.

tests/e2e/tidbcluster/across-kubernetes.go
just1900: We have failed Tikv before failing PD, though it should make no difference to restart a failed tikv pod.

pkg/binlog/binlog.go
just1900: As mentioned in this pr's comment, even though `DialTimeout` has been set, `clientv3.New()` returns immediately.
This seems like the behavior changed after v3.3  https://github.com/etcd-io/etcd/issues/9877#issuecomment-399262512.
Also checkout the comments for field `DialOptions` here. https://pkg.go.dev/go.etcd.io/etcd@v3.3.27+incompatible/clientv3#Config

tests/e2e/tidbcluster/across-kubernetes.go
DanielZhangQD: It's different, the first part is to check that Pods can restart successfully after all TiKV down, and the second part is to check that Pods can restart successfully after all PD down.

just1900: addressed in https://github.com/pingcap/tidb-operator/pull/4382

PR #4379, opened 2022-01-19
Add startUpScriptVersion for pd spec

july2993:
for version >= v.5.0.5, the nslookup version is changed in the image.

if nameserver in `/etc/resolv.conf` contains not only coredns cluster ip, PD pod will fail to start since the startup script use nslookup and the behavior of nslookup is different from general resolver.

ref: https://serverfault.com/a/398854
https://github.com/coredns/coredns/issues/3835#issuecomment-616715938

https://github.com/isc-projects/bind9/blob/808d0b07b00244de735be68ce85b2652e4e0efec/bin/dig/dighost.c#L4128-L4162
```
  if ((msg->rcode == dns_rcode_servfail && !l->servfail_stops) ||
(check_ra && (msg->flags & DNS_MESSAGEFLAG_RA) == 0 && l->recurse))
```
we will hit the `check_ra&&。。` condition here, and nslookup will try the next nameserver, and the next nameserver will return NXDOMAIN(since it is not the coredns)

support add startUpScriptVersion as v1 to use dig to wait dns can be
resolved:
...
pd:
  startUpScriptVersion: "v1"
...

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Manual test 
- deploy cluster using master version
- upgrade operator as version of the PR
- check will not trigger update of PD

---
- set `startUpScriptVersion` as v1
- check PD start success and startup script is using the v1

---
apply some invalid value
```
The TidbCluster "basic" is invalid: spec.pd.startUpScriptVersion: Unsupported value: "v11": supported values: "", "v1"
```

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/manager/member/template.go
+++ b/pkg/manager/member/template.go
@@ -160,14 +160,14 @@ then
 echo "waiting for pd cluster ready timeout" >&2
 exit 1
 fi
-
+{{ if eq .CheckDomainScript ""}}
 if nslookup ${domain} 2>/dev/null
 then
 echo "nslookup domain ${domain}.svc success"
 break
 else
 echo "nslookup domain ${domain} failed" >&2
-fi
+fi {{- else}}{{.CheckDomainScript}}{{end}}
 done
 
 ARGS="--data-dir={{ .DataDir }} \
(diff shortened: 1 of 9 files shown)


pkg/manager/member/template.go
```diff
@@ -160,14 +160,14 @@ then
 echo "waiting for pd cluster ready timeout" >&2
 exit 1
 fi
-
+{{ if eq .CheckDomainScript ""}}
```
DanielZhangQD, line 163: Will this be a new line if the value is ""?

pkg/manager/member/template.go
```diff
@@ -160,14 +160,14 @@ then
...
 if nslookup ${domain} 2>/dev/null
 then
 echo "nslookup domain ${domain}.svc success"
 break
 else
 echo "nslookup domain ${domain} failed" >&2
-fi
+fi {{- else}}{{.CheckDomainScript}}{{end}}
```
DanielZhangQD, line 170: Will the extra space after `fi` cause a rolling update?

pkg/manager/member/template.go
july2993: yes.

pkg/manager/member/template.go
july2993: no.  note  `{{-` with '-' it will remove space before it

DanielZhangQD (approved):
LGTM

csuzhangxc (approved):

PR #3944, opened 2021-04-25
optimize pd upgarde

shonge:
### What problem does this PR solve?

Close #1235 

### Code changes

- [ ] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [ ] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note

```

PR #4392, opened 2022-01-28
Choose a healthy pd to transfer leader when upgrade pd

### LINKED ISSUE #1235: Optimize tranferring pd leader

KanShiori:

Close #1235
Ref: #3944

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [x] Manual test 

Deploy tc with pd's replicas is 5

* Upgrade pd when the leader is pd-4, and check transfer leader to pd-0.
* Upgrade pd when the leader is pd-4 and pd-0 is unhealthy, and check transfer leader to pd-1.
* Upgrade pd when the leader is pd-0 and check transfer leader to pd-4.

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/manager/member/pd_upgrader.go
+++ b/pkg/manager/member/pd_upgrader.go
@@ -112,14 +112,17 @@ func (u *pdUpgrader) upgradePDPod(tc *v1alpha1.TidbCluster, ordinal int32, newSe
 
 	// If current pd is leader, transfer leader to other pd
 	if tc.Status.PD.Leader.Name == upgradePdName || tc.Status.PD.Leader.Name == upgradePodName {
-		var targetName string
+		targetName := ""
+
 		if tc.PDStsActualReplicas() > 1 {
 			targetName = choosePDToTransferFromMembers(tc, newSet, ordinal)
-		} else {
+		}
+
+		if targetName == "" {
 			targetName = choosePDToTransferFromPeerMembers(tc, upgradePdName)
 		}
 
-		if len(targetName) > 0 {
+		if targetName != "" {
 			err := u.transferPDLeaderTo(tc, targetName)
 			if err != nil {
 				klog.Errorf("pd upgrader: failed to transfer pd leader to: %s, %v", targetName, err)
--- a/pkg/manager/member/pd_upgrader.go
+++ b/pkg/manager/member/pd_upgrader.go
@@ -109,25 +109,16 @@ func (u *pdUpgrader) upgradePDPod(tc *v1alpha1.TidbCluster, ordinal int32, newSe
 	tcName := tc.GetName()
 	upgradePdName := PdName(tcName, ordinal, tc.Namespace, tc.Spec.ClusterDomain)
 	upgradePodName := PdPodName(tcName, ordinal)
+
+	// If current pd is leader, transfer leader to other pd
 	if tc.Status.PD.Leader.Name == upgradePdName || tc.Status.PD.Leader.Name == upgradePodName {
 		var targetName string
 		if tc.PDStsActualReplicas() > 1 {
-			targetOrdinal := helper.GetMaxPodOrdinal(*newSet.Spec.Replicas, newSet)
-			if ordinal == targetOrdinal {
-				targetOrdinal = helper.GetMinPodOrdinal(*newSet.Spec.Replicas, newSet)
-			}
-			targetName = PdName(tcName, targetOrdinal, tc.Namespace, tc.Spec.ClusterDomain)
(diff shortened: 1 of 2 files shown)


csuzhangxc (approved):

july2993 (approved):

pkg/manager/member/pd_upgrader.go
```diff
@@ -109,25 +109,16 @@ func (u *pdUpgrader) upgradePDPod(tc *v1alpha1.TidbCluster, ordinal int32, newSe
...
 		if tc.PDStsActualReplicas() > 1 {
-			targetOrdinal := helper.GetMaxPodOrdinal(*newSet.Spec.Replicas, newSet)
-			if ordinal == targetOrdinal {
-				targetOrdinal = helper.GetMinPodOrdinal(*newSet.Spec.Replicas, newSet)
-			}
-			targetName = PdName(tcName, targetOrdinal, tc.Namespace, tc.Spec.ClusterDomain)
-			if _, exist := tc.Status.PD.Members[targetName]; !exist {
-				targetName = PdPodName(tcName, targetOrdinal)
-			}
+			targetName = choosePDToTransferFromMembers(tc, newSet, ordinal)
```
DanielZhangQD, line 117: Should we also check the peer members if no member is available?

pkg/manager/member/pd_upgrader.go
```diff
@@ -146,6 +140,68 @@ func (u *pdUpgrader) transferPDLeaderTo(tc *v1alpha1.TidbCluster, targetName str
...
+	ordinals := helper.GetPodOrdinals(*newSet.Spec.Replicas, newSet)
+
+	genPDName := func(targetOrdinal int32) string {
+		pdName := PdName(tcName, targetOrdinal, tc.Namespace, tc.Spec.ClusterDomain)
+		if _, exist := tc.Status.PD.Members[pdName]; !exist {
+			pdName = PdPodName(tcName, targetOrdinal)
+		}
+		return pdName
+	}
+	pred := func(pdName string) bool {
```
DanielZhangQD, line 159: Is it better to replace `pred` with `healthy`？

pkg/manager/member/pd_upgrader.go
```diff
@@ -146,6 +140,68 @@ func (u *pdUpgrader) transferPDLeaderTo(tc *v1alpha1.TidbCluster, targetName str
...
+		return tc.Status.PD.Members[pdName].Health
+	}
+
+	// set ordinal to max ordinal if ordinal isn't exist
+	if !ordinals.Has(ordinal) {
```
DanielZhangQD, line 164: What's the case here? Does it mean a Pod that is not desired is being upgraded?

pkg/manager/member/pd_upgrader.go
KanShiori: LTGM, can we merge the functions `choosePDToTransferFromPeerMembers` and `choosePDToTransferFromPeerMembers` ?

pkg/manager/member/pd_upgrader.go
KanShiori: `healthy` is a filter, and we can add more conditions

pkg/manager/member/pd_upgrader.go
KanShiori: it's mean that current sts don't contain the pod, and regard the pod as the max ordinal, because all pods in sts haven't been upgraded

pkg/manager/member/pd_upgrader.go
KanShiori: 1. Find the max suitable ordinal in (x, n], because they have been upgraded
2. If no suitable ordinal, find the min suitable ordinal in [0, x) to reduce the count of transfer
3. If no suitable ordinal, find the peer member

KanShiori: Done

Issue #4467, opened 2022-03-10
Consider enable prometheusReloader and reloader by default


just1900:
## Feature Request

The feature of reloading prometheus rules and config files are commonly needed, can we enable it by default rather than configure it in `tidbmonitor.spec`?

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**
enable PrometheusReloader and reloader by default.

**Describe alternatives you've considered:**

- merge prometheusReloader and reloader's function into one container to reload rules and configs.(this is what prometheus-config-reloader is doing now) But we should consider that [reloader](https://github.com/pingcap/monitoring/blob/master/README.md#function2---prometheus-rule-reloader) also provide an ui for rules configuration

**Teachability, Documentation, Adoption, Migration Strategy:**

mikechengwei:
Operator upgrade will cause tidbmonitor to restart. I think this effect is okay.

PR #4520, opened 2022-05-11
enable prometheus reloader rule default

mikechengwei:
#4467
Enable prometheus reloader image default.
### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Enable prometheus reloader image default.
```

### DIFF
--- a/pkg/monitor/monitor/util.go
+++ b/pkg/monitor/monitor/util.go
@@ -726,9 +726,20 @@ func getMonitorGrafanaContainer(secret *core.Secret, monitor *v1alpha1.TidbMonit
 }
 
 func getMonitorPrometheusReloaderContainer(monitor *v1alpha1.TidbMonitor, shard int32) core.Container {
+	image := "quay.io/prometheus-operator/prometheus-config-reloader"
+	version := "v0.49.0"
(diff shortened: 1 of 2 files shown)


pkg/monitor/monitor/util.go
```diff
@@ -726,9 +726,20 @@ func getMonitorGrafanaContainer(secret *core.Secret, monitor *v1alpha1.TidbMonit
 }
 
 func getMonitorPrometheusReloaderContainer(monitor *v1alpha1.TidbMonitor, shard int32) core.Container {
+	image := "quay.io/prometheus-operator/prometheus-config-reloader"
```
DanielZhangQD, line 729: Some users cannot access this image, so after upgrading the TiDB Operator, their tidb monitors will fail to start, I think it's OK that we add the reloader configuration in the examples but not enforce this in the code.

mikechengwei: ok，you‘re right.

PR #3296, opened 2020-09-22
Remove the PodRestarter controller and `tidb.pingcap.com/pod-defer-deleting` annotation

lonng:
### What problem does this PR solve? 

This `PodRestarter` is used to restart a pod by annotating it with `tidb.pingcap.com/pod-defer-deleting`. However, there is no user to use this feature because we recommandate users to restart pods gracefully via instrucations https://docs.pingcap.com/tidb-in-kubernetes/dev/restart-a-tidb-cluster#performing-a-graceful-rolling-restart-to-all-pods-in-a-component.  The `PodRestarter` control will retrieve all pods belongs to the current TiDBCluster instance when every `sync` iteration and delete some of it, I think there is unnecessary cost.

### What is changed and how does it work?

Remove the `PodRestarter`

### Check List 

Tests 

 - Unit test
 - E2E test

Code changes

 - Has Go code change

### Does this PR introduce a user-facing change?:

```release-note
Remove the PodRestarter controller and `tidb.pingcap.com/pod-defer-deleting` annotation
```

PR #4521, opened 2022-05-12
Remove unused or meaningless tests

sangheee:

I was looking into the chaos test.
The [chaos action](https://github.com/pingcap/tidb-operator/blob/master/.github/workflows/chaos.yaml) of tidb-operator was filtering `--ginkgo.focus Restarter`.
However, `Restarter` test was removed from #3296 because of its inefficiency, and in fact, chaos action did not do any meaningful work other than build and deploy.

Looking around, I found some similar tests that don't actually do anything meaningful.
They were always failing because of incorrect syntax or not passing parameters properly.
I think there are unnecessary and even better remove them.

Remove tests related `Restarter` and always throwing fail.

### Code changes

- [ ] Has Go code change
- [X] Has CI related scripts change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [X] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


DanielZhangQD:
@sangheee Thanks for submitting this PR.
I think we can leave the files in the repo now, we may support testing on those environments later.
Is it OK for you?

sangheee:
@DanielZhangQD 
Does it make sense to leave some tests that always fail? 
Even if leave them, I wonder if you have any plans to support these and when will you do that.
There doesn't seem to be any issues with it.

DanielZhangQD:
> @DanielZhangQD Does it make sense to leave some tests that always fail? 
> Even if leave them, I wonder if you have any plans to support these and when will you do that. 
> There doesn't seem to be any issues with it.

@sangheee Sorry that supporting those tests is not our priority yet.
Could you please help create issues to trace this?
Thanks!

PR #4603, opened 2022-06-28
Resize pvc one by one

KanShiori:

* Add `resizingCount` field in volume status
* Before resize, check whether `resizingCount` is zero
* Resize only one PVC one time

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 

Resize tikv volume and check whether resize PVC one by one.

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Resize pvc one by one
```

### DIFF
--- a/pkg/manager/member/pvc_resizer.go
+++ b/pkg/manager/member/pvc_resizer.go
@@ -94,6 +94,8 @@ type componentVolumeContext struct {
 	desiredVolumeQuantity map[v1alpha1.StorageVolumeName]resource.Quantity
 	// actualPodVolumes is the actual status for all volumes
 	actualPodVolumes []*podVolumeContext
+	// observedStatus is current volume status assembled from `actualPodVolumes`
+	observedStatus map[v1alpha1.StorageVolumeName]*v1alpha1.ObservedStorageVolumeStatus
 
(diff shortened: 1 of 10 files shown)


pkg/manager/member/pvc_resizer.go
```diff
@@ -460,16 +502,16 @@ func (p *pvcResizer) resizeVolumes(ctx *componentVolumeContext) error {
 				return err
 			}
 
-			klog.V(2).Infof("PVC %s/%s storage request is updated from %s to %s",
-				pvc.Namespace, pvc.Name, currentRequest.String(), quantityInSpec.String())
+			klog.V(2).Infof("PVC %q of TC %q storage request is updated from %s to %s", pvcID, tcID, currentRequest.String(), quantityInSpec.String())
+			break // resize pvc one by one
```
DanielZhangQD, line 506: With multiple PVC case, is it possible that tikv-1 scales pvc0 and tikv-2 scales pvc1?

DanielZhangQD: What we are trying to do is to scale all the PVCs for one Pod only for each component.

PR #4604, opened 2022-06-29
Resize PVCs pod by pod

KanShiori:
### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 

1. Deploy tc and tikv have three volumes
2. Scale up three volumes
3. Check the resize is pod by pod

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Resize PVCs pod by pod 
```

### DIFF
--- a/pkg/manager/member/pvc_resizer.go
+++ b/pkg/manager/member/pvc_resizer.go
@@ -529,7 +529,6 @@ func (p *pvcResizer) collectAcutalStatus(ns string, selector labels.Selector) ([
 	}
 
 	for _, pod := range pods {
-		pods := pod.DeepCopy()
 		volToPVCs := map[v1alpha1.StorageVolumeName]*corev1.PersistentVolumeClaim{}
 
 		for _, vol := range pod.Spec.Volumes {
--- a/pkg/manager/member/pvc_resizer.go
+++ b/pkg/manager/member/pvc_resizer.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/pingcap/tidb-operator/pkg/apis/label"
 	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
@@ -87,7 +88,10 @@ type podVolumeContext struct {
 }
 
 type componentVolumeContext struct {
-	comp v1alpha1.MemberType
+	comp      v1alpha1.MemberType
+	namespace string
+	name      string
+
 	// label selector for pvc and pod
 	selector labels.Selector
 	// desiredVolumeSpec is the volume request in tc spec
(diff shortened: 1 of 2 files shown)


pkg/manager/member/pvc_resizer.go
```diff
@@ -395,51 +404,75 @@ func (p *pvcResizer) resizeVolumes(ctx *componentVolumeContext) error {
...
+		allResized := true
+
 		for volName, pvc := range podVolume.volToPVCs {
+			pvcID := fmt.Sprintf("%s/%s", pvc.Namespace, pvc.Name)
+
+			// check whether the PVC is resized
 			quantityInSpec, exist := desiredVolumeQuantity[volName]
 			if !exist {
-				klog.Warningf("PVC %s/%s does not exist in desired volumes", pvc.Namespace, pvc.Name)
+				klog.Errorf("Check PVC %q of TC %q resized failed: not exist in desired volumes", pvcID, tcID)
```
DanielZhangQD, line 419: It can be TC or DC.

pkg/manager/member/pvc_resizer.go
```diff
@@ -395,51 +404,75 @@ func (p *pvcResizer) resizeVolumes(ctx *componentVolumeContext) error {
...
+					klog.Infof("PVC %q of TC %q is resizing, request: %s capcaity: %s",
+						pvcID, tcID, currentRequest.String(), currentCapacity.String())
+				} else {
+					klog.V(4).Infof("PVC %q of TC %q is already resized, request: %s",
+						pvcID, tcID, quantityInSpec.String())
+				}
 				continue
 			}
+
+			allResized = false
```
DanielZhangQD, line 447: Why set to false here? no update to the PVC if `cmpVal < 0`, right?

KanShiori: It means the PVC need to resize, so `allResize` is set to false and don't resize other PVCs of other Pods

KanShiori: moved it after checking, PTAL

pkg/manager/member/pvc_resizer.go
```diff
@@ -529,6 +529,7 @@ func (p *pvcResizer) collectAcutalStatus(ns string, selector labels.Selector) ([
 	}
 
 	for _, pod := range pods {
+		pods := pod.DeepCopy()
```
DanielZhangQD, line 532: What's this for?

pkg/manager/member/pvc_resizer.go
```diff
@@ -466,15 +464,17 @@ func (p *pvcResizer) resizeVolumes(ctx *componentVolumeContext) error {
...
 					continue
 				}
 			} else {
-				klog.V(4).Infof("Storage classes lister is unavailable, skip checking volume expansion support for PVC %q of TC %q with storage class %s. This may be caused by no relevant permissions",
-					pvcID, tcID, *pvc.Spec.StorageClassName)
+				klog.V(4).Infof("Storage classes lister is unavailable, skip checking volume expansion support for PVC %q of %q with storage class %s. This may be caused by no relevant permissions",
+					pvcID, id, *pvc.Spec.StorageClassName)
 			}
 
+			allResized = false
```
DanielZhangQD, line 476: Move to L491?

pkg/manager/member/pvc_resizer.go
KanShiori: a mistake..

pkg/manager/member/pvc_resizer.go
KanShiori: It is not a check, so I think it's ok.
```go
			mergePatch, err := json.Marshal(map[string]interface{}{
				"spec": map[string]interface{}{
					"resources": corev1.ResourceRequirements{
						Requests: corev1.ResourceList{
							corev1.ResourceStorage: quantityInSpec,
						},
					},
				},
			})
```

DanielZhangQD: or L495

KanShiori: arrive here, it means we should resize the PVC, and the error of `json.Marshal` can be regarded as failed to resize

DanielZhangQD: But actually, before L495, this will be no use because the code will return, and no chance to check this var.

DanielZhangQD (approved):
LGTM

PR #4608, opened 2022-06-30
fix: error in the log when no pd member fails

hoyhbx:

When failovering a broken PD member, [this line](https://github.com/pingcap/tidb-operator/blob/d5ca267ad940d7037954d2185cbe60a023954f3b/pkg/manager/member/pd_failover.go#L70) gives us an error even if no PD member fails (i.e. `pdDeletedFailureReplicas=0`). When we inspect the error log, we find that the error message “**PD failover replicas (0) reaches the limit (0), skip failover**” is constantly written to the operator log and is very confusing to users. However, there is actually no error if no PD member fails. We believe when `pdDeletedFailureReplicas=0`, the operator should not throw the error message “**PD failover replicas (0) reaches the limit (0), skip failover**“.
This PR provides a straightforward fix for this problem.

We fix it by adding one more condition in the if statement in [this line](https://github.com/pingcap/tidb-operator/blob/d5ca267ad940d7037954d2185cbe60a023954f3b/pkg/manager/member/pd_failover.go#L70) to check whether `pdDeletedFailureReplicas` is larger than 0, as illustrated below:
```golang
if pdDeletedFailureReplicas >= *tc.Spec.PD.MaxFailoverCount && pdDeletedFailureReplicas > 0 {
	klog.Errorf(“PD failover replicas (%d) reaches the limit (%d), skip failover”, pdDeletedFailureReplicas, *tc.Spec.PD.MaxFailoverCount)
	return nil
}
```
### Code changes
- [x] Has Go code change
### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 
This is a simple fix and we suppose no test above is needed.
### Side effects
- [ ] Other side effects: 
### Related changes
### Release Notes

```release-note
NONE
```


pkg/manager/member/pd_failover.go
```diff
@@ -66,7 +66,7 @@ func (f *pdFailover) Failover(tc *v1alpha1.TidbCluster) error {
 	}
 
 	pdDeletedFailureReplicas := tc.GetPDDeletedFailureReplicas()
-	if pdDeletedFailureReplicas >= *tc.Spec.PD.MaxFailoverCount {
+	if pdDeletedFailureReplicas >= *tc.Spec.PD.MaxFailoverCount && pdDeletedFailureReplicas > 0 {
```
DanielZhangQD, line 69: For your case, you have set `Spec.PD.MaxFailoverCount=0`, right? Which means that you want to disable failover for PD, in this case, I think we should update the code here https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/pd_member_manager.go#L241 as below:
`if m.deps.CLIConfig.AutoFailover && tc.Spec.PD.MaxFailoverCount != nil && *tc.Spec.PD.MaxFailoverCount > 0 {`
WDYT?

hoyhbx: @DanielZhangQD Thanks for your reply. Yes, we think that your suggestion is correct, and we have updated the code in the latest commit in this PR now.

DanielZhangQD (approved):
LGTM

KanShiori:
What happens if failover occurs and set the max failover count to 0?

pkg/manager/member/pd_failover.go
KanShiori: What happens if failover occurs and set the max failover count to 0?

pkg/manager/member/pd_member_manager.go
```diff
@@ -238,7 +238,7 @@ func (m *pdMemberManager) syncPDStatefulSetForTidbCluster(tc *v1alpha1.TidbClust
 		return err
 	}
 
-	if m.deps.CLIConfig.AutoFailover {
+	if m.deps.CLIConfig.AutoFailover && tc.Spec.PD.MaxFailoverCount != nil && *tc.Spec.PD.MaxFailoverCount > 0 {
```
KanShiori, line 241: What happens if failover occurs and set the max failover count to 0?

KanShiori: can not arrive the `Recover` func?

DanielZhangQD: The failover Pods will not be scaled in and the user needs to remove the failure member manually in the status.pd then the failover Pods will be scaled in.

DanielZhangQD: @hoyhbx After checking the logic of the other component, I think @KanShiori 's above comment is reasonable, we should be able to recover the previous failover Pods even if we set MaxFailoverCount to 0.

DanielZhangQD: Could you please add the check in L244?

hoyhbx: Sure, I have changed it. @DanielZhangQD

pkg/manager/member/pd_member_manager.go
```diff
@@ -241,7 +241,7 @@ func (m *pdMemberManager) syncPDStatefulSetForTidbCluster(tc *v1alpha1.TidbClust
 	if m.deps.CLIConfig.AutoFailover {
 		if m.shouldRecover(tc) {
 			m.failover.Recover(tc)
-		} else if tc.PDAllPodsStarted() && !tc.PDAllMembersReady() || tc.PDAutoFailovering() {
+		} else if (tc.PDAllPodsStarted() && !tc.PDAllMembersReady() || tc.PDAutoFailovering()) && tc.Spec.PD.MaxFailoverCount != nil && *tc.Spec.PD.MaxFailoverCount > 0 {
```
DanielZhangQD, line 244:
```suggestion
		} else if tc.Spec.PD.MaxFailoverCount != nil && *tc.Spec.PD.MaxFailoverCount > 0 && (tc.PDAllPodsStarted() && !tc.PDAllMembersReady() || tc.PDAutoFailovering()) {
```

KanShiori (approved):

PR #4617, opened 2022-07-08
Fix the issue that backup files may leak after cleanup

KanShiori:

1. Add a field `retryCount`, and retry to clean backup files when one cleanup fails
2. Try to clean all files in one cleanup instead of returning when met any errors.
3. Add a field `backoffEnabled` to enable backoff when a deletion API fails. It may be useful when the deletion API is rate limited.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [x] Manual test 

1. Create a Backup and add 40w files to the bucket.
2. Delete Backup.
3. Check all files are deleted.

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fix the issue that backup files leak after cleanup
```


liubog2008 (approved):
LGTM

cmd/backup-manager/app/util/util.go
```diff
@@ -431,3 +432,22 @@ func GetContextForTerminationSignals(op string) (context.Context, context.Cancel
...
+	var lastErr error
+	for i := 0; i < attempts; i++ {
+		err := fn()
+		if err == nil {
+			return nil
+		}
+
+		lastErr = err
+		if sleep != 0 {
+			time.Sleep(sleep)
```
csuzhangxc, line 448: do we need to support cancel with a context?

KanShiori: add a callback `retriable func(error) bool` to support cancel, so we can cancel with context in `retriable`.

csuzhangxc: but we `RetriableOnAnyError` now?

KanShiori: if the context is done, the delete function will fail fastly, so I think it is ok.

tools/tools.go
```diff
@@ -11,6 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build tools
```
DanielZhangQD, line 14: Remove this?

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1635,6 +1635,12 @@ type CleanOption struct {
 	// PageSize represents the number of objects to clean at a time.
 	// default is 10000
 	PageSize uint64 `json:"pageSize,omitempty"`
+	// RetryCount represents the number of retries in pod when the cleanup fails.
+	// +kubebuilder:default=3
+	RetryCount int `json:"retryCount,omitempty"`
+	// BackoffEnabled represents whether to enable the backoff when a deletion API fails.
+	// It is useful when the deletion API is rate limited.
+	BackoffEnabled bool `json:"backoffEnabled,omitempty"`
```
DanielZhangQD, line 1643: Should we enable this by default?

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -30,6 +30,7 @@ var (
 	// defaultCleanOption is default clean option
 	defaultCleanOption = CleanOption{
 		PageSize:          10000,
+		RetryCount:        3,
```
DanielZhangQD, line 33: Set it to 10 as we want to delete the files as much as possible?

cmd/backup-manager/app/util/util.go
```diff
@@ -431,3 +432,26 @@ func GetContextForTerminationSignals(op string) (context.Context, context.Cancel
...
 	return ctx, cancel
 }
+
+func RetriableOnAnyError(err error) bool {
+	return err != nil
+}
+
+// RetryOnError allows the caller to retry fn in case the error returned by fn.
+// sleep define the interval between two retries.
+func RetryOnError(attempts int, sleep time.Duration, retriable func(error) bool, fn func() error) error {
```
DanielZhangQD, line 442: The `sleep` is always 0?

pkg/apis/pingcap/v1alpha1/backup.go
KanShiori: the backoff of the clean job is 5, so the total retry count is 15, so I think it is enough. And clients can modify the option if needed

cmd/backup-manager/app/util/util.go
KanShiori: Updated

pkg/apis/pingcap/v1alpha1/types.go
KanShiori: It is useful when API is rate limited, and not suitable for all clients. So I think disable by default.

tools/tools.go
KanShiori: Done

pkg/apis/pingcap/v1alpha1/backup.go
DanielZhangQD: But it's more lightweight to retry in the current Pod instead of waiting for another Pod, right?

KanShiori: default to 5

DanielZhangQD (approved):
LGTM

csuzhangxc (approved):

PR #4628, opened 2022-07-14
Mark CRD as chart

nobiit:
Allows to install CRD as a chart, can add it as a dependency or automatically install/upgrade the latest version by helm install tidb-crd pingcap/tidb-crd

### Code changes

- [ ] Has Go code change
- [x] Has CI related scripts change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


DanielZhangQD:
@nobiit What's the benefit to make it a separate chart?
With the current yaml file, we still can manage it with one command line.

nobiit:
Hi @DanielZhangQD , If it is a standalone helm chart, I can declare it as a dependency in another helm chart (eg tidb). And with just one command helm install tidb will install both CRD and tidb, it will help newbies be less confused, make it easier for everyone to update their CRD to the latest version

DanielZhangQD:
CRD is a global resource, and TiDB is not, BTW, tidb-cluster chart has been deprecated since TiDB Operator v1.1, and we now use TidbCluster CR to manage the TiDB cluster.

nobiit:
Hi @DanielZhangQD, I understand that there is a lot of complicated logic that makes you not want to use helm chart because it has many limitations. Why don't we use helm chart to create a fast TiDB cluster for demo/testing purposes and use TidbCluster CR for advanced tasks or production deployment? Is that possible?

Users are not always willing to install a binary file and run it without knowing what it does. And when there is an error, users will be very confused

DanielZhangQD:
> Why don't we use helm chart to create a fast TiDB cluster for demo/testing purposes and use TidbCluster CR for advanced tasks or production deployment? Is that possible?

Users can apply the yamls in the `examples` dir to create a TiDB cluster fast.
BTW, with the helm chart, which fields should be exposed to the users to configure in the values.yaml, with more and more users, we eventually need to expose all the fields in the CR in the values.yaml because different users have different requirements (this is the case with the tidb-cluster chart before), in this case, what's the benefit of using helm chart against the CR?

> Users are not always willing to install a binary file and run it without knowing what it does. And when there is an error, users will be very confused

What does this mean? It's not related to helm chart, right?

Issue #4632, opened 2022-07-17
operator is not able to create pump pods when `spec.pump.resources.request.storage` is missing


hoyhbx:
## Bug Report

**What version of Kubernetes are you using?**

```shell
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-03T13:46:05Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: [version.Info]([[http://version.info/){Major](http://version.info/)%7BMajor)](http://version.info/)%7BMajor](http://version.info/)%7BMajor)):"1", Minor:"22", GitVersion:"v1.22.9", GitCommit:"6df4433e288edc9c40c2e344eb336f63fad45cd2", GitTreeState:"clean", BuildDate:"2022-05-19T19:53:08Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of TiDB Operator are you using?**

```shell
TiDB Operator Version: [version.Info]([[http://version.info/){GitVersion](http://version.info/)%7BGitVersion)](http://version.info/)%7BGitVersion](http://version.info/)%7BGitVersion)):"v1.3.0-45+1470cfb46e1ffb-dirty", GitCommit:"1470cfb46e1ffb8bb86f74ba455865a95b825413", GitTreeState:"dirty", BuildDate:"2022-07-17T16:54:00Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}
```

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

```shell
$ kubectl get sc
NAME                 PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
standard (default)   [rancher.io/local-path](http://rancher.io/local-path)   Delete          WaitForFirstConsumer   false                  40m

$ kubectl get pvc -n {tidb-cluster-namespace}
NAME                        STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
pd-advanced-tidb-pd-0       Bound    pvc-b566858b-bf4e-4e33-b31e-5d7feb7397b1   10Gi       RWO            standard       10m
[...]

PR #4633, opened 2022-07-17
validate `spec.pump.resources.request.storage`

hoyhbx:

We found out that when we specify pump spec with a positive `replicas` but without assigning a valid storage to it, we got rejection from the events saying that `storage` is a required field. However, we did not get error messages or alerts from the operator indicating that `storage` is required. Additionally, we checked the crd and found that `replicas` is the only required field. 

Related issue: https://github.com/pingcap/tidb-operator/issues/4632

We believed it is necessary to check whether `storage` is set when pump spec is not empty. We added a sanity check in the function `validatePumpSpec` to see whether `storage` exists in the resource requests. 

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Add validation to ensure `storage` resource in `spec.pump.resources.request`
```


pkg/apis/pingcap/v1alpha1/validation/validation.go
```diff
@@ -275,6 +275,11 @@ func validateTiDBSpec(spec *v1alpha1.TiDBSpec, fldPath *field.Path) field.ErrorL
 func validatePumpSpec(spec *v1alpha1.PumpSpec, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	allErrs = append(allErrs, validateComponentSpec(&spec.ComponentSpec, fldPath)...)
+	// fix pd spec
```
DanielZhangQD, line 278:
```suggestion
	// fix pump spec
```

DanielZhangQD (approved):
LGTM

KanShiori (approved):

PR #4650, opened 2022-07-27
fix: avoid silent rejection to invalid `podManagementPolicy`

hoyhbx:

This PR solves the problem mentioned in https://github.com/pingcap/tidb-operator/issues/4649. Basically, we believe an invalid `podManagementPolicy` should not be silently replaced by some default values without explicitly telling the user that the input is invalid.

We added a warning to explicitly tell the user that `podManagement` is invalid and the operator will use `Parallel` instead.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


pkg/apis/pingcap/v1alpha1/tidbcluster_component.go
```diff
@@ -118,7 +119,10 @@ func (a *componentAccessorImpl) PodManagementPolicy() apps.PodManagementPolicyTy
 	// unified podManagementPolicy check to avoid check everywhere
 	if policy == apps.OrderedReadyPodManagement {
 		return apps.OrderedReadyPodManagement
+	} else if policy == apps.ParallelPodManagement {
+		return apps.ParallelPodManagement
 	}
+	klog.Warningf("PodManagementPolicy \"%s\" is invalid. Fall back to use \"Parallel\" instead\n", policy)
```
DanielZhangQD, line 125: Usually, we do not add log in this file, as you can see, no klog imported in this file, each function will be called repeatedly during each sync which will flood too many logs, what about adding descriptions in types.go to describe this behavior clearly.

hoyhbx: @DanielZhangQD Thank you for your reply. Yes, I think it makes sense not to put the log in this file. But I suppose we can add an `enum` for kubebuilder in types.go to prevent invalid input. I updated the code in the latest commit.

pkg/apis/pingcap/v1alpha1/tidbcluster_component.go
```diff
@@ -118,6 +118,8 @@ func (a *componentAccessorImpl) PodManagementPolicy() apps.PodManagementPolicyTy
 	// unified podManagementPolicy check to avoid check everywhere
 	if policy == apps.OrderedReadyPodManagement {
 		return apps.OrderedReadyPodManagement
+	} else if policy == apps.ParallelPodManagement {
```
DanielZhangQD, line 121: This is not necessary, right？

DanielZhangQD:
@hoyhbx Please rebase this PR, thanks!

pkg/apis/pingcap/v1alpha1/tidbcluster_component.go
hoyhbx: Yes, if we use `enum` then this is not necessary. I have removed it in the latest commit.

hoyhbx:
> @hoyhbx Please rebase this PR, thanks!

@DanielZhangQD It's done now.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -324,6 +324,7 @@ type TidbClusterSpec struct {
 	StatefulSetUpdateStrategy apps.StatefulSetUpdateStrategyType `json:"statefulSetUpdateStrategy,omitempty"`
 
 	// PodManagementPolicy of TiDB cluster StatefulSets
+	// +kubebuilder:validation:Enum=OrderedReady;Parallel
```
DanielZhangQD, line 327: Does this work if this field is not set explicitly?

hoyhbx: @DanielZhangQD Do you mean we need a default value for this field? If so, what is the default value?

PR #4685, opened 2022-08-24
Fix log level

hoyhbx:

This PR solves the problem mentioned in #4684. Basically, we propose to use ERROR logs that report erroneous user input when rejecting them. In principle, any errors should be logged with ERROR (user error is no exception). Currently, INFO logs are used which makes it very difficult to identify those errors from a large number of INFO logs. This PR fixes the usage of the level of the log.

We changed the level of `klog` from `Infof` to `Errorf` where the log to be printed contains an error (`err` in the code), while `Infof` is used for the log.

### Code changes

- [x] Has Go code change

### Tests

We believe this fix is straightforward since only the log level is changed and no test is needed.

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


liubog2008:
I'm not sure whether you still want to fix it. If so, reopen it.

PR #4707, opened 2022-09-16
Not suspend a component when the component is scaling or upgrading

KanShiori:

If a component is scaling or upgrading, the statefulset may be upgrading, so we can' delete the sts at this time.

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 

Suspend a cluster when tikv is upgrading:
1. Deploy a cluster and upgrade tikv.
2. Suspend this cluster.
3. Check that all pods of this cluster are deleted. and tikv pods are deleted after the upgrade complete

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


pkg/manager/suspender/suspender.go
```diff
@@ -223,6 +223,12 @@ func needsSuspendComponent(cluster v1alpha1.Cluster, comp v1alpha1.MemberType) b
 
 // canSuspendComponent checks whether suspender can start to suspend the component
 func canSuspendComponent(cluster v1alpha1.Cluster, comp v1alpha1.MemberType) (bool, string) {
+	// only support to suspend Normal or Suspend cluster
+	// If the cluster is Upgrading or Scaling, the sts can not be deleted.
+	if !cluster.ComponentIsNormal(comp) && !cluster.ComponentIsSuspending(comp) {
+		return false, "component phase is not Normal or Suspend"
```
csuzhangxc, line 229:
```suggestion
		return false, "component phase is not Normal or Suspending"
```

?

csuzhangxc (approved):

pkg/manager/suspender/suspender.go
KanShiori: the value of `SuspendPhase` is "Suspend" because it contain suspended status and suspending status. So this is "Suspend"

PR #4721, opened 2022-09-23
fix(pd): fix the pd name of tc with across k8s opt

liubog2008:

Now if `spec.acrossK8s` is set but `spec.clusterDomain` is empty in TC, the upgrading process of PD will be blocked because the name of pd in status is mismatched with the return result of function `PdName`.

For now, if `spec.clusterDomain` is empty, the `PdName` will return pod name whether `spec.acrossK8s` is true or false. However, if `spec.acrossK8s` is true, the real pd name generated by start-script is the full domain name.

After this PR merged, `PdName` will also return full domain name when `spec.acrossK8s` is true

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


pkg/manager/member/utils.go
```diff
@@ -140,10 +141,16 @@ func DMMasterPodName(dcName string, ordinal int32) string {
...
+func PdName(tcName string, ordinal int32, namespace string, clusterDomain string, acrossK8s bool) string {
 	if len(clusterDomain) > 0 {
 		return fmt.Sprintf("%s.%s-pd-peer.%s.svc.%s", PdPodName(tcName, ordinal), tcName, namespace, clusterDomain)
 	}
+
+	// clusterDomain is not set
+	if acrossK8s {
+		return fmt.Sprintf("%s.%s-pd-peer.%s.svc", PdPodName(tcName, ordinal), tcName, namespace)
+	}
+
```
KanShiori, line 153: how about adding a comment to reference pd name in the start script?

KanShiori (approved):

pkg/manager/member/utils.go
liubog2008: OK

csuzhangxc (approved):

Issue #4546, opened 2022-06-09
Can't scale in TiKV from 3 to 2.


jaggerwang:
## Bug Report

**What version of Kubernetes are you using?**
1.21

**What version of TiDB Operator are you using?**
1.2

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**
```
$ kubectl get sc
NAME                             PROVISIONER                       RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
alibabacloud-cnfs-nas            nasplugin.csi.alibabacloud.com    Delete          Immediate              true                   119d
alicloud-disk-available          diskplugin.csi.alibabacloud.com   Delete          Immediate              true                   119d
alicloud-disk-efficiency         diskplugin.csi.alibabacloud.com   Delete          Immediate              true                   119d
alicloud-disk-essd               diskplugin.csi.alibabacloud.com   Delete          Immediate              true                   119d
alicloud-disk-logging            diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   68d
alicloud-disk-minio-log          diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   82d
alicloud-disk-minio-monitoring   diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   82d
alicloud-disk-minio-storage      diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   82d
alicloud-disk-monitoring         diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   71d
alicloud-disk-pulsar             diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   119d
alicloud-disk-pulsar-bookie      diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   119d
alicloud-disk-pulsar-zookeeper   diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   119d
alicloud-disk-redis              diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   119d
alicloud-disk-ssd                diskplugin.csi.alibabacloud.com   Delete          Immediate              true                   119d
alicloud-disk-tidb               diskplugin.csi.alibabacloud.com   Retain          WaitForFirstConsumer   true                   119d
[...]

PR #4737, opened 2022-10-10
Failed to scale down tikv instances less than 3

wxbty:
Misleading users when scaling down tikv
  
In the k8s test environment, after deploying tidb, the number of instances of tipd, tidb, and tikv in my cluster is 3, 3, and 3, respectively.
I tried to use the kubectl command or update tidb-cluster.yaml to shrink the cluster to 1, 1, and 3, and the results showed that the pod was successful.

I used the following command or apply tidb-cluster.yaml
kubectl patch -n ${namespace} tc ${cluster_name} --type merge --patch '{"spec":{"tipd":{"replicas":1}}}'
kubectl patch -n ${namespace} tc ${cluster_name} --type merge --patch '{"spec":{"tidb":{"replicas":1}}}'

However, when I try to scale down tikv, there is no change in the pod, there are always 3, and the status is running. I tried again and again, the problem is still there, I think there is something wrong.
This pr has the same problem, but no reply #4546 

I checked the pod logs
tidb-controller-manager output the following logs. which is same as 4546

can't scale in TiKV of TidbCluster [tidb-cluster/basicai], cause the number of up stores is equal to MaxReplicas in PD configuration(3), and the store in Pod basicai-tikv-2 which is going to be deleted is up too

In the [documentation](https://docs.pingcap.com/tidb-in-kubernetes/dev/scale-a-tidb-cluster), found this description 
According to the prompt, I need to modify the value of pd MaxReplicas.
But it's hard to find how to configure MaxReplicas for pd
There is no [configuration document](https://github.com/pingcap/tidb-operator/blob/master/docs/api-references/docs.md#tidbspec) for k8s

Finally found that you can use pd-ctl (do not want to install it separately) or sql command line real-time change to do it
[link](https://docs.pingcap.com/zh/tidb/stable/dynamic-config)

This problem has been bothering me for a long time. I just want to scale down tikv in test environment

It seems that this question will mislead newbies, so I propose two solutions to solve.
1. Add the MaxReplicas field in the pd area of tidb-cluster.yaml, so that it can be changed by kubectl patch ... spec, and the api field adds a description of the limit between fields. If this is allowed, I will mention another pr to solve.
2. If only online changes are allowed due to design or other reasons, then at least the error message is need more detailed, such as the current pr submission

### Code changes

- [ yes ] Has Go code change
- [ no ] Has CI related scripts change

### Tests

Only the prompt has been modified, no test is needed for the time being

### Side effects

no

### Related changes

- [ no ] Need to cherry-pick to the release branch 
- [yes ] Need to update the documentation  https://docs.pingcap.com/tidb-in-kubernetes/dev/scale-a-tidb-cluster
       Tell the user how to configure  MaxReplicas
       【original】
       Note 
      When the number of UP stores is equal to or less than the parameter value of MaxReplicas in the PD configuration, the TiKV components can not be scaled in.  

### Release Notes
no

```release-note

```


x-shadow-man:
> /cc @AstroProfundis @tennix

@AstroProfundis @tennix @wxbty. Thx for your time. Pls feel free to check the PR. Thx a lot. :-)

csuzhangxc (approved):

pkg/manager/member/tikv_scaler.go
```diff
@@ -305,7 +305,7 @@ func (s *tikvScaler) preCheckUpStores(tc *v1alpha1.TidbCluster, podName string,
 		return false
 	} else if upNumber == maxReplicas {
 		if storeState == v1alpha1.TiKVStateUp {
-			errMsg := fmt.Sprintf("can't scale in TiKV of TidbCluster [%s/%s], cause the number of up stores is equal to MaxReplicas in PD configuration(%d), and the store in Pod %s which is going to be deleted is up too", tc.GetNamespace(), tc.GetName(), maxReplicas, podName)
+			errMsg := fmt.Sprintf("can't scale in TiKV of TidbCluster [%s/%s], cause the number of up stores is equal to MaxReplicas in PD configuration(%d), and the store in Pod %s which is going to be deleted is up too. MaxReplicas can be update online using pd-ctl or SQL statements ,as https://docs.pingcap.com/tidb/stable/dynamic-config", tc.GetNamespace(), tc.GetName(), maxReplicas, podName)
```
csuzhangxc, line 308:
```suggestion
			errMsg := fmt.Sprintf("can't scale in TiKV of TidbCluster [%s/%s], cause the number of up stores is equal to MaxReplicas in PD configuration(%d), and the store in Pod %s which is going to be deleted is up too. MaxReplicas can be update online using pd-ctl or SQL statements, as https://docs.pingcap.com/tidb/stable/dynamic-config", tc.GetNamespace(), tc.GetName(), maxReplicas, podName)
```

KanShiori (approved):

PR #4746, opened 2022-10-17
BR: log backup checkpoint ts can be updated when operator restart

### LINKED ISSUE #4745: BR: can not update log backup checkpoint ts when operator restart
## Bug Report

master

**What did you do?**

check log backup checkpoint ts after operator restarted.
checkpoint ts can be updated.
**What did you see instead?**
chekpoint ts was not updated.

WizardXiao:

Closes #4745 

1. The root casuse is the log backup will be added to the tracker only when it is started, but log backup is already running after operator restart. so tracker lose the log backup after restart. 
2. this pr will list all log backup and add them to tracker when tracker init which will be called once in operator. the instance in `operator main --> backup controller --> backup manager --> backup tracker` is only one in operator.
3. On the whole，the main processes of the log backup track:
    a. tracker init will try to find all log backup and add them to the map which key is namespack and cluster.
    b. log backup start will add it to the map
    c. if add log backup to the map successfully, it will start a go routine which has a loop to track log backup's checkpoint ts and will stop when log backup complete.
    d. by the way, add or delete the map has a mutex.
### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
NONE
```


pkg/backup/backup/backup_tracker.go
```diff
@@ -53,11 +55,31 @@ type trackDepends struct {
...
+func (bt *backupTracker) initTrackLogBackupsProgress() {
+	backups, err := bt.deps.Clientset.PingcapV1alpha1().Backups("").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Warningf("list backups error %v, skip track all log backups progress when init, will track when log backup start", err)
+		return
+	}
+	klog.Infof("list backups size %d", len(backups.Items))
+	for _, backup := range backups.Items {
+		if backup.Spec.Mode == v1alpha1.BackupModeLog {
+			err = bt.StartTrackLogBackupProgress(&backup)
```
grovecai, line 77: Do we still need to call this method in another place?

pkg/backup/backup/backup_tracker.go
```diff
@@ -53,11 +55,31 @@ type trackDepends struct {
...
+			if err != nil {
```
grovecai, line 78: It seems it can not be restarted if StartTrackLogBackupProgress failed?

pkg/backup/backup/backup_tracker.go
```diff
@@ -53,11 +55,31 @@ type trackDepends struct {
...
 		statusUpdater: statusUpdater,
 		logBackups:    make(map[string]*trackDepends),
 	}
+	go tracker.initTrackLogBackupsProgress()
+	return tracker
+}
+
+// initTrackLogBackupsProgress lists all log backups and track their progress.
+func (bt *backupTracker) initTrackLogBackupsProgress() {
+	backups, err := bt.deps.Clientset.PingcapV1alpha1().Backups("").List(context.TODO(), metav1.ListOptions{})
```
grovecai, line 69: Can you add retry here? if it failed, it will be a critical error, not just a warning.

WizardXiao: yes, i can, but it may be not a critical error, because if the log backup start as normal, it can be added to tracker.

pkg/backup/backup/backup_tracker.go
WizardXiao: yes, maybe can add retry, but if it is failed here, it will not affect normal started log backup.

pkg/backup/backup/backup_tracker.go
```diff
@@ -95,14 +136,28 @@ func (bt *backupTracker) removeLogBackup(ns, name string) {
...
+		tc, err = bt.deps.Clientset.PingcapV1alpha1().TidbClusters(clusterNamespace).Get(context.TODO(), backup.Spec.BR.Cluster, metav1.GetOptions{})
+		if err != nil {
+			klog.Warningf("get log backup %s/%s tidbcluster %s/%s failed and will retry, err is %v", ns, name, clusterNamespace, backup.Spec.BR.Cluster, err)
+			return err
+		}
+		return nil
+	})
+
 	if err != nil {
+		klog.Errorf("get log backup %s/%s tidbcluster %s/%s failed and will retry, err is %v", ns, name, clusterNamespace, backup.Spec.BR.Cluster, err)
```
WangLe1321, line 160: I think you don't need to print log here, because you have returned error with same information.

pkg/backup/backup/backup_tracker.go
```diff
@@ -95,14 +136,28 @@ func (bt *backupTracker) removeLogBackup(ns, name string) {
...
 		return nil, fmt.Errorf(fmt.Sprintf("get log backup %s/%s tidbcluster %s/%s failed", ns, name, clusterNamespace, backup.Spec.BR.Cluster), err)
```
WangLe1321, line 161: It seems the returned error has problems. You don't need use `fmt.Sprintf` in `fmt.Errorf`, and there is no placeholder for the `err` in the result of `fmt.Sprintf`.

```suggestion
		return nil, fmt.Errorf("get log backup %s/%s tidbcluster %s/%s failed, err %s", ns, name, clusterNamespace, backup.Spec.BR.Cluster, err)
```

WizardXiao: ok

pkg/backup/backup/backup_tracker.go
WizardXiao: yes, the message can be print when return as an error.

WangLe1321 (approved):

grovecai (approved):
LGTM

KanShiori (approved):

PR #4810, opened 2022-12-23
BR: log backup add stopped status to support truncate after stop

WizardXiao:
### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -202,6 +202,15 @@ func IsBackupPrepared(backup *Backup) bool {
 	return condition != nil && condition.Status == corev1.ConditionTrue
 }
 
+// IsBackupStopped returns true if a Backup is stopped.
+func IsBackupStopped(backup *Backup) bool {
```
WangLe1321, line 206: What's the difference between `IsBackupStopped` and `IsLogBackupAlreadyStop`? Can we use `IsLogBackupAlreadyStop` to replace `IsBackupStopped`?

WizardXiao: `IsBackupStopped` is more likely to be in this state, such as command is stop and status is stopped, but command is truncate and status is stopped is not this status, and this is why truncate can run after stopped. `IsLogBackupAlreadyStop` is more likely to don't care about specific subcommands. so they a two function now.

WangLe1321: Can you add some comments to this function to interpret its usage?

WizardXiao: ok

WangLe1321 (approved):

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1972,6 +1972,8 @@ const (
 	BackupInvalid BackupConditionType = "Invalid"
 	// BackupPrepare means the backup prepare backup process
 	BackupPrepare BackupConditionType = "Prepare"
+	// BackupStopped means the backup was stopped, just log backup has this condition
+	BackupStopped BackupConditionType = "Stopped"
```
grovecai, line 1976: When it's stopped, can it resume again? If yes, I prefer use "paused"?

grovecai: If it only use for logback, make it explicit, just like LogbackupPaused

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -202,6 +202,17 @@ func IsBackupPrepared(backup *Backup) bool {
 	return condition != nil && condition.Status == corev1.ConditionTrue
 }
 
+// IsBackupStopped returns true if a Backup is stopped.
+// It means log backup is at stopped status.
+// It used to filter CR update event which is stop command and stopped status, and let it run truncate after log backup stopped which is truncate command and stopped status.
+func IsBackupStopped(backup *Backup) bool {
```
grovecai, line 208: make it specified for logbackup

pkg/apis/pingcap/v1alpha1/types.go
WizardXiao: it can not resume again, we'd better use `Stopped`, because br has `pause` and `resume` command, we can use `paused` when we support  `pause` command in the future.

pkg/apis/pingcap/v1alpha1/backup.go
WizardXiao: ok

grovecai (approved):
LGTM

PR #4844, opened 2023-01-19
tidb status add gethealth info

### LINKED ISSUE #4772: do not lost the error when checking tidb health
https://github.com/pingcap/tidb-operator/blob/e60a3ea2003965fa99f763323618cb1c452ad6e3/pkg/controller/tidb_control.go#L73

​​​​when upgrading tidb, it keep unhealth and can not know what happens

mikechengwei:
Close #4772 .

When tidb sync status info, not ignoring error, and adding message field to  sync error detail info.
### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/manager/member/tidb_member_manager.go
+++ b/pkg/manager/member/tidb_member_manager.go
@@ -1026,13 +1026,13 @@ func (m *tidbMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, set 
 	for id := range helper.GetPodOrdinals(tc.Status.TiDB.StatefulSet.Replicas, set) {
 		name := fmt.Sprintf("%s-%d", controller.TiDBMemberName(tc.GetName()), id)
 		health, err := m.deps.TiDBControl.GetHealth(tc, int32(id))
-		if err != nil {
-			return err
-		}
-
 		newTidbMember := v1alpha1.TiDBMember{
 			Name:   name,
 			Health: health,
+		}
+		if err != nil {
+			klog.V(4).Infof("tidb cluster %s/%s sync status get health error:%v", tc.GetNamespace(), tc.GetName(), err)
+			newTidbMember.Message = err.Error()
 		}
 		oldTidbMember, exist := tc.Status.TiDB.Members[name]
 
--- a/docs/api-references/docs.md
+++ b/docs/api-references/docs.md
@@ -16034,6 +16034,16 @@ string
 <p>Node hosting pod of this TiDB member.</p>
 </td>
 </tr>
+<tr>
+<td>
+<code>message</code></br>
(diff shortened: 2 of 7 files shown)


july2993 (commented):
It is ok to return err in `syncTidbClusterStatus`
the issue is `GetHealth()` DO NOT return the error and also DO NOT log the error when `getBodyOk` return the error(so we lost this error), so if something makes it keep failing to request the status API, we do not know what happens.

The original issue is `/status` keeps blocking(caused by some bug of the pd version), not returning a response. I can not see any log or other thinks about this, and I need to `kubectl exec ..` into the pod the request the status API to make sure that this keeps not health in the tc status.

mikechengwei:
> It is ok to return err in `syncTidbClusterStatus` the issue is `GetHealth()` DO NOT return the error and also DO NOT log the error when `getBodyOk` return the error(so we lost this error), so if something makes it keep failing to request the status API, we do not know what happens.
> 
> The original issue is `/status` keeps blocking(caused by some bug of the pd version), not returning a response. I can not see any log or other thinks about this, and I need to `kubectl exec ..` into the pod the request the status API to make sure that this keeps not health in the tc status.

Ok, I modified the code from your perspective. 
Ref: 
https://github.com/pingcap/tidb-operator/blob/cc2e2ee8adef538e3a0e2a77dc9bed43041d4174/pkg/controller/tidb_control.go#L74

original code is:

https://github.com/pingcap/tidb-operator/blob/dc1508a8a2998c376f7c97fc9d56bb285ba5ebf4/pkg/controller/tidb_control.go#L74

pkg/manager/member/tidb_member_manager.go
```diff
@@ -1049,14 +1049,14 @@ func (m *tidbMemberManager) syncTidbClusterStatus(tc *v1alpha1.TidbCluster, set
...
 		health, err := m.deps.TiDBControl.GetHealth(tc, int32(id))
-		if err != nil {
-			return err
-		}
-
 		newTidbMember := v1alpha1.TiDBMember{
 			Name:   name,
 			Health: health,
 		}
+		if err != nil {
```
csuzhangxc, line 1056: do we need to keep the `err` in `_, err = getBodyOK(httpClient, url)`?

csuzhangxc: why do you revert it in https://github.com/pingcap/tidb-operator/pull/4844/commits/a637f81cfe5e78504bfa02a01a5ed38c1c6a5566

Issue #4805, opened 2022-12-14
Setting Restart policy of backup job to onfailure


charleszheng44:
## Feature Request

**Is your feature request related to a problem? Please describe:**

We found that when we killed a backup pod the backup job went to failed state without retrying. This could happen when we do a node upgrade or a node was down. The reason is that the BackoffLimit is always set to 0 and the RestartPolicy of the pod is never https://github.com/pingcap/tidb-operator/blob/ec8974c534d6beeedc68a9106b82ec45fbab3d90/pkg/backup/backup/backup_manager.go#L660. Unfortunately we can’t modify it for now. Is it possible that we can set it in our Backup CR?

**Describe the feature you'd like:**

Allow setting the restart policy of the backup job to `RestartPolicyOnFailure`

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

charleszheng44:
@csuzhangxc

csuzhangxc:
@WizardXiao will you implement this?

WizardXiao:
ok, i will try to implement this.

charleszheng44:
@WizardXiao Can I know the ETA of this feature?

WizardXiao:
v1.4.4 has support retries for snapshot backups in case of unexpected failures caused by Kubernetes job or pod issues. It takes effect through `spec.backoffRetryPolicy`, Ref https://docs.pingcap.com/tidb-in-kubernetes/stable/backup-restore-cr#general-fields

PR #4883, opened 2023-02-16
fix backup failed when pod was auto restarted by k8s

WizardXiao:

Ref #4805 
### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### k8s may restart backup pod which is not done.
1. drain node, node's pod will be schedule to other node
2. delete pod when job still exist.
### detail test
drain node test
1. backup is running at `worker2`, drain `worker2`, backup restart at `worker3` by k8s.
image
image
2. clean data before run br command to backup data
image
3. backup success
image

delete pod test
1. backup pod was deleted and then job restart a new pod
image
2. clean data before run br command to backup data in the new pod 
image
3. backup success
image
### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
fix backup failed when pod was auto restarted by k8s
```

### DIFF
--- a/cmd/backup-manager/app/backup/backup.go
+++ b/cmd/backup-manager/app/backup/backup.go
@@ -292,11 +292,6 @@ func (bo *Options) brCommandRunWithLogCallback(ctx context.Context, fullArgs []s
 		return fmt.Errorf("cluster %s, wait pipe message failed, errMsg %s, err: %v", bo, errMsg, err)
 	}
 
-	for i := 0; i < 5*60; i++ {
-		klog.Infof("simulate stuck br commond %v for cluster %s", fullArgs, bo)
-		time.Sleep(time.Second * 1)
-	}
-
 	klog.Infof("Run br commond %v for cluster %s successfully", fullArgs, bo)
 	return nil
 }
--- a/cmd/backup-manager/app/backup/backup.go
+++ b/cmd/backup-manager/app/backup/backup.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Masterminds/semver"
 	"github.com/pingcap/errors"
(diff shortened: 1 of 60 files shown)


cmd/backup-manager/app/backup/backup.go
```diff
@@ -112,13 +115,39 @@ func (bo *Options) backupData(
 		go bo.updateProgressFromFile(progressCtx.Done(), backup, progressFile, progressStep, statusUpdater)
 	}
 
+	if backup.Spec.Mode == v1alpha1.BackupModeSnapshot && !bo.isBRCanContinueRunByCheckpoint() {
```
WangLe1321, line 118: Can we check whether the backup is restarted? I worry that we may delete valid backup data. I think we shouldn't clean backup data at first time. For example, user starts a backup with `retain` policy, then user deletes that backup. If user starts a backup with same name, we deletes valid backup data.

WizardXiao: ok, i think this commit can check this and only when backup is already running status, we will clean
https://github.com/pingcap/tidb-operator/pull/4883/commits/ffc46ec7178acc6fd22a7801c69ec1d2d99b6c4a

WizardXiao: i add running status check in https://github.com/pingcap/tidb-operator/pull/4883/commits/acecff3b4d6853b7f5a37e963cd53533aeaed370, PTAL

fengou1 (approved):

cmd/backup-manager/app/backup/backup.go
WizardXiao: after a discuss, i use `!scheduled` as the judging condition and add `restart` condition as a record.

grovecai (approved):
LGTM, pls add e2e test

PR #4907, opened 2023-02-24
feat(scaler): scale out by group and wait until pods are scheduled

liubog2008:

If there are 3 zones and 9 replicas of TiKV with PodSpreadConstraints, pods may not spread in order as expected. Because all pods of TiKV will be created at the same time and then they will not be scheduled in order.

So pods may be scheduled as below, and then TiKV will fail to scale down more than 1 replica.
- zone-a: 0, 1, 2
- zone-b: 3, 4, 5
- zone-c: 6, 7, 8

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


pkg/manager/member/tikv_scaler.go
```diff
@@ -59,31 +60,55 @@ func (s *tikvScaler) ScaleOut(meta metav1.Object, oldSet *apps.StatefulSet, newS
...
 			errs = append(errs, err)
 		} else {
 			finishedOrdinals.Insert(ordinal)
-			updateReplicasAndDeleteSlots = true
 		}
 	}
-	if updateReplicasAndDeleteSlots {
-		setReplicasAndDeleteSlotsByFinished(scalingOutFlag, newSet, oldSet, ordinals, finishedOrdinals)
-	} else {
-		resetReplicas(newSet, oldSet)
```
csuzhangxc, line 84: I'm not sure whether `resetReplicas(newSet, oldSet)` is needed for `return controller.RequeueErrorf("tikv.ScaleOut, cluster %s/%s ready to scale out, wait for next round", tc.GetNamespace(), tc.GetName())` in L122.

csuzhangxc:
Is this feature controlled by `scalePolicy.scaleOutParallelism`?

Issue #4758, opened 2022-10-27
Mechanism to gracefully shut down a Pod in TiDB cluster


luohao:
## Feature Request

**Is your feature request related to a problem? Please describe:**
We have a node drainer service in k8s cluster that rotates k8s nodes(EC2 VMs), and all Pods on the victim node are evicted with the eviction API. Conflicts between tidb-operator initiated Pod rolling upgrade and node drainer initiated Pod eviction will result in availability issues in PD and TiKV.

**Describe the feature you'd like:**
We would like a mechanism for external operators(e.g., cluster admin, node maintenance service etc) to notify tidb-operator that a Pod in the cluster needs to be restarted. tidb-operator should decide **_when_** and **_how_** the Pod will be restarted. 

tidb-operator already has graceful shutdown for TiKV Pod using annotation([proposal](https://github.com/pingcap/tidb-operator/blob/master/docs/design-proposals/2021-11-24-graceful-restart-tikv-pod.md)). We'd like something similar but need it to

1. support all componenets(e.g., PD and TiDB etc): right now Pod controller only syncs TiKV pod.
2. support multi-cluster deployment.

**Describe alternatives you've considered:**
We already have a scheduled maintenance window but with multi-cluster deployment(i.e., a TiDB cluster spans across 3 k8s cluster) it's hard to avoid conflicts(e.g., an ongoing rolling upgrade in cluster-1 and a node drain in cluster-3 may still have conflicts).

**Teachability, Documentation, Adoption, Migration Strategy:**

We plan to integrate this with out node drainer service so that TiDB Pods can be safely evicted from the victim node. This should minimize the disruption caused by maintenance tasks in general.

hanlins:
Hi @luohao, thanks for raising the issue here! I found that in this issue, you're planning to support graceful shutdown for all components, I'm wondering if we could tackle this one by one? IMO the procedure for evicting different components could have subtle differences. Take an example, for a PD instance, we might want to evict it when A). It's no longer the leader, B). A quorum of PD members is up to date and can take requests. Whereas for TiDB, the eviction criteria could be there's no connection affiliated with the pod. Maybe we can break this down, propose solutions respectively and solve them separately? It would also be helpful to dispatch these problems to multiple developers. What do you think?

PR #4948, opened 2023-03-31
Add annotation to graceful shutdown tidb pod

wxiaomou:

Add annotation to graceful shutdown tidb pod
Closes: https://github.com/pingcap/tidb-operator/issues/4758
### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -1380,6 +1380,8 @@ const (
 	EvictLeaderAnnKeyForResize = "tidb.pingcap.com/evict-leader-for-resize"
 	// PDLeaderTransferAnnKey is the annotation key to transfer PD leader used by user.
 	PDLeaderTransferAnnKey = "tidb.pingcap.com/pd-transfer-leader"
+	// TiDBTransferAnnKey is the annotation key to graceful shutdown tidb pod by user.
+	TiDBGracefulShutdownAnnKey = "tidb.pingcap.com/tidb-graceful-shutdown"
 )
 
 // The `Value` of annotation controls the behavior when the leader count drops to zero, the valid value is one of:
@@ -1398,6 +1400,15 @@ const (
 const (
 	TransferLeaderValueNone      = "none"
 	TransferLeaderValueDeletePod = "delete-pod"
+)
+
+// The `Value` of TiDB deletion annotation controls the behavior when the tidb pod got deleted, the valid value is one of:
+//
+// - `none`: doing nothing.
+// - `delete-pod`: delete pod.
+const (
+	TiDBPodDeletionValueNone = "none"
+	TiDBPodDeletionDeletePod = "delete-pod"
 )
 
 type EvictLeaderStatus struct {
--- a/pkg/controller/tidbcluster/pod_control.go
+++ b/pkg/controller/tidbcluster/pod_control.go
@@ -216,6 +216,8 @@ func (c *PodController) sync(key string) (reconcile.Result, error) {
 		return c.syncPDPod(ctx, pod, tc)
 	case label.TiKVLabelVal:
 		return c.syncTiKVPod(ctx, pod, tc)
+	case label.TiDBLabelVal:
+		return c.syncTiDBPod(ctx, pod, tc)
 	default:
 		return reconcile.Result{}, nil
 	}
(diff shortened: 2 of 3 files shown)


wxiaomou:
@hanlins just fix the tests, can you help to triggered the ci again? Thank you so much~

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1380,6 +1380,8 @@ const (
 	EvictLeaderAnnKeyForResize = "tidb.pingcap.com/evict-leader-for-resize"
 	// PDLeaderTransferAnnKey is the annotation key to transfer PD leader used by user.
 	PDLeaderTransferAnnKey = "tidb.pingcap.com/pd-transfer-leader"
+	// TiDBTransferAnnKey is the annotation key to graceful shutdown tidb pod by user.
```
csuzhangxc, line 1383:
```suggestion
	// TiDBGracefulShutdownAnnKey is the annotation key to graceful shutdown tidb pod by user.
```

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -412,6 +414,57 @@ func (c *PodController) syncTiKVPod(ctx context.Context, pod *corev1.Pod, tc *v1
...
+			tc.Status.PD.Phase, tc.Status.TiKV.Phase, tc.Status.TiFlash.Phase,
+			tc.Status.Pump.Phase, tc.Status.TiDB.Phase)
+		return reconcile.Result{RequeueAfter: RequeueInterval}, nil
+	}
+
+	switch value {
+	case v1alpha1.TiDBPodDeletionValueNone:
+	case v1alpha1.TiDBPodDeletionDeletePod:
+	default:
+		klog.Warningf("Ignore unknown value %q of annotation %q for Pod %s/%s", value, v1alpha1.PDLeaderTransferAnnKey, pod.Namespace, pod.Name)
```
csuzhangxc, line 445:
```suggestion
		klog.Warningf("Ignore unknown value %q of annotation %q for Pod %s/%s", value, v1alpha1.TiDBGracefulShutdownAnnKey, pod.Namespace, pod.Name)
```

csuzhangxc (approved):

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -412,6 +414,57 @@ func (c *PodController) syncTiKVPod(ctx context.Context, pod *corev1.Pod, tc *v1
...
+	}
+
+	ns := tc.GetNamespace()
+	tcName := tc.GetName()
+
+	if tc.Status.PD.Phase == v1alpha1.UpgradePhase || tc.Status.PD.Phase == v1alpha1.ScalePhase ||
+		tc.Status.TiKV.Phase == v1alpha1.UpgradePhase || tc.Status.TiKV.Phase == v1alpha1.ScalePhase ||
+		tc.Status.TiFlash.Phase == v1alpha1.UpgradePhase || tc.Status.TiFlash.Phase == v1alpha1.ScalePhase ||
+		tc.Status.Pump.Phase == v1alpha1.UpgradePhase || tc.Status.Pump.Phase == v1alpha1.ScalePhase ||
+		tc.TiDBScaling() {
```
hanlins, line 431: Seems mixed checking directly against `tc.Status` and use methods like `tc.TiDBScaling()`. Is it feasible to:
1. use methods like `tc.PDUpgrading()` or `tc.PDScaling()` for each component instead of using `tc.Status` directly?
2. aggregate the checking for all components into a single method such as `tc.Upgrading()`?

That helps to make it easier for readability or reuse.

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -412,6 +414,57 @@ func (c *PodController) syncTiKVPod(ctx context.Context, pod *corev1.Pod, tc *v1
...
+}
+
+func needDeleteTiDBPod(pod *corev1.Pod) (string, bool) {
+	value, exist := pod.Annotations[v1alpha1.TiDBGracefulShutdownAnnKey]
+	if exist {
+		return value, true
+	}
+
+	return "", false
+}
```
hanlins, line 466: Can we use `value` only? Seems there's no scenario where the `value` is an empty string but `exist` is `true`, we can check whether `value` is empty to learn whether the annotation is set.
```suggestion
func needDeleteTiDBPod(pod *corev1.Pod) string {
	if pod.Annotations == nil {
		return ""
	}
	return pod.Annotations[v1alpha1.TiDBGracefulShutdownAnnKey]
}
```

hanlins (commented):
Thanks for the change @wxiaomou! Have some suggestions for improvements, commented inline.

pkg/controller/tidbcluster/pod_control.go
hanlins: Also curious is it intended to skip `tc.TiDBUpgrading` check for `tidb` only but keep the upgrading check for other components? If it is intended, maybe we can elaborate on the reason in a comment to avoid confusion.

wxiaomou: @hanlins Thanks  for the comments. I think I missed the tc.TiDBUpgrading case, added here and the test case. Also I used the suggested functions instead read tc status.

hanlins (approved):
LGTM, thanks for the change!

Issue #4972, opened 2023-04-14
Check annotation and label diff when sync services 


wxiaomou:
## Bug Report
I added new annotation to pd services and pd headless service, but the service not get updated by the tidb operator.

Seems when we sync [tidb services](https://sourcegraph.com/github.com/pingcap/tidb-operator@master/-/blob/pkg/manager/member/tidb_member_manager.go) we have checked the label and annotation of the new service. Should we do the similar thing for other components as well? 

**What version of Kubernetes are you using?**

**What version of TiDB Operator are you using?**

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

**What's the status of the TiDB cluster pods?**

**What did you do?**

**What did you expect to see?**

**What did you see instead?**

csuzhangxc:
It seems doing a similar thing for other components should be better.  And TiDB service label & annotation check was added in #3892

wxiaomou:
@csuzhangxc Thanks, just pulled a pr https://github.com/pingcap/tidb-operator/pull/4973/files

PR #4973, opened 2023-04-14
Check service annotation and labels

wxiaomou:
close https://github.com/pingcap/tidb-operator/issues/4972
### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/controller/equality.go
+++ b/pkg/controller/equality.go
@@ -93,29 +93,6 @@ func ServiceEqual(newSvc, oldSvc *corev1.Service) (bool, error) {
 	return false, nil
 }
 
-// ServiceAnnLabelEqual compares the new Services's annotation and lables with old Services
-func ServiceAnnLabelEqual(newSvc, oldSvc *corev1.Service) (bool, error) {
-	if newSvc == nil || oldSvc == nil {
-		return false, fmt.Errorf("ServiceAnnLabelEqual: newservice or oldService is nil")
-	}
-	if newSvc.Annotations == nil {
-		newSvc.Annotations = map[string]string{}
-	}
-	if oldSvc.Annotations == nil {
-		oldSvc.Annotations = map[string]string{}
-	}
-	if newSvc.Labels == nil {
-		newSvc.Labels = map[string]string{}
-	}
-	if oldSvc.Labels == nil {
-		oldSvc.Labels = map[string]string{}
-	}
-	delete(oldSvc.Annotations, LastAppliedConfigAnnotation)
-	annoEqual := apiequality.Semantic.DeepEqual(newSvc.Annotations, oldSvc.Annotations)
-	labelEqual := apiequality.Semantic.DeepEqual(newSvc.Labels, oldSvc.Labels)
-	return annoEqual && labelEqual, nil
-}
-
 func IngressV1beta1Equal(newIngress, oldIngres *extensionsv1beta1.Ingress) (bool, error) {
 	oldIngressSpec := extensionsv1beta1.IngressSpec{}
 	if lastAppliedConfig, ok := oldIngres.Annotations[LastAppliedConfigAnnotation]; ok {
--- a/pkg/controller/equality.go
+++ b/pkg/controller/equality.go
@@ -93,6 +93,29 @@ func ServiceEqual(newSvc, oldSvc *corev1.Service) (bool, error) {
 	return false, nil
 }
 
+// ServiceAnnLabelEqual compares the new Services's annotation and lables with old Services
+func ServiceAnnLabelEqual(newSvc, oldSvc *corev1.Service) (bool, error) {
+	if newSvc == nil || oldSvc == nil {
+		return false, fmt.Errorf("ServiceAnnLabelEqual: newservice or oldService is nil")
+	}
+	if newSvc.Annotations == nil {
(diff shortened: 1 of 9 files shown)


csuzhangxc:
@hanlins PTAL

hanlins (commented):
It seems we're copying existing logics
https://github.com/pingcap/tidb-operator/blob/4e9e294ead1a935e7ac5a60f193ba50f25450d06/pkg/manager/member/tidb_member_manager.go#L479-L526
over and over again, can we take one step further, refactor that chunk of code for updating service object into a function and reuse that function in these places?

pkg/controller/equality.go
```diff
@@ -93,6 +93,29 @@ func ServiceEqual(newSvc, oldSvc *corev1.Service) (bool, error) {
 	return false, nil
 }
 
+// ServiceAnnLabelEqual compares the new Services's annotation and lables with old Services
+func ServiceAnnLabelEqual(newSvc, oldSvc *corev1.Service) (bool, error) {
```
hanlins, line 97: Hi @wxiaomou, thanks for the refactoring work! I'm considering can we take one step further, take an additional `runtime.Object` parameter (which is the input for `m.deps.ServiceControl.UpdateService`), and a modifier function to sync cluster IP (`svc.Spec.ClusterIP = oldSvc.Spec.ClusterIP`) in the function signature? Then we can unify the service syncing logic for each component, and `ServiceAnnLabelEqual` becomes something like `SyncComponentService` semantically. What do you think?

wxiaomou: @hanlins refactored and add an unit test for SyncComponentService. Thanks for the review.

wxiaomou: @hanlins I didn't apply this refactor for tidb_manager and pump_manager since they are checking if the service is orphan, do you have any suggestion? We can bypass them or add the orphan check in SyncComponentService. WDYT

hanlins: I think it should be fine, we could leave that as a follow-up improvement to take care of the orphan-checking logic in a separate function.

pkg/controller/service_control.go
```diff
@@ -130,6 +133,53 @@ func (c *realServiceControl) recordServiceEvent(verb, name, kind string, object
 	}
 }
 
+func (c *realServiceControl) SyncComponentService(tc runtime.Object, newSvc, oldSvc *corev1.Service, syncClusterIPFn func(newSvc *corev1.Service)) (*corev1.Service, error) {
```
hanlins, line 136: It seems that, the function parameter `syncClusterIPFn` is only copies `.spec.clusterIP` from old svc object to the new one, it seems a function modifier would be too generic for that purpose. Maybe we could use a boolean to differentiate that? And we wouldn't need to pass a function every time invoking this method, what't do you think?

wxiaomou: @hanlins yes, I think that make sense, will make the change

hanlins (approved):
LGTM

pkg/controller/service_control.go
```diff
@@ -130,6 +133,53 @@ func (c *realServiceControl) recordServiceEvent(verb, name, kind string, object
 	}
 }
 
+func (c *realServiceControl) SyncComponentService(tc runtime.Object, newSvc, oldSvc *corev1.Service, isHeadlessSvc bool) (*corev1.Service, error) {
```
csuzhangxc, line 136: The name of `isHeadlessSvc` is not accurate. e.g both PD services are passing `false`, and not accurate for DM-master & DM-worker services too.

pkg/controller/service_control.go
```diff
@@ -39,6 +41,7 @@ type ServiceControlInterface interface {
 	CreateService(runtime.Object, *corev1.Service) error
 	UpdateService(runtime.Object, *corev1.Service) (*corev1.Service, error)
 	DeleteService(runtime.Object, *corev1.Service) error
+	SyncComponentService(runtime.Object, *corev1.Service, *corev1.Service, bool) (*corev1.Service, error)
```
csuzhangxc, line 44: it's better to add some comments or naming these args.

wxiaomou: added the naming

pkg/controller/service_control.go
wxiaomou: @csuzhangxc good catch! I changed the param name to `setClusterIP` and fix it in dm_master

wxiaomou:
the e2e test failed seems not related to the changes 
`/api/v1/nodes/tidb-operator-worker3?timeout=10s 404 Not Found in 2 milliseconds
`@csuzhangxc

csuzhangxc:
> the e2e test failed seems not related to the changes `/api/v1/nodes/tidb-operator-worker3?timeout=10s 404 Not Found in 2 milliseconds `@csuzhangxc

the CI env is not stable now.

csuzhangxc (approved):

wxiaomou:
@csuzhangxc does the ci issue get fixed? Wondering if we can merge this PR. Thanks~

PR #4979, opened 2023-04-21
check all stores are up before upgrade

Tema:
It is unsafe to start rolling upgrade or delete tikv pods if some of stores in are offline (e.g. due to hardware failure). E.g. common tidb cluster setup has replication factor 3. If one of TiKV nodes is down, evict leader will still succeed, but once the TiKV node gets restarted, some regions will loose quorum. The PR solves the problem for any topology including heterogeneous and cross k8s deployments. 

When TiDB executes rolling restart during upgrade, it will check status of all Stores registered in PD to make sure they are not in DOWN, OFFLINE or DELETE state https://docs.pingcap.com/tidb/stable/tidb-scheduling#information-collection

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fix the issue that TiKV briefly looses quorum when starting upgrade while some of TiKV stores are down.
```


pkg/manager/member/tikv_upgrader.go
```diff
@@ -174,6 +175,10 @@ func (u *tikvUpgrader) Upgrade(meta metav1.Object, oldSet *apps.StatefulSet, new
 			continue
 		}
 
+		if isStable, reason := pdapi.IsClusterStable(controller.GetPDClient(u.deps.PDControl, tc)); !isStable {
+			return controller.RequeueErrorf("cluster is unstable: %s", reason)
+		}
+
```
hanlins, line 181: Isn't the cluster stability already checked above? Probably it's a duplication that we can remove?
https://github.com/pingcap/tidb-operator/blob/7a9fe4126e81691796320cba0cf7115f77c22f49/pkg/manager/member/tikv_upgrader.go#L74

pkg/manager/member/tikv_upgrader.go
```diff
@@ -491,7 +496,7 @@ func getStoreByOrdinal(name string, status v1alpha1.TiKVStatus, ordinal int32) *
 	return nil
 }
 
-func isTiKVReadyToUpgrade(tc *v1alpha1.TidbCluster) (bool, string) {
+func isTiKVReadyToUpgrade(pdControl pdapi.PDControlInterface, tc *v1alpha1.TidbCluster) (bool, string) {
```
hanlins, line 499: Similar to the `IsClusterStable`, maybe we can make some improvements and return the reason only? Another thing is, can we make this method a member function of `tikvUpgrader`?

hanlins (commented):
Thanks @Tema, the checking logic looks good in general. I have two suggestions:
1. It seems `IsClusterStable` is pretty specific to tikv and maybe we can make it a member function of `tikvUpgrader`?
2. Is it possible to add an annotation or something to opt out the cluster stability check? In some situations it might block the revert of an upgrade so we might have to opt out the check.
The change set looks promising to me, cc @csuzhangxc.

pkg/manager/member/tikv_upgrader.go
hanlins: Oh, I meant, in method `isTiKVReadyToUpgrade` we have already checked whether the TiKV nodes are stable
https://github.com/pingcap/tidb-operator/blob/7a9fe4126e81691796320cba0cf7115f77c22f49/pkg/manager/member/tikv_upgrader.go#L510
and as `isTiKVReadyToUpgrade` is invoked above, do we need to double-check the TiKV stability here? Or we can keep the check here but remove the check in `isTiKVReadyToUpgrade`, and leave it as it used to be. What do you think?

hanlins (approved):
LGTM, cc @csuzhangxc

pkg/manager/member/tikv_upgrader.go
```diff
@@ -43,6 +44,7 @@ const (
 	// TODO: change to use minReadySeconds in sts spec
 	// See https://kubernetes.io/blog/2021/08/27/minreadyseconds-statefulsets/
 	annoKeyTiKVMinReadySeconds = "tidb.pingcap.com/tikv-min-ready-seconds"
+	annoKeySkipStoreStateCheck = "tidb.pingcap.com/skip-store-check-for-upgrade"
```
csuzhangxc, line 47: How about changing from `skip: false` to `check: true`?

In other words, renaming the anno key to something like `tidb.pingcap.com/store-check-for-upgrade`, and only take effect if the value is `true`.

BTW, this is only for the TiKV store and not for the TiFlash store, so how about renaming it to include `tikv` in the key name or adding some comments to explain it?

csuzhangxc (approved):

PR #4995, opened 2023-05-06
check all PD peers are healthy before upgrade

Tema:
This PR is identical to https://github.com/pingcap/tidb-operator/pull/4979/files but for PD peer check.

It is unsafe to start rolling upgrade or delete pd pods if some of other pd pods in are not healthy or offline (e.g. due to hardware failure). The PR adds safety during rolling restart for any topology including heterogeneous and cross k8s deployments. 

When TiDB executes rolling restart during upgrade, it will check health of all PDs to make sure they are healthy.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
None
```


pkg/controller/tidbcluster/pod_control.go
```diff
@@ -484,21 +486,6 @@ func needPDLeaderTransfer(pod *corev1.Pod) (string, bool) {
...
-		if pdMember.Health {
-			healthCount++
-		}
-	}
-	for _, pdMember := range tc.Status.PD.PeerMembers {
-		if pdMember.Health {
-			healthCount++
-		}
-	}
-	return healthCount > (len(tc.Status.PD.Members)+len(tc.Status.PD.PeerMembers))/2+1
```
csuzhangxc, line 499: If we have 5 PD members and 1 of them is not healthy, then PD can't be restarted before, but now, it will be blocked when restarting, right?

csuzhangxc: I mean,  safeToRestartPD` is replaced with the new `pdapi.IsPDStable`, but `pdapi.IsPDStable` will block the restarting process if any member is unhealthy, then the previous behavior is changed.

csuzhangxc: Can we keep the previous behavior is `tidb.pingcap.com/pd-check-all-peers-up-before-upgrade` is not set?

hanlins: Second to @csuzhangxc, maybe keep the original `safeToRestartPD`, and when `tidb.pingcap.com/pd-check-quorum-before-upgrade` is set, keep using CR status for quorum check, and use pd client instead when the annotation is set? Less behavior changes means less surprises to us :D

pkg/manager/member/pd_upgrader.go
```diff
@@ -19,13 +19,19 @@ import (
...
 
 	"github.com/pingcap/advanced-statefulset/client/apis/apps/v1/helper"
 	apps "k8s.io/api/apps/v1"
 	"k8s.io/klog/v2"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 )
 
+const (
+	// set this PD clustre annotation to true to fail cluster upgrade if PD loose the quorum during one pod restart
+	annoKeyPDPeersCheck = "tidb.pingcap.com/pd-check-quorum-before-upgrade"
```
hanlins, line 32: Minior comments. Before the change seems we have already performed some quorum checks but using CR status instead of PD client, maybe we can be more specific in the annotation, telling people that when the annotation is set we're using PD client?

csuzhangxc (approved):

hanlins (approved):
Ok reading PD status from CR versus using pd client is effectively the same. Let's get this merged, thanks @Tema!

PR #4279, opened 2021-11-30
*: implements RFC: Add 2021-11-24-graceful-reschedule-tikv-pod.md

july2993:
### What problem does this PR solve?

implements #4277 
Fix #4215 
### What is changed and how does it work?

see docs in #4277 
### Code changes

- [ ] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [x] Manual test 
### test1
1. deploy a cluster (examples/basic/, change replicas of tikv to 3)
2. kubectl annotate pods basic-tikv-0 tidb.pingcap.com/restart=""
3. observe
    1. add evict-leader-scheduler
    2. tikv pod deleted
    3. remove evict-leader-scheduler
Screen Shot 2021-11-30 at 11 42 43 AM

### test2
1. add `tidb.pingcap.com/restart` annotation, check evict-leader-scheduler added
2. remove the annotation and check evict-leader-scheduler removed

- [ ] No code <!
-- If this PR contains no code changes, check this box -->

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [ ] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note
Support evicting region leaders for TiKV and restarting TiKV gracefully through annotations
```

PR #4997, opened 2023-05-08
allow leader eviction annotation to expire

Tema:
https://github.com/pingcap/tidb-operator/pull/4279 has introduced annotation for graceful shutdown of TiKV nodes. Various automations use this annotation to drain nodes gracefully for k8s cluster upgrade and other required node replacements. However, given that tidb-operator acts on these annotations asynchronously, the result of these annotations might might violate maintenance windows defined in the k8s clusters where TiDB cluster running. 

This PR introduced additional annotation to specify expiration time for leader transfer and node deletion, so that external annotation can leverage it to prevent tidb-operator acting on leader-transfer and pod-deletion annotations outside of maintenance window.

### Code changes

- [x] Has Go code change

### Tests
- [x] Unit test 

### Release Notes

```release-note
NONE
```


pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1382,6 +1382,10 @@ const (
 	PDLeaderTransferAnnKey = "tidb.pingcap.com/pd-transfer-leader"
 	// TiDBGracefulShutdownAnnKey is the annotation key to graceful shutdown tidb pod by user.
 	TiDBGracefulShutdownAnnKey = "tidb.pingcap.com/tidb-graceful-shutdown"
+	// TiKVEvictLeaderExpirationTimeAnnKey is the annotation key to expire evict leader annotation. Type: time.RFC3339.
+	TiKVEvictLeaderExpirationTimeAnnKey = "tidb.pingcap.com/tikv-evict-leader-expiration-time"
+	// PDKVEvictLeaderExpirationTimeAnnKey is the annotation key to expire evict leader annotation. Type: time.RFC3339.
```
csuzhangxc, line 1387:
```suggestion
	// PDEvictLeaderExpirationTimeAnnKey is the annotation key to expire evict leader annotation. Type: time.RFC3339.
```

BTW, how about renaming it to `LeaderTransfer` as L1382

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -458,6 +466,34 @@ func (c *PodController) syncTiDBPod(ctx context.Context, pod *corev1.Pod, tc *v1
...
+	if exist {
+		evictionExpirationTime, err := time.Parse(time.RFC3339, timeToExpireAnnValue)
+		if err == nil {
+			if metav1.Now().Time.After(evictionExpirationTime) {
+				klog.Infof("Annotation to evict leader on the Pod %s/%s is expired", pod.Namespace, pod.Name)
+				return true
+			}
+		} else {
+			klog.Warningf("Can't parse %s value %s on %s/%s. Ignore and never expire. Err: ", annKey, evictionExpirationTime, pod.Namespace, pod.Name, err)
+			return true
```
csuzhangxc, line 491: Why `never expire` but `return true` here?

pkg/apis/pingcap/v1alpha1/types.go
csuzhangxc: How about renaming it to `PDLeaderTransferExpirationTimeAnnKey `?

csuzhangxc:
@hanlins PTAL

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1382,6 +1382,10 @@ const (
 	PDLeaderTransferAnnKey = "tidb.pingcap.com/pd-transfer-leader"
 	// TiDBGracefulShutdownAnnKey is the annotation key to graceful shutdown tidb pod by user.
 	TiDBGracefulShutdownAnnKey = "tidb.pingcap.com/tidb-graceful-shutdown"
+	// TiKVEvictLeaderExpirationTimeAnnKey is the annotation key to expire evict leader annotation. Type: time.RFC3339.
+	TiKVEvictLeaderExpirationTimeAnnKey = "tidb.pingcap.com/tikv-evict-leader-expiration-time"
+	// PDLeaderTransferExpirationTimeAnnKey is the annotation key to expire evict leader annotation. Type: time.RFC3339.
+	PDLeaderTransferExpirationTimeAnnKey = "tidb.pingcap.com/pd-evict-leader-expiration-time"
```
hanlins, line 1388: Do we need to differentiate components using different keys here? Seems the annotation is applied directly to pod rather than the shared `TidbCluster` CR, do we still need to keep track of the component information?

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -458,6 +466,34 @@ func (c *PodController) syncTiDBPod(ctx context.Context, pod *corev1.Pod, tc *v1
 	return reconcile.Result{}, nil
 }
 
+func (c *PodController) cleanupLeaderEvictionAnnotations(pod *corev1.Pod, tc *v1alpha1.TidbCluster, annKeys []string) error {
+	for _, ann := range annKeys {
```
hanlins, line 470: Would be nice to be defensive
```suggestion
func (c *PodController) cleanupLeaderEvictionAnnotations(pod *corev1.Pod, tc *v1alpha1.TidbCluster, annKeys []string) error {
	if pod.Annotations == nil {
		return nil
	}
	for _, ann := range annKeys {
```

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -458,6 +466,34 @@ func (c *PodController) syncTiDBPod(ctx context.Context, pod *corev1.Pod, tc *v1
...
+	}
+
+	if _, err := c.deps.PodControl.UpdatePod(tc, pod); err != nil {
+		return perrors.Annotatef(err, "failed delete expired annotations for tc %s/%s", pod.Namespace, pod.Name)
+	}
+	return nil
+}
+
+func (c *PodController) isEvictLeaderExpired(pod *corev1.Pod, annKey string) bool {
+	timeToExpireAnnValue, exist := pod.Annotations[annKey]
```
hanlins, line 481:
```suggestion
func (c *PodController) isEvictLeaderExpired(pod *corev1.Pod, annKey string) bool {
	if pod.Annotations == nil {
		return false
	}
	timeToExpireAnnValue, exist := pod.Annotations[annKey]
```

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -249,6 +249,10 @@ func (c *PodController) syncPDPod(ctx context.Context, pod *corev1.Pod, tc *v1al
 		return reconcile.Result{}, nil
 	}
 
+	if c.isEvictLeaderExpired(pod, v1alpha1.PDLeaderTransferExpirationTimeAnnKey) {
+		return reconcile.Result{}, c.cleanupLeaderEvictionAnnotations(pod, tc, []string{v1alpha1.PDLeaderTransferAnnKey, v1alpha1.PDLeaderTransferExpirationTimeAnnKey})
```
hanlins, line 253: It seems the annotation is cleared during pod sync if the expiration time is reached. Would it be a problem if next time when the operator enters this reconciliation loop(well at that moment it's still expired), the `isEvictLeaderExpired` would return `false` since the annotation is removed, and then it proceeds with the leader eviction?

hanlins (commented):
Thanks for the changes! Some comments are added, might be nice to put one sentence or two in the release note so people could understand the feature added.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1382,6 +1382,10 @@ const (
 	PDLeaderTransferAnnKey = "tidb.pingcap.com/pd-transfer-leader"
 	// TiDBGracefulShutdownAnnKey is the annotation key to graceful shutdown tidb pod by user.
 	TiDBGracefulShutdownAnnKey = "tidb.pingcap.com/tidb-graceful-shutdown"
+	// TiKVEvictLeaderExpirationTimeAnnKey is the annotation key to expire evict leader annotation. Type: time.RFC3339.
+	TiKVEvictLeaderExpirationTimeAnnKey = "tidb.pingcap.com/tikv-evict-leader-expiration-time"
+	// PDLeaderTransferExpirationTimeAnnKey is the annotation key to expire evict leader annotation. Type: time.RFC3339.
```
csuzhangxc, line 1387:
```suggestion
	// PDLeaderTransferExpirationTimeAnnKey is the annotation key to expire transfer leader annotation. Type: time.RFC3339.
```

csuzhangxc (approved):

pkg/apis/pingcap/v1alpha1/types.go
hanlins: Well I would claim that those annotations mentioned above are added by Xiaomou and Hao recently, maybe it's not too late to change the course. I'll leave it to you :)

pkg/controller/tidbcluster/pod_control.go
hanlins: Ok, that makes sense.

hanlins (approved):

PR #5140, opened 2023-07-04
ticdc: upgrade ticdc first if version >= v5.1.0

hi-rustin:
### What problem does this PR solve?

close https://github.com/pingcap/tidb-operator/issues/4966

### What is changed and how does it work?

upgrade ticdc first if version >= v5.1.0

### Code changes

- [x] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [x] Unit test 
- [ ] Manual test 

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [ ] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note
Upgrade ticdc first if its version is bigger than or equal to v5.1.0
```

Issue #5011, opened 2023-05-17
tidb-controller-manager panics if tidb-cluster is configured with AdditionalVolumeMounts


sergiomcalzada:
## Bug Report

**What version of Kubernetes are you using?**

Client Version: v1.24.10
Kustomize Version: v4.5.4
Server Version: v1.24.10

**What version of TiDB Operator are you using?**

TiDB Operator Version: version.Info{GitVersion:"v1.4.4", GitCommit:"4b1c55c9aabf0410e276efcb460e37488f203b66", GitTreeState:"clean", BuildDate:"2023-03-13T06:41:33Z", GoVersion:"go1.19.7", Compiler:"gc", Platfor
m:"linux/amd64"}

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**

N/A

**What's the status of the TiDB cluster pods?**

Controller manager crashloopback
**What did you do?**

Updated the operator from 1.3.7 to 1.4.4

**What did you expect to see?**
The operator continues working

**What did you see instead?**
The operator is on cashloopback 

NOTE: The backupSchedule and the Cluster have that additional volume mounted `tidb/tidb-backup-pvc`. There is a warning about matiching desired volume size, but we didn't changed the size. Indeed there is no size configured there for it

```
W0517 10:43:33.428815       1 phase.go:69] volume tidb/tidb-backup-pvc modification is not allowed: can't match desired volume
E0517 10:43:33.429089       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 650 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x2a88e40?, 0x48c9990})
	k8s.io/apimachinery@v0.19.16/pkg/util/runtime/runtime.go:74 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc000433f90?})
	k8s.io/apimachinery@v0.19.16/pkg/util/runtime/runtime.go:48 +0x75
panic({0x2a88e40, 0x48c9990})
	runtime/panic.go:884 +0x212
github.com/pingcap/tidb-operator/pkg/manager/volumes.observeVolumeStatus({0x34df500, 0xc000641b90}, {0xc000b39500, 0x3, 0x4?}, {0xc001c33ef0, 0x1, 0x1})
[...]

PR #5150, opened 2023-07-06
Add Volume Replacement feature with support for pd, tidb, tikv

anish-db:

Implement a generic feature to make volume changes to tidb via replacement.
Detects volume changes (storageclass change, size change, or number of volumes change)
Notably change in number of volumes such as adding a separate TiKV raft disk is poorly handled by operator today.

Deletes replicas one by one and replaces with new disks that use the new volume parameters in a graceful way similar to how scale in is implemented for components today. (e.g for TiKV it deletes stores and waits for transition).
Adds one spare replica (for PD & TiKV) for handling load while the current one is being deleted.

PR Implements support for TiDB, PD, TiKV components.

More details of problem described in https://docs.google.com/document/d/1WpAYVcF1fXpcuVH6mFlz9v0fbLG3EA9r2gsDgFdqW8E/edit#heading=h.sv3296iop7bh

Detailed List of changes:
* Refactor some volume comparing utils from `pod_vol_modifier.go` & `pvc_modifier.go` in to `vol_compare_utils.go` for reuse in this feature as well as some new functions to compare volume changes in pods, etc.
* Explicitly detect extra volumes (or missing volumes) when comparing statefulset and throw an error as unsupported for pvc_modifier's use case. (Which was previously ignored silently)
* Add a `tidb.pingcap.com/replace-disk` annotation that can be used to trigger a pvc replace implemented in `pkg/controller/tidbcluster/pod_control.go`.
  - TiKV
    1. Get Store ID for Pod via Status/Label. Ensure label is set with store id, this is important to avoid races during scale down later.
    2. Ensure all TiKVs are currently ready and stable.
    3. If store is Up, delete it via PD and wait for it to become Tombstone
    4. Delete all PVCs associated with pod, then delete the pod itself.
    5. StatefulSet will automatically. recreate a new replacement which will come up on its own.
  - TiDB: If other dependent components & TiDB instances are stable, go ahead and delete the pod's PVCs & the pod itself.
  - PD:
    1. Get Member ID for this PD Pod via label
    2. If this pod is leader, transfer leader to next available member.
    3. If this pod is still part of PD members & all PD's are ready, delete it as a member
    4. Delete PVCs & the Pod, let statefulset recreate it.
* Add a VolReplaceInProgress Status field to tidbcluster definition, for TiKV, TiDB & PD components. + supporting functions for it in component_status.go
* Add a new PvcReplacer component that orchestrates new feature with two functions `UpdateStatus()` & `Sync()`
  - In `tidb_cluster_control.go` `UpdateStatus()` is run before component member manager syncs, and `Sync()` is run after.
  - `UpdateStatus()` is responsible for detecting storage volume changes and sets the `VolReplaceInProgress` status field for the respective component.
  - Each supported component's member manager Sync() does the following when a `VolReplaceInProgress` is set:
    + For TiKV and PD an extra desired replica is configured, which triggers ScaleOut
    + The StatefulSet's update strategy is set to OnDelete, so that config+volume changes can be rolled out atomically (by deleting pod with annotation)
     + Any actual config changes to pod spec is suppressed from being updated by member manager itself (bypassing it's upgrade routines)
  - PvcReplacer's `Sync()` waits until scaling is complete, then deletes statefulset orphaning pods & waits for member manager to recreate it (with new config), then applies the `tidb.pingcap.com/replace-disk` annotation to each pod one by one.
*  Add a `tidb.pingcap.com/tikv-no-active-store-since` annotation for TiKV pods (set by pkg/controller/pod_control.go) when a TiKV pod has a valid store id in label but no active stores from PD (via status tikv stores fields). Used to detect Tombstone stores that have been GC'ed too soon (due to a potential PD early GC race when stores are deleted immediately after creation)
* Handle an edge case in TiKV scale in where a pod's store id can be missing from PD and also not available as tombstone (due to PD GC race). Use no-active-store-since annotation above to assume tombstone'd after some waiting time and scale it in.
* Move `MetaManager.Sync()` in `tidb_cluster_control.go` which sets tikv store labels to run before component member manager Sync's so that label is updated, even if member managers error.

### Code changes

- [X] Has Go code change

### Tests

- [X] Unit test 
- [ ] E2E test 
- [X] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Support a generic disk replacement mechanism for making pvc changes behind a new feature (default off) for TiDB, TiKV & PD components.
```

### Manual Test

Here's a procedure to manually test this feature on a test cluster:

Enable feature by setting `features: [VolumeReplacing=true]` in `charts/tidb-operator/values.yaml`
Start a local operator using `./hack/local-up-operator.sh`

Apply the basic tidb cluster but with 3 tikv replicas:
```
cp examples/basic/tidb-cluster.yaml /tmp/basic.yaml
patch /tmp/basic.yaml << EOF
33c33
<     replicas: 1
---
>     replicas: 3
EOF
kubectl --context kind-kind apply -f /tmp/basic.yaml
```

Once cluster is up, apply an update adding a second volume with a config change to be rolled out atomically:
```
cp /tmp/basic.yaml /tmp/twodisk.yaml
patch /tmp/twodisk.yaml << EOF
47a48,53
>       raft-engine:
>         dir: /var/lib/raftlog/raftdb
>     storageVolumes:
>       - name: raftdb-volume
>         storageSize: 1Gi
>         mountPath: /var/lib/raftlog
EOF
kubectl --context kind-kind apply -f /tmp/twodisk.yaml
```

Verify that new temporary tikv replica is created, new volumes created by deleting & replace, and spare replica cleaned up.
Takes about ~10min for test cluster.


pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+		errStr := fmt.Sprintf("Number of volumes mismatch desired %d vs sts %d for cluster %s", len(ctx.desiredVolumes), len(sts.Spec.VolumeClaimTemplates), ctx.ComponentID())
+		if errOnNewVol {
+			newVolErr = errors.New(errStr)
+		} else {
+			klog.Warningf(errStr)
+		}
+		return false, newVolErr
+	}
+
+	return true, nil
```
hanlins, line 189: It seems in this function, the returned error is only `newVolErr`. I think instead of passing in `errOnNewVol` as a flag to differentiate the behavior on emiting errors or not, maybe it's simpler to let the caller of this function to decide whether to propagate the error or not. That said, we can do `isSynced, err := p.utils.IsStatefulSetSynced(ctx, ctx.sts)` if the error is needed or `isSynced, _ := p.utils.IsStatefulSetSynced(ctx, ctx.sts)` if error can be ignored, what do you think?

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+}
+
+func (u *volCompareUtils) IsPodSyncedForReplacement(ctx *componentVolumeContext, pod *corev1.Pod) (bool, error) {
+	// Does not check underlying StorageClass for modifications, only matches PVC specs for use by pvc replacer.
+	ns := pod.Namespace
+	podPvcCount := 0
+	for i := range pod.Spec.Volumes {
+		vol := &pod.Spec.Volumes[i]
+		pvc, err := u.getPVC(ns, vol)
```
hanlins, line 198:
```suggestion
	for _, vol := range pod.Spec.Volumes {
		pvc, err := u.getPVC(ns, &vol)
```

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+func getStorageSize(r corev1.ResourceList) resource.Quantity {
+	return r[corev1.ResourceStorage]
+}
+
+func ignoreNil(s *string) string {
+	if s == nil {
+		return ""
+	}
+	return *s
+}
```
hanlins, line 372: Can we use `pointer.StringDeref` instead? See `"k8s.io/utils/pointer"`

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+		podPvcCount++
+		desired := getDesiredVolumeByName(ctx.desiredVolumes, v1alpha1.StorageVolumeName(vol.Name))
+		if desired == nil {
+			// Extra un-desired volume in pod, needs to be removed.
+			return false, nil
+		}
+		desiredScName := desired.GetStorageClassName()
+		// If desired sc name is unspecified , any sc used by pvc is okay. (Default storage class).
+		if desiredScName != "" && ignoreNil(pvc.Spec.StorageClassName) != desiredScName {
```
hanlins, line 213: Should we also return `true` if desiredScName is set to empty string? If let's say a sc name is changed from a non-empty sc name to the empty string, I think user's intention is to switch from the original sc to the default sc, right? And maybe we should propagate the sc name to pvc?

hanlins (commented):
Hi @anish-db, thanks for your patience! I had an initial round of review, I'm still reading into the details but I don't want to put you in the wait for so long. Maybe we can start from these initial comments before I finish the review entirely? And don't hesitate to let me know if you have some different ideas in these comments, thanks!

pkg/features/features.go
```diff
@@ -50,6 +51,10 @@ const (
 	// VolumeModifying controls whether allow to modify volumes
 	// NOTE: volume resize is always allowed even if this feature is disabled
 	VolumeModifying string = "VolumeModifying"
+
+	// VolumeReplacing controls whether to replace whole volumes by deleting and recreating on changes.
+	// tidb, tikv & pd supported.
+	VolumeReplacing string = "VolumeReplacing"
```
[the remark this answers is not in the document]
anish-db: It should be fine to enable both, but replacing takes precedence, elaborated on the other comment.

pkg/controller/tidbcluster/tidb_cluster_control.go
```diff
@@ -281,6 +293,14 @@ func (c *defaultTidbClusterControl) updateTidbCluster(tc *v1alpha1.TidbCluster)
 		}
 	}
 
+	// Replace volumes if necessary. Note: if enabled, takes precedence over pvcModifier.
+	if features.DefaultFeatureGate.Enabled(features.VolumeReplacing) {
```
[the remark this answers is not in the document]
anish-db: After https://github.com/pingcap/tidb-operator/commit/df734da0eb0c73d71f09944d8f31d604631d4bfa I think resizing is turned on by default anyways.

I've tested after this commit, and replacer is designed to take precedence over modification if enabled. Basically the Sync here will keep requeuing until all replacement is complete, so it will never give pvcModifier below a chance. When replacement is done, modifier will have nothing to do. Without precedence pvc modifier may keep failing for changes it can't handle and never let replacer run.

This means if a modification would have been enough, replacement feature will still replace instead of resizing, this is a downside to note. Maybe in future we can coordinate if pvc modifier can handle it let it do instead of replace, but didn't complicate it in this PR.

So it should be fine to turn on both, but only one will take effect. Didn't want to throw an error in case we default modification feature later. Let me know what you think.

pkg/manager/volumes/pvc_replacer.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return errutil.NewAggregate(errs)
+}
+
+func (p *pvcReplacer) Sync(tc *v1alpha1.TidbCluster) error {
+	components := tc.AllComponentStatus()
+	errs := []error{}
+
+	for _, comp := range components {
```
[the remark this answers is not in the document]
anish-db: Done.

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	pods, err := u.getPodsOfComponent(tc, comp)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx.pods = pods
+	ctx.sts = sts
+	ctx.shouldEvict = comp == v1alpha1.TiKVMemberType
```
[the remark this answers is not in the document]
anish-db: Several of the functions in this file where refactored here for reuse from the pvc_modifier feature.
Copied it as it is from existing code uses it here:
https://github.com/pingcap/tidb-operator/blob/899409c2f2858b6aa8e2f9a46db4607e32ee6fc4/pkg/manager/volumes/pvc_modifier.go#L73
https://github.com/pingcap/tidb-operator/blob/899409c2f2858b6aa8e2f9a46db4607e32ee6fc4/pkg/manager/volumes/pvc_modifier.go#L144

Didn't want to modify existing code much, so left it as it is in this. (lmk if you prefer to still modify it with change to pvc_modifier)

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+		if size.Cmp(desired.Size) != 0 {
+			return false, nil
+		}
+		scName := volTemplate.Spec.StorageClassName
+		if !isStorageClassMatched(desired.StorageClass, ignoreNil(scName)) {
+			return false, nil
+		}
+	}
+	// TODO: verify new behaviour is fine with pvc_modifier, previously number of volumes not checked.
```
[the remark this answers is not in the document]
anish-db: Removed TODO, but see above comment.

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+
+func (u *volCompareUtils) IsStatefulSetSynced(ctx *componentVolumeContext, sts *appsv1.StatefulSet, errOnNewVol bool) (bool, error) {
+	var newVolErr error = nil
+	for _, volTemplate := range sts.Spec.VolumeClaimTemplates {
+		volName := v1alpha1.StorageVolumeName(volTemplate.Name)
+		size := getStorageSize(volTemplate.Spec.Resources.Requests)
+		desired := getDesiredVolumeByName(ctx.desiredVolumes, volName)
+		if desired == nil {
+			// TODO: verify behaviour is correctly intended, this was previously being ignored. Now explicit error in
+			// pvc modifier..
```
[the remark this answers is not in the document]
anish-db: Removed the TODO, but I left this as a note to call out during review.
This function was originally used by existing pvc modifier here:
https://github.com/pingcap/tidb-operator/blob/899409c2f2858b6aa8e2f9a46db4607e32ee6fc4/pkg/manager/volumes/pvc_modifier.go#L201

We are now checking change in number of volumes and marking it as an error for pvc modifier (since it cannot handle it). Previously it would silently ignore it and allow other updates to continue. So this is a new behavior change I wanted to call out to make sure it's okay.

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+			errStr := fmt.Sprintf("volume %s in sts for cluster %s does not exist in desired volumes", volName, ctx.ComponentID())
+			if errOnNewVol {
+				newVolErr = errors.New(errStr)
```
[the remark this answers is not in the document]
anish-db: There are two users of this function pvc_modifer and pvc_replacer.
For pvc_modifier change in number of disks cannot be handled so is an error for it.
For replacer it is okay.

I've added a note for this function explaining this and also changed to always return an error and handle on caller side like you've suggested.

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,385 @@
...
+		scName := volTemplate.Spec.StorageClassName
+		if !isStorageClassMatched(desired.StorageClass, ignoreNil(scName)) {
+			return false, nil
+		}
+	}
+	// TODO: verify new behaviour is fine with pvc_modifier, previously number of volumes not checked.
+	if len(sts.Spec.VolumeClaimTemplates) != len(ctx.desiredVolumes) {
+		errStr := fmt.Sprintf("Number of volumes mismatch desired %d vs sts %d for cluster %s", len(ctx.desiredVolumes), len(sts.Spec.VolumeClaimTemplates), ctx.ComponentID())
+		if errOnNewVol {
+			newVolErr = errors.New(errStr)
```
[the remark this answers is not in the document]
anish-db: Done.

pkg/manager/volumes/vol_compare_utils.go
anish-db: Done. Also added comments in places where it is ignored.

pkg/manager/volumes/vol_compare_utils.go
anish-db: Applied.

pkg/manager/volumes/vol_compare_utils.go
anish-db: If desiredScName is blank, the pvc storage class will still not be blank but use whatever is the default storage class configured in the k8s cluster. It is a little tricky to find out which is default and use it to compare because the default might change later and trigger a change in disks with operator.

Existing pvc modifier also uses this behaviour (ignore changes if desired sc is blank), e.g.:
https://github.com/pingcap/tidb-operator/blob/899409c2f2858b6aa8e2f9a46db4607e32ee6fc4/pkg/manager/volumes/phase.go#L200
https://github.com/pingcap/tidb-operator/blob/899409c2f2858b6aa8e2f9a46db4607e32ee6fc4/pkg/manager/volumes/pod_vol_modifier.go#L403

So I think safer behavior is to not change storage class if it is not explicitly specified.

pkg/manager/volumes/vol_compare_utils.go
anish-db: This was old code's function so I left it untouched, but let me know if you prefer to update it with existing uses also to StringDeref

https://github.com/pingcap/tidb-operator/blob/899409c2f2858b6aa8e2f9a46db4607e32ee6fc4/pkg/manager/volumes/pvc_modifier.go#L110

anish-db (commented):
Thanks for your comments @hanlins , replied inline.

pkg/manager/volumes/vol_compare_utils.go
anish-db: Actually had to undo this, I had to use i to avoid this linter error:

```
pkg/manager/volumes/vol_compare_utils.go:183:28: G601: Implicit memory aliasing in for loop. (gosec)
		pvc, err := u.getPVC(ns, &vol)
```

pointer to loop variable Issue might not apply in this case, but safer to avoid and appease the linter.

anish-db:
Pushed commit for lint fixes (forgot to push earlier)

hanlins (approved):
Thanks for the change! Had a discussion with @anish-db, overall LGTM, targeting `v1.5.1`. Some potential followups:
1. Reject changes on TiFlash storage components in validation webhook or support volume replacing for TiFlash in a future release
2. Decouple volume modifying and resizing features and provide an option to turn-off volume modifying so replacer/modifier don't coexist and we don't need to rely on volume replacer taking precedence.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -1394,6 +1398,8 @@ const (
 	TiKVEvictLeaderExpirationTimeAnnKey = "tidb.pingcap.com/tikv-evict-leader-expiration-time"
 	// PDLeaderTransferExpirationTimeAnnKey is the annotation key to expire transfer leader annotation. Type: time.RFC3339.
 	PDLeaderTransferExpirationTimeAnnKey = "tidb.pingcap.com/pd-evict-leader-expiration-time"
+	// ReplaceDiskAnnKey is the annotation key to replace disks used by pod.
+	ReplaceDiskAnnKey = "tidb.pingcap.com/replace-disk"
```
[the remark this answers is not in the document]
anish-db: I've replaced all references in this PR for disk to volume so it's consistent everywhere.

pkg/apis/pingcap/v1alpha1/tidbcluster.go
```diff
@@ -558,7 +558,11 @@ func (tc *TidbCluster) PDStsDesiredReplicas() int32 {
 	if tc.Spec.PD == nil {
 		return 0
 	}
-	return tc.Spec.PD.Replicas + tc.GetPDDeletedFailureReplicas()
+	var spareReplaceReplicas int32 = 0
+	if tc.Status.PD.VolReplaceInProgress {
+		spareReplaceReplicas = 1
```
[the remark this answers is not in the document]
anish-db: I've expanded the comments on the feature at `charts/tidb-operator/values.yaml` including this behaviour. Lmk if there is a better place to document it.

pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -0,0 +1,371 @@
...
+func (u *volCompareUtils) IsStatefulSetSynced(ctx *componentVolumeContext, sts *appsv1.StatefulSet) (bool, error) {
+	// Note: The only errors that IsStatefulSetSynced returns is when number of disks change. This cannot be handled by
+	// pvc_resizer so is an error for it, however pvc_replacer can handle it and will ignore all errors returned here
+	// and only rely on the boolean return (which must be set to false if number of disks change on top of error)
+	for _, volTemplate := range sts.Spec.VolumeClaimTemplates {
+		volName := v1alpha1.StorageVolumeName(volTemplate.Name)
+		size := getStorageSize(volTemplate.Spec.Resources.Requests)
+		desired := getDesiredVolumeByName(ctx.desiredVolumes, volName)
+		if desired == nil {
+			return false, fmt.Errorf("volume %s in sts for cluster %s does not exist in desired volumes", volName, ctx.ComponentID())
```
csuzhangxc, line 161: May this returned error block the reconciliation if the additional volume is used?

pkg/controller/tidbcluster/pod_control.go
```diff
@@ -235,6 +240,14 @@ func (c *PodController) getPDClient(tc *v1alpha1.TidbCluster) pdapi.PDClient {
 }
 
 func (c *PodController) syncPDPod(ctx context.Context, pod *corev1.Pod, tc *v1alpha1.TidbCluster) (reconcile.Result, error) {
+	result, err := c.syncPDPodForLeaderTransfer(ctx, pod, tc)
+	if err != nil || result.Requeue || result.RequeueAfter > 0 {
+		return result, err
+	}
+	return c.syncPDPodForReplaceVolume(ctx, pod, tc)
```
csuzhangxc, line 247: How about checking the feature gate here?

pkg/manager/member/tikv_scaler.go
```diff
@@ -191,6 +191,9 @@ func (s *tikvScaler) scaleInOne(tc *v1alpha1.TidbCluster, skipPreCheck bool, upT
 	for _, store := range tc.Status.TiKV.Stores {
 		if store.PodName == podName {
 			state := store.State
+			if pod.Labels[label.StoreIDLabelKey] == "" {
+				return deletedUpStore, fmt.Errorf("StoreID not yet updated on pod label")
+			}
```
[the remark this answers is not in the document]
anish-db: I'm not sure if there is some edge case where store id in label won't match store id in tc status. If that happens I don't want to block it, as tc status value is more authoritative. The label is only required because we need to confirm tombstone store afterwards. (tc.status won't have store id once deleted). This is for the case 3 situation below.
So I feel better safe to not block deletion if store mismatches somehow (we are deleting & fixing. here anyways)

pkg/manager/volumes/vol_compare_utils.go
anish-db: Please read the Note in comments above (~line 153).
This function is used by both pvc replacer and pvc modifier. In replacer error is ignored because differences will be fixed by it. For pvc modifier it is intentional error that we should block reconciliation.

If user does not have replacer enabled and has requested additional volume, pvc modifier cannot make the change and I feel better to explicitly block reconciliation with error than silently ignoring additional volume and rolling out half a change for user.

pkg/manager/volumes/pvc_replacer.go
```diff
@@ -0,0 +1,213 @@
...
+			continue
+		}
+		ctx, err := p.utils.BuildContextForTC(tc, comp)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("build ctx used by replacer sync for %s/%s:%s failed: %w", tc.Namespace, tc.Name, comp.MemberType(), err))
+			continue
+		}
+		err = p.replaceVolumes(ctx)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("replace volumes for %s failed: %w", ctx.ComponentID(), err))
```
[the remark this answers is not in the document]
anish-db: Done.

pkg/manager/volumes/pvc_replacer.go
```diff
@@ -0,0 +1,213 @@
...
+	ns := ctx.sts.Namespace
+	name := ctx.sts.Name
+
+	// Ignore errors as they only indicate change in number of volume which replacer can handle.
+	isSynced, _ := p.utils.IsStatefulSetSynced(ctx, ctx.sts)
+	if isSynced {
+		return nil
+	}
+	if utils.StatefulSetIsUpgrading(ctx.sts) {
+		return fmt.Errorf("component sts %s/%s is upgrading", ctx.sts.Name, ctx.sts.Namespace)
```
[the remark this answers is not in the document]
anish-db: Fixed

pkg/controller/tidbcluster/pod_control.go
csuzhangxc: OK

pkg/manager/volumes/vol_compare_utils.go
csuzhangxc: But if the user does not want to use the volume replace feature, the previous volume modify feature will be blocked here. If other words, the previous behavior is broken.

anish-db: Yes this is actually a change in existing behavior, but I believe the existing behavior was also problematic.

I'd called this out earlier in review here: https://github.com/pingcap/tidb-operator/pull/5150/commits/4f3ef146cb15d16861e05ecc2de255414f785f34#r1275584130
And wasn't sure which way pingcap folks preferred.

Today pvc modifer would silently ignore extra volumes, which can be problematic. Fundamentally tidb operator does not have proper support for changing number of volumes before this PR.

Can you provide some use case when replacer feature is disabled that users will want to add a new volume?
New volume cannot be used without a config change. Today operator will push config change without the volume (silently ignoring volume) and will cause pods to crash because missing volume/path. Due to this user cannot make new volume anyways today, so I thought safer to block the reconciliation instead of silently ignoring it and causing pod crashes.

If you feel better to keep old behavior or have some use case where you think it's needed, I can revert to old behavior, let me know.

csuzhangxc: #5011 may be one use case.

anish-db: It doesn't seem like the user was trying to change additionalMounts in that issue. Existing additionalMounts won't be a problem in this PR, changing additionalMounts is what is problematic.

I think we should decide what is the expected behavior if user changes additionalMount for an existing cluster without enabling replace feature. Today's behavior is to ignore the mount but rollout everything else , which I believe is rarely user's intent and can cause issues. Proposed behavior is to explicitly block upgrade as it is explicitly unsupported instead of doing a partial update

Either case, should not affect the replace feature. So if you prefer old behavior I can split this function and revert for existing modifier's behavior. Let me know.

csuzhangxc: If the user changes the additionalMount without enabling replace feature, then TiDB Operator blocks it should be OK.

But if the user changes the main PVC (e.g size) but with additionalMount, then TiDB Operator should not block when checking this additionalMount.

anish-db: That should still be fine here?
This line only blocks if desired = nil, but a volume was defined in current spec. (I.e storageVolumes was defined).
Note that additionalVolume/mounts are not used anywhere in this logic, only storageVolumes field.

If additional volumes/mounts or storage volumes are used , this code won't block as long as they are not being changed. If only size is being changed line 163 below will catch it return a difference without error thereby not blocking it.

liubog2008: `additionalVolumes` are append into `sts.spec.podTemplate.spec.volumes` but not `sts.spec.volumeClaimTemplates`.

The `storageVolume` deletion is not well handled in previous logic. There are two problems:
1. STS will always be recreated if it's not synced. Some PVCs maybe left if `storageVolumes` are deleted.
2. If only `storageVolume` are deleted, STS is synced.

There are three choice to fix this behavior:
1. Throw a clear err to notify users that changes of `storageVolumes` is unsupported if this feature is not enabled.
2. Fix both problem 1 and 2, clean up left PVCs of deleted `storageVolumes`.
3. Fix the problem 2. And just announce that PVCs of deleted `storageVolumes` will be left because of safety.
 
For me, option 1 > 3 > 2. So I think it's enough to add a more clear comment and return a more clear error.
@csuzhangxc What do you think about?

csuzhangxc: @liubog2008 Thanks for your detailed explanation, I agree with your suggestion ( 1 > 3 > 2).
@anish-db How about making the error messages more explicit as it's unsupported?

anish-db: Sure, I've expanded the error warning to be more prominent. I've added this under pvc_modifier.go where this function gets used and the error is handled. New Error reads like:

`change in number of volumes is not supported without VolumeReplacing feature. Reconciliation is blocked due to :`

Followed by either of the two error conditions returned by this function.

hanlins (approved):
I had a round of validation, the volume modifier feature is not impacted and works as expected. I tested tuning iops of TiKV from 400 to 4200. The original volume info is follows:
```console
➜  op kubectl -n tidb3839733 get pvc
NAME                       STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS        AGE
o11y-vector-agent          Bound    pvc-0f50fbdc-4641-4661-a10a-b21e7b723ee0   2Gi        RWO            o11y-metric-agent   20m
pd-db-pd-0                 Bound    pvc-e9b20e8e-1ac2-42f3-b5bb-c5fe241ecef1   10Gi       RWO            pd-gp3-4000-400     20m
tidb-slowlog-db-tidb-0     Bound    pvc-1dbca99c-845b-4d1a-8133-358a5b6339c8   30Gi       RWO            auditlog            17m
tidb3839733-vmagent        Bound    pvc-56d364d6-4671-471b-9097-578705d16934   4Gi        RWO            o11y-metric-agent   20m
tidbmonitor-db-monitor-0   Bound    pvc-5b6a2bb1-b85d-4f1e-82dc-7f7fcc2bdee9   50Gi       RWO            cloud-ssd           20m
tikv-db-tikv-0             Bound    pvc-3362ae49-99b3-492c-bb9e-21057b9d1639   10Gi       RWO            tikv-gp3-4000-400   18m
tikv-db-tikv-1             Bound    pvc-d221dce1-7a3c-4f80-9b0b-316064af8bb9   10Gi       RWO            tikv-gp3-4000-400   18m
tikv-db-tikv-2             Bound    pvc-0fbc500f-d4fc-4d14-aa17-8ed05795968d   10Gi       RWO            tikv-gp3-4000-400   18m
➜  op kubectl get pv pvc-0fbc500f-d4fc-4d14-aa17-8ed05795968d -oyaml | grep volumeHandle
    volumeHandle: vol-0a9867efab0e1723f
```
The original iops is like following on aws console:
image
I replaced the operator image with a dev build:
```console
➜  op kubectl -n tidb-admin get deploy tidb-controller-manager -oyaml | grep "image:"
        image: xxx.dkr.ecr.us-west-2.amazonaws.com/tidbcloud/tidb-operator:v0.1
```
After changing iops using volume modifier, I found the iops for the corresponding image is changed:
image
image
Then I checked the TC is ready as expected and the image is indeed the dev build wIe used previously:
```console
➜  op kubectl -n tidb3839733 get tc db
NAME   READY   PD                                                                                     STORAGE   READY   DESIRE   TIKV                                                                                     STORAGE   READY   DESIRE   TIDB                                                                                     READY   DESIRE   AGE
db     True    385595570414.dkr.ecr.us-west-2.amazonaws.com/tidbcloud/pd:v6.5.1-20230304-1677915816   10Gi      1       1        385595570414.dkr.ecr.us-west-2.amazonaws.com/tidbcloud/tikv:v6.5.1-20230304-1677915816   10Gi      3       3        385595570414.dkr.ecr.us-west-2.amazonaws.com/tidbcloud/tidb:v6.5.1-20230304-1677915816   1       1        141m
➜  op kubectl -n tidb-admin get deploy tidb-controller-manager -oyaml | grep 'image:'
        image: 200683210682.dkr.ecr.us-west-2.amazonaws.com/tidbcloud/tidb-operator:v0.1
```
This wraps the manual validation process, that result looks promising. cc @csuzhangxc.

pkg/controller/tidbcluster/tidb_cluster_control.go
```diff
@@ -174,6 +178,23 @@ func (c *defaultTidbClusterControl) updateTidbCluster(tc *v1alpha1.TidbCluster)
...
+			return err
+		}
+	}
+
+	// syncing the labels from Pod to PVC and PV, these labels include:
+	//   - label.StoreIDLabelKey
+	//   - label.MemberIDLabelKey
+	//   - label.NamespaceLabelKey
+	// Also updates the TiKV pod annotation for label.AnnTiKVNoActiveStoreSince if store is missing.
+	if err := c.metaManager.Sync(tc); err != nil {
```
csuzhangxc, line 193: > Move MetaManager.Sync() in tidb_cluster_control.go which sets tikv store labels to run before component member manager Sync's so that label is updated, even if member managers error.

Could it cause a deadlock if meta sync failed as the member manager can't sync some necessary operations?

csuzhangxc:
rest LGTM

pkg/controller/tidbcluster/tidb_cluster_control.go
anish-db: It's a fair concern, I've gone through the metamanger Sync to see what errors it could possibly throw, and they are usually errors from kubernetes operations which should be safe to retry without blocking on a member sync / deadlock.
https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/meta/meta_manager.go#L39
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/pod_control.go#L107
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/pv_control.go#L121
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/pvc_control.go#L144

The only concerning portion that stood out to me is that there is a pdclient call:
https://github.com/pingcap/tidb-operator/blob/master/pkg/controller/pod_control.go#L126C34-L126C34

Just to be safe I've moved this after pd member manager, Other than that I can't see any errors that would depend on member manager to resolve. The only new dependency is that this runs before tikv member manager.

csuzhangxc: #5140 will change the sync order, and TiCDC may need to be synced before PD

anish-db: Ah ok, I've changed the code to make this call twice now. Once before all the member managers but without blocking on error. And the second after all the member managers which blocks on error (existing old location). Should be safe to do this twice since it's idempotent, and the first one will now ignore error to avoid deadlock.  Do you think this is okay?

Issue #5213, opened 2023-07-29
Support custom version strings on images with operator.


anish-db:
## Feature Request

**Is your feature request related to a problem? Please describe:**
We adapt the official pingcap images thru our internal docker image pipeline to produce images with custom version tags that don't follow the tidb version string format of `v6.5.3` etc.  e.g. our version strings look like `tidb_2023-06-27_00.43.39Z_master_123abcde_123456789`

Operator makes various assumptions to gate functionality based on wether a component is new enough past some version or not. For this it tries to parse the `.spec.version` field from the tidb cluster definition and errors out in our case like:

```
E0729 00:22:23.727017       1 utils.go:464] Parse version tidb_2023-06-27_00.43.39Z_master_123abcde_123456789 failure, error: Invalid Semantic Version
W0729 00:22:23.927754       1 tidb_member_manager.go:1102] parse tidb verson 'tidb_2023-06-27_00.43.39Z_master_123abcde_123456789' failed, skip setting store labels for TiKV of TiDB cluster tidb/devdev. err: Invalid Semantic Version
```

This makes a lot of error logs & prevents us from using various functionality and potentially incorrect usage at various operator code paths:
https://github.com/pingcap/tidb-operator/blob/34c19d4b977df6afe5f3a60db17905f71a2fc1cf/pkg/manager/member/tidb_member_manager.go#L1144
https://github.com/pingcap/tidb-operator/blob/34c19d4b977df6afe5f3a60db17905f71a2fc1cf/pkg/manager/member/utils.go#L463
https://github.com/pingcap/tidb-operator/blob/34c19d4b977df6afe5f3a60db17905f71a2fc1cf/pkg/manager/member/ticdc_scaler.go#L236

**Describe the feature you'd like:**
We would like to be able to specify an override version string for informing operator what tidb versions we are using which is separate from the actual image tag used for container images.

**Describe alternatives you've considered:**
N/A

**Teachability, Documentation, Adoption, Migration Strategy:**
Explained above.

cc: @coderplay @gegao-db

csuzhangxc:
adding a new field as the desired version to override the version string from the image tag should work. But in some cases, if we need to know the current version of a Pod or StatefulSet, then we may need to get this information from HTTP APIs or store this information in the annotation of Pod/Statefulset.

anish-db:
Yes, I think using version over HTTP Api in some cases would be fine since they would still be accurate in our case.

adamf-db:
[...]

PR #5233, opened 2023-08-08
add support for sha256 container image specifications

adamf-db:
The operator doesn't work when the image spec includes a sha256 image reference. 

For security reasons, I much prefer to use the sha256 method of retrieving container images, rather then tags. Eg, we want to do `pingcap/tidb@sha256:deadbeef....` instead of `pingcap/tidb:v7.1.0`. However, this breaks the operator because the operator tests if it can set labels in https://github.com/pingcap/tidb-operator/blob/master/pkg/manager/member/tidb_member_manager.go#L1142 by checking if the version extracted from the container image spec is greater than `v6.3.0`. 

When I use the sha256 method of container image specification this code reports that we can't set labels and bails on all the code after that check.

This change makes it so that any valid sha256 is considered a greater version number then any other, similar to the logic for handling `latest` or `nightly`. 

The logic in the semantic version comparison is changed to support sha256 values. A value is consider a sha256 if `hex.DecodeString` returns no error and the length of the string is 64. 

### Code changes

- [X] Has Go code change

### Tests

- [X] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [X] Other side effects: 
There's a potential that people are using 6.3.0 or early with sha256 container image specs.

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/manager/member/tidb_member_manager_test.go
+++ b/pkg/manager/member/tidb_member_manager_test.go
@@ -2799,6 +2799,19 @@ func TestTiDBMemberManagerSetServerLabels(t *testing.T) {
 			setCount: 0,
 		},
 		{
+			name:        "sha256 version string",
+			tidbVersion: "d7f62aab6315b4378cbbfaaaaaaaaaaaaaaaa90ecaf7e0f22d2225304822ee2e",
+			members: []Member{
+				{
+					node: "node-1",
+				},
+				{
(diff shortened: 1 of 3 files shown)


csuzhangxc:
@adamf-db how about adding version stings as #5213

PR #5279, opened 2023-09-05
[WIP] Remove setting gc life time among dumpling

hongyunyan:
When we do dumpling in cloud env, root has no permissions to run the sql such as `select variable_value from tidb where variable_name =  tikv_gc_life_time`, which leading dumpling failed directly.

If we do dumpling not in cloud env, we also don't need the logical about setting gc life time in tidb-operator, because we already have the similar logical in dumpling code(use pd client to set instead of running the sql)

Thus we need to remove the logical about setting gc life time in tidb-operator.

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [x] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


breezewish:
How about using a field to control this behavior? Otherwise it may affect existing users.

breezewish (approved):

Issue #5204, opened 2023-07-26
Operator will be stuck on rollback when reverting a change in-between a partial component upgrade.


anish-db:
## Bug Report

If we try to rollback a change mid-way while it was upgrading (say some but not all replicas upgraded), then operator will get stuck in upgrade phase and not actually rollback the updated replicas.
Noticed in PD & TiKV, but likely affects tidb and other components.

**What version of Kubernetes are you using?**
1.26

**What version of TiDB Operator are you using?**
1.4.5 but can also reproduce on test environment with latest operator built from HEAD

**What did you do?**
Steps to repro using dev operator:

`./hack/local-up-operator.sh`

Start a test tidb cluster with 3 tikv replicas:

```
cp examples/basic/tidb-cluster.yaml /tmp/basic.yaml
patch /tmp/basic.yaml << EOF
33c33
<     replicas: 1
---
>     replicas: 3
EOF
kubectl --context kind-kind apply -f /tmp/basic.yaml
```

Prepare a variation with a small tikv config change to trigger a test update.
```
cp /tmp/basic.yaml /tmp/basic2.yaml
patch /tmp/basic2.yaml << EOF
45c45
<         max-open-files: 256
---
>         max-open-files: 255
EOF
```

Apply the update:
`kubectl --context kind-kind apply -f /tmp/basic2.yaml`

Wait for a few seconds and watch till the first pod (basic-tikv-2) updates and restarts and then immediately rollback:

`kubectl --context kind-kind apply -f /tmp/basic.yaml`

**What did you expect to see?**
Operator should rollback basic-tikv-2

**What did you see instead?**
Operator remains stuck with tikv in upgrade phase, and basic-tikv-2 remains stuck in different version.

```
% kubectl --context kind-kind get tc basic -o json | jq '.status.tikv.phase'
"Upgrade"
% kubectl --context kind-kind get sts basic-tikv -o json | jq '.status'
{
  "availableReplicas": 3,
  "collisionCount": 0,
  "currentReplicas": 2,
  "currentRevision": "basic-tikv-888479b96",
  "observedGeneration": 4,
  "readyReplicas": 3,
  "replicas": 3,
  "updateRevision": "basic-tikv-888479b96",
  "updatedReplicas": 2
}
% kubectl --context kind-kind get pod basic-tikv-2 -o json | jq '.metadata.labels["controller-revision-hash"]'
"basic-tikv-565c5bdbf6"
[...]

PR #5291, opened 2023-09-15
Remove early return during upgrade that might block rollbacks.

anish-db:
Fix operator so it properly rollbacks pods when a partial upgrade is rolled back.
Issue described in https://github.com/pingcap/tidb-operator/issues/5204

Remove the early return when updateRevision and currentRevision match.
This was purely for optimization, and instead now let the operator check each individual pod matches
before confirming that the Upgrade() function can return as completed.

### Code changes

- [X] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fix an issue that prevents rollback of partially in-flight updates to pods
```

### DIFF
--- a/pkg/manager/member/dm_master_upgrader.go
+++ b/pkg/manager/member/dm_master_upgrader.go
@@ -62,10 +62,6 @@ func (u *masterUpgrader) gracefulUpgrade(dc *v1alpha1.DMCluster, oldSet *apps.St
 		return nil
 	}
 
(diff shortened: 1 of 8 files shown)


anish-db:
Looks like it needs more test fixes, closing for now.

csuzhangxc:
@anish-db do you need any help about this PR?

anish-db:
Re-sending this PR as https://github.com/pingcap/tidb-operator/pull/5345

PR #5381, opened 2023-11-09
Make PD and TiKV wait until local IP address matches the one published to external DNS

smineyev81:
Currently PD and TiKV pods await for dns name assigned to them to be resolvable by external DNS but they do not check if that IP matches IP assigned to them which causes problems on pod restart when DNS has old IP which might not match new IP. Which leads to false-positive: PD/TiKv pods are ready from k8s point of view but not accessible by dns name assigned to them.

This change adds extra check to DP and TiKv startup scripts that makes sure that IP address received from external DNS matches the IP on a host (essentially `nslookup` result should match one of the IPs returned by `hostname -I`)

### Code changes

- [x] Has Go code change

### Tests
- [x] Unit test
- [ ] E2E test
- [ ] Manual test
- [ ] No code

### Side effects
- [ ] Other side effects:

### Related changes

### Release Notes
NONE


csuzhangxc (commented):
If we update the start scripts directly and once we upgrade the TiDB Operator, all existing components will be restarted, this is not acceptable in some scenes.

smineyev81:
> If we update the start scripts directly and once we upgrade the TiDB Operator, all existing components will be restarted, this is not acceptable in some scenes.

addressed with moving new feature under feature flag which is exposed via new field in TidbCluster.Spec.WaitForDnsNameIpMatchOnStartup

pkg/manager/member/startscript/v2/common.go
```diff
@@ -37,6 +37,65 @@ then
...
+
+	componentCommonWaitForDnsIpMatchScript = `
+elapseTime=0
+period=1
+while true; do
+    sleep ${period}
+    elapseTime=$(( elapseTime+period ))
+
+    if [[ ${elapseTime} -ge ${waitThreshold} ]]; then
+        echo "waiting for pd cluster ready timeout" >&2
```
csuzhangxc, line 51: `pd` is not `common` as this script is used both for PD and TiKV.

csuzhangxc (commented):
please execute `hack/update-all.sh` to update CRDs to fix the CI

pkg/manager/member/startscript/v2/common.go
smineyev81: fixed

smineyev81:
> please execute `hack/update-all.sh` to update CRDs to fix the CI

regenerated and pushed

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -375,6 +375,10 @@ type TidbClusterSpec struct {
 
 	// PreferIPv6 indicates whether to prefer IPv6 addresses for all components.
 	PreferIPv6 bool `json:"preferIPv6,omitempty"`
+
+	// WaitForDnsNameIpMatchOnStartup indicates whether PD and TiKV has to wait
+	// until local IP address matches the one published to external DNS
+	WaitForDnsNameIpMatchOnStartup bool `json:"waitForDnsNameIpMatchOnStartup,omitempty"`
```
csuzhangxc, line 381: How about adding a comment to indicate it is only supported in start script v2 now?

pkg/manager/member/startscript/v2/tikv_start_script.go
```diff
@@ -101,9 +114,18 @@ done
 {{- end }}
 `
 
+	tikvWaitForDnsIpMatchSubScript = `
+componentDomain={{ .AdvertiseHost }}
+waitThreshold={{ .KVStartTimeout }}
+nsLookupCmd="getent ahosts $componentDomain | sed -n 's/ *STREAM.*//p'"
```
csuzhangxc, line 120: why not use `dig` as PD?

smineyev81: `dig` is not available on tikv image

pkg/apis/pingcap/v1alpha1/types.go
smineyev81: done

csuzhangxc: I didn't see any new commit after this comment.

smineyev81: > I didn't see any new commit after this comment.

sorry, push did not go through first time.
now changes are there

csuzhangxc: you need to re-generate OpenAPI spec.

smineyev81: what do you think if I change `WaitForDnsNameIpMatchOnStartup bool` with more generic field `StartupScriptFeatureFlags []string `json:"startupScriptFeatureFlags, omitempty"`` ? This will allow us to enable features without extra fields to CRD in future.

csuzhangxc:
> /run-all-tests

do not be surprised if some E2E tests failed as the CI environment is not stable now

csuzhangxc (approved):

PR #5393, opened 2023-11-16
Fix issue when using VolumeReplace feature that may mark wrong PD member ID labels.

anish-db:
Fix an issue Introduced in https://github.com/pingcap/tidb-operator/pull/5150 by running metaManager.Sync() once before the member syncs status.

Specifically in rare cases (AFAIK only when using volume replace) there can be a race condition.
`controller/pod_control.go` can run right after a pd member is deleted + pod is deleted + new pod comes back all before pd member manager has had a chance to run it's `syncTidbClusterStatus` function
This leaves `tc.Status.PD.Members` in an old state without removing the member that was deleted.
This can cause controller/pod_control.go to run with an out of date stale tc.Status.PD.Members and use the wrong member id to apply as pod label for PD.
Consequently causing more bugs when that pd pod is restarted (leader eviction missed), or disk replaced (pvc deleted without deleting pd member), or scaled in (without deleting pd member), etc.

I believe this race condition can only trigger with disk replace as it is rare for pd member delete + pod delete + new pod with same name to happen quickly in any other case. The deletion today happens in `tidbcluster/pod_control.go` which runs in a separate thread than member managers/metamanager/`controller/pod_control.go`, hence the race conditions.

Revert the extra metamanager.Sync() introduced in https://github.com/pingcap/tidb-operator/pull/5150 to go back to old behavior with only one run
Instead run UpdateMetaInfo for the specific pod being scaled in by tikv_scaler which is the only dependency that was needed/introduced. Since scaler will run only after tikv's `syncTidbClusterStatus` , updating the labels now is safe.

Some extra changes to improve safety:
 * Add error messages when pod_control detects mismatched store/member labels on pod. No action taken, but will help debug bad state.
 * In disk replace for PD, check all PD's are available before triggering pd leader transfer (avoid leader transfer to bad PD)

### Code changes

- [X] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fix an issue where VolumeReplacement feature on PD can cause stale PD member labels resulting in potential PD members left without cleanup or PD restarts without leader eviction.
```

### DIFF
--- a/pkg/manager/member/tikv_scaler.go
+++ b/pkg/manager/member/tikv_scaler.go
@@ -191,7 +191,7 @@ func (s *tikvScaler) scaleInOne(tc *v1alpha1.TidbCluster, skipPreCheck bool, upT
 	// update it once here (to avoid a dependency on metaManager to sync it first instead)
 	pod, err = s.deps.PodControl.UpdateMetaInfo(tc, pod)
 	if err != nil {
-		klog.Errorf("tikvScaler.ScaleIn: failed to update pod MetaInfo for %s", podName)
+		klog.Errorf("tikvScaler.ScaleIn: failed to update pod MetaInfo for, pod %s, cluster %s/%s", podName, ns, tcName)
 		return deletedUpStore, nil
 	}
 
--- a/pkg/manager/member/tikv_scaler.go
+++ b/pkg/manager/member/tikv_scaler.go
@@ -187,6 +187,14 @@ func (s *tikvScaler) scaleInOne(tc *v1alpha1.TidbCluster, skipPreCheck bool, upT
 		return deletedUpStore, fmt.Errorf("tikvScaler.ScaleIn: failed to pass up stores check , pod %s, cluster %s/%s", podName, ns, tcName)
 	}
 
+	// Below code depends on tikv StoreIDLabelKey & AnnTiKVNoActiveStoreSince to be correctly updated, so manually
+	// update it once here (to avoid a dependency on metaManager to sync it first instead)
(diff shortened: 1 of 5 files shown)


anish-db:
Leaving this in draft, still have to verify if this covers all cases

anish-db:
This is now ready for review

pkg/manager/member/tikv_scaler.go
```diff
@@ -187,6 +187,14 @@ func (s *tikvScaler) scaleInOne(tc *v1alpha1.TidbCluster, skipPreCheck bool, upT
 		return deletedUpStore, fmt.Errorf("tikvScaler.ScaleIn: failed to pass up stores check , pod %s, cluster %s/%s", podName, ns, tcName)
 	}
 
+	// Below code depends on tikv StoreIDLabelKey & AnnTiKVNoActiveStoreSince to be correctly updated, so manually
+	// update it once here (to avoid a dependency on metaManager to sync it first instead)
+	pod, err = s.deps.PodControl.UpdateMetaInfo(tc, pod)
+	if err != nil {
+		klog.Errorf("tikvScaler.ScaleIn: failed to update pod MetaInfo for %s", podName)
```
csuzhangxc, line 194: How about logging in with ns and tcName too?

anish-db: Good suggestion, added.

csuzhangxc (approved):

PR #5454, opened 2023-12-12
Add sleep before shutdown backup container

wxiaomou:

Add sleep before shutdown backup container
### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/images/tidb-backup-manager/entrypoint.sh
+++ b/images/tidb-backup-manager/entrypoint.sh
@@ -17,6 +17,8 @@ set -e
 
 # The trap command is to make sure the sidecars are terminated when the jobs are finished
 cleanup() {
+    echo "Sleeping for 30 seconds before exit..."
+    sleep 30
     touch /tmp/pod/main-terminated
 }
 


images/tidb-backup-manager/entrypoint.sh
```diff
@@ -17,6 +17,8 @@ set -e
 
 # The trap command is to make sure the sidecars are terminated when the jobs are finished
 cleanup() {
+    echo "Sleeping for 30 seconds before exit..."
```
nkg-, line 20: Can you please add a comment motivating the sleep.

wxiaomou: @nkg- added

BornChanger (approved):

WangLe1321 (commented):
I think you should wait in your sidecar instead of backup-manager container. It blocks normal exit without sidecar.

WangLe1321 (changes requested):
I think you should wait in your sidecar instead of backup-manager container. It blocks normal exit without sidecar.

wxiaomou:
@WangLe1321 our logging sidecar will flash log by batch, the timeout is set to 5s, if we reduce the sleep to 10s does it work for you? Current cleanup pod disappear within 5s, it makes it very hard to discover it.

WangLe1321:
> @WangLe1321 our logging sidecar will flash log by batch, the timeout is set to 5s, if we reduce the sleep to 10s does it work for you? Current cleanup pod disappear within 5s, it makes it very hard to discover it.

@wxiaomou you can only sleep in the clean command. I don't want to influence all the backup pods.

nkg-:
> > @WangLe1321 our logging sidecar will flash log by batch, the timeout is set to 5s, if we reduce the sleep to 10s does it work for you? Current cleanup pod disappear within 5s, it makes it very hard to discover it.
> 
> @wxiaomou you can only sleep in the clean command. I don't want to influence all the backup pods.

Backup pods have the same need right. Some time to publish logs externally. Now I'm not sure how much time is sufficient. @wxiaomou : Do we need 30 seconds. Or is 10 seconds sufficient.

wxiaomou:
@nkg- 10 second should be good
@WangLe1321 I can remove the sleep after     `touch /tmp/pod/main-terminated` but we need 10 second sleep before it to give our logging sidecar enough time to upload complete logs. This applied to all br pods.

WangLe1321 (approved):

PR #5457, opened 2023-12-13
br: fix backup stuck due to init pod creating stuck

### LINKED ISSUE #5456: backup is stuck when the init pod creating stuck and exceed ttl
## Bug Report

v1.5

**What did you do?**

I create a `VolumeBackup`, but the init pod creating is stuck. The init pod is always in `PodInitializing` state. After 15 minutes, it's deleted by job ttl.

the `VolumeBackup` is failed

**What did you see instead?**

the `VolumeBackup` is still running.

WangLe1321:

Closes #5456 

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


pkg/backup/backup/backup_manager.go
```diff
@@ -244,7 +244,7 @@ func (bm *backupManager) checkVolumeBackupInitializeJobRunning(backup *v1alpha1.
 		// all the volume snapshots has created
 		return nil
 	}
-	if !v1alpha1.IsVolumeBackupInitialized(backup) || v1alpha1.IsVolumeBackupInitializeFailed(backup) {
+	if v1alpha1.IsVolumeBackupInitializeFailed(backup) {
```
nkg-, line 247: Noob question. checkVolumeBackupInitializeJobRunning is used to detect init job failures. Based on the "err" returned from this function. IF IsVolumeBackupInitializeFailed == true, then why we return nil, rather than returning an error.

nkg-, line 247: Let me understand the removal of IsVolumeBackupInitialized. Now if the init job is stuck and eventually ttled, then  this method won't find the job, and return "init job deleted before ..." error here https://github.com/pingcap/tidb-operator/blob/6ec83b038432fd5c64458e883d7c12d02e83b5d3/pkg/backup/backup/backup_manager.go#L268. Right?

WangLe1321: No, the job still exists but it has failed status due to ttl. So we will find job failed and set backup failed here https://github.com/pingcap/tidb-operator/blob/6ec83b038432fd5c64458e883d7c12d02e83b5d3/pkg/backup/backup/backup_manager.go#L292

In addition, I verified the change by unit test. We can't remove the condition directly because in the first reconcile, the backup job is not created and we will get a 'not found' error. So I modified the condition.

WangLe1321: Because the aim of this function is to detect init job failure and set backup CR to `IsVolumeBackupInitializeFailed`. So if `IsVolumeBackupInitializeFailed == true`, it means we don't need this check. If we return error here, it can block the main logic of the backup controller.

BornChanger (approved):

BornChanger (approved):

PR #5572, opened 2024-03-13
ebs br: add resilience to single TiKV crash

BornChanger:
### What problem does this PR solve?

This PR provides the resilience to a single TiKV crash due to wal corruption at restore

### What is changed and how does it work?

 Add a new attribute `spec.tolerateSingleTiKVOutage` to restore CR. When it's set to true, restore to the tc can tolerate a single TiKV outage due to wal log corruption issue.

### Code changes

- [x] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [x] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note

```

PR #5569, opened 2024-03-11
Fail restore if any warmup job failed for volume-snapshot restores

michaelmdeng:
### What problem does this PR solve?

Currently restore only checks if all warmup jobs completed (success or fail) before continuing w/ later steps of restore. Instead, if any warmup jobs fail, we should fail the restore.

Observed behavior:

* Failed warmup job conditions:

```yaml
...
conditions:
  - lastProbeTime: "2024-03-10T21:05:24Z"
    lastTransitionTime: "2024-03-10T21:05:24Z"
    message: Job has reached the specified backoff limit
    reason: BackoffLimitExceeded
    status: "True"
    type: Failed
failed: 5
ready: 0
startTime: "2024-03-10T20:53:09Z"
uncountedTerminatedPods: {}
```

* Results in `WarmUpComplete` in restore

```yaml
commitTs: "448239596192667254"
conditions:
  - lastTransitionTime: "2024-03-10T20:51:49Z"
    status: "True"
    type: Scheduled
  - lastTransitionTime: "2024-03-10T20:51:57Z"
    status: "True"
    type: Running
  - lastTransitionTime: "2024-03-10T20:52:11Z"
    status: "True"
    type: VolumeComplete
  - lastTransitionTime: "2024-03-10T20:52:57Z"
    status: "True"
    type: WarmUpStarted
  - lastTransitionTime: "2024-03-10T22:02:49Z"
    status: "True"
    type: WarmUpComplete
phase: WarmUpComplete
progresses:
  - lastTransitionTime: "2024-03-10T20:52:11Z"
    progress: 100
    step: Volume Restore
timeCompleted: null
timeStarted: "2024-03-10T20:51:57Z"
```

### What is changed and how does it work?

When any warmup jobs fail, fail the entire restore. Verified this in a testing workload by artificially causing warmup to fail and observing `Restore` (and consequently, `VolumeRestore`) put into `Failed` state

### Code changes

- [x] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [x] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note

```

PR #5621, opened 2024-04-16
Fail restore if warmup fails only during `check-wal-only` strategy

michaelmdeng:

#5569 updates volume-snapshot restore process to fail the entire restore if any warmup job failed. We use this to quickly check the viability of restores and terminate restore processing early if a corruption is detected.

#5572 updates volume-snapshot restore process to enable recovery from a corruption in a single TiKV through manual cluster operations. We use this in a full restore in case we encounter a corruption during this process.

These features are in conflict w/ each other. If we want to perform a full restore and use single TiKV recovery in the event of corruption, we cannot fail the restore during warmup and instead need to complete warmup stage and progress to restarting TiKVs. If we only want to check the viability of a restore, we are ok w/ failing the restore and not progressing to any further steps. Thus, we gate this failure behavior only behind the `check-wal-only` strategy.

Gate restore failure on warmup failure only for `check-wal-only` warmup strategy.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/backup/restore/restore_manager.go
+++ b/pkg/backup/restore/restore_manager.go
@@ -1298,7 +1298,7 @@ func (rm *restoreManager) waitWarmUpJobsFinished(r *v1alpha1.Restore) error {
 	for _, job := range jobs {
 		finished := false
 		for _, condition := range job.Status.Conditions {
-			if condition.Type == batchv1.JobFailed {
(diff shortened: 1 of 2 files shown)


YuJuncen (approved):
Rest LGTM

BornChanger (approved):

PR #5622, opened 2024-04-16
Make warmup failure/skip on corruption configurable

michaelmdeng:

Currently warmup script exits w/ error when it encounters corruption. Similar to #5621, this behavior makes sense for checking viability of backups, but conflicts w/ desire to perform a full restore and recover.

When we perform a full restore, if restore encounters corruption, warmup will fail and will cancel the actual warmup of the volumes. Desired behavior is to log the corruption but continue the warmup operation, so that, when complete, if there is only a single corrupt TiKV, we can attempt to recover using functionality added in #5585.

Change updates warmup script to only exit/cancel warmup when encountering corruption when explicitly enabled. This is only enabled for `check-wal-only` strategy. Otherwise, we only log the corruption but continue with normal warmup.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/images/ebs-warmup/warmup-steps.sh
+++ b/images/ebs-warmup/warmup-steps.sh
@@ -31,6 +31,7 @@ Supported flags:
     --block <files_to_be_warmed_up_by_fio>
     --fs <files_to_be_warmed_up_by_filesystem>
     --debug enable \`set -x\`
+    --exit-on-corruption
 EOF
 }
 
@@ -53,7 +54,16 @@ cleanup() {
 trap cleanup EXIT
 
 operation=none
+exit_on_corruption=false
 bg_works=""
+
+for arg in "$@"; do
+    case $arg in
+        --exit-on-corruption) exit_on_corruption=true
+            ;;
+    esac
+done
+
 while [ $# -gt 0 ]; do
     case $1 in
         --help | -h) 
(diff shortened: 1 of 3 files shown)


images/ebs-warmup/warmup-steps.sh
```diff
@@ -53,7 +54,16 @@ cleanup() {
 trap cleanup EXIT
 
 operation=none
+exit_on_corruption=false
 bg_works=""
+
+for arg in "$@"; do
+    case $arg in
+        --exit-on-corruption) exit_on_corruption=true
```
YuJuncen, line 62: We may also need to match this flag before line 79 or the script may exit due to the unrecognized flag (Maybe parse this flag in the `while` loop will be fine?):

https://github.com/pingcap/tidb-operator/pull/5622/files#diff-113f3c9592cd00b3ff52744edb177373f178c3c7a0ed08607af5558a7f54d76aR79-R80

YuJuncen (commented):
Rest LGTM, thanks!

images/ebs-warmup/warmup-steps.sh
michaelmdeng: ah you're right, this will exit in the next while loop if provided .

however the while loop below does both arg parsing and does the work of WAL checking, so parsing there requires explicit ordering of `--exit-on-corruption` flag prior to any other args. while we do specify the correct ordering when generating warmup args, would prefer to not rely on it.

will keep the parsing here and then just no-op on the `exit-on-corruption` flag in the while loop below

YuJuncen (approved):

BornChanger (approved):

PR #5634, opened 2024-04-24
br: clean volumes when restore volume failed

### LINKED ISSUE #5638: Volume leaked if volume restore failed in restore volume step
## Bug Report

v1.5

**What did you do?**

execute a volume restore, but the restore failed due to restore volume failed.

some volumes are restored, these volumes can be cleaned by tidb-operator automatically

**What did you see instead?**

the restored volumes are still reserved in aws console

WangLe1321:

Add cleaning volumes function when restore volume failed. Then we can avoid volume leak

Closes #5638

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

manual test steps:

1. create a cluster
2. create a volume backup
3. create a restore cluster
4. create a volume restore using the volume backup in step 2. in this step, restore volume is successful and the tikv pods are started
5. edit the restore CR, modify its status from `VolumeComplete` to `Failed` to mock restore volume failed scenario
6. edit the tc CR, remove the annotation `tidb.pingcap.com/tikv-volumes-ready` to block tikv creation, then delete the tikv statefulset to detach the EBS volumes
7. wait for the volumes detached and deleted by tidb-operator

image

image

image

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/restore.go
+++ b/pkg/apis/pingcap/v1alpha1/restore.go
@@ -137,7 +137,7 @@ func IsRestoreVolumeFailed(restore *Restore) bool {
 		!IsRestoreVolumeComplete(restore)
 }
 
-// IsCleanVolumeComplete returns true if a Restore for volume is Failed
+// IsCleanVolumeComplete returns true if restored volumes are cleaned
 func IsCleanVolumeComplete(restore *Restore) bool {
 	_, condition := GetRestoreCondition(&restore.Status, CleanVolumeComplete)
 	return condition != nil && condition.Status == corev1.ConditionTrue
--- a/pkg/apis/pingcap/v1alpha1/restore.go
+++ b/pkg/apis/pingcap/v1alpha1/restore.go
@@ -130,6 +130,19 @@ func IsRestoreVolumeComplete(restore *Restore) bool {
 	return condition != nil && condition.Status == corev1.ConditionTrue
 }
 
+// IsRestoreVolumeFailed returns true if a Restore for volume is Failed
+func IsRestoreVolumeFailed(restore *Restore) bool {
+	return restore.Spec.Mode == RestoreModeVolumeSnapshot &&
+		IsRestoreFailed(restore) &&
+		!IsRestoreVolumeComplete(restore)
+}
+
+// IsCleanVolumeComplete returns true if a Restore for volume is Failed
+func IsCleanVolumeComplete(restore *Restore) bool {
+	_, condition := GetRestoreCondition(&restore.Status, CleanVolumeComplete)
+	return condition != nil && condition.Status == corev1.ConditionTrue
+}
+
 // IsRestoreWarmUpStarted returns true if all the warmup jobs has successfully started
 func IsRestoreWarmUpStarted(restore *Restore) bool {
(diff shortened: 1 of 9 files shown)


BornChanger (approved):

csuzhangxc (approved):

pkg/apis/pingcap/v1alpha1/restore.go
```diff
@@ -130,6 +130,19 @@ func IsRestoreVolumeComplete(restore *Restore) bool {
 	return condition != nil && condition.Status == corev1.ConditionTrue
 }
 
+// IsRestoreVolumeFailed returns true if a Restore for volume is Failed
+func IsRestoreVolumeFailed(restore *Restore) bool {
```
nkg-, line 134: Trying to understand this what exactly does IsRestoreVolumeFailed entail.  

We should only delete volumes, if restore fails in the first step (ie RestoreVolume which creates the volumes), and before creating PV/PVCs which attach volumes to ec2 nodes. Its the volumes which are never attached to an ec2 node (via PVCs) which cause a leak. After PVCs are created, they are tracked and deleted by csi driver. 

Also If they restore fails during warmup, or restore_data, its actually useful to investigate the tikvs on what happened. If the volumes are deleted, then it would be hard to investigate.

pkg/backup/util/aws_ebs.go
```diff
@@ -204,6 +205,68 @@ func (e *EC2Session) DeleteSnapshots(snapIDMap map[string]string, deleteRatio fl
 	return nil
 }
 
+func (e *EC2Session) DeleteVolumes(volumeIDs []string) error {
```
nkg-, line 208: We should only delete volumes, which are unattached. How is the validated here.

nkg- (commented):
THanks for working on it. Left a couple of comments.

pkg/backup/util/aws_ebs.go
WangLe1321: If the volume is attached, the delete request will fail. It means, we can't delete an attached volume. In my test, if the tikv pod is running, the volume can't be deleted. This is the reason why I deleted the tikv statefulset in my test steps.

pkg/apis/pingcap/v1alpha1/restore.go
WangLe1321: We create pv/pvc and warm up volumes and restore data **only** if the restore CR has the status RestoreVolumeComplete. And if restore volume failed, the status of restore CR is Failed. So this condition can ensure the restore is failed at restore volume step.

nkg-: Got it. So `!IsRestoreVolumeComplete(restore)` is the condition which ensure restorevolume step has failed.

pkg/backup/util/aws_ebs.go
nkg-: Got it. Basically we rely in aws api contract to prevent attached volumes from getting deleted. `Deletes the specified EBS volume. The volume must be in the available state (not attached to an instance).`

https://docs.aws.amazon.com/cli/latest/reference/ec2/delete-volume.html

nkg- (commented):
Thanks for the responses.

PR #5637, opened 2024-04-25
[BR]initialize status metric to zero

wxiaomou:

Currently when the federated manager restart we lost the metrics, and the new status metric will start with value 1. In our dashboard we are measuring success/fail backups by increase() function, current implementation will cause we lost the back up status when federated manager restarts. 

This PR set the status counter to 0 before it emit the first metric. 
### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```


pkg/controller/fedvolumebackup/fed_volume_backup_control.go
```diff
@@ -130,6 +131,17 @@ func (c *defaultBackupControl) updateVolumeBackupMetrics(volumeBackup *v1alpha1.
 	}
 }
 
+// add zero to the backup status counters; this sets the backup counter metric to zero when the federated manager restarts.
+func (c *defaultBackupControl) addZeroToBackupStatusMetric(volumeBackup *v1alpha1.VolumeBackup) {
```
nkg-, line 135: nit: s/addZeroToBackupStatusMetric/initBackupStatusMetrics. or reset.

pkg/controller/fedvolumebackup/fed_volume_backup_control.go
```diff
@@ -130,6 +131,17 @@ func (c *defaultBackupControl) updateVolumeBackupMetrics(volumeBackup *v1alpha1.
...
+	ns := volumeBackup.Namespace
+	tcName := volumeBackup.GetCombinedTCName()
+	metrics.FedVolumeBackupStatusCounterVec.WithLabelValues(ns, tcName, string(v1alpha1.VolumeBackupRunning)).Add(0)
```
WangLe1321, line 138: it's better to use `GetMetricWithLabelValues()` to create a new counter if it doesn't exist instead of adding a zero value to the counter.

pkg/controller/fedvolumebackup/fed_volume_backup_control.go
WangLe1321: I think `initBackupStatusMetrics` is better, we should only do initialization.

wxiaomou: done

pkg/controller/fedvolumebackup/fed_volume_backup_control.go
wxiaomou: done

WangLe1321 (approved):

WangLe1321 (approved):

BornChanger (approved):

PR #5665, opened 2024-06-20
Add: tc level flag to enablePVCReplace feature

rajsuvariya:

https://github.com/pingcap/tidb-operator/blob/9ef26f88bf75338cef427b44aa6303237ddcdbd6/pkg/features/features.go#L55C9-L57C44
PVC replace feature is currently enabled / disabled at tidb-operator level, which might not be comforting situation as a single tidb-operator can be handling multiple tidbclusters, so this PR enables tidbcluster level flag to enable / disable this feature. This is similar to another similar feature `enablePVReclaim` flag

Added a new flag in CRD `enablePVCReplace` with a false as default value. If this flag is enabled then only tidb-operator would perform the PVC replacement process.

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
ACTION REQUIRED: Introduces a new feature flag at a tidb-cluster level for enablePVCReplace, this works the same as VolumeReplacing but only enables specified tidb-cluster for VolumeReplacing. Enabling either the tidb-cluster level feature flag or the operator level feature flag will turn on this feature for a given tidb-cluster.
```


pkg/manager/volumes/pvc_replacer.go
```diff
@@ -73,6 +73,11 @@ func (p *pvcReplacer) getVolReplaceStatusForComponent(tc *v1alpha1.TidbCluster,
 }
 
 func (p *pvcReplacer) UpdateStatus(tc *v1alpha1.TidbCluster) error {
+	if !tc.IsPVCReplaceEnabled() {
```
csuzhangxc, line 76: After this PR, if the user wants to enable PVC replace feature for a TidbCluster, both feature-gate in TiDB Operator and TidbCluster should be enabled, right?

If the feature-gate in TiDB Operator is enabled, and then upgrade TiDB Operator to a version with this PR, then this feature is disabled, this may be unexpected for these users.

rajsuvariya: Yes, that's why I have added in release notes. Also I believe this is not released feature and there is no documentation of this feature so customers are highly unlikely to be using this feature.

csuzhangxc: @anish-db WDYT

anish-db, line 76: Instead of disabling here, can we do it at the same place as the existing feature flag, just so both pieces of code disabling this are next to each other?

The gate for that is here: https://github.com/pingcap/tidb-operator/blob/9ef26f88bf75338cef427b44aa6303237ddcdbd6/pkg/controller/tidbcluster/tidb_cluster_control.go#L183C41-L183C65
and here: 
https://github.com/pingcap/tidb-operator/blob/9ef26f88bf75338cef427b44aa6303237ddcdbd6/pkg/controller/tidbcluster/tidb_cluster_control.go#L309

anish-db: Yes, LGTM to controlling this per cluster, we where thinking of adding something like this too (Since we may want to selectively disable if we want to use resize feature instead of replace).

Slightly related, when using these features we noticed a few edge cases where it would be useful to customize the "spareReplaceReplicas" which is currently default 1, I wanted to send a PR to make that customizable, just FYI will try to send one for it soon as I get a chance. But that will also be a similar parameter to this in CRD.

anish-db (changes requested):

pkg/manager/volumes/pvc_replacer.go
anish-db: Here's the proposal for the other change about spareReplaceReplicas that I mentioned: https://github.com/pingcap/tidb-operator/pull/5666

rajsuvariya: I am okay to move the check however I kept it inside considering it declutters the tidb_cluster_control.go file and it follows a similar approach as `enbalePVReclaim`. Feature gate probably makes sense at tidb_cluster_control.go level as if the feature itself is disabled why should we even go to pvc_cleaner?
https://github.com/pingcap/tidb-operator/blob/9ef26f88bf75338cef427b44aa6303237ddcdbd6/pkg/manager/member/pvc_cleaner.go#L75

Let me know if you want me to still move it to tidb_cluster_control.go?

anish-db: personally I feel easier to read the code if both disables are in the same place since they functionally do the same thing. but no strong preference eitherways. I'll leave it to @csuzhangxc

rajsuvariya: @csuzhangxc please check and let me know if this is fine or do you want me to move the check at tidb_cluster_control.go?

csuzhangxc: I prefer to move the check to tidb_cluster_control.go.

BTW, can we define the behavior as:
- if one of these feature-gates is enabled, this feature is enabled
- only if both of them are disabled, this feature is disabled

rajsuvariya: Ok I will move the check

However 2nd part doesn't make sense when we want to enable TiDB operator to handle volume replace feature but allow only if the TiDB cluster has this feature flag enabled.

Consider a scenario where 1 tidb-operator is managing 5 tidb clusters (1, 2, 3, 4, 5). We want to enable volume replace only for cluster 3. If we enable the feature gate it would enable it for all the clusters. 

Also can you let me know the reason why this has not been moved to GA? can you share the issues which needs to be fixed? 

One such issue I know is handling of TiFlash is not enabled which I will raise PR in next week.

anish-db: I think what @csuzhangxc is suggesting would give a path to eventually deprecate the feature-gate and only rely on this new tc-level setting. I'd be fine with dropping the feature-gate entirely in favor of this new tc-level setting alone if that's an option but may cause compatibility issues for anyone else relying on flag, will let Xuecheng decide.

TiCDC is one other component I did not get a chance to add support for VolumeReplace in my original PR

rajsuvariya: @anish-db I am aligned on the removal of the feature gate once it's GA. I would love to add support for both TiFlash and TiCDC in next PR. I hope that's fine.

rajsuvariya: @anish-db @csuzhangxc After giving a second thought, the suggestion from @csuzhangxc makes sense.

It makes this new feature gate backward compatible. And in our scenario where we want to enable it for specific clusters we can keep the feature gate disabled and only enable tc level flag. 

I have made the changes please review again

csuzhangxc (approved):

rajsuvariya:
I will raise another PR, due to CLA agreement

rajsuvariya:
Raised another PR #5670

PR #5684, opened 2024-07-16
Add generation to all component statuses

ari-e:
tidb-operator currently does not propagate the CR object meta's generation to the status. This is problematic because it means external systems cannot always tell whether the status corresponds to the current version of the CRD spec, or if it is an old status. For example consider the series of events
```
process A updates CR spec
process B reads CR status
tidb-operator reconciliation triggers on the object and updates status
```
Here, process B will think that the status corresponds to the updated spec as written by process A, when in reality it is a stale status belonging to a previous version of the CR. This is exacerbated if the operator has high latency or is briefly unavailable due to deploys or other reasons.

Each component status within the overarching TidbClusterStatus now has a `generation` field in addition to its other fields like `phase`. Whenever the `phase` is updated by the respective controllers, the generation is also set to be equal to the current CR's generation. Kubernetes increments this generation anytime the spec for the CR is updated, so this provides a consistent way to indicate the status reflects the current version of the CR spec.

Note this practice of passing the generation through to the status is standard practice and is how Kubernetes deployments work, and Kubernetes source code even indicates that [all custom operators should do this](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/registry/customresourcedefinition/strategy.go#L92-L99).

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [x] Breaking backward compatibility 
TidbCluster now registers status as a subresource, meaning changes to status do not increment the CR's generation number.
- [x] Other side effects: 
CRD changes need to be deployed to clusters

### Related changes

### Release Notes

```release-note
TidbCluster CRD now registers `status` as a subresource of the CR, so updates to status no longer update the `generation` meta field of the CR.

TiDB Operator now sets `observedGeneration` field on each component status anytime a component's `phase` is updated.
```

### DIFF
--- a/docs/api-references/docs.md
+++ b/docs/api-references/docs.md
@@ -9592,6 +9592,16 @@ MemberPhase
 </tr>
 <tr>
 <td>
+<code>observedGeneration</code></br>
+<em>
+int64
+</em>
+</td>
+<td>
+</td>
+</tr>
+<tr>
+<td>
 <code>statefulSet</code></br>
 <em>
 <a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#statefulsetstatus-v1-apps">
@@ -11174,6 +11184,16 @@ MemberPhase
 </tr>
 <tr>
 <td>
+<code>observedGeneration</code></br>
+<em>
+int64
+</em>
+</td>
+<td>
+</td>
+</tr>
+<tr>
+<td>
 <code>statefulSet</code></br>
 <em>
 <a href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#statefulsetstatus-v1-apps">
@@ -12382,6 +12402,16 @@ bool
 <a href="#memberphase">
 MemberPhase
 </a>
+</em>
+</td>
+<td>
+</td>
+</tr>
+<tr>
+<td>
+<code>observedGeneration</code></br>
+<em>
+int64
 </em>
 </td>
 <td>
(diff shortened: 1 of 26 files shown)


csuzhangxc:
Thanks for your contribution, but as our `status` is not a sub-resources of the CR now. when we updating the status, the generation in the metadata will also be updated. then this observed generation in status may be stale again. does this work in your case?

ari-e:
Hi @csuzhangxc thanks for the comment. Is not registering status as a subresource intentional? I can easily convert it to a subresource but then that would of course affect backwards incompatibility. I still think that's the right thing to do though as it's more canonical k8s.

csuzhangxc:
How about using the name of `observedGeneration` instead of `generation`?

csuzhangxc:
will it break any existing clusters after chaning it to subresources?

csuzhangxc:
> > will it break any existing clusters after changing it to subresources?
> 
> The only behavior change should be this difference with the generation number. So if someone is depending on generation being incremented whenever the status changes that will break. But that seems like depending on undefined behavior to me. Generation is meant for spec changes.
> 
> This could be included in the next major version release if that's a concern.

I still think the change of subresource is too critical. Can you make this change only to one CRD and do more tests to ensure it works correctly, especially for upgrading from a previous version of TiDB Operator.

BTW, the `e2e-example` case failed.

ari-e:
@csuzhangxc thanks for the feedback. I'll do some more testing and fix the e2e. And again this can go in the next major version release if the compatibility change is a big concern. I'm busy with other projects for the next week or so but will pick this up after.

PR #5696, opened 2024-07-25
[PD] add option to wait a certain time before start pd

wxiaomou:
### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -572,6 +572,10 @@ type PDSpec struct {
 	// +kubebuilder:default=30
 	StartTimeout int `json:"startTimeout,omitempty"`
 
+	// Wait time before pd get started
+	// +kubebuilder:default=0
+	InitWaitTime int `json:"initWaitTime,omitempty"`
+
 	// Mode is the mode of PD cluster
 	// +optional
 	// +kubebuilder:validation:Enum:="";"ms"
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -572,7 +572,8 @@ type PDSpec struct {
 	// +kubebuilder:default=30
 	StartTimeout int `json:"startTimeout,omitempty"`
 
-	// Wait time before pd get started
+	// Wait time before pd get started. This wait time is to allow the new DNS record to propagate,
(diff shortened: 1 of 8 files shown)


pkg/manager/member/startscript/v2/pd_start_script.go
```diff
@@ -181,6 +187,8 @@ const (
 	pdWaitForDnsIpMatchSubScript = `
 componentDomain=${PD_DOMAIN}
 waitThreshold={{ .PDStartTimeout }}
+initWaitTime={{ .PDInitWaitTime }}
+sleep initWaitTime
```
csuzhangxc, line 191: This will cause existing PD to be restarted after we upgraded TiDB Operator if `WaitForDnsNameIpMatch` has been enabled before.

BTW, what is the problem without this `sleep`?

wxiaomou: @csuzhangxc Thanks for  the reminder, the reason we want  to add this option is: currently we do have  the DNS resolved IP check with current pod ip during pd start script. But we saw an racing condition that the check passed but the pd still hit the wrong ip during it restarts. The reason is we have a set of core  dns pods, maybe some of them got the new record, but some does not. In this case we still hit the wrong IP even after the dns resolved IP check passed. Adding this option, we can force a wait of 2*dns_ttl time before pd start to  avoid such racing condition.

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -572,6 +572,10 @@ type PDSpec struct {
 	// +kubebuilder:default=30
 	StartTimeout int `json:"startTimeout,omitempty"`
 
+	// Wait time before pd get started
+	// +kubebuilder:default=0
+	InitWaitTime int `json:"initWaitTime,omitempty"`
```
csuzhangxc, line 577: How about adding a comment to explain this works with `WaitForDnsNameIpMatch`?

wxiaomou: @csuzhangxc added

csuzhangxc (approved):

abebts2000:
@csuzhangxc - can we cherry pick this PR to branch release-1.6 as we are not upgrading to 1.6 any time soon?

PR #4682, opened 2022-08-23
br: add log backup, include start, truncate, stop

WizardXiao:
### What problem does this PR solve?

Ref https://github.com/pingcap/tidb-operator/issues/4645
### What is changed and how does it work?

**The operation of log backup start, truncate, stop.**
1. open start, truncate, stop.
   1.1 start: there is no `logTruncateUntil` and `logStop` is false
   1.2 truncate: there is `logTruncateUntil` and `logStop` is false
   1.3 stop: `logStop` is true
   1.4 the priority is stop > truncate > start, such as, there are both `logStop` is true and `logTruncateUntil`, it will run stop.
   1.5 the relation of command is truncate/stop can run parallelly, but they need wait start complete.
2. the status of log backup and its commands.
    2.1. the whole log backup status and its commands status are separated, whole log backup status is still the info in `BackupStatus`, the commands status is in the map `logSubCommandStatuses` whose key is command type and value is the command's status.
    2.2. the commands status update will selectively modify the whole log backup status, such as start running, stop complete, truncate complete. the details is in [updateLogBackupStatus](https://github.com/pingcap/tidb-operator/pull/4682/files#diff-87b9f36ed20d95aff7bf42e5692e5f0fad07d7244b6eb097d82fc6923fe3096eR160).
    2.3. we just record the last time run the command.
3. the end of start, truncate, stop, if there is no error. just pay attention to the whole log status is enough.
   3.1 start: there will be a `commitTs` in `status`.
   3.2 truncate:  status's `logSuccessTruncateUntil`  >= spec's `logTruncatingUntil`.
   3.3 stop: the log backup will be `complete`.

**The technical implementation details.**
1. all of the log backup reconcile will create a job and start a pod like before.
2. the `updateBackup` in controller will just skip the running command update, other update will handle by `sync`.
3. the `sync` is only based on the actual whole log status to determine whether it is necessary to skip, requeue, or mark completed.
4. the `sync` will check the state of the job and delete old job to control the number of jobs which is only one at most for the same command, and make sure the same command is running one by one.
5. the `status updater` will update the whole log status and the command status together, and it will not use the spec's info to update status, because spec's info may be modified by the user at any time.
6. the commands will be converted to different commands to the `backup-manager` in the running pod.

### Code changes

- [x] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [x] Unit test 
- [x] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Breaking backward compatibility 
[...]

PR #5710, opened 2024-08-15
br: a more straight forward operator interface

### LINKED ISSUE #5699: pitr: add support to log backup subcommand of status/pause/resume
## Feature Request

currently, tidb operator does not support status/pause/resume log backup subcommands, while the pitr kernel supports them. br operator needs to expose the interfaces.

RidRisR:

Closes #5699 

1. Use a more straight forward operator interface to replace the old one.
2. support pause and resume pitr task

Now, the crd to control log backup would loops like
```spec:
  backupMode: log
  logSubcommand: log-start/log-stop/log-pause
  logTruncateUntil: "xxxxxxxxx"
```
(If no subcommand is provided, the default value is `log-start`)

If the task is not created before:
It could transfer to `running` state using `log-start`

If the task is running:
It could transfer to `paused` state using `log-pause`
It could transfer to `stop` state using `log-stop`

If the task is paused:
It could transfer to `running` state using `log-start`
It could transfer to `stop` state using `log-stop`

If the task is stopped:
This is the end state and can't transfer state again

As long as the task is created, user could set `logTruncateUntil` and truncate would be sync after subcommand is done

截屏2024-09-02 17 50 46

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
1. Use a more straight forward operator interface to replace the old one.
Now, the crd to control log backup would loops like
`spec:
  backupMode: log
  logSubcommand: `log-start`/`log-stop`/`log-pause`
  logTruncateUntil: "xxxxxxxxx"
`
(If no subcommand is provided, the default value is `log-start`)

2. Support pause and resume pitr task
Now, users could pause and resume pitr task by using `log-pause` and `log-start`
```

### DIFF
--- a/cmd/backup-manager/app/backup/manager.go
+++ b/cmd/backup-manager/app/backup/manager.go
@@ -472,8 +472,9 @@ func (bm *Manager) performLogBackup(ctx context.Context, backup *v1alpha1.Backup
 		resultStatus, reason, err = bm.stopLogBackup(ctx, backup)
 	case string(v1alpha1.LogTruncateCommand):
 		resultStatus, reason, err = bm.truncateLogBackup(ctx, backup)
+	//TODO: (Ris) support more funcs
 	case string(v1alpha1.LogResumeCommand):
-		klog.Errorf("Log resume command is not supported yet")
+		resultStatus, reason, err = bm.resumeLogBackup(ctx, backup)
 	case string(v1alpha1.LogPauseCommand):
 		klog.Errorf("Log pause command is not supported yet")
 	default:
@@ -551,6 +552,46 @@ func (bm *Manager) startLogBackup(ctx context.Context, backup *v1alpha1.Backup) 
 		TimeStarted:   &metav1.Time{Time: started},
 		TimeCompleted: &metav1.Time{Time: finish},
 		CommitTs:      &ts,
+	}
+	return updateStatus, "", nil
+}
+
+// resumeLogBackup resume log backup.
+func (bm *Manager) resumeLogBackup(ctx context.Context, backup *v1alpha1.Backup) (*controller.BackupUpdateStatus, string, error) {
+	started := time.Now()
+	backupFullPath, err := util.GetStoragePath(backup)
+	if err != nil {
+		klog.Errorf("Get backup full path of cluster %s failed, err: %s", bm, err)
+		return nil, "GetBackupRemotePathFailed", err
+	}
+	klog.Infof("Get backup full path %s of cluster %s success", backupFullPath, bm)
+
+	updatePathStatus := &controller.BackupUpdateStatus{
+		BackupPath: &backupFullPath,
+	}
+
+	// change Prepare to Running before real backup process start
+	if err := bm.StatusUpdater.Update(backup, &v1alpha1.BackupCondition{
+		Command: v1alpha1.LogResumeCommand,
+		Type:    v1alpha1.BackupRunning,
+		Status:  corev1.ConditionTrue,
+	}, updatePathStatus); err != nil {
+		return nil, "UpdateStatusFailed", err
+	}
+
+	// run br binary to do the real job
+	backupErr := bm.doResumeLogBackup(ctx, backup)
+
+	if backupErr != nil {
+		klog.Errorf("Resume log backup of cluster %s failed, err: %s", bm, backupErr)
+		return nil, "ResumeLogBackuFailed", backupErr
+	}
+	klog.Infof("Resume log backup of cluster %s to %s success", bm, backupFullPath)
+
+	finish := time.Now()
+	updateStatus := &controller.BackupUpdateStatus{
+		TimeStarted:   &metav1.Time{Time: started},
+		TimeCompleted: &metav1.Time{Time: finish},
 	}
 	return updateStatus, "", nil
 }
--- a/cmd/backup-manager/app/backup/manager.go
+++ b/cmd/backup-manager/app/backup/manager.go
@@ -472,6 +472,10 @@ func (bm *Manager) performLogBackup(ctx context.Context, backup *v1alpha1.Backup
 		resultStatus, reason, err = bm.stopLogBackup(ctx, backup)
 	case string(v1alpha1.LogTruncateCommand):
 		resultStatus, reason, err = bm.truncateLogBackup(ctx, backup)
+	case string(v1alpha1.LogResumeCommand):
+		klog.Errorf("Log resume command is not supported yet")
+	case string(v1alpha1.LogPauseCommand):
+		klog.Errorf("Log pause command is not supported yet")
 	default:
 		return fmt.Errorf("log backup %s unknown log subcommand %s", bm, bm.SubCommand)
 	}
--- a/cmd/backup-manager/app/backup/manager.go
+++ b/cmd/backup-manager/app/backup/manager.go
@@ -472,11 +472,10 @@ func (bm *Manager) performLogBackup(ctx context.Context, backup *v1alpha1.Backup
 		resultStatus, reason, err = bm.stopLogBackup(ctx, backup)
 	case string(v1alpha1.LogTruncateCommand):
 		resultStatus, reason, err = bm.truncateLogBackup(ctx, backup)
-	//TODO: (Ris) support more funcs
 	case string(v1alpha1.LogResumeCommand):
 		resultStatus, reason, err = bm.resumeLogBackup(ctx, backup)
 	case string(v1alpha1.LogPauseCommand):
-		klog.Errorf("Log pause command is not supported yet")
+		resultStatus, reason, err = bm.pauseLogBackup(ctx, backup)
 	default:
 		return fmt.Errorf("log backup %s unknown log subcommand %s", bm, bm.SubCommand)
 	}
@@ -559,23 +558,13 @@ func (bm *Manager) startLogBackup(ctx context.Context, backup *v1alpha1.Backup) 
 // resumeLogBackup resume log backup.
 func (bm *Manager) resumeLogBackup(ctx context.Context, backup *v1alpha1.Backup) (*controller.BackupUpdateStatus, string, error) {
 	started := time.Now()
-	backupFullPath, err := util.GetStoragePath(backup)
-	if err != nil {
-		klog.Errorf("Get backup full path of cluster %s failed, err: %s", bm, err)
-		return nil, "GetBackupRemotePathFailed", err
-	}
-	klog.Infof("Get backup full path %s of cluster %s success", backupFullPath, bm)
-
-	updatePathStatus := &controller.BackupUpdateStatus{
-		BackupPath: &backupFullPath,
-	}
 
 	// change Prepare to Running before real backup process start
 	if err := bm.StatusUpdater.Update(backup, &v1alpha1.BackupCondition{
 		Command: v1alpha1.LogResumeCommand,
 		Type:    v1alpha1.BackupRunning,
 		Status:  corev1.ConditionTrue,
-	}, updatePathStatus); err != nil {
+	}, nil); err != nil {
 		return nil, "UpdateStatusFailed", err
 	}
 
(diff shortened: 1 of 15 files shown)


pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -321,13 +337,21 @@ func ParseLogBackupSubcommand(backup *Backup) LogSubCommandType {
...
+	switch backup.Spec.LogSubcommand {
+	case "log-start":
+		if IsLogBackupAlreadyPaused(backup) {
+			return LogResumeCommand
+		}
+		return LogStartCommand
+	case "log-stop":
 		return LogStopCommand
-	}
-	if backup.Spec.LogTruncateUntil != "" {
```
csuzhangxc, line 327: Is it needed to compatible with the old fields?

RidRisR: No, user shouldn't mix up these two interface to avoid mistakes

csuzhangxc: So the old existing Backup user MUST update their pipeline to use these new fields, right?

RidRisR: Yes, we don't want handle corner case like:
````
Spec:
    logSubcommand: log-start
    logStop: true
```
BTW, I am fresh to operator. Would there be any problem?
I tested updating operator online while a pitr task is running.
Seems there wouldn't be any problem.

WangLe1321: We need to keep compatible with the old fields, because tidbcloud depends on current implementation. I think we can make old fields have higher priority. For example:
```
if backup.Spec.LogStop
    return LogStopCommand
if backup.Spec.LogTruncateUntil != "" && backup.Spec.LogTruncateUntil != backup.Status.LogSuccessTruncateUntil
    return LogTruncateCommand

switch backup.Spec.LogSubcommand
...

// default command is log start
return LogStartCommand
```

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -201,6 +201,22 @@ func IsBackupScheduled(backup *Backup) bool {
...
 
+// HaveTruncateUntil returns true if a Backup has truncate until set
+func HaveTruncateUntil(backup *Backup) bool {
+	if backup.Spec.Mode != BackupModeLog {
+		return false
+	}
+	return backup.Spec.LogTruncateUntil != ""
+}
+
+func IsBackupPaused(backup *Backup) bool {
```
WangLe1321, line 212:
```suggestion
func IsLogBackupPaused(backup *Backup) bool {
```

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -201,6 +201,22 @@ func IsBackupScheduled(backup *Backup) bool {
...
+	if backup.Spec.Mode == BackupModeLog {
+		return IsLogBackupSubCommandOntheCondition(backup, BackupPaused)
```
WangLe1321, line 214: Seems log backup sub command condition has no `BackupPaused` status

pkg/controller/backup/backup_controller.go
```diff
@@ -230,6 +230,10 @@ func (c *Controller) updateBackup(cur interface{}) {
 		return
 	}
 
+	if newBackup.Spec.Mode == v1alpha1.BackupModeLog && newBackup.Spec.LogSubcommand == string(v1alpha1.LogStartCommand) && v1alpha1.IsBackupPaused(newBackup) {
+		newBackup.Spec.LogSubcommand = string(v1alpha1.LogResumeCommand)
```
WangLe1321, line 234: I don't know why add this hack logic. We shouldn't modify the `spec` fields even the user inputs wrong value.

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -201,6 +201,22 @@ func IsBackupScheduled(backup *Backup) bool {
...
+	}
+	_, condition := GetBackupCondition(&backup.Status, BackupPaused)
+	return condition != nil && condition.Status == corev1.ConditionTrue
```
WangLe1321, line 217: This only means the log backup is ever paused, not means it is paused. I guess you want to check if the log backup is paused. Only `backup.Status.Phase` represents the current status.

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -392,3 +416,13 @@ func IsLogBackupAlreadyTruncate(backup *Backup) bool {
 func IsLogBackupAlreadyStop(backup *Backup) bool {
 	return backup.Spec.Mode == BackupModeLog && backup.Status.Phase == BackupStopped
 }
+
+// IsLogBackupAlreadyStop return whether log backup has already paused.
```
WangLe1321, line 420:
```suggestion
// IsLogBackupAlreadyPaused return whether log backup has already paused.
```

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -392,3 +416,13 @@ func IsLogBackupAlreadyTruncate(backup *Backup) bool {
...
+func IsLogBackupAlreadyPaused(backup *Backup) bool {
+	return backup.Spec.Mode == BackupModeLog && backup.Status.Phase == BackupPaused
+}
+
+// IsLogBackupAlreadyRestart return whether log backup has already resumed.
```
WangLe1321, line 425:
```suggestion
// IsLogBackupAlreadyRunning return whether log backup has already resumed.
```

BornChanger:
For the state machine figure, please also add the scenarios where the state change due to gc-ttl

pkg/apis/pingcap/v1alpha1/backup.go
RidRisR: But it do have, after adding the pause stage and command in this PR

pkg/controller/backup/backup_controller.go
RidRisR: deleted

pkg/apis/pingcap/v1alpha1/backup.go
RidRisR: deleted

pkg/apis/pingcap/v1alpha1/backup.go
RidRisR: deleted

pkg/apis/pingcap/v1alpha1/backup.go
RidRisR: Seems `backup.Spec.LogTruncateUntil != backup.Status.LogSuccessTruncateUntil` is unnecessary
The user should be allowed to truncate multiple times.

RidRisR:
> For the state machine figure, please also add the scenarios where the state change due to gc-ttl. @RidRisR

we can't modify advancer time out setting now. therefore, the gc-ttl will only time out after 24 hours, it seems too long for a test.

cmd/backup-manager/app/backup/manager.go
```diff
@@ -582,6 +616,37 @@ func (bm *Manager) stopLogBackup(ctx context.Context, backup *v1alpha1.Backup) (
...
 }
 
+// pauseLogBackup stops log backup.
+func (bm *Manager) pauseLogBackup(ctx context.Context, backup *v1alpha1.Backup) (*controller.BackupUpdateStatus, string, error) {
+	started := time.Now()
+
+	// change Prepare to Running before real backup process start
+	if err := bm.StatusUpdater.Update(backup, &v1alpha1.BackupCondition{
+		Command: v1alpha1.LogPauseCommand,
+		Type:    v1alpha1.BackupRunning,
```
csuzhangxc, line 626: Should we need to set `Paused` here?

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -2177,6 +2179,9 @@ type BackupSpec struct {
 	// Default is current timestamp.
 	// +optional
 	CommitTs string `json:"commitTs,omitempty"`
+	// Subcommand is the subcommand for BR, such as start, stop, pause etc.
+	// +optional
+	LogSubcommand string `json:"logSubcommand,omitempty"`
```
csuzhangxc, line 2184: How about mark it with `+kubebuilder:validation:Enum`?

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -2397,6 +2404,12 @@ const (
 	LogTruncateCommand LogSubCommandType = "log-truncate"
 	// LogStopCommand is the stop command of log backup.
 	LogStopCommand LogSubCommandType = "log-stop"
+	// LogPauseCommand is the pause command of log backup.
+	LogPauseCommand LogSubCommandType = "log-pause"
+	// LogResumeCommand is the resume command of log backup.
+	LogResumeCommand LogSubCommandType = "log-resume"
+	// LogUnknownCommand is the unknown command of log backup.
+	LogUnknownCommand LogSubCommandType = "log-unknown"
```
csuzhangxc, line 2412: is `log-unknown` set by users? or just an invalid value.

cmd/backup-manager/app/backup/manager.go
RidRisR: The running state here means the `command` is running rather than the `log backup` is running.

pkg/apis/pingcap/v1alpha1/types.go
RidRisR: The latter one, just in case that the user used some unexpected input, maybe `+kubebuilder:validation:Enum` is a better choice

pkg/apis/pingcap/v1alpha1/types.go
RidRisR: Great idea.

pkg/apis/pingcap/v1alpha1/backup.go
```diff
@@ -316,18 +313,60 @@ func NeedNotClean(backup *Backup) bool {
...
+			subCommand = LogResumeCommand
+		} else {
+			subCommand = LogStartCommand
+		}
+	case LogStopCommand:
+		subCommand = LogStopCommand
+	case LogPauseCommand:
+		subCommand = LogPauseCommand
+	default:
+		return LogUnknownCommand
```
csuzhangxc, line 344: if set `LogSubcommand` as `""`, what the subcommand will be used, `log-unkown` or `log-start`.

RidRisR: Let me make it more clear:

1. 
Actually we don't need the "" as a subcommand. The only reason we keep it is for compatible with the old version interface.
The old usage is demonstrated here: https://github.com/pingcap/tidb-operator/pull/4682
TL;DR
In the old version, we use this CRD to start a new task:

```
Spec:
# empty field
```

Therefore, `LogSubcommand == ""` means `LogStartCommand`

2.
I add `LogUnknownCommand` for handling unexpected input from users, but now we use `+kubebuilder:validation:Enum` instead. This field is useless now. I keep it just for a double insurance.

3.
Now I refactored the code like

```
if backup.Spec.LogSubcommand == "" {
        if backup.Spec.LogStop || IsLogBackupAlreadyStop(backup) {
            subCommand = LogStopCommand
        } else if IsLogBackupAlreadyPaused(backup){
            subCommand = LogResumeCommand
        } else {
			subCommand = LogStartCommand
		}
    } else {
        switch backup.Spec.LogSubcommand {
        case LogStartCommand:
            if IsLogBackupAlreadyPaused(backup) {
                subCommand = LogResumeCommand
            } else {
                subCommand = LogStartCommand
            }
        case LogStopCommand:
            subCommand = LogStopCommand
        case LogPauseCommand:
            subCommand = LogPauseCommand
        default:
            return LogUnknownCommand
        }
    }
```
I suppose it could be more clear now.

WangLe1321 (approved):

csuzhangxc (approved):

PR #5720, opened 2024-08-26
BR: support Azure blob storage sas token

tennix:

Previously, tidb-operator only supports access key and AAD authentication for Azure blob storage. The [SAS (Shared Access Signature)](https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview) token is also commonly used, especially for cross Azure tenant scenario. The BR kernel already supports SAS token authentication. This PR adds SAS token support in tidb-operator.

On Azure, the storage account is a namespace for Azure storage. In this respect, it's similar to S3 and GCS bucket. Though Azure blob storage has another hierarchy `container`. So the storage account should also be in the `Backup` CR spec. The BR kernel should also re-organize the storage account as normal argument.

I add two new fields (`StorageAccount`, `SasToken`) in `AzblobStorageProvider` and concatenate the br command arguments with these two new fields. Now the authentication for Azure is as follows:
* If `Backup.Spec.StorageProvider.Azblob.SasToken` is specified, then use SAS token authentication
* If `Backup.Spec.StorageProvider.Azblob.StorageAccount` is specified, then use this to set backup-manager pod env. Otherwise get storage account from secret as before.
* If sas token is unspecified, then try to use access key or AAD authentication.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

I've tested this PR manually in local kind environment with Azure blob storage SAS token. The backup job finished successfully. The storage account and sas token are all set in `Backup` CR spec without specifying `spec.azblob.secretName`
image

I've also tested backup via access key with following two cases:
* Specifying a secret with only `AZURE_STORAGE_KEY` and specifying the storage account in the `spec.azblob.storageAccount`
* Specifying both `AZURE_STORAGE_KEY` and `AZURE_STORAGE_ACCOUNT` in secret, but not specifying `spec.azblob.storageAccount`

Both cases succeeded as expected.

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Support Backup & Restore using Azure blob storage SAS token authentication
```

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -2013,6 +2013,8 @@ type AzblobStorageProvider struct {
 	// azblob service account credentials.
 	SecretName string `json:"secretName,omitempty"`
 	// StorageAccount is the storage account of the azure blob storage
+	// If this field is set, then use this to set backup-manager env
+	// Otherwise retrieve the storage account from secret
 	StorageAccount string `json:"storageAccount,omitempty"`
 	// SasToken is the sas token of the storage account
 	SasToken string `json:"sasToken,omitempty"`
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -2012,6 +2012,10 @@ type AzblobStorageProvider struct {
 	// SecretName is the name of secret which stores the
 	// azblob service account credentials.
 	SecretName string `json:"secretName,omitempty"`
+	// StorageAccount is the storage account of the azure blob storage
+	StorageAccount string `json:"storageAccount,omitempty"`
+	// SasToken is the sas token of the storage account
+	SasToken string `json:"sasToken,omitempty"`
 	// Prefix of the data path.
 	Prefix string `json:"prefix,omitempty"`
 }
--- a/pkg/backup/util/remote.go
+++ b/pkg/backup/util/remote.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/url"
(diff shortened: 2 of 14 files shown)


pkg/backup/util/remote.go
```diff
@@ -634,6 +637,16 @@ func newGcsStorageOptionForFlag(conf *gcsConfig, flag string) []string {
...
 	path := fmt.Sprintf("azure://%s/", path.Join(conf.container, conf.prefix))
+	values := url.Values{}
+	if conf.storageAccount != "" {
+		values.Add("account-name", conf.storageAccount)
+	}
+	if conf.sasToken != "" {
+		values.Add("sas-token", conf.sasToken)
+	}
+	if v := values.Encode(); v != "" {
+		path = path + "?" + v
```
tennix, line 648: Ref: https://docs.pingcap.com/tidb/stable/backup-and-restore-storages#uri-examples

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -2012,6 +2012,10 @@ type AzblobStorageProvider struct {
 	// SecretName is the name of secret which stores the
 	// azblob service account credentials.
 	SecretName string `json:"secretName,omitempty"`
+	// StorageAccount is the storage account of the azure blob storage
+	StorageAccount string `json:"storageAccount,omitempty"`
```
csuzhangxc, line 2016: How about adding some comments for `If Backup.Spec.StorageProvider.Azblob.StorageAccount is specified, then use this to set backup-manager pod env. Otherwise get storage account from secret as before`

csuzhangxc (approved):

PR #5101, opened 2023-06-28
opt: do not evict leader when changing PVC size

csuzhangxc:
### What problem does this PR solve?

Do not evict leader when changing (increasing) PVC size, as if the storage size is not enough, the leader eviction will be blocked (never finished)

**NOTE**: if changing size and rolling restart TiKV at the same time, leader eviction is still triggered.

### Code changes

- [ ] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
  - only change storage size, no evict leader happen
  - only change IOPS/throughput, evict leader takes effect
  - change both IOPS/throughput&storage size, no evict leader happen
- [ ] No code 

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [ ] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note
Skip evicting leader for TiKV when changing PVC size
```

PR #5826, opened 2024-10-31
TidbCluster: disable evict leader for TiKV EBS modification

csuzhangxc:

evict leader for TiKV may take a long time, and the performance drop without leader-eviction is very low.

So we decide to disable this leader-eviction behavior for TiKV EBS modification.

**NOTE**: do not upgrade TiDB Operator when some leader-eviction (with EBS modification, but do not include EBS size modification, ref #5101) is happening as after this PR no "remove evict-leader-scheduler" will be performed then some evict-leader-scheduler may be kept.

### Code changes

- [ ] Has Go code change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
disable leader-eviction for TiKV EBS modification
```

### DIFF
--- a/pkg/manager/volumes/vol_compare_utils.go
+++ b/pkg/manager/volumes/vol_compare_utils.go
@@ -70,7 +70,7 @@ type componentVolumeContext struct {
 	tc     *v1alpha1.TidbCluster
 	status v1alpha1.ComponentStatus
 
-	// always be true now
+	// always be false now
 	// as we think there is no need to evict leader for AWS EBS modification
 	shouldEvict bool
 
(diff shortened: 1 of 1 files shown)


pkg/manager/volumes/vol_compare_utils.go
```diff
@@ -70,6 +70,8 @@ type componentVolumeContext struct {
 	tc     *v1alpha1.TidbCluster
 	status v1alpha1.ComponentStatus
 
+	// always be true now
```
fgksgf, line 73:
```suggestion
	// always be false now
```

fgksgf (approved):

PR #5992, opened 2024-12-20
refactor(utils): add hasher pkg and add more uts for toml

liubog2008:
- move `GenerateHash` to hasher pkg
- add more UTs for toml pkg

### DIFF
--- a/pkg/utils/hasher/hasher.go
+++ b/pkg/utils/hasher/hasher.go
@@ -1,3 +1,17 @@
+// Copyright 2024 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


fgksgf (approved):

pkg/utils/hasher/hasher.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/pingcap/tidb-operator/apis/core/v1alpha1"
+	hashutil "github.com/pingcap/tidb-operator/third_party/kubernetes/pkg/util/hash"
+)
+
+// GenerateHash takes a TOML string as input, unmarshals it into a map,
+// and generates a hash of the resulting configuration. The hash is then
+// encoded into a safe string format and returned.
+// If the order of keys in the TOML string is different, the hash will be the same.
+func GenerateHash(tomlStr v1alpha1.ConfigFile) (string, error) {
```
csuzhangxc, line 33: As this function only works for TOML, should we include `toml` in the name?

liubog2008: LGTM, but may add it when a new similar function is added?

csuzhangxc (approved):

PR #6011, opened 2024-12-26
feat(pdgroup): migrate to task v3

liubog2008:
- use task v3 to org tasks of pd group
- refactor status task
- support currentReplicas and updateReplicas

### DIFF
--- a/apis/core/v1alpha1/common_types.go
+++ b/apis/core/v1alpha1/common_types.go
@@ -288,15 +288,15 @@ type GroupStatus struct {
 	Replicas int32 `json:"replicas"`
 
 	// ReadyReplicas is the number of Instances created for this ComponentGroup with a Ready Condition.
-	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
+	ReadyReplicas int32 `json:"readyReplicas"`
 
 	// CurrentReplicas is the number of Instances created by the Group controller from the Group version
 	// indicated by currentRevision.
-	CurrentReplicas int32 `json:"currentReplicas,omitempty"`
+	CurrentReplicas int32 `json:"currentReplicas"`
 
-	// UpdatedReplicas is the number of Instances created by the Group controller from the Group version
+	// UpdateReplicas is the number of Instances created by the Group controller from the Group version
 	// indicated by updateRevision.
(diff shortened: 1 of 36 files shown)


apis/core/v1alpha1/common_types.go
```diff
@@ -288,15 +288,15 @@ type GroupStatus struct {
...
 	// CurrentReplicas is the number of Instances created by the Group controller from the Group version
 	// indicated by currentRevision.
-	CurrentReplicas int32 `json:"currentReplicas,omitempty"`
+	CurrentReplicas int32 `json:"currentReplicas"`
 
-	// UpdatedReplicas is the number of Instances created by the Group controller from the Group version
+	// UpdateReplicas is the number of Instances created by the Group controller from the Group version
 	// indicated by updateRevision.
-	UpdatedReplicas int32 `json:"updatedReplicas,omitempty"`
+	UpdateReplicas int32 `json:"updateReplicas"`
```
fgksgf, line 299: Why change this? StatefulSet also use `updatedReplicas` instead of `updateReplicas`
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/stateful-set-v1/#StatefulSetStatus

fgksgf (changes requested):

pkg/controllers/pdgroup/tasks/status.go
```diff
@@ -15,82 +15,129 @@
...
+
+		replicas, readyReplicas, updateReplicas, currentReplicas := calcReplicas(state.PDSlice(), state.CurrentRevision, state.UpdateRevision)
+
+		// all instances are updated
+		if updateReplicas == replicas {
+			// update current revision
+			state.CurrentRevision = state.UpdateRevision
+			// update status of pdg version
+			// TODO(liubo02): version of a group is hard to understand
+			// We need to change it to a more meaningful field
```
fgksgf, line 87: We need a field to indicate if a group is upgrading.

liubog2008: prefer condition

apis/core/v1alpha1/common_types.go
liubog2008: OK, I'll change back

csuzhangxc (approved):

PR #6020, opened 2025-01-02
fix(manager): declare status port in tikv pods

omar-mohamed-khallaf:
When scraping pods with victoriametrics, it automatically drops targets (pods in this case) that don't have the port to scrape declared in its `spec.ports`.

Declared the status port in the pod spec.

### Code changes

- [x] Has Go code change

### Tests
- [ ] Unit test
- [ ] E2E test
- [ ] Manual test
- [ ] No code

### Side effects

- [ ] Other side effects

### Related changes

### Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/manager/member/tikv_member_manager.go
+++ b/pkg/manager/member/tikv_member_manager.go
@@ -641,6 +641,11 @@ func getNewTiKVSetForTidbCluster(tc *v1alpha1.TidbCluster, cm *corev1.ConfigMap)
 				ContainerPort: v1alpha1.DefaultTiKVServerPort,
 				Protocol:      corev1.ProtocolTCP,
 			},
+			{
+				Name:          "status",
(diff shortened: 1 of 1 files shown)


omar-mohamed-khallaf:
They will be recreated one by one, right?
What should I do?

csuzhangxc:
> They will be recreated one by one, right? What should I do?

YES, one by one. But this is still not acceptable for some users.

Is there any config item in victoriametrics can achieve the goal, otherwise, I think a new config item needs to be added in TiDB Operaotr

omar-mohamed-khallaf:
For victoriametrics there's none that I'm aware of, I think it's intentional to drop those pods.

csuzhangxc:
Can you set `enableNamedStatusPort` for TiKV to add this port to Pod spec?

omar-mohamed-khallaf:
Thank you, this is exactly what I needed.
I will close the PR then.

PR #6048, opened 2025-01-17
fix(tidb): move bootstrap sql to cluster spec

liubog2008:
- move bootstrap sql ref to cluster spec because it is only worked when tidb is bootstrapped.

### DIFF
--- a/apis/core/v1alpha1/cluster_types.go
+++ b/apis/core/v1alpha1/cluster_types.go
@@ -17,6 +17,7 @@ package v1alpha1
 import (
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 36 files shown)


apis/core/v1alpha1/cluster_types.go
```diff
@@ -75,6 +76,11 @@ type ClusterSpec struct {
 	// Whether enable the TLS connection between TiDB cluster components.
 	TLSCluster *TLSCluster `json:"tlsCluster,omitempty"`
 
+	// BootstrapSQL refer to a configmap which contains the bootstrap SQL file with the key `bootstrap-sql`,
```
fgksgf, line 79:
```suggestion
	// BootstrapSQL refers to a configmap which contains the bootstrap SQL file with the key `bootstrap-sql`,
```

csuzhangxc (approved):

PR #6069, opened 2025-02-10
Volume Replacement causes high latency/jitter to the clients due to excessive IO and CPU usage

aravndr:
Volume Replacement causes high latency/jitter to the clients due to excessive IO and CPU usage  because schedule.store-limit.xyz.remove-peer limit is being set to unlimited. Effectively the limit becomes num_available_tikv_pods *  schedule.store-limit.xyz.add-peer. The problem is the TIKV store is still serving active leaders causing a high latency impact to clients.

This fix evicts the leaders from the TIKV pod/store before proceeding with volume replacement on the pod.

Ref: https://github.com/tikv/pd/issues/4099

Evict region leaders from the TIKV store before proceeding with volume replacement

### Code changes

- [X ] Has Go code change

### Tests

- [X] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Fix high latency/jitter due to excessive IO and CPU usage during Volume Replacement.
```

### DIFF
--- a/pkg/manager/volumes/pvc_replacer.go
+++ b/pkg/manager/volumes/pvc_replacer.go
@@ -163,11 +163,25 @@ func isVolumeReplacing(pod *corev1.Pod) bool {
 	return exist
 }
 
-func (p *pvcReplacer) startVolumeReplace(pod *corev1.Pod) error {
+func (p *pvcReplacer) startVolumeReplace(ctx *componentVolumeContext, pod *corev1.Pod) error {
 	pod = pod.DeepCopy()
 
 	if pod.Annotations == nil {
 		pod.Annotations = map[string]string{}
+	}
+
+	// try to evict leader
+	isEvicted := isLeaderEvictedOrTimeout(ctx.tc, pod)
+	if !isEvicted {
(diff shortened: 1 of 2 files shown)


anish-db:
Discussed offline w/ Aravind, but leaving some code pointers here to make this fix more generic. This fix should apply to all cases where operator does a DeleteStore, which should probably go more generically for disk replace cases here: https://github.com/pingcap/tidb-operator/blob/f8ebb9c7636149910de3e19b12bc34470b31ae90/pkg/controller/tidbcluster/pod_control.go#L572

And also for tikv scale down (which can be impacted by the same bug and is actually the scenario mentioned in https://github.com/tikv/pd/issues/4099): https://github.com/pingcap/tidb-operator/blob/f8ebb9c7636149910de3e19b12bc34470b31ae90/pkg/manager/member/tikv_scaler.go#L210

csuzhangxc (approved):

anish-db:
@csuzhangxc we likely need to modify this PR to move the fix more generally into pod_control.go based on my above comment

PR #6076, opened 2025-02-13
feat(cluster): add feature gates

liubog2008:
- add feature gates support


pkg/features/features.go
```diff
@@ -0,0 +1,119 @@
...
+
+	Register(c *v1alpha1.Cluster)
+	Deregister(ns, name string)
+}
+
+type gates struct {
+	lock sync.Mutex
+
+	features map[string]featureSet
+}
```
fgksgf, line 52: In scenarios with frequent reads and infrequent writes, sync.Map significantly reduces contention through lock-free read operations and partitioned write optimizations, demonstrating superior performance compared to a conventional map combined with sync.Mutex.

pkg/features/features.go
```diff
@@ -0,0 +1,119 @@
...
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	key := types.NamespacedName{
+		Namespace: c.Namespace,
+		Name:      c.Name,
+	}.String()
+
+	fs, ok := g.features[key]
+	if !ok || fs.uid != c.UID || fs.generation != c.Generation {
```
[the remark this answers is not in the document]
liubog2008: if the generation is changed, enabled feature gates may be changed.

pkg/features/features.go
liubog2008: I prefer `sync.RWMutex`. We need two `sync.Map` in this case and I feel it's complex and premature optimized. Maybe we can optimize it after we find any performance issues.

PR #6083, opened 2025-02-18
feat(status): add synced and ready condition

liubog2008:
- add Ready and Synced condition
- refactor Suspended condition task
- split status task to multiple tasks and only update status once.
- rename Health condition to Ready


pkg/updater/actor.go
```diff
@@ -97,7 +97,7 @@ func (act *actor[T, O, R]) ScaleInUpdate(ctx context.Context) (bool, error) {
 	}
 	obj := act.update.Del(name)
 
-	isUnavailable := !obj.IsHealthy() || !obj.IsUpToDate()
+	isUnavailable := !obj.IsReady() || !obj.IsUpToDate()
```
fgksgf, line 100: Still use `IsUpToDate()`? Should not be `Synced()`?

liubog2008: IsUpToDate means the instance is updated. Synced means subresources of the instance are updated

csuzhangxc (approved):

PR #6091, opened 2025-02-21
fix(prestop): fix prestop hook of tikv

liubog2008:
- fix prestop hook of tikv
- add e2e case of prestop hook


pkg/controllers/tikv/tasks/pod.go
```diff
@@ -291,14 +290,15 @@ func buildPrestopCheckScript(cluster *v1alpha1.Cluster, tikv *v1alpha1.TiKV) str
 	sb.WriteString("/prestop-checker")
 	sb.WriteString(" -pd ")
 	sb.WriteString(cluster.Status.PD)
-	sb.WriteString(" -addr ")
-	sb.WriteString(kvcfg.GetAdvertiseClientURLs(tikv))
+	sb.WriteString(" -tikv-status-addr ")
```
csuzhangxc, line 293: do we need to set `mode` here?

liubog2008: I'll set it in the next PR

csuzhangxc (approved):

PR #6095, opened 2025-02-25
feat(ticdc): support prestop hook of ticdc

liubog2008:
- support prestop hook of ticdc


pkg/controllers/ticdc/tasks/pod.go
```diff
@@ -37,6 +39,8 @@ import (
 
 const (
 	metricsPath = "/metrics"
+
+	defaultGracefulShutdownSeconds = 30
```
csuzhangxc, line 43: Do we need to use the default timeout (10m) as v1?

liubog2008: When I try to delete the whole cluster, deletion of ticdc will be blocked and the prestop hook cannot pass.

liubog2008: I'll fix it in another PR

csuzhangxc (approved):

PR #6153, opened 2025-04-14
fix(instance): fix and refactor ready condition

liubog2008:
- fix ready condition when suspending
- refactor task to sync ready condition
- add more details about why unready
- add common `Health` state
- add common `IsPodTerminating` state


pkg/controllers/common/interfaces.go
```diff
@@ -246,13 +246,38 @@ type (
...
+
+type HealthState interface {
+	// It means the instance is health to serve.
+	// Normally, it's from PD or api exposed by the instance.
+	// Now the operator checks it just like a liveness/readiness prober.
+	// It may be removed if all components support /ready api in the future.
+	// And then we can use pod's ready condition directly to check whether the instance
+	// is ready.
+	// But now, we still probe health from the operator.
+	IsHealth() bool
```
fgksgf, line 278:
```suggestion
	// IsHealthy means the instance is healthy to serve.
	// Normally, it's from PD or api exposed by the instance.
	// Now the operator checks it just like a liveness/readiness prober.
	// It may be removed if all components support /ready api in the future.
	// And then we can use pod's ready condition directly to check whether the instance
	// is ready.
	// But now, we still probe health from the operator.
	IsHealthy() bool
```

pkg/controllers/common/interfaces.go
```diff
@@ -246,13 +246,38 @@ type (
...
+}
+
+type HealthStateUpdater interface {
+	SetHealth()
```
fgksgf, line 282:
```suggestion
	SetHealthy()
```

pkg/controllers/pd/tasks/state.go
```diff
@@ -33,27 +31,35 @@ type state struct {
 	pod     *corev1.Pod
 	pds     []*v1alpha1.PD
 
+	// Pod cannot be updated when call DELETE API, so we have to set this field to indicate
+	// the underlay pod has been deleting
+	isPodTerminating bool
+
 	statusChanged bool
+
+	health bool
```
fgksgf, line 40:
```suggestion
	healthy bool
```

fgksgf (approved):

Issue #6136, opened 2025-04-01
Unexpected topology spreading of tidb


liubog2008:
# reproduce steps
- create tidb with 2 replicas and evenly spread in 3 topologies
- rolling update by change config
- scale out to 3 and rolling update at the same time
- tidb will spread in (1 1 2) or (1 1 3)

liubog2008:
After https://github.com/pingcap/tidb-operator/pull/6161, this issue is partially fixed

N means new and O means old instance.
 
- initial spread
  - O2, O3
- scale out to 3 and rolling update
  - N1, O2, O3, N1 (before https://github.com/pingcap/tidb-operator/pull/6161)
  - N1, O2, O3, N2 (after the fix https://github.com/pingcap/tidb-operator/pull/6161)
- O2 is unavailable, so O2 will be updated
  - N1, N2, O3, N2
- scale in O3

liubog2008:
After https://github.com/pingcap/tidb-operator/pull/6203. This issue is fixed

- initial spread
  - O1, O2, O3
-  rolling update with max surge 1
  - O1, O2, O3, N1
- next instance
  - N1, O2, O3, N1 (before https://github.com/pingcap/tidb-operator/pull/6203, update O1 to N1)
  - O2, O3, N1, N2 (after https://github.com/pingcap/tidb-operator/pull/6203, scale in O1 and scale out N2)

PR #6161, opened 2025-04-18
fix(tidb): fix topology spread issue of tidb

liubog2008:
- prefer evenly spread updated instances if there are more than one choices of topology. 

See https://github.com/pingcap/tidb-operator/issues/6136

It is not fully fixed because we always prefer scale in unavailable instances which may be not the best choice for topology spread.

### DIFF
--- a/pkg/utils/topology/scheduler.go
+++ b/pkg/utils/topology/scheduler.go
@@ -16,6 +16,7 @@ package topology
 
 import (
 	"fmt"
+	"slices"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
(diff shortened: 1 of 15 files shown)


pkg/utils/topology/scheduler.go
```diff
@@ -90,6 +91,9 @@ type topologyScheduler struct {
 }
 
 func (s *topologyScheduler) Add(name string, t v1alpha1.Topology) {
+	if _, ok := s.nameToIndex[name]; ok {
+		s.Del(name)
+	}
```
fgksgf, line 96: Add some comments?

Issue #232, opened 2018-12-17
pd and tikv happended concurrent upgrading at some times


xiaojingchen:
when upgrading tidb cluster, found that the pd and tikv 's upgrade phase is setted to `upgrade` at same time. this cause the upgrading is not serial.

PR #6180, opened 2025-04-28
Save backup manifests for snapshot backup (#232)

### LINKED ISSUE #6179: Save cluster manifests as part of snapshot backup
## Feature Request

It is useful to save the cluster manifests as part of the backup so that the restore process can inspect the backup cluster's state directly from the backup rather than querying the backup cluster's metadata. This is especially important in a disaster scenario, where the backup cluster may not be available.

This capability already exists for EBS volume snapshot backups, as the cluster manifests are strictly necessary for the restore process. However, these manifests would be useful for logical snapshot backups as well.

Other alternatives would require querying the manifests directly from the backup cluster, which is prohibitive in a disaster recovery scenario

michaelmdeng:

Closes #6179 

Saves cluster manifests to external storage as part of the snapshot backup process, similar to EBS volume snapshot backups.

While EBS backups perform this step after the init job and prior to each data-plane's individual backup job (in fact each data plane will upload manifest to the same location, so only last-write-wins manifest will be backed up). For logical snapshot backup, there is no init job and there is only a single backup job that orchestrates the entire backup -- we choose to add the manifest backup step prior to the overall backup job. Similarly, this will only backup the cluster manifests from a single AZ, the control-plane AZ that performs the backup job, but this is sufficient for our use-cases.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [x] Other side effects: 

Adds additional external storage requirements for snapshot backup, additional step prior to the BR process that could fail.

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/backup/backup/backup_manager.go
+++ b/pkg/backup/backup/backup_manager.go
@@ -723,6 +723,11 @@ func (bm *backupManager) makeBRBackupJob(backup *v1alpha1.Backup) (*batchv1.Job,
 		}
 		args = append(args, fmt.Sprintf("--mode=%s", v1alpha1.BackupModeVolumeSnapshot))
 	default:
+		klog.Infof("Backup cluster manifests for snapshot backup %s/%s", ns, name)
+		err = bm.backupManifests(backup, tc)
+		if err != nil {
+			return nil, "BackupManifestsFailed", fmt.Errorf("Backup cluster manifests failed for snapshot backup %s/%s, ignoring error: %v", ns, name, err)
+		}
 		args = append(args, fmt.Sprintf("--mode=%s", v1alpha1.BackupModeSnapshot))
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/backup/backup/backup_manager.go
```diff
@@ -723,6 +723,11 @@ func (bm *backupManager) makeBRBackupJob(backup *v1alpha1.Backup) (*batchv1.Job,
 		}
 		args = append(args, fmt.Sprintf("--mode=%s", v1alpha1.BackupModeVolumeSnapshot))
 	default:
+		klog.Infof("Backup cluster manifests for snapshot backup %s/%s", ns, name)
```
BornChanger, line 726: please add a switch (by default false) to control if manifest backup is needed for snapshot backup since this change add more items to backup bucket.

Issue #6171, opened 2025-04-28
[Umbrella] PDMS support in v2


liubog2008:
- [ ] design doc
- [ ] TSO
  - [ ] api
  - [ ] controller
  - [ ] e2e
- [ ] Scheduler
  - [ ] api
  - [ ] controller
  - [ ] e2e

PR #6184, opened 2025-05-06
feat(tso): support tso component

liubog2008:
https://github.com/pingcap/tidb-operator/issues/6171

- add tso/tsogroup controller
- refactor some common tasks
- add basic e2e test

TODO:
- transfer primary when rolling update
- support change between normal/ms mode
- probe health by api


fgksgf:
`pkg/action/upgrader.go` should also be modified

fgksgf (approved):

PR #6193, opened 2025-05-09
fix(features): reimplement verify api and rename to use

liubog2008:
- reimplement `Verify` and rename to `Use`

Verify cannot ensure all running tasks of one reconcile use a same version of feature gates.

- A controller calls `Verify` and OK to run next tasks.
- `Register` is called and feature gates are changed.
- Some tasks may use a different version of feature gates and do some unexpected things.

After this change, all controllers will wait until all tasks use the outdated feature gates finished.


liubog2008:
Prefer https://github.com/pingcap/tidb-operator/pull/6199

PR #6200, opened 2025-05-16
fix(br): rename backup-manager to tidb-backup-manager

liubog2008:
- rename backup-manager to tidb-backup-manager to use same a name of image, cmd and build stage.


charts/tidb-operator/values.yaml
```diff
@@ -55,7 +55,7 @@ operator:
 
 # Backup manager image configuration
 backupManager:
-  image: pingcap/backup-manager:latest
+  image: pingcap/tidb-backup-manager:latest
```
wuhuizuo, line 58: Yes, before value is wrong.

fgksgf (approved):

PR #6226, opened 2025-05-30
feat(pkg): support modifying feature gates

liubog2008:
- support modifying feature gates


pkg/controllers/common/task.go
```diff
@@ -152,16 +125,51 @@ func TaskSuspendPod(state PodState, c client.Client) task.Task {
...
+
+	for i, f := range fs {
+		if c.Spec.FeatureGates[i].Name != f {
+			return false
+		}
+	}
+
+	return true
+}
+
```
fgksgf: Do not need to sort?

pkg/utils/k8s/revision/controller_revision.go
```diff
@@ -172,9 +172,13 @@ func getPatch(obj client.Object, gvk schema.GroupVersionKind) ([]byte, error) {
 	spec := raw["spec"].(map[string]any)
 	template := spec["template"].(map[string]any)
 	specCopy["template"] = template
+	if f, ok := spec["features"]; ok {
+		specCopy["features"] = f
+	}
 	template["$patch"] = "replace"
 	objCopy["spec"] = specCopy
 	patch, err := json.Marshal(objCopy)
+
```
fgksgf, line 181: Will this make existing clusters restart?

pkg/reloadable/common.go
```diff
@@ -72,3 +74,29 @@ func convertAnnotations(ls map[string]string) map[string]string {
...
+	}
+
+	for i := range a {
+		if string(a[i]) != string(b[i]) {
+			return false
+		}
+	}
+
+	return true
+}
```
fgksgf: Need to sort?

pkg/controllers/common/task.go
liubog2008: it's ordered by users.

pkg/utils/k8s/revision/controller_revision.go
liubog2008: revision will be changed.

PR #6267, opened 2025-06-24
set `gc.ratio-threshold` to `-1` when performing pitr restore

YuJuncen:
The `gc.ratio-threshold` may not be persisted and might cause data loss after the restoration when a TiKV restart during PiTR.

This PR will make operator set the `gc.ratio-threshold` to config map when encountering a pitr restore task.

Below is the timesequnce:

```mermaid
sequenceDiagram
    participant MemberManager as Member Manager
    participant ConfigMap
    participant RestoreCR as Restore CR
    participant Job
    participant TiKVCluster as TiKV Cluster

    RestoreCR-->>RestoreCR: Created by user
    par Wait for ConfigMap update
        RestoreCR->>ConfigMap: Waits for gc.ratio-threshold = -1
    and Member Manager notices CR
        MemberManager->>MemberManager: Notices Restore CR
        MemberManager->>ConfigMap: Updates gc.ratio-threshold to -1 (inplace)
    end
    
    ConfigMap-->>RestoreCR: gc.ratio-threshold = -1
    RestoreCR->>Job: Schedules Job
    Job->>TiKVCluster: Set gc.ratio-threshold to -1 by SQL
    Job->>RestoreCR: Job finishes successfully
    Job->>TiKVCluster: Set gc.ratio-threshold back by SQL
    
    MemberManager->>MemberManager: Notices Restore CR finished
    MemberManager->>ConfigMap: Recovers ConfigMap (inplace)
```

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test 
- [ ] E2E test 
- [x] Manual test 

Before Start:
```console

    config-file: |
      [import]
        num-threads = 8
    startup-script: "#!/bin/sh\n\n# This script is used to start tikv containers in
      kubernetes cluster\n\n# Use DownwardAPIVolumeFiles to store informations of
      the cluster:\n# https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api\n#\n#
      \  runmode=\"normal/debug\"\n#\n\nset -uo pipefail\n\nANNOTATIONS=\"/etc/podinfo/annotations\"\n\nif
      [[ ! -f \"${ANNOTATIONS}\" ]]\nthen\n    echo \"${ANNOTATIONS} does't exist,
      exiting.\"\n    exit 1\nfi\nsource ${ANNOTATIONS} 2>/dev/null\n\nrunmode=${runmode:-normal}\nif
      [[ X${runmode} == Xdebug ]]\nthen\n\techo \"entering debug mode.\"\n\ttail -f
      /dev/null\nfi\n\n# Use HOSTNAME if POD_NAME is unset for backward compatibility.\nPOD_NAME=${POD_NAME:-$HOSTNAME}\nARGS=\"--pd=http://${CLUSTER_NAME}-pd:2379
```

After Start:
```console
    config-file: |
      [gc]
        ratio-threshold = -1.0

      [import]
        num-threads = 8
    config-file-overlay: |
      [gc]
        ratio-threshold = -1.0
    startup-script: "#!/bin/sh\n\n# This script is used to start tikv containers in
      kubernetes cluster\n\n# Use DownwardAPIVolumeFiles to store informations of
```

- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note
Now, `gc.ratio-threshold` will be persisted during restoring.
```


pkg/backup/restore/restore_manager.go
```diff
@@ -66,6 +69,80 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+		tc.Spec.TiKV.Config = v1alpha1.NewTiKVConfig()
+	}
+	val := tc.Spec.TiKV.Config.Get(TiKVConfigGCThreshold)
+	if val != nil {
+		v := val.MustFloat()
+		originalCfg.GCRatioThreshold = &v
+	}
+	tc.Status.TiKV.PiTRStatus.OriginConfigMap = originalCfg
+	tc.Spec.TiKV.Config.Set(TiKVConfigGCThreshold, -1.0)
+	_, err := rm.deps.TiDBClusterControl.Update(tc)
```
BornChanger: is this api a synced one?

pkg/backup/restore/restore_manager.go
```diff
@@ -66,6 +69,80 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+	}
+	return true
+}
+
+func pitrTaskIsDone(restore *v1alpha1.Restore) bool {
+	if v1alpha1.IsRestoreComplete(restore) {
+		return true
+	}
+	if idx, cond := v1alpha1.GetRestoreCondition(&restore.Status, v1alpha1.RestoreFailed); idx != -1 {
+		if time.Since(cond.LastTransitionTime.Time) > 4*time.Hour {
```
BornChanger: what's the rational here?

pkg/backup/restore/restore_manager.go
```diff
@@ -66,6 +69,80 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
 	return rm.statusUpdater.Update(restore, condition, nil)
 }
 
+func (rm *restoreManager) pitrEnable(tc *v1alpha1.TidbCluster) error {
+	if tc.Status.TiKV.PiTRStatus.Active {
+		return nil
+	}
+	s := v1alpha1.ConfigUpdateStrategyInPlace
+	tc.Spec.TiKV.ConfigUpdateStrategy = &s
```
3pointer: should we reset this ConfigUpdateStrategy back after Update finished?

pkg/backup/restore/restore_manager.go
```diff
@@ -194,6 +275,15 @@ func (rm *restoreManager) syncRestoreJob(restore *v1alpha1.Restore) error {
 		return fmt.Errorf("restore %s/%s get job %s failed, err: %v", ns, name, restoreJobName, err)
 	}
 
+	if restore.Spec.Mode == v1alpha1.RestoreModePiTR {
+		// Note: perhaps better to reschedule here and wait the cluster config applied.
```
BornChanger, line 291: why not reconcile?

pkg/backup/restore/restore_manager.go
YuJuncen: A failed restore may not block GC forever.

pkg/backup/restore/restore_manager.go
YuJuncen: Because BR sets this value, so it is safe to start BR here without wating configuration applied. Also it is hard to making sure the configuration was applied in this context. As configuration reload won't actually trigger a reconcile for `Restore` CR.

pkg/backup/restore/restore_manager.go
YuJuncen: This might cause TiKV restart if we do so? Also notice that it is not easy to know whether TiKV configuration was changed in this context.

pkg/backup/restore/restore_manager.go
YuJuncen: I asked @WangLe1321 and he verifies that there will only be one reconiler running per CR. So this should be fine.

WangLe1321: It's better to wait tikv config applied to the config map, or the config may lost temporarily if tikv pod restarted between br set config and tikv-controller applied config.

pkg/backup/restore/restore_manager.go
```diff
@@ -66,6 +69,152 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+		klog.InfoS("restore-manager pitr disable, restore tikv config from current configmap",
+			"tidbcluster", klog.KObj(tc), "currentConfigUpdateStrategy", tc.Spec.TiKV.ConfigUpdateStrategy)
+		cfgMap, err := rm.configMapOfTiKV(tc)
+		if err != nil {
+			return fmt.Errorf("failed to get configmap of tikv for tidbcluster %s/%s, err: %v", tc.Namespace, tc.Name, err)
+		}
+		tikvConfigMap := tc.Spec.TiKV.Config
+
+		if cfgMap.Get(TiKVConfigGCThreshold) != tikvConfigMap.Get(TiKVConfigGCThreshold) {
+			return nil
```
WangLe1321: we should return error here. return `nil` means the operation is successful

pkg/backup/restore/restore_manager.go
```diff
@@ -66,6 +69,152 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+			"tidbcluster", klog.KObj(tc), "originConfigMap", tc.Status.TiKV.PiTRStatus.OriginConfigMap)
+		threshold := tc.Status.TiKV.PiTRStatus.OriginConfigMap.GCRatioThreshold
+		if threshold != nil {
+			tc.Spec.TiKV.Config.Set(TiKVConfigGCThreshold, threshold)
+		} else {
+			tc.Spec.TiKV.Config.Del(TiKVConfigGCThreshold)
+		}
+		tc.Status.TiKV.PiTRStatus.OriginConfigMap = nil
+		_, err := rm.deps.TiDBClusterControl.Update(tc)
+		return err
```
WangLe1321, line 96: we should return an error even if `err == nil` since the whole operation is not finished here.

pkg/backup/restore/restore_manager.go
```diff
@@ -66,6 +69,152 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+	for _, restore := range restores {
+		if !pitrTaskIsDone(restore) {
+			return false
+		}
+	}
+	return true
+}
+
+func pitrTaskIsDone(restore *v1alpha1.Restore) bool {
+	if v1alpha1.IsRestoreComplete(restore) {
```
WangLe1321: If the restore is in `Complete` or `Failed` phase, the restore won't be reconciled. So maybe we can't watch a restore is finished in reconcile.

YuJuncen: Will there be a periodical reconcile? Or it can be tricky to fix the cluster config...

pkg/backup/restore/restore_manager.go
YuJuncen: This might be tricky to be tested. Maybe we can have a try...

pkg/backup/restore/restore_manager.go
```diff
@@ -66,9 +73,213 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+	tikvConfigMap := tc.Spec.TiKV.Config
+	cfgMap, err := rm.configMapOfTiKV(tc)
+	if err != nil {
+		return err
+	}
+
+	// When parsing TOML, it treats -1 as int64...
+	floatValueEquals := func(v1, v2 *config.Value) bool {
+		if v1 == nil || v2 == nil {
+			return v1 == v2
```
3pointer: Is there any chance that both v1/v2 is nil?

pkg/controller/restore/restore_controller.go
```diff
@@ -251,13 +266,73 @@ func (c *Controller) updateRestore(cur interface{}) {
...
+func (k parsedRestoreKey) String() string {
+	return fmt.Sprintf("%s/%s:%s/%s",
+		k.targetClusterNamespace, k.targetClusterName, k.restoreNamespace, k.restoreName)
+}
+
+func (k parsedRestoreKey) IsBR() bool {
+	return k.targetClusterNamespace != "" && k.targetClusterName != ""
+}
+
+func parseRestoreKey(s string) (parsedRestoreKey, error) {
```
3pointer: do we need handle old format restore here? to keep the compatibility?

3pointer (commented):
rest LGTM

pkg/controller/restore/restore_controller.go
YuJuncen: The old format should be `keyspace/name`, which should already be handled here:

https://github.com/YuJuncen/tidb-operator/blob/ddf91b00eca30c55209b701e42ce3c1625a64b84/pkg/controller/restore/restore_controller.go#L299-L311

pkg/backup/restore/restore_manager.go
YuJuncen: For now there should be such scenario. But this check can be nice to have.

3pointer (approved):

pkg/backup/restore/restore_manager.go
```diff
@@ -66,9 +73,216 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+		return controller.RequeueErrorf("config reset, waiting for configmap updated")
+	}
+
+	if tc.Status.TiKV.PiTRStatus.State == v1alpha1.PiTRStateWaitingForConfig {
+		klog.InfoS("restore-manager pitr enable, ensure tikv config is reconciled", "tidbcluster", klog.KObj(tc))
+		if err := rm.ensureConfigMapReconileDone(tc); err != nil {
+			return err
+		}
+		tc.Status.TiKV.PiTRStatus.State = v1alpha1.PiTRStateRunning
+		_, err := rm.deps.TiDBClusterControl.Update(tc)
```
csuzhangxc: What's the expected behavior if the user updates the TidbCluster CR again?

pkg/backup/restore/restore_manager.go
```diff
@@ -66,9 +73,216 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+			v := val.MustFloat()
+			originalCfg.GCRatioThreshold = &v
+		}
+		tc.Status.TiKV.PiTRStatus.TiKVConfigUpdateStrategy = tc.Spec.TiKV.ConfigUpdateStrategy
+
+		s := v1alpha1.ConfigUpdateStrategyInPlace
+		tc.Spec.TiKV.ConfigUpdateStrategy = &s
+		tc.Status.TiKV.PiTRStatus.OriginConfigMap = originalCfg
+		tc.Spec.TiKV.Config.Set(TiKVConfigGCThreshold, -1.0)
+		_, err := rm.deps.TiDBClusterControl.Update(tc)
```
csuzhangxc: What's the expected behavior if the user updates the TidbCluster CR again?

WangLe1321 (approved):

pkg/backup/restore/restore_manager.go
```diff
@@ -66,9 +73,216 @@ func (rm *restoreManager) UpdateCondition(restore *v1alpha1.Restore, condition *
...
+	deps *controller.Dependencies
+}
+
+func NewPiTRManager(deps *controller.Dependencies) *PiTRManager {
+	return &PiTRManager{
+		deps: deps,
+	}
+}
+
+func (rm *PiTRManager) Enable(tc *v1alpha1.TidbCluster) error {
```
liubog2008, line 85: It's not reasonable to change TC spec in other controllers. There are many race issues if TC spec is changed by other controllers.

For example, clients may diff tikv config and find it's changed, and then update TC to change it back.

liubog2008 (changes requested):
Do not manage one field by many different owners. Users may also change the config field in TC spec directly and it's hard to resolve the conflict.

pkg/backup/restore/restore_manager.go
YuJuncen: If they updates the `gc.ratio-threshold` metric, the modification will be overridden by BR after BR finishes. But user generally won't touch this config term (if they did so, restore might be inconsistent) so I think this can be acceptable. Anyway though without this PR this problem already exists in kernel.

pkg/manager/utils/config.go
```diff
@@ -123,6 +122,10 @@ func confirmNameByData(existing, desired *corev1.ConfigMap, dataEqual bool) {
 		desired.Name = existing.Name
 	}
 	if !dataEqual && existing.Name == desired.Name {
-		desired.Name = fmt.Sprintf("%s-new", desired.Name)
+		// When restoring, there will be a overlay configuration directly applied to configmap.
```
liubog2008: Why?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -317,7 +319,15 @@ func (m *tikvMemberManager) syncTiKVConfigMap(tc *v1alpha1.TidbCluster, set *app
 		}
 	}
 
-	err = mngerutils.UpdateConfigMapIfNeed(m.deps.ConfigMapLister, tc.BaseTiKVSpec().ConfigUpdateStrategy(), inUseName, newCm)
+	updateStrategy := tc.BaseTiKVSpec().ConfigUpdateStrategy()
+	klog.InfoS("PiTR Status was modified, the configuration will be applied `InPlace`.", "cluster", klog.KObj(tc))
+
+	err = mngerutils.UpdateConfigMapIfNeed(m.deps.ConfigMapLister, updateStrategy, inUseName, newCm)
```
liubog2008: Why `UpdateConfigMapIfNeed` is not changed?

pkg/manager/member/tikv_member_manager.go
```diff
@@ -317,7 +319,15 @@ func (m *tikvMemberManager) syncTiKVConfigMap(tc *v1alpha1.TidbCluster, set *app
 		}
 	}
 
-	err = mngerutils.UpdateConfigMapIfNeed(m.deps.ConfigMapLister, tc.BaseTiKVSpec().ConfigUpdateStrategy(), inUseName, newCm)
+	updateStrategy := tc.BaseTiKVSpec().ConfigUpdateStrategy()
+	klog.InfoS("PiTR Status was modified, the configuration will be applied `InPlace`.", "cluster", klog.KObj(tc))
```
liubog2008: Confusing log

pkg/manager/member/tikv_member_manager.go
```diff
@@ -1110,3 +1120,86 @@ func (m *FakeTiKVMemberManager) Sync(tc *v1alpha1.TidbCluster) error {
...
+	}
+
+	// Filter restores that belong to this cluster and are PiTR mode
+	hasActivePiTRRestore := false
+	for _, restore := range restores {
+		if restore.Spec.BR != nil &&
+			restore.Spec.BR.Cluster == tc.Name &&
+			restore.Spec.BR.ClusterNamespace == tc.Namespace &&
+			restore.Spec.Mode == v1alpha1.RestoreModePiTR &&
+			!isRestoreDone(restore) {
```
liubog2008, line 1135: When is the `gc.ratio-threshold` changed back? What will happen if there are many restores(I'm not sure whether it's allowed)?

YuJuncen: When the restore finishes, then this overlay no more applies to configmap. The config map will be replaced by the configuration defined in `tikv.spec`.

pkg/manager/member/tikv_member_manager.go
```diff
@@ -1110,3 +1116,94 @@ func (m *FakeTiKVMemberManager) Sync(tc *v1alpha1.TidbCluster) error {
...
+	cm.Data["config-file"] = string(overlayAppliedConfigContent)
+	return nil
+}
+
+// isRestoreDone checks if a restore is completed or failed
+func isRestoreDone(restore *v1alpha1.Restore) bool {
+	if v1alpha1.IsRestoreComplete(restore) {
+		return true
+	}
+	if idx, _ := v1alpha1.GetRestoreCondition(&restore.Status, v1alpha1.RestoreFailed); idx != -1 {
```
WangLe1321: restore failed is a final state. we can treat as complete.

pkg/manager/utils/config.go
```diff
@@ -27,6 +27,30 @@ import (
 func updateConfigMap(old, new *corev1.ConfigMap) (bool, error) {
 	dataEqual := true
 
+	// When doing point-in-time-restoring or something, we will put an "overlay" over the tikv configuration.
+	// This should also be applied to the new one. (usually from spec)
+	applyOverlay := func(old, new *corev1.ConfigMap, key string) ([]byte, bool, error) {
```
liubog2008, line 49: why get overlay from the "old" and then overlay the "new" cfg?

liubog2008: Oh, I understand. But maybe still need more comments about why "use old overlay to overlay the new config"

liubog2008 (approved):

pkg/manager/member/tikv_member_manager.go
YuJuncen: This restore might be retried from the checkpoint? cc @Leavrth

csuzhangxc (approved):

PR #6243, opened 2025-06-12
opt: check PD member ready when upgrading

csuzhangxc:
### What problem does this PR solve?

Use PD ready API to check if PD members are ready during upgrading/updating, ref https://github.com/tikv/pd/pull/8749

### Code changes

- [x] Has Go code change
- [ ] Has CI related scripts change

### Tests

- [ ] Unit test 
- [ ] E2E test 
- [x] Manual test 
  - upgrade PD from 8.5.0 to 8.5.1 then to 8.5.2 (without mTLS enabled) successfully, and check log
    ```
    I0612 03:40:35.851728       1 pdapi.go:820] ready API is not found, assuming PD is ready
    ```
  - upgrade PD from 8.5.0 to 8.5.1 then to 8.5.2 (without mTLS enabled) successfully, and check log
     ```
     I0612 06:40:23.794992       1 tidb_cluster_controller.go:142] TidbCluster: default/basic, still need sync: tidbcluster: [default/basic]'s pd member: [basic-pd-1] is not ready, error: pd upgrader: failed to check if pd member basic-pd-1 is ready: Get "https://basic-pd-1.basic-pd-peer.default.svc:2379/pd/api/v2/ready": context deadline exceeded (Client.Timeout exceeded while awaiting headers), requeuing
     ```
- [ ] No code 

### Side effects

- [ ] Breaking backward compatibility 
- [ ] Other side effects: 

### Related changes

- [ ] Need to cherry-pick to the release branch 
- [ ] Need to update the documentation 

### Release Notes

Please refer to [Release Notes Language Style Guide](https://github.com/pingcap/tidb-operator/blob/master/docs/release-note-guide.md) before writing the release note.

```release-note

```

PR #6282, opened 2025-07-04
Add a feature gate to use PD's ready API as the readiness probe

fgksgf:

Port #6243 

### Code changes

- [x] Has Go code change

### Tests

- [ ] Unit test 
- [x] E2E test 
- [ ] Manual test 
- [ ] No code 

### Side effects

- [ ] Other side effects: 

### Related changes

### Release Notes

```release-note

```

### DIFF
--- a/pkg/controllers/pd/tasks/pod.go
+++ b/pkg/controllers/pd/tasks/pod.go
@@ -20,6 +20,7 @@ import (
 	"path"
 	"path/filepath"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -29,6 +30,7 @@ import (
 	metav1alpha1 "github.com/pingcap/tidb-operator/api/v2/meta/v1alpha1"
 	coreutil "github.com/pingcap/tidb-operator/pkg/apiutil/core/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/client"
+	"github.com/pingcap/tidb-operator/pkg/compatibility"
 	"github.com/pingcap/tidb-operator/pkg/features"
 	"github.com/pingcap/tidb-operator/pkg/image"
 	"github.com/pingcap/tidb-operator/pkg/overlay"
@@ -247,8 +249,12 @@ func newPod(cluster *v1alpha1.Cluster, pd *v1alpha1.PD, g features.Gates, cluste
 	if !g.Enabled(metav1alpha1.DisablePDDefaultReadinessProbe) {
 		pod.Spec.Containers[0].ReadinessProbe = buildPDReadinessProbe(coreutil.PDClientPort(pd))
 	}
+
 	if g.Enabled(metav1alpha1.UsePDReadyAPI) {
-		pod.Spec.Containers[0].ReadinessProbe = buildPDReadinessProbeWithReadyAPI(cluster, coreutil.PDClientPort(pd))
+		version, err := semver.NewVersion(pd.Spec.Version)
+		if err == nil && compatibility.Check(version, compatibility.PDReadyAPI) {
+			pod.Spec.Containers[0].ReadinessProbe = buildPDReadinessProbeWithReadyAPI(cluster, coreutil.PDClientPort(pd))
+		}
 	}
 
 	if pd.Spec.Overlay != nil {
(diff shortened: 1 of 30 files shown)


Makefile
```diff
@@ -52,36 +52,36 @@ deploy: release
...
 
-	$(DEEPCOPY_GEN) \
+	cd $(ROOT)/api && GOWORK=off $(DEEPCOPY_GEN) \
 		--output-file=zz_generated.deepcopy.go \
 		--go-header-file=$(BOILERPLATE_FILE) \
-		$(API_PATH)/...
+		github.com/pingcap/tidb-operator/api/v2/...
 
-	$(REGISTER_GEN) \
+	GOWORK=off $(REGISTER_GEN) \
```
liubog2008: why?

Makefile
```diff
@@ -239,7 +242,7 @@ bin/kind:
 .PHONY: bin/license-eye
 LICENSE_EYE = $(BIN_DIR)/license-eye
 bin/license-eye:
-	if [ ! -f $(LICENSE_EYE) ]; then $(ROOT)/hack/download.sh go_install $(LICENSE_EYE) github.com/apache/skywalking-eyes/cmd/license-eye v0.6.0; fi
+	if [ ! -f $(LICENSE_EYE) ]; then $(ROOT)/hack/download.sh go_install $(LICENSE_EYE) github.com/apache/skywalking-eyes/cmd/license-eye 049742de2276515409e3109ca2a91934053e080d; fi
```
liubog2008, line 242: why?

pkg/controllers/pd/tasks/pod.go
```diff
@@ -247,6 +250,13 @@ func newPod(cluster *v1alpha1.Cluster, pd *v1alpha1.PD, g features.Gates, cluste
 		pod.Spec.Containers[0].ReadinessProbe = buildPDReadinessProbe(coreutil.PDClientPort(pd))
 	}
 
+	if g.Enabled(metav1alpha1.UsePDReadyAPI) {
+		version, err := semver.NewVersion(pd.Spec.Version)
```
liubog2008: always handle this err. Just panic to ensure this err will not be ignored silently.

pkg/controllers/pd/tasks/pod.go
```diff
@@ -265,3 +275,40 @@ func buildPDReadinessProbe(port int32) *corev1.Probe {
...
+		cacert := path.Join(v1alpha1.DirPathClusterTLSPD, corev1.ServiceAccountRootCAKey)
+		cert := path.Join(v1alpha1.DirPathClusterTLSPD, corev1.TLSCertKey)
+		key := path.Join(v1alpha1.DirPathClusterTLSPD, corev1.TLSPrivateKeyKey)
+		command = append(command, "--cacert", cacert, "--cert", cert, "--key", key)
+	}
+
+	return &corev1.Probe{
+		InitialDelaySeconds: defaultReadinessProbeInitialDelaySeconds,
+		ProbeHandler: corev1.ProbeHandler{
+			Exec: &corev1.ExecAction{
```
liubog2008, line 309: Why not using http probe directly ？ It means we have to ensure that `curl` exists.

pkg/pdapi/v1/client.go
```diff
@@ -613,3 +619,32 @@ func (c *pdClient) GetTSOLeader(ctx context.Context) (string, error) {
 
 	return primary, nil
 }
+
+func (c *pdClient) GetMemberReady(ctx context.Context, url string) (bool, error) {
```
liubog2008: also use semver to check the version and do not depend on `404`

Makefile
fgksgf: https://github.com/apache/skywalking-eyes/pull/202

Makefile
fgksgf: Otherwise，`make check` will fail on my PC

pkg/controllers/pd/tasks/pod.go
fgksgf: If TLS is not enabled, we can use http probe. However, when TLS is enabled, client cert must be provided and we have to use exec action.

pkg/controllers/pd/tasks/pod.go
fgksgf: But we should avoid to use `panic`

pkg/pdapi/v1/client.go
fgksgf: Done

pkg/controllers/pd/tasks/pod.go
liubog2008: The version has been validated. It should not panic here.

If the error is not handled and our validation rule is not worked well. The best solution is panic to avoid undefined behaviour.

Makefile
fgksgf: Revert

Issue #6254, opened 2025-06-19
TiDB Operator may create more replicas than desired


fgksgf:
## Bug Report

**What version of Kubernetes are you using?**

1.31

**What version of TiDB Operator are you using?**

v2

**What did you do?**

Create a TiKVGroup with 1 replicas

**What did you expect to see?**
Only one tikv pods should be created.

**What did you see instead?**
There were two tikv pods, and operator can not delete the extra one:

![Image](https://github.com/user-attachments/assets/e49363be-feef-4c47-b04e-0e9f50903c2b)

PR #6287, opened 2025-07-08
fix(updater): fix scale issues because of read-after-write inconsistency

liubog2008:
- Fix some scale issues because of  read-after-write inconsistency
  - Controllers do not comply with read-after-write consistency because kube clients read from cache
  - e.g. Scale out more than expected
- Add `2s` watch delay for e2e to ensure all test cases are stable.
- Fix unsync PD config map after PD is bootstrapped.

This PR use `pkg/utils/tracker` to generate stable names when scaling out. If cache is not up to date, updater will try to create an existing instance and then fail because of `AlreadyExists` error. 

Fixes https://github.com/pingcap/tidb-operator/issues/6254

### DIFF
--- a/pkg/updater/actor.go
+++ b/pkg/updater/actor.go
@@ -19,9 +19,9 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/types"
 
-	"github.com/go-logr/logr"
 	"github.com/pingcap/tidb-operator/api/v2/core/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/client"
 	"github.com/pingcap/tidb-operator/pkg/runtime"
--- a/pkg/updater/selector_test.go
+++ b/pkg/updater/selector_test.go
@@ -17,12 +17,13 @@ package updater
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/pingcap/tidb-operator/api/v2/core/v1alpha1"
 	coreutil "github.com/pingcap/tidb-operator/pkg/apiutil/core/v1alpha1"
 	"github.com/pingcap/tidb-operator/pkg/runtime"
 	"github.com/pingcap/tidb-operator/pkg/runtime/scope"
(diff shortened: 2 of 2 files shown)


hack/lib/e2e.sh
```diff
@@ -28,7 +28,8 @@ source $ROOT/hack/lib/image.sh
 
 OUTPUT_DIR=$ROOT/_output
 KIND=$OUTPUT_DIR/bin/kind
-KUBECTL=$OUTPUT_DIR/bin/kubectl
+KUBECTL=kubectl
```
fgksgf: Why

pkg/controllers/pdgroup/tasks/boot.go
```diff
@@ -33,6 +38,46 @@ func TaskBoot(state *ReconcileContext, c client.Client) task.Task {
...
+		pds := state.InstanceSlice()
+		for _, pd := range pds {
+			if _, ok := pd.Annotations[v1alpha1.AnnoKeyInitialClusterNum]; !ok {
+				continue
+			}
+			if err := delBootAnnotation(ctx, c, pd); err != nil {
+				return task.Fail().With("cannot del boot annotation after pd is available: %v", err)
+			}
+		}
+
```
fgksgf, line 50: Add some comments to explain why we need to delete the annotaion

pkg/utils/tracker/tracker.go
```diff
@@ -0,0 +1,141 @@
...
+	I client.Object,
+] interface {
+	Track(owner O, items ...I) Allocator
+}
+
+type tracker[
+	O client.Object,
+	I client.Object,
+] struct {
+	as map[string]Allocator
```
fgksgf: Need a mutex?

hack/lib/e2e.sh
liubog2008: fixed

pkg/controllers/pdgroup/tasks/boot.go
liubog2008: done

fgksgf (approved):

PR #6311, opened 2025-07-17
fix(tikv): evict all leaders before deleting tikv

liubog2008:
- evict all leaders before deleting tikv

### DIFF
--- a/pkg/controllers/tikv/tasks/ctx.go
+++ b/pkg/controllers/tikv/tasks/ctx.go
@@ -63,18 +63,21 @@ func TaskContextInfoFromPD(state *ReconcileContext, cm pdm.PDClientManager) task
 		state.SetRegionCount(s.RegionCount)
 		state.SetStoreBusy(s.IsBusy)
 
-		// TODO: cache evict leader scheduler info, then we don't need to check suspend here
-		if coreutil.ShouldSuspendCompute(state.Cluster()) {
-			return task.Complete().With("cluster is suspending")
-		}
+		// TODO: cache evict leader scheduler info
 		scheduler, err := state.PDClient.Underlay().GetEvictLeaderScheduler(ctx, state.Store.ID)
 		if err != nil {
-			return task.Fail().With("pd is unexpectedly crashed: %w", err)
+			return task.Fail().With("pd is unexpectedly crashed: %v", err)
 		}
 		if scheduler != "" {
 			state.LeaderEvicting = true
 		}
 
-		return task.Complete().With("get store info")
+		return task.Complete().With("get store info, state: %s, leader count: %v, region count: %v, busy: %v, leader evicting: %v",
+			state.GetStoreState(),
+			state.GetLeaderCount(),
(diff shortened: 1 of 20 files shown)


pkg/controllers/tikv/tasks/ctx.go
```diff
@@ -63,18 +63,21 @@ func TaskContextInfoFromPD(state *ReconcileContext, cm pdm.PDClientManager) task
...
 
-		// TODO: cache evict leader scheduler info, then we don't need to check suspend here
-		if coreutil.ShouldSuspendCompute(state.Cluster()) {
-			return task.Complete().With("cluster is suspending")
-		}
+		// TODO: cache evict leader scheduler info
 		scheduler, err := state.PDClient.Underlay().GetEvictLeaderScheduler(ctx, state.Store.ID)
 		if err != nil {
-			return task.Fail().With("pd is unexpectedly crashed: %w", err)
+			return task.Fail().With("pd is unexpectedly crashed: %v", err)
```
fgksgf, line 72: why change this

pkg/controllers/tikv/tasks/offline.go
```diff
@@ -23,23 +23,34 @@ import (
...
-		if state.PDClient == nil {
-			return task.Fail().With("pd client is not registered")
+		var reason string
+		delTime := state.Object().GetDeletionTimestamp()
+		switch {
+		// leaders evicted
+		case state.LeaderEvicting && state.GetLeaderCount() == 0:
+			reason = "leaders have been all evicted"
+		case !delTime.IsZero() && delTime.Add(defaultLeaderEvictTimeout).Before(time.Now()):
+			reason = "leader eviction timeout"
```
fgksgf, line 43: Use the constant defined in `api`? It seems it's duplicated with `syncLeadersEvictedCond`?

pkg/controllers/tikv/tasks/ctx.go
liubog2008: %w is not supported ..

pkg/controllers/tikv/tasks/offline.go
liubog2008: which one?

tests/e2e/utils/tidb/tidb.go
```diff
@@ -327,7 +393,15 @@ func AreAllTiDBHealthy(cli client.Client, dbg *v1alpha1.TiDBGroup) error {
 	return nil
 }
 
-func AreAllTiFlashHealthy(cli client.Client, flashg *v1alpha1.TiFlashGroup) error {
+func AreAllTiFlashHealthy(cli client.Client, flashg *v1alpha1.TiFlashGroup, checkGroup bool) error {
+	if checkGroup {
+		if err := cli.Get(context.TODO(), client.ObjectKeyFromObject(flashg), flashg); err != nil {
+			return err
+		}
```
fgksgf, line 400: In what case, we need to set `checkGroup` to false?

PR #6325, opened 2025-07-25
feat(log): support use sidecar to tail log

liubog2008:
- tail cannot handle the signal TERM when it's PID is 1
- use sidecar to avoid losing logs

### DIFF
--- a/pkg/apis/pingcap/v1alpha1/types.go
+++ b/pkg/apis/pingcap/v1alpha1/types.go
@@ -984,7 +984,10 @@ type LogTailerSpec struct {
 	corev1.ResourceRequirements `json:",inline"`
 
 	// If true, we use native sidecar feature to tail log
+	// It requires enable feature gate "SidecarContainers"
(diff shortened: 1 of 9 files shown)


csuzhangxc (approved):

pkg/apis/pingcap/v1alpha1/types.go
```diff
@@ -982,6 +982,11 @@ type TiProxySpec struct {
 // +k8s:openapi-gen=true
 type LogTailerSpec struct {
 	corev1.ResourceRequirements `json:",inline"`
+
+	// If true, we use native sidecar feature to tail log
+	// See https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/
+	// +optional
```
fgksgf: Add a comment, the sidecar-container requires k8s version 1.x

liubog2008: fixed

Issue #6419, opened 2025-09-01
tikv rolling restart blocked


Lily2025:
## Bug Report

**What version of Kubernetes are you using?**

**What version of TiDB Operator are you using?**
v2.0.0-beta.8 

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**
no matter

**What's the status of the TiDB cluster pods?**
39 * tikv

**What did you do?**
update the config of tikv

**What did you expect to see?**
rolling restart can succeed

**What did you see instead?**
tikv rolling restart blocked

![Image](https://github.com/user-attachments/assets/7bba1948-9858-4f33-82bb-7e6b34695d0c)

PR #6427, opened 2025-09-05
fix(tikv): cannot retry if pod is ready but not available

liubog2008:
- cannot requeue if tikv is not ready but available


liubog2008:
Fix https://github.com/pingcap/tidb-operator/issues/6419

PR #6441, opened 2025-09-09
feat(e2e): add workload to ensure no down time when tikv is rolling update

liubog2008:
Add a workload to ensure that tikv has no down time.

NOTE: I'm not sure whether it always works in some corner cases


cmd/testing-workload/workload.go
```diff
@@ -35,6 +35,8 @@ func Workload(db *sql.DB) error {
 
 	db.SetMaxIdleConns(maxConnections)
 	db.SetMaxOpenConns(maxConnections)
+	db.Exec("set global max_execution_time = 1000")
+	db.Exec("set global tidb_backoff_weight = 1")
```
fgksgf: add comments for these statements?

Issue #6433, opened 2025-09-08
tiflash query report error during tiflash cn rolling restart


Lily2025:
## Bug Report

**What version of Kubernetes are you using?**
v2.0.0-beta.8

**What version of TiDB Operator are you using?**
no matter

**What storage classes exist in the Kubernetes cluster and what are used for PD/TiKV pods?**
no matter

**What's the status of the TiDB cluster pods?**
normal

**What did you do?**
1、run tiflash query
2、modify tiflash cn config and tiflash cn rolling restart

**What did you expect to see?**
tiflash query run normally

**What did you see instead?**
tiflash query report error

[2025/09/08 07:44:50.094 +00:00] [WARN] [local_mpp_coordinator.go:839] ["other error"] [keyspaceName=10643894232434485134] [txnStartTS=460670111241994252] [storeAddr=db-fc7681af-compute-tiflash-lxeyct.db-fc7681af-compute-tiflash-peer.tidb1738483010403163:3930] [mpp-version=3] [task-id=23] [error="other error for mpp stream: Code: 0, e.displayText() = DB::Exception: Receiver state: ERROR, error message: Receive cancel request from TiDB, e.what() = DB::Exception,"]

liubog2008:
Also see https://github.com/pingcap/tidb-operator/issues/6364

liubog2008:
Fixed by https://github.com/pingcap/tidb-operator/pull/6450

PR #6450, opened 2025-09-12
fix(tiflash): fix rolling update issue for tiflash

liubog2008:
- Add a feature gates to support terminating log tail.
- Fix tiflash ready condition
- Add e2e to ensure that tiflash is always available


JaySon-Huang:
Related to https://github.com/pingcap/tidb-operator/issues/6433

fgksgf (approved):

PR #6487, opened 2025-10-09
fix(updater): fix defer deleting tikv cannot be canceled

liubog2008:
- combine `cancel offline` and `update` to avoid scaling out an outdated instance
- del `defer deletion` annotation when set offline to allow the deletion can be canceled.
- support poll `Removed` store
- refactor offline store task to ensure removed store has completed condition.

### DIFF
--- a/api/core/v1alpha1/common_types.go
+++ b/api/core/v1alpha1/common_types.go
@@ -74,8 +74,8 @@ const (
 const (
 	// ReasonOfflineProcessing means the store is being offlined.
 	ReasonOfflineProcessing = "Processing"
-	// ReasonOfflineCancelling means the store is being offlined.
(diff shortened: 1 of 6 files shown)


api/core/v1alpha1/common_types.go
```diff
@@ -74,9 +74,13 @@ const (
 const (
 	// ReasonOfflineProcessing means the store is being offlined.
 	ReasonOfflineProcessing = "Processing"
+	// ReasonOfflineCancelling means the store is being offlined.
+	ReasonOfflineCancelling = "Cancelling"
```
fgksgf:
```suggestion
	// ReasonOfflineCanceling means the store is being offlined.
	ReasonOfflineCanceling = "Canceling"
```

Issue #6530, opened 2025-11-03
Fix no inplace update way


liubog2008:
We expected no inplace update can act like deployment. However, a new instance will be created immediately after the outdated one is marked as deleting.

The expected behavior is waiting until the outdated one is fully deleted.

Now there are at most two extra instances created which is more than maxSurge(1).

PR #6533, opened 2025-11-04
fix(updater): optimize no inplace update

liubog2008:
Fix https://github.com/pingcap/tidb-operator/issues/6530


pkg/controllers/ticdcgroup/tasks/updater.go
```diff
@@ -113,7 +113,7 @@ func TaskUpdater(state *ReconcileContext, c client.Client, af tracker.AllocateFa
 			WithDelHooks(topoPolicy).
 			WithUpdateHooks(topoPolicy).
 			WithScaleInPreferPolicy(
-				topoPolicy,
+				topoPolicy.PolicyScaleIn(),
 			).
```
fgksgf, line 117: Add `WithUpdatePreferPolicy` for ticdc?

Issue #6461, opened 2025-09-22
Update pd member event is lost


liubog2008:
Poller sent `MODIFIED` event but no `update member` log. A PD instance stucks in the unready state.

liubog2008:
Poll events of all controllers may be lost. TSO and TiKV also meet same issue.

PR #6572, opened 2025-12-05
fix(timanager): fix lost event issue

liubog2008:
Fix https://github.com/pingcap/tidb-operator/issues/6461

- reset default resync period to 0
- use uuid as resource version
- add ut to replay the issue

## What happens

- Informer will ignore events if rv is changed or `isSyncing` is true. See https://github.com/kubernetes/client-go/blob/v0.32.6/tools/cache/shared_informer.go#L657-L676 and https://github.com/kubernetes/client-go/blob/v0.32.6/tools/cache/shared_informer.go#L779-L795
- `isSyncing` will be false only when `shouldResync` is called. See https://github.com/kubernetes/client-go/blob/v0.32.6/tools/cache/shared_informer.go#L837
- `shouldResync` will be called in reflector at `resyncPeriod` later after informer is started.
- If `addListener` is called after informer is started, `shouldResync` will be false at first resync. And then `isSyncing` will be true at second resync.
- If resync period is 0. `isSyncing` will always be true.


csuzhangxc (approved):

fgksgf (approved):

liubog2008:
cherry-pick release-2.0

PR #6627, opened 2025-12-25
feat(cluster): support using FQDN

liubog2008:
- support using FQDN in the tidb cluster

### DIFF
--- a/pkg/configs/pd/config_test.go
+++ b/pkg/configs/pd/config_test.go
@@ -67,6 +67,10 @@ func TestValidate(t *testing.T) {
 
 func TestOverlay(t *testing.T) {
 	cluster := &v1alpha1.Cluster{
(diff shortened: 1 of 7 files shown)


pkg/controllers/tsogroup/builder.go
```diff
@@ -56,6 +55,9 @@ func (r *Reconciler) NewRunner(state *tasks.ReconcileContext, reporter task.Task
 			common.TaskStatusPersister[scope.TSOGroup](state, r.Client),
 		),
 
+		// register tso client
+		tasks.TaskContextTSOClient(state, r.TSOClientManager),
```
tennix, line 59: Why re-order the task, does the order matter?

liubog2008: If the group is deleting, register may always fail. It blocks the deletion of this tso group.

PR #6631, opened 2025-12-26
feat(client): support ignore diff annotation

liubog2008:
- support `pingcap.com/ignore-diff` to avoid unexpected diff produced by mutating webhook.

## Why

Local dns webhook of aliyun will change `spec.dnsPolicy` to `None` when pod is creating.
And the webhook doesn't handle the `UPDATE` operation. It breaks the apply op if we set `spec.hostNetwork: true` and `spec.dnsPolicy: ClusterFirstWithHostNet`.
And we cannot unset the `spec.dnsPolicy` because the webhook depends on this field to inject the dns config.

We cannot ensure that all mutating webhooks work properly and it's also hard to filter all immutable fields in client side. That's why we add this annotation to ignore diff.

## How to use

```yaml
# group cr
spec:
  template:
    spec:
      overlay:
        pod:
          metadata:
            annotations:
              pingcap.com/ignore-diff: "spec.dnsPolicy"
```

### DIFF
--- a/pkg/client/opt.go
+++ b/pkg/client/opt.go
@@ -21,7 +21,7 @@ import (
 )
 
 // AnnoKeyIgnoreDiff defines the fields list to ignore difference which may be produced by mutating webhook.
-// Fields are seperated by comma. Example: "spec.xxx,spec.yyy"
+// Fields are separated by comma. Example: "spec.xxx,spec.yyy"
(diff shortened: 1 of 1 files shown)


srstack (approved):

tennix (approved):
LGTM

fgksgf (approved):

PR #6644, opened 2026-01-08
feat(crd): support applying crd by operator

liubog2008:
- support applying crds directly

### DIFF
--- a/charts/tidb-operator/templates/deployment.yaml
+++ b/charts/tidb-operator/templates/deployment.yaml
@@ -44,14 +44,14 @@ spec:
       - command:
         - /tidb-operator
         args:
-          - --backup-manager-image={{ template "image" (tuple .Values.backupManager.image $.Chart.AppVersion) }}
-          - --default-prestop-checker-image={{ template "image" (tuple .Values.prestopChecker.image $.Chart.AppVersion) }}
-          - --default-resource-syncer-image={{ template "image" (tuple .Values.resourceSyncer.image $.Chart.AppVersion) }}
-        {{- if eq .Values.skipCRDsManagement true -}}
-          - --skip-crds-management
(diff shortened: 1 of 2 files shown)


cmd/tidb-operator/main.go
```diff
@@ -109,6 +114,8 @@ func main() {
 			"kube clients's read-after-write inconsistency(always read from cache).")
 	flag.Var(&image.PrestopChecker, "default-prestop-checker-image", "Default image of prestop checker.")
 	flag.Var(&image.ResourceSyncer, "default-resource-syncer-image", "Default image of resource syncer.")
+	flag.BoolVar(&allowEmptyOldVersion, "allow-empty-old-version", false,
+		"allow crd version is empty, if false, cannot update crd which has no version annotation")
```
fgksgf, line 119: Need a flag to control whether or not apply CRDs automatically

cmd/tidb-operator/main.go
```diff
@@ -144,6 +155,11 @@ func main() {
 	kubeconfig.UserAgent = client.DefaultFieldManager
 	kubefeat.MustInitFeatureGates(kubeconfig)
 
+	if err := applyCRDs(ctx, kubeconfig, allowEmptyOldVersion); err != nil {
+		setupLog.Error(err, "failed to apply crds")
+		os.Exit(1)
```
fgksgf: How about just logging the error

cmd/tidb-operator/main.go
liubog2008: why? I hope to ensure the CRDs are in right version.

cmd/tidb-operator/main.go
liubog2008: same as the above.

PR #6680, opened 2026-01-26
Add operator build info metric

SuperQ:
Add a new metric to the tidb-operator that exposes the build info as labels. This allows for monitoring of what versions are deployed across multiple installs.


liubog2008:
thanks for your contribution

PR #6815, opened 2026-04-10
feat: support mount sa token explicitly if automount is false

liubog2008:
- If automountServiceAccountToken is set to false, br and discovery should mount sa token explicitly


liubog2008:
e	/test pull-e2e-kind-basic

csuzhangxc (approved):

PR #6829, opened 2026-04-16
tiproxy: add annotation `tiproxy-graceful-shutdown-delete-delay-seconds` to gracefully mark unhealthy before deleting the pods

YangKeao:
# Background

I'm designing the graceful restarting of TiProxy in cloud environment.

The intended behavior is:

1. Use a big `maxSurge`.
2. Patch the existing `TiProxyGroup` to enable a long graceful shutdown delete delay.
3. Restart.
4. Old TiProxy instances are first marked unhealthy, then kept alive for a while before the old pods are actually deleted.

This is mainly for cloud load balancers: existing long-lived connections can continue to work on the old TiProxy (with big enough `target_health_state.unhealthy.draining_interval_seconds` for AWS and disable `ConnectionDrainEnabled` for aliyun), while new connections should be sent to the new TiProxy instances.

We cannot rely on changing `terminationGracePeriodSeconds` for existing pods, because that would itself require restarting them. So this PR adds a controller-side graceful delete flow for TiProxy.

# Design

1. Add a new spec field `spec.template.spec.gracefulShutdownDeleteDelaySeconds` to `TiProxyGroup` / `TiProxy`.
2. This field is treated as reloadable, so patching it does **not** trigger a rolling restart by itself.
3. When a `TiProxy` object is being deleted and this field is set to a positive value:
    1. operator first tries to call `POST /api/debug/health/unhealthy`
    2. if the API is not supported (`404`), operator falls back to sending `SIGTERM` to the TiProxy process by `pods/exec`
    3. only after TiProxy is confirmed unhealthy, operator writes `core.pingcap.com/tiproxy-graceful-shutdown-begin-time` on the pod and starts the delete-delay timer
    4. after the timer expires, operator deletes the pod
4. If TiProxy cannot be marked unhealthy, operator will keep retrying and will **not** start the delete-delay timer.
5. When the whole `Cluster` is deleting, this graceful delay is skipped and the pod is deleted directly.

This design keeps the user-facing control in `spec`, avoids changing `terminationGracePeriodSeconds`, and supports both new TiProxy versions (with unhealthy API) and older ones (with `SIGTERM` fallback).

# Usage

Patch an existing `TiProxyGroup`, so old TiProxy pods will be kept for a while after they are marked unhealthy:

```bash
kubectl --context "$CONTEXT" -n "$NS" patch tiproxygroup pg --type merge -p '{
  "spec": {
    "template": {
      "spec": {
        "gracefulShutdownDeleteDelaySeconds": 20
      }
    }
  }
}'

Then trigger a rolling restart, for example by changing config or image. With a large maxSurge, new TiProxy pods can come up first, and old TiProxy pods will only be deleted after entering graceful shutdown and waiting for
the configured delay.

### CODE REVIEW

File: pkg/controllers/tiproxy/tasks/finalizer.go [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+			retryAfter, err := drainOrDeletePod(ctx, c, tiproxy, pod)
+			if err != nil {
+				return task.Fail().With("cannot delete pod of tiproxy: %v", err)
+			}
+			if retryAfter > 0 {
+				return task.Retry(retryAfter).With("wait for tiproxy pod to be deleted")
+			}
+		}
+
+		res, _ := task.RunTask(ctx, common.TaskInstanceFinalizerDel[scope.TiProxy](state, c, tiproxyFinalizerSubresourceLister))
```
liubog2008, line 56: Add a new task and do not run another task in this task. If task is not fail, next task will be run.

pkg/controllers/tiproxy/tasks/finalizer.go
```diff
@@ -0,0 +1,114 @@
...
+var tiproxyFinalizerSubresourceLister = common.NewSubresourceLister(
+	common.NewSubresource[corev1.ConfigMapList](),
+	common.NewSubresource[corev1.PersistentVolumeClaimList](),
+)
+
+func TaskFinalizerDel(state common.ObjectState[*v1alpha1.TiProxy], c client.Client) task.Task {
+	return task.NameTaskFunc("FinalizerDel", func(ctx context.Context) task.Result {
+		tiproxy := state.Object()
+
+		pod, err := apicall.GetPod[scope.TiProxy](ctx, c, tiproxy)
```
liubog2008, line 42: move `common.TaskContextPod[scope.TiProxy](state, r.Client)` before this task and no need to get pod again.

pkg/controllers/tiproxy/tasks/finalizer.go
```diff
@@ -0,0 +1,114 @@
...
+				return task.Retry(retryAfter).With("wait for tiproxy pod to be deleted")
+			}
+		}
+
+		res, _ := task.RunTask(ctx, common.TaskInstanceFinalizerDel[scope.TiProxy](state, c, tiproxyFinalizerSubresourceLister))
+		return res
+	})
+}
+
+func drainOrDeletePod(ctx context.Context, c client.Client, tiproxy *v1alpha1.TiProxy, pod *corev1.Pod) (time.Duration, error) {
```
liubog2008, line 61: How to notify the tiproxy that it is terminating?

pkg/controllers/tiproxy/tasks/finalizer.go
YangKeao: Done.

pkg/controllers/tiproxy/tasks/finalizer.go
YangKeao: 1. Use the `unhealthy` API in TiProxy (https://github.com/pingcap/tiproxy/pull/1141 not merged yet).
2. If the current version of TiProxy doesn't have the API, use the `kill xxx` to send SIGTERM, which requires the TiProxy to have a large enough `graceful-wait-before-shutdown` to continue to work during the termination.

We can set the `TiProxyGroup.spec.template.spec.gracefulShutdownDeleteDelaySeconds` to 24h and set the tiproxy config `graceful-wait-before-shutdown` to 25h, so though the tiproxy is not actually internally graceful shutdown, the connection should be quited gracefully (in 24h).

pkg/controllers/tiproxy/tasks/finalizer.go
YangKeao: Done.

YangKeao:
/hold

Merge after https://github.com/pingcap/tiproxy/pull/1141

pkg/tiproxyapi/v1/client.go
```diff
@@ -76,9 +79,26 @@ func NewTiProxyClient(addr string, timeout time.Duration, tlsConfig *tls.Config)
...
+	//nolint:bodyclose,gosec // bodyclose: has been handled; gosec: URL is constructed from trusted internal config
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer httputil.DeferClose(resp.Body)
+
 	// NOTE: we don't check the response body here.
-	return err == nil, err
+	return resp.StatusCode < http.StatusBadRequest, nil
```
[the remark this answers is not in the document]
YangKeao: `DeferClose` will close it.

PR #6835, opened 2026-04-20
feat(metrics): add per-instance abnormal_instance gauge for stuck-state alerts

fgksgf:
## Summary

Add a single Prometheus gauge `tidb_operator_abnormal_instance` that flags managed instances stuck in an abnormal state.

```
tidb_operator_abnormal_instance{
    namespace, cluster, component, group, instance,
    condition="Synced"|"Ready",
}
```

- `1` when the named condition is `False`, `0` otherwise.
- Series stays present while the operator manages the instance, removed when the instance is finalized.

Recommended alerting form (Prometheus owns the duration logic):

```yaml
- alert: TiKVInstanceAbnormalTooLong
  expr: tidb_operator_abnormal_instance{component="tikv", condition="Synced"} == 1
  for: 30m
```

### Why two condition values

- `Synced=False`: operator has more work to do (template mismatch, pod missing, scale-in pending). Persistent value indicates the operator cannot make the pod converge.
- `Ready=False`: pod is up but cannot serve (PD unreachable, leader eviction stuck, store offline). Persistent value indicates a degraded data plane even when the rollout looks fine.

Both warrant attention; one metric with a `condition` label keeps the schema small and leaves room for additional conditions.

## Implementation

- The gauge is updated inside the existing `TaskInstanceConditionSynced` and `TaskInstanceConditionReady` common tasks. PD / TiKV / TiDB / TiFlash / TiCDC / TiProxy / Scheduler / Scheduling / TSO / ResourceManager / Router / TiKVWorker all expose both samples automatically.
- `ClearInstanceConditionMetrics` is internalized into the existing `TaskInstanceFinalizerDel`. Future instance controllers using the standard finalize task get cleanup for free.
- Helpers live in `pkg/metrics` next to the gauge definition.

## Test plan

- [x] `make lint` — 0 issues
- [x] `make unit` — all packages pass
- [x] Unit tests cover Observe paths for `True` / `False` / absent for both conditions plus the finalizer-clear path

### DIFF
--- a/pkg/controllers/common/task_finalizer.go
+++ b/pkg/controllers/common/task_finalizer.go
@@ -24,6 +24,7 @@ import (
 	utilerr "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/pingcap/tidb-operator/v2/pkg/client"
+	"github.com/pingcap/tidb-operator/v2/pkg/metrics"
 	"github.com/pingcap/tidb-operator/v2/pkg/runtime"
 	"github.com/pingcap/tidb-operator/v2/pkg/runtime/scope"
 	"github.com/pingcap/tidb-operator/v2/pkg/utils/k8s"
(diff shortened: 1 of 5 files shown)


pkg/controllers/common/instance_unsynced_metric.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// ObserveInstanceUnsynced refreshes the unsynced gauge based on the instance's
+// Synced condition.
+//
+//   - Synced=False: gauge = time.Since(LastTransitionTime). A growing value
+//     means the operator cannot make the pod converge to the desired template
+//     (rolling restart stuck, scale-in stuck, manual delete stuck, ...).
+//   - Synced=True or absent: the series is dropped.
+func ObserveInstanceUnsynced(obj client.Object, conds []metav1.Condition) {
```
liubog2008: move to metrics pkg？

PR #6839, opened 2026-04-21
fix(metrics): prevent abnormal_instance gauge leaks on deletion paths

fgksgf:
## Summary

Follow-up to #6835. On a dev EKS we saw the `tidb_operator_abnormal_instance` gauge retain `value=1` for 8 instances whose CR / Pod / parent Group had all been deleted, which would have produced false-positive `metric == 1 for: 30m` alerts and grown label cardinality across each cluster lifecycle.

Root cause is two independent paths:

1. **Seven instance builders keep Cond* tasks inside the `CondObjectIsDeleting` IfBreak block** (pd/tidb/tiproxy/ticdc/tso/scheduling/tikvworker). `TaskInstanceFinalizerDel` clears the gauge via `ClearInstanceConditionMetrics`, but the `TaskInstanceConditionReady` that follows immediately re-populates it with `value=1` because the pod is gone (`PodNotCreated`). `TaskStatusPersister` at the tail then fails with `NotFound`, confirming the CR is already deleted. `tikv` / `tiflash` already had the correct shape - only `FinalizerDel` in the deletion branch. This PR aligns the remaining seven.

2. **Force-delete paths that strip finalizers bypass `FinalizerDel` entirely.** `kubectl patch --type=merge -p '{\"metadata\":{\"finalizers\":null}}'` is an operationally common action when unwedging a stuck instance (exactly the state this alert exists for), so relying solely on the finalize-time cleanup is insufficient. Add a watch-level `DELETE` handler on every instance kind that writes the gauge, so the series clears the moment the API server GCs the object. Covers force-delete, operator-down-during-delete, and any other path that reaches the informer without going through `FinalizerDel`.

### Changes

- `pkg/controllers/{pd,tidb,tiproxy,ticdc,tso,scheduling,tikvworker}/builder.go`: remove `TaskInstanceConditionSynced/Ready/Running` + `TaskStatusPersister` from the `CondObjectIsDeleting` branch.
- `pkg/metrics/cleanup_handler.go`: new `RegisterAbnormalInstanceCleanup` that attaches a `DeleteFunc` on each instance kind's shared informer.
- `cmd/tidb-operator/main.go`: wire the registration into manager setup.
- `pkg/metrics/abnormal_instance.go`: update `ClearInstanceConditionMetrics` doc to describe both cleanup call sites.
- `tests/e2e/utils/metrics/metrics.go`: add generic `FetchOperatorMetric` helper.
- `tests/e2e/metrics/abnormal_instance_leak.go`: two cases on TiKV - graceful delete via the group, and force-delete via finalizer strip after `Delete()` sets `DeletionTimestamp`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...` on modified packages
- [x] `golangci-lint run` on modified packages - 0 issues
- [x] Unit tests green on `pkg/metrics` and controller `tasks` packages
- [ ] E2E `Abnormal Instance Gauge Leak` against a kind cluster
- [ ] Dev EKS soak: deploy this image, confirm current 8 leaked series disappear after operator restart, then `kubectl patch --type=merge -p '{"metadata":{"finalizers":null}}'` a TiProxy CR and confirm its series is cleared within seconds

### DIFF
--- a/cmd/tidb-operator/main.go
+++ b/cmd/tidb-operator/main.go
@@ -294,6 +294,10 @@ func setup(ctx context.Context, mgr ctrl.Manager) error {
 	tsocm.Start(ctx)
 	rmcm.Start(ctx)
 
+	if err := metrics.RegisterAbnormalInstanceCleanup(ctx, mgr); err != nil {
+		return fmt.Errorf("register abnormal-instance cleanup: %w", err)
+	}
+
 	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
 		return fmt.Errorf("unable to set up health check: %w", err)
 	}
--- a/pkg/controllers/pd/builder.go
+++ b/pkg/controllers/pd/builder.go
@@ -39,11 +39,6 @@ func (r *Reconciler) NewRunner(state *tasks.ReconcileContext, reporter task.Task
 		task.IfBreak(common.CondObjectIsDeleting[scope.PD](state),
 			tasks.TaskDeleteMember(state),
 			common.TaskInstanceFinalizerDel[scope.PD](state, r.Client, common.DefaultInstanceSubresourceLister),
-			// TODO(liubo02): if the finalizer has been removed, no need to update status
-			common.TaskInstanceConditionSynced[scope.PD](state),
-			common.TaskInstanceConditionReady[scope.PD](state),
-			common.TaskInstanceConditionRunning[scope.PD](state),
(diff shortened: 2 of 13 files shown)


fgksgf:
Closing in favor of a smaller refactor: a single TaskObserveInstance in pkg/controllers/common that handles both observe (when obj exists) and clear (when obj is nil, i.e. watch DELETE), mirroring the shape of TaskTrack. Reviewer feedback: observe and clear should live together. New PR coming.

PR #6849, opened 2026-04-23
fix(tikv): ensure store-id label is set before scale-in to prevent tombstone matching failure

zgcbj:
Problem
When scaling in a TiKV node, the scale-in flow relies on matching the pod's store-id label against TombstoneStores to confirm the store has been safely removed:

if pod.Labels[label.StoreIDLabelKey] != storeID {
    // skip — mismatch, scale-in stuck here forever
    continue
}

However, UpdateMetaInfo (which writes the store-id label to the pod) only looks up tc.Status.TiKV.Stores. Once the store transitions to Tombstone, it is moved out of Stores into TombstoneStores, so UpdateMetaInfo can no longer find the storeID to write. If the pod label was never written before this transition, the tombstone matching check always fails and the scale-in gets permanently stuck.

Root Cause
The store-id label may not be set on the pod when scale-in is triggered (e.g., due to a timing issue or controller restart). Since UpdateMetaInfo only queries tc.Status.TiKV.Stores, once the store becomes a tombstone, the label can never be patched and the scale-in flow is blocked indefinitely.

Fix
In scaleInOne, when the store is still present in tc.Status.TiKV.Stores (Up/Offline state) and the pod is missing the store-id label, proactively call UpdateMetaInfo to write the label immediately. The existing RequeueError at the end of this branch ensures the next reconcile loop picks up the updated pod from the informer cache, at which point the tombstone matching can succeed normally.


liubog2008:
@zgcbj please change base to `release-1.x`

zgcbj:
> @zgcbj please change base to `release-1.x`

I've checked the release-1.x branch, and the changes here should have resolved this issue. My PR is probably no longer necessary.
https://github.com/pingcap/tidb-operator/blob/a0439462232b3cf4805dbe20e7e50703e0ce56b3/pkg/manager/member/tikv_scaler.go#L206

PR #6864, opened 2026-04-27
feat: support of PodDisruptionBudgets

### LINKED ISSUE #4221: Support PDB for k8s cluster version upgrade
## Feature Request

We need to limit the unavailable pod number when upgrading k8s clusters' version, especially for multi k8s clusters deployment

Use PDB to limit the unavailable pod number

We have hundreds of k8s clusters, so Ops need to upgrade in parallel to save time.

### LINKED ISSUE #6901: Support PodDisruptionBudget in v2
In order to allow safe background maintenance of cluster nodes it is useful to deploy [PodDisruptionBudget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) resources to control how fast Kubernetes can act on a cluster.

Currently we have resources like this:

```
TIKVGroup -> TiKV -> Pod
```

But because [PDB maxUnavailable only works on core resources](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors) we can't simply implement the Scale subresource on TiKV.

Instead we need to be able to have an intermediate resource like this:

```
TIKVGroup -> TiKV -> StatefulSet -> Pod
```

This will work by allowing the PDB to match all instances within a TiKVGroup.

Here's a simple demo config that shows how this would work:

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    app: podinfo
  name: test-0
  namespace: pdb-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: podinfo
  template:
    metadata:
      labels:
        app: podinfo
    spec:
      containers:
      - image: ghcr.io/stefanprodan/podinfo:latest
        name: podinfo
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    app: podinfo
  name: test-1
  namespace: pdb-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: podinfo
  template:
    metadata:
      labels:
        app: podinfo
    spec:
      containers:
      - image: ghcr.io/stefanprodan/podinfo:latest
        name: podinfo
---
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    app: podinfo
  name: test-2
  namespace: pdb-test
spec:
  replicas: 1
  selector:
    matchLabels:
      app: podinfo
  template:
    metadata:
      labels:
        app: podinfo
    spec:
      containers:
      - image: ghcr.io/stefanprodan/podinfo:latest
        name: podinfo
---
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: test-statefulset
  namespace: pdb-test
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      app: podinfo
```

As you can see by the PDB status all 3 individual StatefulSet instances are taken into account:

```console
$ kubectl describe pdb test-statefulset 
Name:             test-statefulset
Namespace:        pdb-test
Max unavailable:  1
Selector:         app=podinfo
Status:
    Allowed disruptions:  1
    Current:              3
    Desired:              2
    Total:                3
Events:                   <none>
```

SuperQ:
In order for PodDisruptionBudgets to work on various groups of Pods we need to implement the Scale subresource at the direct owner of the Pod.

Otherwise the PDB will return errors like:

```yaml
status:
  conditions:
    - lastTransitionTime: "2026-04..."
      message: tikvs.core.pingcap.com does not implement the scale subresource
      reason: SyncFailed
      status: "False"
      type: DisruptionAllowed
```

Followup to https://github.com/pingcap/tidb-operator/pull/6213

Closes: https://github.com/pingcap/tidb-operator/issues/6901
Closes: https://github.com/pingcap/tidb-operator/issues/4221

### DIFF
--- a/api/core/v1alpha1/pd_types.go
+++ b/api/core/v1alpha1/pd_types.go
@@ -97,6 +97,7 @@ type PDList struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
 // +kubebuilder:resource:categories=instance
 // +kubebuilder:selectablefield:JSONPath=`.spec.cluster.name`
 // +kubebuilder:printcolumn:name="Cluster",type=string,JSONPath=`.spec.cluster.name`
@@ -238,6 +239,12 @@ type PDSpec struct {
 	// A same pd cluster will use a same subdomain
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="subdomain is immutable"
 	Subdomain string `json:"subdomain"`
+
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1
+	// +default:value=1
+	// +optional
+	Replicas *int32 `json:"replicas"`
 
 	// PDTemplateSpec embedded some fields managed by PDGroup
 	PDTemplateSpec `json:",inline"`
--- a/api/core/v1alpha1/resourcemanager_types.go
+++ b/api/core/v1alpha1/resourcemanager_types.go
@@ -80,6 +80,7 @@ type ResourceManagerList struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:categories=instance
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
 // +kubebuilder:selectablefield:JSONPath=`.spec.cluster.name`
 // +kubebuilder:printcolumn:name="Cluster",type=string,JSONPath=`.spec.cluster.name`
 // +kubebuilder:printcolumn:name="Synced",type=string,JSONPath=`.status.conditions[?(@.type=="Synced")].status`
@@ -196,6 +197,12 @@ type ResourceManagerSpec struct {
 	// A same ResourceManager cluster will use a same subdomain
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="subdomain is immutable"
 	Subdomain string `json:"subdomain"`
+
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1
+	// +default:value=1
+	// +optional
+	Replicas *int32 `json:"replicas"`
 
 	// ResourceManagerTemplateSpec embedded some fields managed by ResourceManagerGroup
 	ResourceManagerTemplateSpec `json:",inline"`
--- a/api/core/v1alpha1/router_types.go
+++ b/api/core/v1alpha1/router_types.go
@@ -79,6 +79,7 @@ type RouterList struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
 // +kubebuilder:resource:categories=instance
 // +kubebuilder:selectablefield:JSONPath=`.spec.cluster.name`
 // +kubebuilder:printcolumn:name="Cluster",type=string,JSONPath=`.spec.cluster.name`
@@ -193,6 +194,12 @@ type RouterSpec struct {
 	// A same Router cluster will use a same subdomain
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="subdomain is immutable"
 	Subdomain string `json:"subdomain"`
+
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1
+	// +default:value=1
+	// +optional
+	Replicas *int32 `json:"replicas"`
 
 	// RouterTemplateSpec embedded some fields managed by RouterGroup
 	RouterTemplateSpec `json:",inline"`
--- a/api/core/v1alpha1/scheduler_types.go
+++ b/api/core/v1alpha1/scheduler_types.go
@@ -85,6 +85,7 @@ type SchedulerList struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
 // +kubebuilder:resource:categories=instance
 // +kubebuilder:selectablefield:JSONPath=`.spec.cluster.name`
 // +kubebuilder:printcolumn:name="Cluster",type=string,JSONPath=`.spec.cluster.name`
@@ -209,6 +210,12 @@ type SchedulerSpec struct {
 	// A same Scheduler cluster will use a same subdomain
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="subdomain is immutable"
 	Subdomain string `json:"subdomain"`
+
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1
+	// +default:value=1
+	// +optional
+	Replicas *int32 `json:"replicas"`
 
 	// SchedulerTemplateSpec embedded some fields managed by SchedulerGroup
 	SchedulerTemplateSpec `json:",inline"`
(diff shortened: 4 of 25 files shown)


liubog2008:
It seems not reasonable to add `/scale` for instance crds. Could you comment your PDB spec?

SuperQ:
Sure, the PDB is pretty simple / standard.

```yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: test-tikv
  namespace: test
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      app.kubernetes.io/component: tikv
      app.kubernetes.io/instance: test
```

My understanding is that the PodDisruptionBudget acts on the Pod resource and looks for the Pod's owner.

So, this example:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: test-tikv-326vs
  namespace: test
  labels:
    app.kubernetes.io/component: tikv
    app.kubernetes.io/instance: test
    ...
  ownerReferences:
  - apiVersion: core.pingcap.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: TiKV
    name: test-tikv
    uid: ...
```

So when Kubernetes tries to gather all the _Pods_ matching the label spec it looks at the owner references and finds that TiKV is the owner, no mention of TiKVGroup.

So the way things are now, the PDB can't work.

liubog2008:
https://kubernetes.io/docs/tasks/run-application/configure-pdb/#arbitrary-controllers-and-selectors

It seems that `maxUnavailable` is not supported for current implementation. I'll dig more about this issue.

Anyway, this PR cannot fix the issue.

SuperQ:
Oof, that's an unfortunate restriction.

So the only real option is that the TiDB Operator will need to create pods via an indirect method, for example each Pod needs to have an individual StatefulSet.

SuperQ:
If you would like, I can file a tracking issue detailing the need here.

liubog2008:
Now operator manages `TIKVGroup -> TiKV -> Pod` ownership. However, PDB cannot find TiKVGroup from pod's owner. It can be enhanced by kubernetes easily. But now, operator v2 cannot support PDB.

liubog2008:
What we can do now is only adding the leader eviction scheduler to one pod even if there are more than one pod is deleting.

SuperQ:
What needs to be done is have an intermediate Pod controller like this:

```
TIKVGroup -> TiKV -> StatefulSet -> Pod
```

Happy to help implement this.

liubog2008:
> What needs to be done is have an intermediate Pod controller like this:

> TIKVGroup -> TiKV -> StatefulSet -> Pod
> Happy to help implement this.

It's useless. PDB needs to get replicas from pod's owner.

SuperQ:
It does work, I have filed https://github.com/pingcap/tidb-operator/issues/6901 with some details on how multiple StatefulSets can be combined into a single PDB.

In this new design, the Pod's owner is the StatefulSet, which supports maxUnavailable. This is how and why we implemented this in the [valkey-operator](https://github.com/valkey-io/valkey-operator/blob/main/docs/valkeynode-design.md).

SuperQ:
Now that we have the details we need in the issue, I am going to close this PR.

api/core/v1alpha1/pd_types.go
```diff
@@ -97,6 +97,7 @@ type PDList struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=.status.selector
```
liubog2008: Remove selectorpath. It's useless.

liubog2008: A new field Replicas should be actually added into `Spec` and `Status`

liubog2008:
I have verified the PDB in my own env.

```
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"policy/v1","kind":"PodDisruptionBudget","metadata":{"annotations":{},"name":"tikv-pdb","namespace":"default"},"spec":{"maxUnavailable":1,"selector":{"matchLabels":{"pingcap.com/component":"tikv"}}}}
  creationTimestamp: "2026-05-20T01:47:48Z"
  generation: 1
  name: tikv-pdb
  namespace: default
  resourceVersion: "12004810"
  uid: 5d6d5d3a-ec36-4d9a-904b-1a8c89ec8cbd
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      pingcap.com/component: tikv
status:
  conditions:
  - lastTransitionTime: "2026-05-20T01:47:48Z"
    message: ""
    observedGeneration: 1
    reason: SufficientPods
    status: "True"
    type: DisruptionAllowed
  currentHealthy: 3
  desiredHealthy: 2
  disruptionsAllowed: 1
  expectedPods: 3
  observedGeneration: 1
```

There are still some issues of this PR:

- Only `spec.replicas` path is needed. Both status and selector are useless. (But controller tools need status path to generate crd)
- New field `spec.replicas` should be added.
- `spec.replicas` should be optional, default to 1 and cannot be changed.

SuperQ:
Updated. I think we can probably remove the Scale subresource from the `FooGroupSpec`, since they don't directly create Pods.

api/core/v1alpha1/pd_types.go
```diff
@@ -60,7 +60,7 @@ type PDGroupList struct {
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas,selectorpath=.status.selector
+// +kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
```
liubog2008: Not here. It's Group's scale

api/core/v1alpha1/pd_types.go
```diff
@@ -239,13 +240,22 @@ type PDSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="subdomain is immutable"
 	Subdomain string `json:"subdomain"`
 
+	// +kubebuilder:validation:Minimum=0
```
liubog2008: Minimum should also be 1

api/core/v1alpha1/pd_types.go
```diff
@@ -239,13 +240,22 @@ type PDSpec struct {
...
+
 	// PDTemplateSpec embedded some fields managed by PDGroup
 	PDTemplateSpec `json:",inline"`
 }
 
 type PDStatus struct {
 	CommonStatus `json:",inline"`
 
+	// Replicas is the number of replicas currently available.
+	Replicas *int32 `json:"replicas,omitempty"`
```
liubog2008: It's useless, remove it

liubog2008:
@SuperQ I left some comments.

> Updated. I think we can probably remove the Scale subresource from the FooGroupSpec, since they don't directly create Pods.

The scale of group(e.g. TiKVGroup) is for HPA and VPA. It has been used. The scale of instance(e.g. TiKV) is only for compatibility of PDB

api/core/v1alpha1/pd_types.go
SuperQ: Oh, I thought we needed the Status Replicas so the PDB could count them. I guess not.

api/core/v1alpha1/pd_types.go
SuperQ: Ahh, I confused this comment, I thought you wanted `Replicas` on the `FooStatus` struct as well.

api/core/v1alpha1/pd_types.go
```diff
@@ -239,6 +240,12 @@ type PDSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="subdomain is immutable"
 	Subdomain string `json:"subdomain"`
 
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=1
+	// +default:value=1
+	// +optional
+	Replicas *int32 `json:"replicas"`
```
liubog2008, line 247: Add a comment about this field. "It's only designed for compatibility with PDB and cannot be changed"

liubog2008, line 247: add omitempty in json

liubog2008:
Still two comments are left. Thanks @SuperQ

SuperQ:
Should we also add a kubebuilder validation like:

```
  // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="replicas is immutable"
```

SuperQ:
Bah! I didn't mean to close this PR.

SuperQ:
Opened https://github.com/pingcap/tidb-operator/pull/6912

PR #6883, opened 2026-05-09
feat: add DM (Data Migration) support

tennix:
## Summary

- Add CRD API types for `DMGroup`, `DM`, `DMWorkerGroup`, `DMWorker` with TLS support from the start
- Add runtime/scope types and config packages (`pkg/configs/dm`, `pkg/configs/dmworker`) for dm-master and dm-worker
- Add four controllers following the Group+Instance pattern:
  - **DMGroup**: manages dm-master replicas with embedded etcd bootstrap, headless + internal service, rolling update
  - **DM**: dm-master instance controller with health check (`GET /api/v1/cluster/info`), ConfigMap, DataVolume PVC, pod lifecycle
  - **DMWorkerGroup**: manages dm-worker replicas with headless service, `maxSurge=0/maxUnavailable=1` for pod name stability
  - **DMWorker**: dm-worker instance controller with health check (`GET /status`), ConfigMap, RelayVolume PVC, pod lifecycle
- Register all controllers in `cmd/tidb-operator/main.go`
- Add example manifests `examples/basic/07-dm.yaml` and `examples/basic/08-dmworker.yaml`

## Architecture notes

- **Bootstrap**: dm-master uses `AnnoKeyInitialClusterNum` annotation on instances to compute `initial-cluster`; `TaskBoot` removes the annotation once all initial instances are ready. Bootstrap state is derived from existing instances (absence of annotation = bootstrapped).
- **dm-worker join**: `join` address is derived from `<DMGroupRef.Name>-dm-master` internal service at the default DM port (8261).
- **Pod name stability**: DMWorker uses `maxSurge=0, maxUnavailable=1` to preserve pod names required by dm-master's `lastBound` etcd history.
- **No PD dependency**: dm-master uses embedded etcd only; no PDClientManager needed.

### Manual test

- [x] Apply `examples/basic/07-dm.yaml` — verify DMGroup creates headless + internal service and DM pods
- [x] Verify dm-master bootstrap: `AnnoKeyInitialClusterNum` annotation removed after all initial replicas ready
- [x] Apply `examples/basic/08-dmworker.yaml` — verify DMWorkerGroup creates headless service and DMWorker pods
- [x] Verify dm-worker joins dm-master via `<dmgroup>-dm-master` service
- [x] Scale DMGroup replicas 1→3 — verify join mode used for new instances (no boot annotation)
- [x] Scale DMWorkerGroup replicas 1→3 — verify rolling update with maxUnavailable=1
- [x] Verify TLS: enable `spec.tlsCluster` on Cluster, confirm TLS mounts appear in pods
- [x] Delete DMGroup — verify all DM instances and services are cleaned up

### E2E & Unit test

Both unit test and e2e test cases are included

🤖 Generated with [Claude Code](https://claude.com/claude-code)


api/core/v1alpha1/dmworker_types.go
```diff
@@ -0,0 +1,226 @@
...
+	UpdateStrategy UpdateStrategy `json:"updateStrategy,omitempty"`
+	// Config defines the config file of DM worker
+	Config ConfigFile `json:"config,omitempty"`
+
+	// Security defines security config
+	Security *Security `json:"security,omitempty"`
+
+	// RelayVolume defines the persistent volume for DM worker's relay log storage.
+	// This volume is required for dm-worker to persist relay logs.
+	RelayVolume Volume `json:"relayVolume"`
```
xuanyu66: @GMHDBJD If we haven't enabled the relay log, does this disk not need to be mounted? Please confirm. If that's the case, this field should be changed to optional.

GMHDBJD: no need

xuanyu66: @tennix Maybe we should make this field optional

xuanyu66: The others LGTM

xuanyu66 (approved):

pkg/controllers/dmworker/tasks/cm.go
```diff
@@ -0,0 +1,78 @@
...
+		cfg := dmworkercfg.Config{}
+		decoder, encoder := toml.Codec[dmworkercfg.Config]()
+		if err := decoder.Decode([]byte(state.DMWorker().Spec.Config), &cfg); err != nil {
+			return task.Fail().With("dm-worker config cannot be decoded: %v", err)
+		}
+
+		dw := state.DMWorker()
+		dmMasterSvcName := fmt.Sprintf("%s-dm-master", dw.Spec.DMGroupRef.Name)
+		dmMasterAddr := coreutil.ServiceHost(state.Cluster(), dmMasterSvcName) +
+			":" + strconv.Itoa(int(v1alpha1.DefaultDMPort))
```
fgksgf, line 46: This hard-codes the DM master join port to `DefaultDMPort`, but the referenced `DMGroup` service is created with `coreutil.DMGroupPort(dmg)` and the API allows users to customize `spec.template.spec.server.ports.port`. With a non-default DM master port, DMWorkers will still render `join = ...:8261` and fail to register.

Please resolve the referenced `DMGroup` and use its configured port here, or reject/custom-validate such configurations so the Service port and worker config cannot diverge.

- HuaxiClaw

pkg/controllers/dm/tasks/pod.go
```diff
@@ -0,0 +1,209 @@
...
+
+	dataMountPath := v1alpha1.VolumeMountDMDataDefaultPath
+	for i := range dm.Spec.DataVolume.Mounts {
+		mount := &dm.Spec.DataVolume.Mounts[i]
+		if mount.Type == v1alpha1.VolumeMountTypeDMData && mount.MountPath != "" {
+			dataMountPath = mount.MountPath
+			break
+		}
+	}
+	mounts = append(mounts, corev1.VolumeMount{
```
fgksgf, line 114: The built-in DM data mount preserves only `mountPath` and drops `subPath`, while the same `VolumeMount` API supports `subPath` and the additional-volume path below passes it through. If a user configures `dataVolume.mounts[]` with `type: data` and `subPath`, the rendered Pod silently ignores it.

Please carry the selected mount's `SubPath` into this `corev1.VolumeMount` as well.

- HuaxiClaw

pkg/controllers/dmworker/tasks/pod.go
```diff
@@ -0,0 +1,204 @@
...
+
+		relayMountPath := v1alpha1.VolumeMountDMWorkerRelayDefaultPath
+		for i := range relayVol.Mounts {
+			mount := &relayVol.Mounts[i]
+			if mount.Type == v1alpha1.VolumeMountTypeDMWorkerRelay && mount.MountPath != "" {
+				relayMountPath = mount.MountPath
+				break
+			}
+		}
+		mounts = append(mounts, corev1.VolumeMount{
```
fgksgf, line 115: Same issue as the DM data volume path: the built-in relay mount only keeps `mountPath` and silently drops `subPath`, even though `relayVolume.mounts[]` uses the shared `VolumeMount` type and additional volumes preserve `SubPath` below.

Please propagate the selected relay mount's `SubPath` into this `corev1.VolumeMount`.

- HuaxiClaw

fgksgf (changes requested):
## Code Review Summary

**Verdict: Changes requested**

I found a few correctness issues in the new DM / DMWorker controllers that should be addressed before merge. The main concerns are that DMWorker config generation ignores customized DM master ports, and built-in volume mounts silently drop `subPath` while the same API supports it for additional volumes.

Validation: targeted Go tests for the new DM/DMWorker config and controller packages passed locally:

```text
go test ./pkg/configs/dm ./pkg/configs/dmworker ./pkg/controllers/dm/... ./pkg/controllers/dmgroup/... ./pkg/controllers/dmworker/... ./pkg/controllers/dmworkergroup/... -count=1
```

Note: `go test ./tests/validation` is not runnable from the v2 module in this checkout because that package is outside the main module.

- HuaxiClaw

fgksgf:
Two additional issues not yet covered:

**1. Duplicate "Ready" print column names (dm_types.go:59,63 / dmworker_types.go:64,68)**

Both `DMGroup` and `DMWorkerGroup` have two `printcolumn` entries with the same name `"Ready"`:
```
// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.readyReplicas`
// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
```
kubectl will render two columns with identical headers. Rename the first to `ReadyReplicas` (or similar) in both CRDs.

**2. Hardcoded service name convention (dmworker/tasks/cm.go:43)**

```go
dmMasterSvcName := fmt.Sprintf("%s-dm-master", dw.Spec.DMGroupRef.Name)
```

This duplicates the naming convention from `InternalServiceName()` in `pkg/controllers/dmgroup/tasks/util.go`. If the convention changes, this will silently break. Should import and reuse the shared function, or move the naming logic to a shared package like `apiutil/core`.

pkg/controllers/dm/tasks/ctx.go
```diff
@@ -0,0 +1,94 @@
...
+
+		var info struct {
+			Name string `json:"name"`
+		}
+		if err := json.Unmarshal(body, &info); err != nil {
+			return task.Complete().With("cannot parse dm-master cluster info response: %v", err)
+		}
+
+		state.SetHealthy()
+		state.MemberID = info.Name
```
fgksgf: This parses `name` from `/api/v1/cluster/info`, but the DM OpenAPI response for that endpoint is `GetClusterInfoResponse` with `cluster_id` and `topology`, not a `name` field: https://pkg.go.dev/github.com/pingcap/tiflow/dm/openapi#GetClusterInfoResponse. `json.Unmarshal` will still succeed, so the task marks the instance healthy while `state.MemberID` stays empty and `DM.Status.MemberID` is never populated.

If the status is meant to track the dm-master member ID/name, this should call an endpoint that returns master members or drop the `MemberID` update. Otherwise this field gives a false signal that the controller synced member identity.

- HuaxiClaw

pkg/controllers/dmworker/tasks/cm.go
```diff
@@ -0,0 +1,83 @@
...
+		dw := state.DMWorker()
+
+		dmg := &v1alpha1.DMGroup{}
+		if err := c.Get(ctx, client.ObjectKey{Name: dw.Spec.DMGroupRef.Name, Namespace: dw.Namespace}, dmg); err != nil {
+			return task.Fail().With("cannot get dmgroup for dm-worker config: %v", err)
+		}
+
+		dmMasterSvcName := coreutil.DMGroupInternalServiceName(dw.Spec.DMGroupRef.Name)
+		dmMasterAddr := coreutil.ServiceHost(state.Cluster(), dmMasterSvcName) +
+			":" + strconv.Itoa(int(coreutil.DMGroupPort(dmg)))
```
fgksgf: This dependency still needs a defined consistency contract. The worker config now resolves the referenced `DMGroup` and renders `join` with `DMGroupPort(dmg)`, while the DMGroup internal Service is also created with `DMGroupPort(dmg)`. However, the `DMWorker` controller only watches `DMWorker`, owned resources, and `Cluster` generation changes. If `spec.template.spec.server.ports.port` on the referenced DMGroup changes, the Service can be reconciled to the new port while existing DMWorker ConfigMaps keep the old `join` until another event happens to enqueue them.

The existing cross-component refs I found avoid this by keeping the internal Service port stable (for example TiKVWorker/TSO use the default internal service port and `targetPort` by name), or by propagating the dependency through Cluster status that consumers already watch (PD URL). Please either make the DM master port immutable/validated, keep the internal DM master service port stable, or add a DMGroup watch/index for DMWorkers that reference it.

- HuaxiClaw

fgksgf (commented):
Thanks for the update. The previous comments about the DM master port, built-in volume `subPath`, duplicate print column names, duplicated service-name convention, and the invalid `memberID` status update are addressed.

One new issue remains: DMWorker config generation now depends on the referenced `DMGroup`, but the DMWorker controller does not watch that object, so DMGroup port changes will not reliably reconcile existing DMWorkers. I left an inline comment.

Validation performed locally:

```text
go test ./pkg/configs/dm ./pkg/configs/dmworker ./pkg/controllers/dm/... ./pkg/controllers/dmgroup/... ./pkg/controllers/dmworker/... ./pkg/controllers/dmworkergroup/... -count=1
```

I also checked the regenerated DMGroup/DMWorkerGroup CRDs and confirmed the group print columns now have only one `Ready` column plus `ReadyReplicas`.

- HuaxiClaw

pkg/controllers/dmgroup/tasks/updater.go
```diff
@@ -0,0 +1,164 @@
...
+const (
+	defaultUpdateWaitTime = time.Second * 30
+)
+
+func TaskUpdater(state *ReconcileContext, c client.Client, af tracker.AllocateFactory) task.Task {
+	return task.NameTaskFunc("Updater", func(ctx context.Context) task.Result {
+		logger := logr.FromContextOrDiscard(ctx)
+		dmg := state.DMGroup()
+
+		checker := action.NewUpgradeChecker[scope.DMGroup](c, state.Cluster(), logger)
```
fgksgf, line 48: This wires DMGroup version upgrades through the generic UpgradeChecker, but the default upgrade policy does not know the dm-master component. For clusters with the default/empty upgradePolicy, NewDefault reaches the default branch in pkg/action/upgrader.go (unknown component: dm-master), CanUpgrade returns false, and this task keeps retrying before it ever calls the updater. A DM version bump therefore cannot roll out once status.version is set.

The RFC says DM master has no PD dependency and does not need an UpgradeChecker, so please either skip this checker for DMGroup or add dm-master/dm-worker to the default policy as no-constraint components. The updater tests should assert that a version change can proceed instead of expecting SRetry.

- HuaxiClaw

fgksgf (approved):
The previous DMWorker port consistency issue is addressed by keeping the internal DM master service port stable, and the DM/DMWorker version-upgrade gate is addressed by adding dm-master/dm-worker to the default no-precondition branch with updated tests.

I rechecked the incremental diff from c00e9ce29 to f3e876ed8 and ran the relevant local tests:

```bash
go test ./pkg/configs/dm ./pkg/configs/dmworker ./pkg/controllers/dm/... ./pkg/controllers/dmgroup/... ./pkg/controllers/dmworker/... ./pkg/controllers/dmworkergroup/... ./pkg/action -count=1
```

The current pull-e2e failure is still real, but the failing specs are TiDB/TiKV label/leader-eviction cases and do not look related to this DM change.

- HuaxiClaw

PR #6904, opened 2026-05-18
Pause DDL during eligible TiDB upgrades

tennix:
## What is changed

This PR adds TiDB smooth-upgrade coordination to TiDB Operator v1 classic TiDB upgrades on `release-1.x`:

- call TiDB `POST /upgrade/start` before eligible TiDB rolling version upgrades begin
- wait for the `200` start response before allowing TiDB pods to roll
- persist the active pause window in `TidbCluster` annotations
- call `POST /upgrade/finish` after the upgraded TiDB StatefulSet is fully updated, Normal, and members are healthy
- clear smooth-upgrade annotations only after finish succeeds
- skip unsupported and auto-supported smooth-upgrade version pairs

## Design notes

The implementation follows `docs/design-proposals/2026-05-14-pause-ddl-during-tidb-upgrade.md`.

Important behavior:

- `/upgrade/start` uses a 30s client timeout because TiDB waits up to 10s for the DDL owner to sync upgrading state before returning success.
- TiDB's success responses are JSON string bodies, so the client decodes them before matching `success!` and duplicate-success responses.
- Active pause state is stored in annotations:
  - `tidb.pingcap.com/smooth-upgrade-ddl-paused`
  - `tidb.pingcap.com/smooth-upgrade-source-version`
  - `tidb.pingcap.com/smooth-upgrade-target-version`
  - `tidb.pingcap.com/smooth-upgrade-started-at`
- Stale/mismatched annotations are handled by idempotently calling `/upgrade/finish`, deleting the annotation keys, and requeueing.

## Tests

```text
PASS git diff --check
PASS go test ./pkg/controller ./pkg/manager/member ./tests/e2e/util/proxiedtidbclient -count=1
```

### DIFF
--- a/pkg/manager/member/smooth_upgrade.go
+++ b/pkg/manager/member/smooth_upgrade.go
@@ -1,0 +1,276 @@
+// Copyright 2026 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package member
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/pingcap/tidb-operator/pkg/apis/pingcap/v1alpha1"
+	"github.com/pingcap/tidb-operator/pkg/controller"
+	"github.com/pingcap/tidb-operator/pkg/util/cmpver"
+
+	apps "k8s.io/api/apps/v1"
+	"k8s.io/klog/v2"
+)
+
+type smoothUpgradeSupport string
+
+const (
+	smoothUpgradeUnsupported      smoothUpgradeSupport = "Unsupported"
+	smoothUpgradeAutoSupported    smoothUpgradeSupport = "AutoSupportedNoop"
+	smoothUpgradeSwitchControlled smoothUpgradeSupport = "SwitchControlled"
+
+	annSmoothUpgradeDDLPaused     = "tidb.pingcap.com/smooth-upgrade-ddl-paused"
+	annSmoothUpgradeSourceVersion = "tidb.pingcap.com/smooth-upgrade-source-version"
+	annSmoothUpgradeTargetVersion = "tidb.pingcap.com/smooth-upgrade-target-version"
+	annSmoothUpgradeStartedAt     = "tidb.pingcap.com/smooth-upgrade-started-at"
+)
+
+func detectTiDBVersionUpgrade(oldSet, newSet *apps.StatefulSet) (string, string, bool) {
+	oldImage, ok := mainContainerImage(oldSet, v1alpha1.TiDBMemberType.String())
+	if !ok {
+		return "", "", false
+	}
+	newImage, ok := mainContainerImage(newSet, v1alpha1.TiDBMemberType.String())
+	if !ok {
+		return "", "", false
+	}
+	source := imageTag(oldImage)
+	target := imageTag(newImage)
+	return source, target, source != "" && target != "" && source != target
+}
+
+func classifySmoothUpgrade(source, target string) smoothUpgradeSupport {
+	if !validTiDBVersion(source) || !validTiDBVersion(target) {
+		return smoothUpgradeUnsupported
+	}
+
+	sourceBefore710, _ := cmpver.Compare(source, cmpver.Less, "v7.1.0")
+	if sourceBefore710 {
+		return smoothUpgradeUnsupported
+	}
+
+	if isAutoSmoothUpgradePair(source, target) {
+		return smoothUpgradeAutoSupported
+	}
+	if isSwitchControlledSmoothUpgradePair(source, target) {
+		return smoothUpgradeSwitchControlled
+	}
+	return smoothUpgradeUnsupported
+}
+
+func isAutoSmoothUpgradePair(source, target string) bool {
+	return (versionEqual(source, "v7.1.0") && (versionEqual(target, "v7.1.1") || versionEqual(target, "v7.2.0") || versionEqual(target, "v7.3.0"))) ||
+		(versionEqual(source, "v7.1.1") && (versionEqual(target, "v7.2.0") || versionEqual(target, "v7.3.0"))) ||
+		(versionEqual(source, "v7.2.0") && versionEqual(target, "v7.3.0"))
+}
+
+func isSwitchControlledSmoothUpgradePair(source, target string) bool {
+	sourceIn712To720 := versionGTE(source, "v7.1.2") && versionLT(source, "v7.2.0")
+	targetIn712To720 := versionGTE(target, "v7.1.2") && versionLT(target, "v7.2.0")
+	sourceGTE740 := versionGTE(source, "v7.4.0")
+	targetGTE740 := versionGTE(target, "v7.4.0")
+	return (sourceIn712To720 && targetIn712To720) || ((sourceIn712To720 || sourceGTE740) && targetGTE740)
+}
+
+func versionEqual(version, target string) bool {
+	gte, err := cmpver.Compare(version, cmpver.GreaterOrEqual, target)
+	if err != nil || !gte {
+		return false
+	}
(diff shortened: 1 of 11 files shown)


pkg/manager/member/smooth_upgrade.go
```diff
@@ -0,0 +1,295 @@
...
+		(versionEqual(source, "v7.1.1") && (versionEqual(target, "v7.2.0") || versionEqual(target, "v7.3.0"))) ||
+		(versionEqual(source, "v7.2.0") && versionEqual(target, "v7.3.0"))
+}
+
+func isSwitchControlledSmoothUpgradePair(source, target string) bool {
+	sourceIn712To720 := versionGTE(source, "v7.1.2") && versionLT(source, "v7.2.0")
+	targetIn712To720 := versionGTE(target, "v7.1.2") && versionLT(target, "v7.2.0")
+	sourceGTE740 := versionGTE(source, "v7.4.0")
+	targetGTE740 := versionGTE(target, "v7.4.0")
+	return (sourceIn712To720 && targetIn712To720) || ((sourceIn712To720 || sourceGTE740) && targetGTE740)
```
fgksgf, line 93: This classifier checks only whether both versions fall into supported ranges, but it does not require `target > source`. Since `detectTiDBVersionUpgrade` only checks that the image tag changed, rollback/downgrade pairs such as `v7.5.0 -> v7.4.0` or `v7.1.3 -> v7.1.2` can be treated as switch-controlled smooth upgrades.

That would call `/upgrade/start` before rolling to the older image and later call `/upgrade/finish` against the target version, which may not support or expect this flow. Please explicitly skip non-increasing version changes before classifying auto/switch-controlled smooth upgrades.

- HuaxiClaw

pkg/manager/member/smooth_upgrade.go
```diff
@@ -0,0 +1,295 @@
...
+	if support := classifySmoothUpgrade(source, target); support != smoothUpgradeSwitchControlled {
+		klog.Warningf("TidbCluster: [%s/%s] TiDB smooth upgrade is %s for version pair %s -> %s, skip DDL pause API", tc.Namespace, tc.Name, support, source, target)
+		return nil
+	}
+
+	ordinal, err := chooseHealthyTiDBOrdinal(tc)
+	if err != nil {
+		return err
+	}
+	if err := u.deps.TiDBControl.StartUpgrade(tc, ordinal); err != nil {
```
fgksgf, line 216: `/upgrade/start` is executed before the durable annotation marker is persisted. If `StartUpgrade` succeeds but `patchTidbClusterAnnotations` fails, the reconcile returns an error with TiDB already in upgrade mode and no persisted state indicating that `/upgrade/finish` is owed. On the next reconcile, `isSmoothUpgradePaused(tc)` is still false, so the controller retries `/upgrade/start` instead of recovering/finishing the already-started pause.

Please make this sequence recoverable, e.g. persist a durable intent before calling the TiDB API, or add explicit duplicate-start/no-annotation recovery that can call `/upgrade/finish` and abort instead of treating the duplicate start as a normal successful start.

- HuaxiClaw

fgksgf (changes requested):
## Code Review Summary

**Verdict: Changes requested**

The overall smooth-upgrade flow is well covered for the happy path, but I found two recovery/correctness issues that can leave TiDB in upgrade mode unexpectedly: reverse/downgrade version changes can be classified as switch-controlled smooth upgrades, and `/upgrade/start` is called before the durable annotation marker is persisted.

Validation: the targeted test command from the PR description passes locally:

```text
go test ./pkg/controller ./pkg/manager/member ./tests/e2e/util/proxiedtidbclient -count=1
```

- HuaxiClaw

pkg/manager/member/smooth_upgrade.go
tennix: tidb does not support version downgrade, so I think it's unnecessary to handle cluster downgrade

fgksgf:
## Additional issues

**1. Error format string argument order** (`pkg/controller/tidb_control.go:397,399`)

```go
return fmt.Errorf("tidb upgrade %s error response %q:%v URL: %s", op, bodyText, res.StatusCode, url)
```

Format specifiers `%q:%v` map to `bodyText, res.StatusCode`, producing `"body":400`. HTTP convention is status before body. Suggested fix:

```go
return fmt.Errorf("tidb upgrade %s error response %d %q URL: %s", op, res.StatusCode, bodyText, url)
```

Same issue on line 399 (unexpected response).

**2. Missing unit test for stale/mismatched annotation recovery path**

`ensureSmoothUpgradeStarted` (smooth_upgrade.go:715-729) handles stale annotations: finish stale upgrade, clear annotations, requeue. This is the most complex recovery scenario but has no unit test. `TestTiDBUpgraderSmoothUpgradeStart` only tests the happy path and the already-active-annotation skip.

**3. Annotation constants duplicated in e2e test without sync guard**

`tests/e2e/tidbcluster/tidbcluster.go:3523-3527` re-declares the 3 annotation key constants from `smooth_upgrade.go`. If keys change in the implementation, e2e tests silently break against stale keys. Either export the constants or add a cross-reference comment in both locations.

**4. `imageTag()` edge cases untested**

`imageTag()` handles digest references (`@sha256:...`), missing tags, and empty tags. `smooth_upgrade_test.go` only tests it indirectly via `detectTiDBVersionUpgrade` with valid images. Missing direct coverage for:
- `pingcap/tidb@sha256:abc` -> `""`
- `pingcap/tidb` (no tag) -> `""`
- `pingcap/tidb:` (empty tag) -> `""`

- HuaxiClaw

pkg/manager/member/smooth_upgrade.go
tennix: The current code is self-correcting via `StartUpgrade`'s idempotency. If `patchTidbClusterAnnotations` fails after a successful `StartUpgrade`, the next reconcile finds `isSmoothUpgradePaused` false, calls `StartUpgrade` again, and TiDB returns HTTP 200 with the duplicate-upgrading body (`"It's a duplicated operation and the cluster is already in upgrading state."`). `tidb_control.go` treats that as a nil error (same path as a fresh success), so annotations are set and the patch is retried. The loop converges whenever the patch succeeds.

The design doc also covers this explicitly: *"Duplicate start/finish HTTP 200 responses are success and should not block recovery after client timeout or operator restart."*

A write-ahead annotation before the API call would not change the convergence behaviour — it would just add another annotation key and another round-trip. Happy to add it if you think the clarity gain is worth it, but the current flow is correct.

pkg/manager/member/smooth_upgrade.go
tennix: Fixed. Added a `target <= source` guard in `classifySmoothUpgrade` that returns `Unsupported` before the auto/switch-controlled checks, so downgrade pairs like `v7.5.0 -> v7.4.0` and `v7.1.3 -> v7.1.2` are skipped. Two test cases added to `TestSmoothUpgradeMatrix` to cover this.

pkg/manager/member/smooth_upgrade.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if isSmoothUpgradePaused(tc) {
+		if smoothUpgradeAnnotationsMatch(tc, source, target) {
+			return nil
+		}
+		ordinal, err := chooseHealthyTiDBOrdinal(tc)
+		if err != nil {
+			return err
+		}
+		if err := u.deps.TiDBControl.FinishUpgrade(tc, ordinal); err != nil {
```
fgksgf, line 199: This mismatch branch treats every active annotation that does not match the current old/new StatefulSet image pair as stale and calls `/upgrade/finish` immediately. That is unsafe if the user changes `spec.version` while a smooth-upgrade rolling update is still in progress: the annotation may describe the first target, while `newSet` now describes the second target. This code would finish and clear the pause before the first rolling update has completed, briefly allowing user DDL in a mixed-version cluster before the next reconcile starts the new pause.

Please distinguish stale state from an active pause for an in-progress previous target. For example, keep the existing annotation and block/defer the new version change until the annotated target has finished, or only run this recovery after verifying the cluster is not still rolling.

- HuaxiClaw

fgksgf (approved):
LGTM. The downgrade/rollback classification issue has been addressed, and the duplicate `/upgrade/start` recovery path is acceptable given TiDB's idempotent duplicate-start response and the retrying annotation patch.

Validated locally:

```text
git diff --check
go test ./pkg/controller ./pkg/manager/member ./tests/e2e/util/proxiedtidbclient -count=1
```

I also checked the current failing Prow jobs; the failures are Docker Hub 429 pull-rate-limit errors while building e2e images, not caused by this PR.

- HuaxiClaw

PR #6925, opened 2026-06-03
fix: preserve legacy DM app label

xuanyu66:
## Summary
- Preserve `app.kubernetes.io/name=dm-cluster` on DM and DMWorker pods for legacy monitoring collectors.
- Add shared k8s app label constants and focused pod label assertions.

## Test plan
- [x] `go test ./pkg/controllers/dm/tasks ./pkg/controllers/dmworker/tasks ./pkg/utils/k8s`
- [x] IDE lints checked for edited files

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/utils/k8s/pod.go
+++ b/pkg/utils/k8s/pod.go
@@ -23,6 +23,12 @@ import (
 	"github.com/pingcap/tidb-operator/api/v2/core/v1alpha1"
 )
 
+const (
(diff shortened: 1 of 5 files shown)


tennix (approved):
LGTM

pkg/utils/k8s/pod.go
liubog2008, line 72: use component to determine label value in this function directly?

PR #6928, opened 2026-06-04
fix(tiproxy): sync graceful delay before rollout

YangKeao:
For `tiproxy-graceful-shutdown-delete-delay-seconds` annotation, it'd be better to sync it to the `TiProxy` CR before rolling restart them, so that the old `TiProxy` CR without this annotation can keep the behavior.


YangKeao:
It's a strange logic (I agree), so I'll close it for now.

PR #6954, opened 2026-06-17
Expose BR operation IDs in Backup and Restore status

RidRisR:

BR external-storage locks and long-running BR jobs can be hard to triage from the Backup/Restore CR alone. Operators often need the BR operation ID from pod logs before they can correlate a Kubernetes Backup or Restore with BR-side lock metadata and diagnostics.

This PR exposes the recent BR operation context on the owning CR so users can inspect the CR status first instead of hunting through backup-manager logs.

This PR adds a bounded `status.brOperations` list to both Backup and Restore status. Each observed operation records:

- `operationID`
- `command`
- `observedAt`
- `startedAt`, when BR emits it

The list is capped at the latest 10 operations. Re-observing the same operation ID refreshes the existing entry instead of appending a duplicate.

backup-manager now streams BR stdout and stderr through a shared line handler for the in-scope paths and parses real BR structured logs without forcing JSON log format. It supports:

- BR default text logs such as `["BR operation started"] [operation_id=...] [command=...]`
- JSON BR logs with `message` / `msg` / `Message`
- error/fatal detection from both legacy text `[ERROR]` lines and JSON `level=error/fatal` lines

Observed `BR operation started` events update the owning Backup or Restore status best-effort. Status update failures are logged but do not change the BR command result.

Current scope intentionally exposes only local BR operation history through `brOperations`; it does not add a separate remote lock-blocker status field.

The implementation covers:

- Backup log truncate BR command observation
- Restore BR command observation
- shared BR log parsing
- shared BR stdout/stderr line handling and stderr-preserving error output
- Backup and Restore status update helpers
- generated deepcopy, CRD schema, and API reference updates
- design doc and local E2E validation

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test
- [x] E2E test
- [ ] Manual test
- [ ] No code

Validation run:

```sh
env GOCACHE=/tmp/go-build go test -count=1 ./cmd/backup-manager/app/backup ./cmd/backup-manager/app/restore ./cmd/backup-manager/app/brlog ./cmd/backup-manager/app/util ./pkg/controller
./hack/verify-crd.sh
./hack/verify-openapi-spec.sh
git diff --check
```

Additional focused validation after preserving BR terminal log format:

```sh
env GOCACHE=/tmp/go-build go test -count=1 ./cmd/backup-manager/app/brlog ./cmd/backup-manager/app/backup ./cmd/backup-manager/app/restore ./cmd/backup-manager/app/util
env GOCACHE=/tmp/go-build go test -count=1 ./cmd/backup-manager/app/brlog ./cmd/backup-manager/app/backup ./cmd/backup-manager/app/restore ./cmd/backup-manager/app/util ./pkg/controller -run 'TestUpdateBackupStatus|TestUpdateRestoreStatus|TestUpdateBROperations|TestParseLine|TestIsErrorLine|TestBRCommandRun|TestBackupLogTruncateObserver|TestRestore'
```

Local k3s E2E validation was performed.

Notes:

- `verify-openapi-spec.sh` prints existing API rule warnings in this repository, but exits successfully and reports no generated OpenAPI diff.

### Side effects

- [ ] Other side effects:

### Related changes

The design doc is included in this PR.

### Release Notes

```release-note
Expose recent BR operation IDs in Backup and Restore status.
```


cmd/backup-manager/app/util/util.go
```diff
@@ -520,6 +522,130 @@ func ParseRestoreProgress(line string) (step, progress string) {
...
+)
+
+// BRLogLineHandler handles one line read from BR stdout or stderr.
+type BRLogLineHandler func(stream BRLogStream, line string)
+
+// RunBRCommandWithLineHandler runs BR and streams stdout/stderr lines to lineHandler.
+//
+// The returned error preserves legacy diagnostic behavior: stderr is included in
+// errMsg in full, while stdout contributes only BR error/fatal/panic lines.
+func RunBRCommandWithLineHandler(
```
YuJuncen, line 574: This adds a more general BR output reader, but `backup.brCommandRunWithLogCallback` still keeps a very similar stdout/stderr reader.

That can make the rules drift over time:

- some status fields are updated from stdout only
- some are updated from both stdout and stderr
- some are only updated when a specific runner is used

Could we move the old backup path to this helper too, or make this helper cover the old stdout-only callback mode? It would be easier to keep BR output reading and log-based status updates behind one entry point.

cmd/backup-manager/app/brlog/parser.go
```diff
@@ -0,0 +1,243 @@
...
+)
+
+// Event is one parsed BR structured-log event.
+type Event struct {
+	Type      EventType
+	Operation *v1alpha1.BROperation
+}
+
+// ParseLine extracts a BR operation or external-storage-lock event from one BR log line.
+func ParseLine(line string) Event {
```
YuJuncen: This adds a fairly complete BR text/json log parser, but the current use case is only to handle `"BR operation started"` and read `operation_id`.

This feels a bit heavy if the parser is not meant to be reused by other BR log readers.

Could we choose one direction here?

- If this is only for this event, use a narrower extractor. For example, first match `BR operation started`, then extract `operation_id`, `command`, and `operation_started_at` with small regexes.
- If we want to keep the parser, please define its boundary more clearly and make the field parsing contract general enough for future BR log parsing.

Leavrth (approved):

tennix (approved):

PR #6964, opened 2026-06-24
tiproxy: revive graceful scaling-in pods when scaling out TiProxy

djshow832:
### Background

Scaling in a TiProxy pod may take 24h. If the operator assigns a new pod when scaling out, there may be many extra pods in a frequent auto-scaling workload, which is a waste.
The goal is to revive graceful scaling in pods when scaling out with the best effort.

### Changes

The restarting/upgrading pods can't be revived, so the steps should be different from those of scaling in pods.

Graceful scale-in steps:
1. Set `spec.offline=true`. No `deletionTimestamp` because it may be reused.
2. Call the TiProxy API to mark it unhealthy.
3. Mark `graceful-shutdown-begin-time` on the pod.
4. Delete the CR if connections are drained or the delay reaches.

Scale-out steps:
1. Choose an updated pod to revive.
2. Set `spec.offline=false`.
3. Call the TiProxy API to clear the unhealthy status.
4. If clearing succeeds, clear `graceful-shutdown-begin-time`.
5. If clearing fails (health API 404), set `spec.offline=true` and mark the tiproxy with `tiproxy-revive-abandoned`.

Follow-ups in HPA:
- Skip the pods with `graceful-shutdown-begin-time` when calculating the average CPU/memory/network.

Notes:
- There may be a race between scale-out and deleting pods, and a deleted pod may be revived. It will be solved in reconciliation by recreating a new pod or reviving another pod.
- To avoid outdated TiProxy pods from being revived in a scale-in -> upgrade -> scale-out path, revival filters outdated TiProxy pods.
- The TiProxy pods that don't support the health API can't be revived. If so, abandon the revival and reset states.

Other refactors:
- Rename `IsStore()` to `SupportsOffline()`.
- Move some graceful restart logic into public files.

### E2E tests
- revive draining TiProxy pods on scale-out instead of creating new ones
- create new TiProxy pods on scale-out when drained instances are no longer revivable
- scale in and out TiProxy when health override API is unsupported


pkg/controllers/tiproxy/tasks/offline_scalein.go
```diff
@@ -0,0 +1,138 @@
...
+func TaskReviveFromScaleIn(state State, c client.Client) task.Task {
+	return task.NameTaskFunc("ReviveFromScaleIn", func(ctx context.Context) task.Result {
+		logger := logr.FromContextOrDiscard(ctx)
+		tiproxy := state.Object()
+		if coreutil.IsOffline[scope.TiProxy](tiproxy) {
+			return task.Complete().With("tiproxy is still marked offline")
+		}
+
+		pod := state.Pod()
+		if pod == nil || pod.Annotations[v1alpha1.AnnoKeyTiProxyGracefulShutdownBeginTime] == "" {
```
fgksgf, line 121: Blocking: this annotation check can permanently skip compensation if scale-out races with drain initialization. `drainPodForGracefulShutdown` calls `MarkUnhealthy` before persisting `AnnoKeyTiProxyGracefulShutdownBeginTime`. If that pod update fails/conflicts, or `spec.offline` is flipped back to `false` between those two operations, this branch returns `Complete` and the health override remains unhealthy; every later reconciliation takes the same branch, so the supposedly revived pod can remain NotReady indefinitely. Please persist a separate operator-owned drain/override intent before the external side effect, or otherwise make cancellation clear an operator-applied override even when the begin-time annotation is absent. Moving the begin-time annotation before `MarkUnhealthy` is not equivalent because HPA consumers treat it as the authoritative signal that draining has actually begun. Please also add a test for `offline=false` with an unhealthy override but no begin-time annotation.

fgksgf:
### Blocking

**`MarkUnhealthy` (PUT) has no `isHealthOverrideUnsupported` fallback, unlike `ClearHealthOverride` (DELETE)**

This PR establishes an "unsupported -> abandon" path for DELETE (`abandonRevive`), but the PUT side in `ensureTiProxyMarkedUnhealthy` has none. On a TiProxy build without health override:

- `IsHealthy()` returns true (GET exists on legacy builds)
- `MarkUnhealthy()` fails with 404/405
- `ensureTiProxyMarkedUnhealthy` returns false
- `drainPodForGracefulShutdown` returns `DefaultRequeueAfter` forever

The begin-time annotation is never written, the delete-delay clock never starts, the CR is never deleted — **scale-in hangs indefinitely** for anyone who opted in via `graceful-shutdown-delete-delay-seconds`.

This also means the e2e case `scale in and out TiProxy when health override API is unsupported` does not exercise what it claims: it only passes because `manuallyTriggerTiProxyUnhealthy` sends a signal to the pods, which makes `IsHealthy()` return false so `ensureTiProxyMarkedUnhealthy` takes the early-return branch. Production has no such external actor. Suggest detecting 404/405 on PUT and falling straight through to the delete-delay timer, then dropping the manual-signal step from that e2e so it actually covers the operator path.

### Should fix

**`TaskReviveFromScaleIn` returning `Retry` does not stop the rest of the pipeline**

`blockTask.sync` in `pkg/utils/task/v3/task.go` only breaks on `SFail`; `SRetry`/`SWait` fall through. So while revive is still retrying (health API flaky), `TaskConfigMap` / `TaskPod` / `TaskServerLabels` still run in the same reconcile. `TaskPod` builds `expected := newPod(...)`, which carries no drain annotation, then does `c.Apply(expected)` + `state.SetPod(expected)`.

`client.Apply` is SSA with `FieldManager: "tidb-operator"`, `Force: true`, while the annotation was written with `c.Update`. If both land under the same manager name, the apply takes ownership and drops the annotation — a **non-racy, deterministic** way to reach the permanently-unhealthy state described in the comment above. Wrapping the revive in `task.IfBreak` would close this independently of how the intent gets persisted.

**`TaskDeleteOfflinedTiProxy` deletes without a precondition**

The PR description acknowledges the scale-out/delete race. Adding `client.Preconditions{UID, ResourceVersion}` turns "a just-revived instance gets deleted" into a cheap conflict retry.

**`newTiProxyAPIClient` duplicates `TaskContextInfoFromTiProxy`**

`ctx.go` already constructs the client and stores it in `state.TiProxyClient`, and both the offline-drain branch and `TaskReviveFromScaleIn` run after it. Today each reconcile of a draining instance re-reads the TLS secret via `apicall.GetClientTLSConfig` two extra times (`ensureTiProxyMarkedUnhealthy` and `tiProxyConnectionsDrained` each build their own client). Reuse `state.TiProxyClient` in those two branches; keep `newTiProxyAPIClient` only for `TaskDrainPodForDelete`, which runs before the context task.

**`pkg/updater` hardcodes a TiProxy-specific annotation**

`FilterReviveAbandoned` in `selector.go` references `AnnoKeyTiProxyReviveAbandoned` directly. A generic `FilterAnnotationAbsent[R](key string)` with the key passed from `tiproxygroup/tasks/updater.go` keeps the generic updater unaware of TiProxy.

**`FilterPolicy`'s nil contract is unreachable**

The doc says "Return nil to abstain and keep the current allowed set", but both implementations return `[]R{}` on empty. The nil branch in `applyFilters` is dead, and leaving it invites a future `return nil` that silently no-ops the filter. Suggest dropping the nil semantics and letting `applyFilters` do `allowed = f.Filter(allowed)`.

**The offline branch swaps `tasks.TaskStatus` for `common.TaskStatusPersister`**

`tasks.TaskStatus` is what syncs `status.observedGeneration` and `status.updateRevision`. Setting `spec.offline=true` bumps `generation`, so for the whole drain window (up to 24h per the description) `observedGeneration` stays stale and `IsUpToDate()` is false. It recovers after revive, but the CR status is misleading to users and to anything reading those fields meanwhile. Intentional?

### Nits

- `TaskOfflineScaleInDrain`'s `pod == nil -> task.Wait()` is unreachable: `CondOfflineScaleInDrainComplete` already breaks on `pod == nil` earlier in the runner.
- `Instance.SupportsOffline` doc comment still says `(TiKV/TiFlash)`. TiProxy now returns true too, and its value is annotation-driven rather than static — worth calling out.
- `spec.offline` is user-writable but `TiProxyNewer` now sets `Offline: ptr.To(false)` and ships it through SSA, so the operator owns the field and manual edits get reverted. Please say so in the field comment (or add validation).
- `reviveTestHealthServer.port()` hand-rolls `host:port` parsing; `net.SplitHostPort(u.Host)` does it.
- Tests only cover 404 from `ClearHealthOverride`. 405 is arguably the likelier response from a build where the route exists but the method does not, and it is the other arm of `isHealthOverrideUnsupported`.
- `abandonRevive` sends a merge patch pinned to `resourceVersion`; the 409 path currently ends in `task.Fail()` and is untested. Worth confirming that is intended.
- In `TestExecutorScaleInUpdateMarksOfflineDuringRollingReplace`, the first `for _, name := range []string{...}` loop only issues Gets with no assertion — dead weight.

### Verified, no issue

- Switching `cancelOfflineSelector` from `beingOffline.List()[0]` to `NewSelector(PreferPriority)` is behavior-preserving for TiKV/TiFlash: `List()` is name-sorted and `PreferPriority` returns an empty set when no priority annotation is present, so `Choose` falls back to `allowed[0]`.
- `httputil.Errorf` returns `Error` by value, so `errors.As(err, &httpErr)` with the value type resolves.
- `finalizer_test.go` does not reference the renamed `drainOrDeletePod`, so the move does not break the build.

fgksgf:
Follow-up on the `SupportsOffline` / defer-delete change.

### The `defer-delete` check only covers deferred cleanup, so it is incomplete

`TiProxySupportsOfflineBeforeDelete` infers "this is a rolling restart, delete directly" from the presence of `AnnoKeyDeferDelete`. That signal does not carry the delete reason. `deleteInstance` has three call sites, and **only one of them writes that annotation beforehand**:

| call site | writes `defer-delete` first? |
|---|---|
| `Cleanup()` | yes (via `deferDelete`) |
| `scaleInOutdated(..., deferDel=false)` | **no** |
| `ScaleInUpdate()` | no |

So the heuristic covers at most one of three paths — and not the one that breaks.

**Concrete counterexample:**

1. One outdated TiProxy is not running.
2. The topology update policy picks a *different*, still-running outdated TiProxy. No difference in priority annotations is needed: `updatePolicies` appends `b.updatePreferPolicies` last (`builder.go:72-77`) and `Choose` walks `slices.Backward(s.ps)`, so the topology policy outranks `PreferNotRunning`.
3. `Update(ctx, true)` then hits `actor.go:319-329`:

```go
if unavailable && !chosen.IsNotRunning() {
    if act.noInPlaceUpdate {
        if _, err := act.scaleInOutdated(ctx, name, false); err != nil {  // deferDel=false
```

TiProxy rolling restart activates exactly this branch — `TaskUpdater` sets `noUpdate = true` when `needRestart`, which becomes `WithNoInPaceUpdate(true)`.

4. That path never adds `defer-delete`, yet still calls `deleteInstance`.
5. `SupportsOffline()` therefore returns true and the instance gets `offline=true`, violating the "rolling restart deletes directly" design and putting an outdated pod into a drain window that can be up to 24h.

### Suggested minimal fix: let the updater carry the delete reason explicitly

- Add `WithDirectDeleteOutdated(bool)` to `Builder`, default `false`.
- TiProxyGroup sets it to `true`.
- With it set, `scaleInOutdated(..., false)` and `Cleanup()` delete the CR directly.
- `ScaleInUpdate()` keeps offline-before-delete, so genuine scale-in still gets the graceful drain and stays revivable.
- TiKV/TiFlash do not set it — `Offline` and `CancelOffline` behavior is completely unchanged.
- Drop the `defer-delete` check from `TiProxySupportsOfflineBeforeDelete`. The delete reason should not be reverse-engineered from a CR annotation.

This has to be a builder option rather than just a `deleteInstance` parameter, because the distinction is two-dimensional: the call site knows *replace vs scale-in* (per-call), while component semantics decide *whether replace should offline first* (per-component — TiKV must evacuate regions even on rolling restart, TiProxy must not). Neither dimension alone expresses it.

The `update` / `outdated` split is computed by `split()` from the revision, so it is a fact rather than a heuristic — a much sounder basis than annotation sniffing.

### `TestDeleteInstanceGracefulTiProxyDeferDeleteDeletesCR` does not prove a Delete happened

```go
err := act.deleteInstance(context.Background(), runtime.FromTiProxy(obj))
require.NoError(t, err)
require.Nil(t, cli.lastDeleteOptions.PropagationPolicy)
```

If the `setOffline` branch is taken, `c.Delete` is never called, `lastDeleteOptions` stays at its zero value, and `PropagationPolicy` is nil — the assertion passes either way. The template came from `TestDeleteInstanceDoesNotOrphanTiProxyDependents`, where a delete is guaranteed and the assertion is meaningful; it does not survive the copy.

`deleteOptionRecorderClient` already records options, so adding a `deleteCalls int` counter and asserting `deleteCalls == 1` is the cheapest fix and localizes failures better than asserting the object is gone. Please also add the "outdated immediate-delete without `defer-delete`" case, which is the path above and is currently uncovered.

### The helper can then be dropped entirely

After removing the `defer-delete` check, `TiProxySupportsOfflineBeforeDelete` has only two conditions left, and the second one is already dead:

```go
if annotations[v1alpha1.AnnoKeyTiProxyReviveAbandoned] == v1alpha1.AnnoValTrue {
    return false
}
```

`abandonRevive` sets `spec.offline=true` **and** writes that annotation in the same patch. `deleteInstance` guards the offline branch with `!obj.IsOffline()`, which is already false by then, so on that path the check never changes the outcome.

The only way to reach it is if an abandoned instance is first flipped back to `offline=false` by `cancelOneOfflining` and then scaled in again — but `FilterAnnotationAbsent(AnnoKeyTiProxyReviveAbandoned)` exists precisely to keep abandoned instances from being picked for revive. So either this is defensive redundancy, which is fine but deserves a comment saying so, or `FilterAnnotationAbsent` has a hole, in which case that is the thing to fix rather than papering over it here. Either way it needs a test that pins the intent — `TestTiProxySupportsOfflineBeforeDelete` currently only checks the function's own truth table, not whether the combination can actually be reached through `deleteInstance`.

With both conditions gone the helper collapses to a single call to `GracefulOfflineScaleInEnabled`, so it can be dropped and `SupportsOffline()` can call `GracefulOfflineScaleInEnabled` directly, removing one exported function from `pkg/runtime/graceful_offline.go`.

Verified against head `9bad0d8`: the package tests pass, but the path above is not exercised by any of them.

fgksgf:
Re-reviewed at head `657dc4f1`.

`WithDirectDeleteOutdated` landed correctly. I checked the truth table: `outdated=true` with the option unset still enters the offline branch, so TiKV/TiFlash behavior is unchanged, and all three `deleteInstance` call sites now pass the right value. The new tests cover the counterexample path (`TestScaleInOutdatedDirectDeletesGracefulTiProxyWithoutDeferDelete`) and the reverse direction (`TestDeleteInstanceGracefulTiProxyScaleInMarksOffline` asserting `deleteCalls == 0`). `TiProxySupportsOfflineBeforeDelete` is gone. That part is good.

### The resolved thread's race is still present in the code

The review thread on `offline_scalein.go` is marked resolved, but its metadata says the code never changed:

```json
{"resolved": true, "outdated": false, "path": "pkg/controllers/tiproxy/tasks/offline_scalein.go", "line": 121}
```

`isOutdated: false` means GitHub still anchors the comment to a line that has not been touched — had the code changed, the thread would have been flagged outdated automatically. Confirmed by reading the file:

`graceful_shutdown.go:111-117` still performs the external side effect before persisting anything:

```go
if !ensureTiProxyMarkedUnhealthy(ctx, state, c, logger, tpClient) {   // external API first
    return task.DefaultRequeueAfter, nil
}
startAt = time.Now()
if err := markGracefulShutdownBeginTime(ctx, c, pod, startAt); err != nil {  // persisted after
    return 0, err
}
```

And the compensation still keys off the begin-time annotation in two places, both reading the same signal that is missing in exactly the failure case — `offline_scalein.go:52` (the `IfBreak` condition) and `offline_scalein.go:121` (inside the task).

I traced the follow-on path to confirm there is no closure: `TaskContextInfoFromTiProxy`'s `IsHealthy()` returns false, so `SetHealthy()` is never called, `TaskInstanceConditionReady` reports `ReasonInstanceNotHealthy`, and `tasks.TaskStatus` returns `Retry` on not-ready. The instance requeues forever down a path that never calls `ClearHealthOverride`. Since the health override lives in the TiProxy process and `TaskPod` only recreates the pod when `!reloadable.CheckTiProxyPod`, nothing in the operator breaks the loop.

### The trigger is more likely than "two rare events coinciding"

`markGracefulShutdownBeginTime` (`graceful_shutdown.go:162-169`) does a full-object `c.Update` on a pod read from the informer cache, with no conflict retry:

```go
newPod := pod.DeepCopy()
newPod.Annotations[v1alpha1.AnnoKeyTiProxyGracefulShutdownBeginTime] = startAt.Format(time.RFC3339Nano)
return c.Update(ctx, newPod)
```

Pod status is updated frequently by kubelet (probe transitions, container state), so a stale cached `resourceVersion` yielding a 409 is one of the most ordinary write conflicts in an operator, not an exotic failure. The error propagates as `return 0, err` → `TaskOfflineScaleInDrain` returns `task.Fail()` → the runner errors out and requeues with backoff. A scale-out arriving during that backoff window is enough. So this is "a routine write conflict" plus "an event the feature exists to handle frequently", which deserves a higher risk rating than the thread's resolution implies.

### On the fix: persisting the intent looks unavoidable, but it can be cheap

I looked for a smaller equivalent. Widening the condition to `offline == false && !IsHealthy()` for unconditional compensation seems attractive, but it collides with the 404 ambiguity in `isHealthOverrideUnsupported`: a `DELETE` returning 404 because there is no override to clear is indistinguishable from 404 because the route does not exist, so a merely not-yet-ready healthy instance would get mis-classified and pushed through `abandonRevive`. So persisting an operator-owned intent really is needed.

It does not have to be an API field, though. An annotation on the **TiProxy CR** is enough — the CR is operator-owned, and `abandonRevive` (`graceful_shutdown.go:66-84`) already patches it via `client.RawPatch(types.MergePatchType, ...)`, so the plumbing exists. Reordering to:

1. patch the drain-intent annotation on the CR
2. `MarkUnhealthy`
3. write the pod begin-time

keeps the begin-time semantics intact for HPA consumers (it still means "draining has actually begun") while making the revive decision read a signal that is written before the side effect. No API type or CRD change.

### Test gap

The case from the original thread — `offline=false`, override set, begin-time annotation absent — still has no test at all. If this is deferred, please add it even as a `t.Skip` placeholder so the gap is visible rather than implicit.

### CI

`pull-e2e` is currently failing on `657dc4f1`; everything else (verify, lint, unit, build, image) is green. I checked the 8 most recent PRs and their `pull-e2e` runs are pass or pending, with no horizontal failure, so this is most likely specific to this PR rather than infrastructure flake — worth reading the log before reaching for `/retest`. I cannot read it myself: the prow GCS bucket rejects anonymous access.

Ordering the blockers: the e2e failure gates mergeability outright, so it comes before the question of whether to accept the race as a follow-up. If the team does decide to defer the race, please reopen the thread or record it in the PR description — a `resolved` marker on an unchanged line is easy to mistake for a fix.

fgksgf (approved):

PR #6966, opened 2026-06-29
feat: PD mode switching

csuzhangxc:
## Summary

Implements controller-driven PD mode switching between normal mode and microservice mode.

This PR makes `PDGroup.spec.template.spec.mode` mutable and reconciles mode changes through the PDGroup controller. The workflow detects whether a switch is active, waits for the required PD/TSO topology state, rolls PD instances to the target mode, and records progress with PDGroup status and conditions.

It also adds TSOGroup protection while PD is in, entering, or leaving MS mode so the cluster does not lose the required TSO capacity during a switch.

### API and status

- Allows `PDTemplateSpec.mode` to be changed after creation.
- Adds `PDGroup.status.mode` as the observed PD mode after all desired PD instances have converged.
- Adds PDGroup `ModeSwitching` condition helpers and reasons for blocked, switching, complete, and no-op states.
- Adds TSOGroup `PDMSProtected` condition helpers and reasons.
- Regenerates deepcopy and CRD manifests.

### PDGroup mode switch workflow

- Adds `TaskModeSwitch` to the PDGroup reconcile path.
- Treats initial creation with `mode=ms` as normal creation, not as a mode switch.
- Detects whether an existing PDGroup is actually switching by checking:
  - desired PDGroup template mode
  - observed PDGroup status mode
  - active `ModeSwitching` condition
  - existing PD instance modes
- Blocks mode switching only when the cluster does not have exactly one PDGroup.
- Allows multiple TSOGroup, SchedulingGroup, and ResourceManagerGroup objects instead of treating them as unsupported topology.
- Before entering MS mode, waits for:
  - at least one TSOGroup
  - every TSOGroup to be healthy and up to date
  - the cluster-level TSO health API client to be registered and healthy
- Marks switching state only when status actually changes, avoiding repeated no-op status writes.
- Clears the `ModeSwitching` condition and updates `status.mode` once all PD instances reach the target mode, current revision, Ready, and Synced.
- Uses bounded retry for blocked switch paths so progress does not depend only on Kubernetes watch events.

### TSOGroup PDMS protection

- Adds `TaskPDMSProtection` to the TSOGroup reconcile path.
- Treats a cluster as MS-related if any PDGroup desired mode, observed mode, active mode-switch condition, or PD instance mode involves MS.
- Clamps effective TSOGroup replicas from 0 to 1 while PD involves MS mode, without mutating the user spec.
- Protects deletion of the last available TSOGroup while PD involves MS mode.
- Allows deletion of extra TSOGroup objects so topology can converge.
- Handles concurrent TSOGroup deletion by checking for another non-deleting TSOGroup instead of only counting total TSOGroup objects.
- Retries protected deletion reconciles so finalizer removal can continue once protection is no longer needed.

### Controller event wiring

- PDGroup controller watches TSOGroup changes and enqueues PDGroups in the same cluster because TSO readiness can unblock a mode switch.
- PDGroup controller no longer watches SchedulingGroup or ResourceManagerGroup changes for PDMS dependencies because those resources no longer block mode switching.
- TSOGroup controller watches PDGroup changes and enqueues TSOGroups in the same cluster.
- These watches intentionally include status changes because readiness, observed mode, and mode-switch conditions can unblock the workflow.

### PD client mode detection

- Updates PD client mode selection to use the shared PD topology helper.
- This lets PD client behavior account for desired mode, observed mode, active mode-switch conditions, and existing PD instance modes.

## Tests

- `make codegen crd runtimegen`
- `go test ./pkg/controllers/pdgroup/tasks ./pkg/controllers/tsogroup/tasks ./pkg/timanager/pd ./pkg/pdms`
- `go test ./pkg/controllers/pdgroup/... ./pkg/controllers/tsogroup/...`
- `go test ./pkg/controllers/tsogroup/...`
- `go test ./pkg/controllers/...`
- `go test ./pkg/... ./cmd/...`
- `cd tests/validation && go test ./...`
- `go test ./tests/e2e/pd`
- `make lint`

## Notes

- This PR does not add a cross-resource admission webhook. The enforcement is implemented in controllers because the repository currently has CRD validation but no webhook implementation for this cross-resource invariant.
- `PDGroup.spec.replicas=0` is not given special mode-switch behavior. A zero-PD topology is invalid for both normal and MS mode, so the switch workflow does not add extra complexity for it.
- Multiple TSOGroups are allowed. The controller checks that all listed TSOGroups are ready and up to date, while the TSO health API check uses the existing cluster-keyed TSO client manager.


api/core/v1alpha1/pd_types.go
```diff
@@ -218,6 +215,22 @@ type PDPorts struct {
...
 	CommonStatus `json:",inline"`
 	GroupStatus  `json:",inline"`
+
+	// Mode is the actual PD mode after all desired PD instances have converged.
+	Mode PDMode `json:"mode,omitempty"`
+	// ModeTransition describes an active PD mode switch.
+	ModeTransition *PDModeTransition `json:"modeTransition,omitempty"`
+}
+
+type PDModeTransition struct {
```
liubog2008: change to condition

api/core/v1alpha1/pd_types.go
```diff
@@ -218,6 +215,22 @@ type PDPorts struct {
 type PDGroupStatus struct {
 	CommonStatus `json:",inline"`
 	GroupStatus  `json:",inline"`
+
+	// Mode is the actual PD mode after all desired PD instances have converged.
+	Mode PDMode `json:"mode,omitempty"`
```
liubog2008, line 220: condition is enough

pkg/pdms/state.go
```diff
@@ -0,0 +1,96 @@
+// Copyright 2024 PingCAP, Inc.
```
liubog2008: this pkg seems useless?

pkg/controllers/pdgroup/tasks/mode_switch.go
```diff
@@ -0,0 +1,229 @@
+// Copyright 2024 PingCAP, Inc.
```
liubog2008, line 1: why a mode switch task? I think pd only need to check whether there is tso or not

pkg/controllers/tsogroup/builder.go
```diff
@@ -39,6 +39,14 @@ func (r *Reconciler) NewRunner(state *tasks.ReconcileContext, reporter task.Task
...
 		task.IfBreak(common.CondObjectIsDeleting[scope.TSOGroup](state),
+			tasks.TaskPDMSProtection(state, r.Client),
+			task.IfBreak(tasks.CondPDMSProtected(state),
+				common.TaskGroupConditionReady[scope.TSOGroup](state),
+				common.TaskGroupConditionSynced[scope.TSOGroup](state),
+				common.TaskStatusRevisionAndReplicas[scope.TSOGroup](state),
+				common.TaskStatusPersister[scope.TSOGroup](state, r.Client),
+				tasks.TaskPDMSProtectedRetry(),
+			),
 			common.TaskGroupFinalizerDel[scope.TSOGroup](state, r.Client),
```
liubog2008, line 51: Check in group finalizer directly.

pkg/controllers/tsogroup/builder.go
```diff
@@ -39,6 +39,14 @@ func (r *Reconciler) NewRunner(state *tasks.ReconcileContext, reporter task.Task
 		common.TaskContextSlice[scope.TSOGroup](state, r.Client),
 
 		task.IfBreak(common.CondObjectIsDeleting[scope.TSOGroup](state),
+			tasks.TaskPDMSProtection(state, r.Client),
+			task.IfBreak(tasks.CondPDMSProtected(state),
+				common.TaskGroupConditionReady[scope.TSOGroup](state),
+				common.TaskGroupConditionSynced[scope.TSOGroup](state),
+				common.TaskStatusRevisionAndReplicas[scope.TSOGroup](state),
+				common.TaskStatusPersister[scope.TSOGroup](state, r.Client),
+				tasks.TaskPDMSProtectedRetry(),
```
liubog2008, line 49: Why?

pkg/controllers/pdgroup/tasks/ctx.go
```diff
@@ -29,9 +29,10 @@ import (
 type ReconcileContext struct {
 	State
 
-	Members []Member
+	Members  []Member
+	PDClient pdm.PDClient
```
liubog2008: Why add it back

api/core/v1alpha1/pd_types.go
csuzhangxc: https://github.com/pingcap/tidb-operator/pull/6966/commits/3d98d44698a693df8d45f6b7dfa32961c700a0ba

api/core/v1alpha1/pd_types.go
csuzhangxc: If we remove `status.mode`, the condition would need to carry more semantics than just `ModeSwitching=True/False`, such as the last converged mode or switching direction.

pkg/pdms/state.go
csuzhangxc: needed

pkg/controllers/pdgroup/tasks/mode_switch.go
csuzhangxc: Put code into a single task

pkg/controllers/pdgroup/tasks/ctx.go
csuzhangxc: removed https://github.com/pingcap/tidb-operator/pull/6966/commits/523a390951337352f14829d591ea84235f3b951c

pkg/controllers/tsogroup/builder.go
csuzhangxc: If we remove it, a PDMS-protected deleting TSOGroup may stop reconciling after status update and stay stuck until another watch event happens.

pkg/timanager/pd/pd.go
```diff
@@ -147,22 +160,19 @@ func CacheKeysFunc(c client.Client) func(cluster *v1alpha1.Cluster) ([]string, e
...
-// If there are any PD in ms mode, use ms mode client
+// If PD desired mode, actual mode, transition, or instances involve ms, use ms mode client.
 func getPDMode(c client.Client, cluster *v1alpha1.Cluster) (v1alpha1.PDMode, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
 	defer cancel()
 
 	mode := v1alpha1.PDModeNormal
 
-	pdgs, err := apicall.ListGroups[scope.PDGroup](ctx, c, cluster.Namespace, cluster.Name)
+	s, err := pdms.GetState(ctx, c, cluster.Namespace, cluster.Name)
```
liubog2008: no need to get all pdms components.

pkg/controllers/pdgroup/tasks/mode_switch.go
```diff
@@ -0,0 +1,209 @@
...
+
+		if !modeSwitchActive(pdg, state.PDSlice(), target) {
+			changed := coreutil.SetStatusCondition[scope.PDGroup](pdg, *coreutil.ModeSwitchNotNeeded())
+			if changed {
+				state.SetStatusChanged()
+			}
+			return task.Complete().With("PD mode switch is not needed")
+		}
+
+		info, err := pdms.GetState(ctx, c, pdg.Namespace, pdg.Spec.Cluster.Name)
```
liubog2008: should not GetState in the task. ContextXXX task will fetch these objects

pkg/controllers/tsogroup/tasks/pdms_protection.go
```diff
@@ -0,0 +1,101 @@
...
+type PDMSProtectionState interface {
+	State
+	SetPDMSProtected(bool)
+	PDMSProtected() bool
+}
+
+func TaskPDMSProtection(state *ReconcileContext, c client.Client) task.Task {
+	return task.NameTaskFunc("PDMSProtection", func(ctx context.Context) task.Result {
+		tg := state.Object()
+		msState, err := pdms.GetState(ctx, c, tg.Namespace, tg.Spec.Cluster.Name)
```
liubog2008: Should not get state in the task. Use objects fetched by ContextXXX tasks.

pkg/pdms/state.go
```diff
@@ -0,0 +1,99 @@
...
+		pdg.Status.Mode == v1alpha1.PDModeMS ||
+		modeSwitchingConditionActive(pdg)
+}
+
+func modeSwitchingConditionActive(pdg *v1alpha1.PDGroup) bool {
+	cond := meta.FindStatusCondition(pdg.Status.Conditions, v1alpha1.CondModeSwitching)
+	return cond != nil && cond.Status == metav1.ConditionTrue
+}
+
+func TSOGroupProtected(ctx context.Context, c client.Client, tg *v1alpha1.TSOGroup) (bool, error) {
```
liubog2008: Move to apiutil. If kubernetes client is needed, move to apicall

pkg/pdms/state.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	for _, pd := range s.PDs {
+		if pd.Spec.Mode == v1alpha1.PDModeMS {
+			return true
+		}
+	}
+	return false
+}
+
+func PDGroupInvolvesMS(pdg *v1alpha1.PDGroup) bool {
```
liubog2008: move to apiutil

pkg/pdms/state.go
```diff
@@ -0,0 +1,99 @@
...
+	return &State{
+		PDGroups:              pdgs,
+		PDs:                   pds,
+		TSOGroups:             tgs,
+		SchedulingGroups:      sgs,
+		ResourceManagerGroups: rmgs,
+	}, nil
+}
+
+func (s *State) InvolvesMS() bool {
```
liubog2008: move to apiutil

pkg/pdms/state.go
```diff
@@ -0,0 +1,99 @@
...
+
+type State struct {
+	PDGroups              []*v1alpha1.PDGroup
+	PDs                   []*v1alpha1.PD
+	TSOGroups             []*v1alpha1.TSOGroup
+	SchedulingGroups      []*v1alpha1.SchedulingGroup
+	ResourceManagerGroups []*v1alpha1.ResourceManagerGroup
+}
+
+func GetState(ctx context.Context, c client.Client, namespace, cluster string) (*State, error) {
```
liubog2008: Recheck whether the GetState is needed.

pkg/controllers/tsogroup/tasks/pdms_protection.go
```diff
@@ -0,0 +1,101 @@
...
+	return task.CondFunc(func() bool { return state.PDMSProtected() })
+}
+
+func TaskPDMSProtectedRetry() task.Task {
+	return task.NameTaskFunc("PDMSProtectedRetry", func(ctx context.Context) task.Result {
+		return task.Retry(defaultUpdateWaitTime).With("TSOGroup deletion is protected by PDMS mode")
+	})
+}
+
+func effectiveReplicas(ctx context.Context, c client.Client, tg *v1alpha1.TSOGroup) (replicas int32, protected bool, err error) {
```
liubog2008: the protected is not used anywhere. Remove the return value `protected`

pkg/timanager/pd/pd.go
csuzhangxc: changed this path to only list PDGroups and PD instances for PD mode detection,

pkg/controllers/pdgroup/tasks/mode_switch.go
csuzhangxc: changed. The required PDMS dependency objects are fetched by TaskContextPDMSDependencies and then consumed from reconcile state

pkg/controllers/tsogroup/tasks/pdms_protection.go
csuzhangxc: changed to use `TaskContextPDMSProtection`

pkg/pdms/state.go
csuzhangxc: moved

pkg/pdms/state.go
csuzhangxc: moved

pkg/pdms/state.go
csuzhangxc: moved

pkg/pdms/state.go
csuzhangxc: The generic pdms.GetState package is no longer needed after splitting pure predicates into apiutil, client helpers into apicall, and controller-specific fetches into Context tasks.

pkg/controllers/tsogroup/tasks/pdms_protection.go
csuzhangxc: Removed

pkg/controllers/pdgroup/tasks/mode_switch.go
```diff
@@ -0,0 +1,206 @@
...
+func blockModeSwitch(state *ReconcileContext, reason, msg string) {
+	pdg := state.PDGroup()
+	state.SetModeSwitchBlocked(true)
+	changed := coreutil.SetStatusCondition[scope.PDGroup](pdg, *coreutil.ModeSwitching(reason, msg))
+	if changed {
+		state.SetStatusChanged()
+	}
+}
+
+func unsupportedTopology(
```
liubog2008: Why？ It looks useless？

csuzhangxc: https://github.com/pingcap/tidb-operator/pull/6966/commits/edab6871a034968fd393b4742dab4e42c5c10cc3

liubog2008 (approved):

PR #6995, opened 2026-07-12
feat: support placement policy

liubog2008:

This PR adds PlacementPolicy support for keyspace placement rules.

It also fixes the CI and e2e failures found on the PR. In particular, updating a TiKVGroup to `placement.exclusive: true` should not restart TiKV pods. The exclusive placement change only needs to update PD store labels and placement rules.

- Add PlacementPolicy API, validation, controller reconciliation, and PD placement rule sync.
- Generate PD placement rules for normal and exclusive TiKVGroup targets.
- Sync TiKVGroup placement labels to PD stores.
- Treat TiKV `Spec.Placement` as reloadable so changing `exclusive` does not recreate TiKV pods.
- Add unit coverage for reloadable TiKV placement changes.
- Add e2e coverage to verify the TiKV pod UID does not change when a TiKVGroup becomes exclusive.
- Fix generated docs/license/test issues reported by CI.

### Check List

Tests:

```bash
make lint
make unit
make verify/license
go test ./tests/e2e/... -run '^$'
go test ./pkg/controllers/placementpolicy ./pkg/pdapi/v1 ./pkg/reloadable ./pkg/controllers/tikv/tasks
```


fgksgf (commented):
Reviewing current head `fa4ca507151711afcf50ba75d177af51764ab531`, I found four new issues; the first three are correctness problems that should be fixed before merge. The nil-pointer panic caused by `selector: {type: Keyspace}` already has an unresolved Copilot thread, so I did not duplicate it. This remains a COMMENT review; I did not approve or request changes.

pkg/apiutil/core/v1alpha1/placement.go
```diff
@@ -0,0 +1,34 @@
...
+package coreutil
+
+import "fmt"
+
+func PlacementTiKVGroupLabelValue(namespace, cluster, group string) string {
+	return fmt.Sprintf("%s.%s.%s", namespace, cluster, group)
+}
+
+func PlacementPolicyGroupID(policyName string) string {
+	return "k%" + policyName
```
fgksgf: [P1] Every bundle ID generated here contains `%`, but `DeletePlacementRuleBundle` applies `PathEscape` only once. The PD v8.5.2 [placement-rule router](https://github.com/tikv/pd/blob/v8.5.2/server/api/router.go#L178-L203) does not enable `UseEncodedPath`, so mux extracts the parameter from the already-decoded `URL.Path`; the [DELETE handler](https://github.com/tikv/pd/blob/v8.5.2/server/api/rule.go#L487-L501) then calls `PathUnescape` again. Consequently, `k%p` fails the second decode with an invalid URL escape, and `reconcileDelete` cannot remove the finalizer, leaving the PlacementPolicy permanently stuck in deletion. Please use a URL-safe group ID that does not contain `%`, and add PD-compatible deletion and finalizer tests using the actual generated group ID.

pkg/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,504 @@
...
+	if !ok {
+		err := fmt.Errorf("pd client is not ready")
+		return ctrl.Result{}, r.updateStatus(ctx, policy, err)
+	}
+
+	if err := apicall.EnsureFinalizer(ctx, r.Client, policy); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	bundle, err := r.buildBundle(ctx, policy)
```
fgksgf: [P1] There is no check for another PlacementPolicy in the same namespace and cluster before this policy is built and synced. Each policy gets a different group ID, but every bundle uses index 99 with `Override: true`. PD orders groups with the same index by GroupID, and a later override group removes the preceding groups. Therefore, when two policies select the same keyspace, both can report `Synced=True` even though only the lexicographically later policy participates in scheduling. This also violates the conflict-rejection requirement in step 3 of the RFC. Please implement deterministic conflict ownership using canonical numeric keyspace IDs, enqueue sibling policies on create/update/delete, and test concurrent creation plus convergence after the owner is deleted.

api/core/v1alpha1/placement_policy_types.go
```diff
@@ -0,0 +1,128 @@
...
+	// Selector to select key range by keyspace
+	*PlacementPolicyKeyspaceSelector `json:",inline"`
+}
+
+type PlacementPolicyKeyspaceSelector struct {
+	// Keyspaces selects PD keyspace IDs. Database and table selectors are reserved
+	// for a later version because their key ranges are unstable today.
+	// +listType=set
+	// +kubebuilder:validation:MinItems=1
+	Keyspaces []string `json:"keyspaces,omitempty"`
```
fgksgf: [P1] `keyspaces` is an unconstrained string set, so Kubernetes accepts `1` and `01` as distinct elements. However, `MakeRegionBound` parses both as numeric keyspace ID 1, while the rule ID preserves the original string. This produces two rules with different IDs for the same key range, and PD satisfies each rule's `Count` independently; for example, two `count: 3` rules require six replicas. Please accept only canonical decimal IDs at admission time, normalize and deduplicate before rule generation and cross-policy conflict detection, and add validation/controller tests for `1` together with `01`.

pkg/pdapi/v1/client.go
```diff
@@ -338,6 +385,123 @@ func (c *pdClient) SetStoreLabels(ctx context.Context, storeID string, labels ma
...
+	}
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return err
+	}
+	return fmt.Errorf("failed to delete store label %s from store %s: %s", labelKey, storeID, body)
+}
+
+func (c *pdClient) SetPlacementRuleBundle(ctx context.Context, bundle *PlacementRuleGroupBundle, partial bool) error {
+	apiURL := fmt.Sprintf("%s/%s?partial=%t", c.url, placementRulePrefix, partial)
```
fgksgf: [P2] PD's SetPlacementRules handler checks only whether the `partial` query key is present; it does not parse the boolean value. As a result, `partial=false` is still treated as a partial update, so this exported client method cannot perform the requested full replacement. The current controller only passes true, so its present call path is unaffected, but the new client contract is incorrect. Please append the query parameter only when `partial == true`, omit it entirely for false, and test both branches.

tennix (changes requested):
## Hermes Agent Review

Found one blocking issue in the new `PlacementPolicy` API/controller path.

`spec.rules[].selector.keyspaces` is currently optional in the generated CRD: the selector schema only requires `type`, while `keyspaces` has `minItems` only when present. Because `PlacementPolicySelector` embeds `*PlacementPolicyKeyspaceSelector`, a valid object like:

```yaml
rules:
- name: voters
  role: voter
  count: 3
  selector:
    type: Keyspace
```

leaves the embedded pointer nil. The controller then evaluates `rule.Selector.Keyspaces`, which dereferences that nil embedded pointer and panics during reconciliation instead of rejecting the object or setting a failed status.

Please make `keyspaces` required for `type: Keyspace` (for example by avoiding the pointer embedding or adding CEL/required validation), and also guard `buildBundle` against a nil keyspace selector so malformed/admitted objects cannot crash the reconciler.

I verified locally:
- `git diff --check main...HEAD` passed
- `go test ./pkg/controllers/placementpolicy ./pkg/pdapi/v1 ./pkg/reloadable ./pkg/controllers/tikv/tasks ./pkg/apiutil/core/v1alpha1 -count=1` passed
- `go test . -run TestPlacementPolicy -count=1` in `tests/validation` passed, but the current validation tests don't cover the missing-`keyspaces` case.

tennix:
Additional notes after a second pass (not duplicating the existing inline threads):

1. **Implementation emits raw + txn keyspace ranges, while the RFC/test plan says txn-only.**
   `pkg/utils/keyspace.MakeRegionBound` returns both `raw` and `txn` ranges, and `buildBundle` emits rules for all returned ranges. However the RFC says the controller should use the same txn keyspace range format as `pd-ctl keyspace set-placement`, and the generated key range is the txn keyspace range. If raw-mode keyspace placement is intended too, please update the RFC/API expectations; otherwise the controller should only emit txn ranges and cover that in tests.

2. **E2E placement-rule checks only compare rule IDs.**
   `WaitForPlacementRuleExists` reduces actual/expected rules to sorted IDs, while the expected e2e rules only set IDs. This would miss wrong `group_id`, key ranges, role/count, or label constraints (`k/tikvgroup` vs `$k/tikvgroup`, wrong values). Since this feature’s correctness depends on those fields, please compare normalized full rule contents in e2e (similar to the controller unit-test normalization).

api/core/v1alpha1/placement_policy_types.go
```diff
@@ -0,0 +1,128 @@
...
+	Keyspace *PlacementPolicyKeyspaceSelector `json:"keyspace,omitempty"`
+}
+
+type PlacementPolicyKeyspaceSelector struct {
+	// Keyspaces selects PD keyspace IDs. Database and table selectors are reserved
+	// for a later version because their key ranges are unstable today.
+	// +listType=set
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:items:Pattern=`^(0|[1-9][0-9]*)$`
+	IDs []string `json:"ids,omitempty"`
```
fgksgf: [P1] Require ids so an empty selector cannot delete existing rules

`MinItems=1` is evaluated only when `ids` is present, but this field is optional in the generated schema, so `selector: { keyspace: {} }` passes admission. `buildRules` then produces an empty expected set, `SetPlacementRuleGroupRulesByIDPrefix` deletes every existing rule owned by this policy, and reconciliation reports success. Please make `ids` required and defensively reject an empty slice in the controller so an omitted selector cannot silently remove placement constraints.

pkg/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,457 @@
...
+		) {
+			kvg := event.Object.(*v1alpha1.TiKVGroup)
+			r.enqueuePoliciesForTiKVGroup(ctx, queue, kvg)
+		},
+		UpdateFunc: func(ctx context.Context, event event.TypedUpdateEvent[client.Object],
+			queue workqueue.TypedRateLimitingInterface[reconcile.Request],
+		) {
+			oldKVG := event.ObjectOld.(*v1alpha1.TiKVGroup)
+			kvg := event.ObjectNew.(*v1alpha1.TiKVGroup)
+			if coreutil.TiKVStorePlacementExclusive(oldKVG.Spec.Template.Spec.Placement) ==
```
fgksgf, line 124: [P1] Reconcile TiKVGroup deletion before draining its stores

This filter drops a deletionTimestamp-only update whenever exclusive is unchanged. A referenced TiKVGroup then enters its finalizer, deletes and offlines its TiKV instances, and waits for their stores to reach Removed, while this policy still requires peers on k/tikvgroup=<that group>. PD RuleChecker does not bypass that label constraint when replacing an offline peer. If this is the only target group, no serving replacement store matches, so the stores cannot reach zero regions and the TiKV and TiKVGroup finalizers can remain pending indefinitely.

This is distinct from the known follow-up case where a referenced TiKVGroup is already NotFound. Please either clean or exclude the policy rules before draining the group, or enforce and document that the policy must be removed first.

docs/rfcs/260707-placement-policy.md
```diff
@@ -0,0 +1,272 @@
...
+
+PlacementPolicy rules target both normal and exclusive TiKVGroups through the shared `k/tikvgroup` constraint. They also add `$k/exclusive notIn ["__null__"]` only to opt in to PD's exclusive-label matching. `__null__` must not be used as a real store label value; as a nonexistent sentinel, it keeps the constraint effectively non-filtering while still mentioning the exclusive label key.
+
+### PlacementPolicy Controller
+
+The controller reconciles `PlacementPolicy` as follows:
+
+1. Resolve `spec.cluster.name` and all `spec.groupRefs`.
+2. Validate every `groupRef` points to a `TiKVGroup` in the same cluster.
+3. Build keyspace ranges using the same txn keyspace range format as `pd-ctl keyspace set-placement`.
```
fgksgf: [P2] Align the documented key range with the implementation

This step and the later detailed section say that only the txn keyspace range is generated, while `MakeRegionBound` returns both raw and txn ranges and the controller creates a PD rule for each. The E2E expectation also contains both rule IDs, whereas the RFC example shows only the raw rule. Please document both generated ranges (and show both in the example), or change the implementation if txn-only is the intended contract.

pkg/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,457 @@
...
+
+	if !policy.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, r.reconcileDelete(ctx, policy)
+	}
+
+	if err := apicall.EnsureFinalizer(ctx, r.Client, policy); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	pdc, ok := r.PDClientManager.Get(timanager.PrimaryKey(policy.Namespace, policy.Spec.Cluster.Name))
```
fgksgf, line 229: [P2] Reconcile ignores `Cluster.spec.paused`. Other v2 controllers gate reconciliation with `task.IfBreak(common.CondClusterIsPaused(state))` (e.g. `pkg/controllers/tikv/builder.go`), but this controller keeps writing placement rules to PD while the cluster is paused or suspended. Please fetch the Cluster and skip the PD sync in those states to keep the convention consistent.

pkg/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.PlacementPolicy{}).
+		Watches(&v1alpha1.Cluster{}, r.ClusterEventHandler()).
+		Watches(&v1alpha1.TiKVGroup{}, r.TiKVGroupEventHandler()).
+		WithOptions(controller.Options{RateLimiter: k8s.RateLimiter}).
+		Complete(r)
+}
+
+func (r *Reconciler) ClusterEventHandler() handler.TypedEventHandler[client.Object, reconcile.Request] {
```
fgksgf, line 83: [P3] The Cluster watch only handles Create and Delete. When a Cluster that has a finalizer is deleted, setting `deletionTimestamp` arrives as an Update event; the Delete event fires only after all finalizers are cleared. A PlacementPolicy that is being deleted during that window can only discover `clusterDeletingOrGone` through error-backoff retries ("pd client is not ready"), which slows down cascading deletion. Consider adding an UpdateFunc that enqueues policies when `deletionTimestamp` transitions from zero to non-zero.

pkg/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,457 @@
...
+		return ctrl.Result{}, r.updateStatus(ctx, policy, fmt.Errorf("failed to build placement rules: %w", err))
+	}
+
+	underlay := pdc.Underlay()
+	rulePrefix := coreutil.PlacementPolicyRuleIDPrefix(policy.Name)
+	if err := syncPolicyRules(ctx, underlay, rulePrefix, expected); err != nil {
+		return ctrl.Result{}, r.updateStatus(ctx, policy, err)
+	}
+
+	return ctrl.Result{}, r.updateStatus(ctx, policy, nil)
```
fgksgf, line 246: [P3] Successful reconciles return an empty Result, so PD-side drift is never repaired until the next Kubernetes event. If someone deletes or edits rules under the `tidb-operator` group via pd-ctl, the policy stays `Synced=True` indefinitely. Store labels have timanager cache events as a backstop, but placement rules have none. Consider returning `RequeueAfter` on success for periodic re-sync.

cmd/tidb-operator/main.go
```diff
@@ -330,6 +331,14 @@ func addIndexer(ctx context.Context, mgr ctrl.Manager) error {
 		return err
 	}
 
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &v1alpha1.PlacementPolicy{}, "spec.cluster.name",
```
fgksgf, line 334: [P3] This `spec.cluster.name` field indexer is registered but never used: `enqueuePoliciesForCluster` and `enqueuePoliciesForTiKVGroup` list all policies in the namespace and filter in Go. Either list with `client.MatchingFields{"spec.cluster.name": ...}` in the event handlers, or drop this registration.

fgksgf (commented):
Second-pass review focused on points not covered by existing threads. Verified against PD v8.5.2 source that the core mechanics are sound: `$`-prefixed exclusive label semantics (`isExclusiveLabel`/`MatchLabelConstraints`), the `notIn ["__null__"]` trick, label key format, `DELETE /store/{id}/label`, `/regions/keyspace/id/{id}`, and `:` in rule IDs are all valid. The findings below are new, mostly minor.

pkg/controllers/placementpolicy/controller.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+
+	rules := []pdapi.PlacementRule{}
+	for _, rule := range policy.Spec.Rules {
+		if rule.Selector.Keyspace == nil {
+			return nil, fmt.Errorf("rule %q selector.keyspace is required", rule.Name)
+		}
+
+		role := rule.Role
+		if role == "" {
```
fgksgf: [P3] Both fallbacks are dead code: `role` is required with enum `voter` and `count` is required with `minimum: 1` at admission time, so neither can be empty/zero here. Silently defaulting `count` to 3 would also mask bugs if the validation is ever relaxed. Consider removing the fallbacks or returning an error instead.

pkg/utils/keyspace/keyspace.go
```diff
@@ -0,0 +1,168 @@
...
+	EndKeyHex   string
+}
+
+func MakeRegionBound(str string) ([]KeyRange, error) {
+	keyspaceID64, err := strconv.ParseUint(str, 10, 32)
+	if err != nil {
+		return nil, err
+	}
+	keyspaceID := uint32(keyspaceID64)
+	if keyspaceID < DefaultKeyspaceID || keyspaceID > MaxValidKeyspaceID {
```
fgksgf, line 37: [P3] `keyspaceID < DefaultKeyspaceID` is always false: `keyspaceID` is uint32 and `DefaultKeyspaceID` is 0, so only the upper-bound check is effective. Consider dropping the dead condition.

tests/e2e/utils/waiter/placement_policy.go
```diff
@@ -0,0 +1,573 @@
...
+
+	return false
+}
+
+func WaitForKeyspaceCreated(ctx context.Context, pdc pdapi.PDClient, name string, timeout time.Duration) (*pdapi.KeyspaceMeta, error) {
+	var lastErr error
+	var meta *pdapi.KeyspaceMeta
+	if err := wait.PollUntilContextTimeout(ctx, Poll, timeout, true, func(ctx context.Context) (bool, error) {
+		var err error
+		meta, err = pdc.CreateKeyspace(ctx, name, nil)
```
fgksgf, line 244: [P3] Retrying `CreateKeyspace` inside the poll loop can make this flaky: if the first create succeeds but the response is lost (e.g. client timeout), every retry returns an already-exists error until the waiter times out with a misleading message. Consider treating already-exists as success, or falling back to fetching the keyspace by name on error.

pkg/utils/keyspace/keyspace.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil, fmt.Errorf("invalid keyspace id %d. It must be in the range of [%d, %d]",
+			keyspaceID, DefaultKeyspaceID, MaxValidKeyspaceID)
+	}
+
+	bound := makeRegionBound(keyspaceID)
+	return []KeyRange{
+		{
+			Type:        "raw",
```
rleungx, line 45: Is it necessary to consider raw KV?

liubog2008: Not necessary.

fgksgf: The current head still emits both raw and txn rules: MakeRegionBound returns both ranges, buildRules writes both, and the RFC and E2E now explicitly expect both. That does not match the "Not necessary" decision above, and the API has no mode selector. Upstream pd-ctl defaults keyspace placement to txn and requires --raw to target raw KV. If raw KV is out of scope, please remove the raw rule; otherwise please clarify that both modes are intentional and why they should always share the same placement.

pkg/pdapi/v1/client.go
```diff
@@ -83,10 +98,18 @@ type PDClient interface {
...
+	// ListPlacementRulesByGroupIDPrefix lists placement rules with the specified ID prefix in a group.
+	ListPlacementRulesByGroupIDPrefix(ctx context.Context, groupID, idPrefix string) ([]PlacementRule, error)
+	// GetPlacementRuleBundle gets a PD placement rule bundle by group ID.
+	GetPlacementRuleBundle(ctx context.Context, groupID string) (*PlacementRuleGroupBundle, error)
 	// GetTombStoneStores lists all tombstone stores of the cluster.
 	// GetTombStoneStores() (*StoresInfo, error)
 	// GetStore gets a TiKV/TiFlash store for a specific store id of the cluster.
 	GetStore(ctx context.Context, storeID string) (*StoreInfo, error)
+	// ListRegions lists all regions of the cluster.
+	ListRegions(ctx context.Context) (*RegionsInfo, error)
```
fgksgf, line 110: [P3] `ListRegions` looks like a leftover of the same cleanup that removed `SetPlacementRuleBundle`/`DeletePlacementRuleGroup`/`DeletePlacementRuleBundle`/`ListPlacementRuleBundles`: it is only defined and unit-tested here, with no caller in the controller or e2e (`ListKeyspaceRegions` is the one actually used). Consider dropping it as well.

fgksgf (commented):
Re-reviewed the update (a5749cd, ceee296). The paused/suspend gate, MatchingFields usage, `ids` required, dead fallbacks, and the unused bundle APIs are all addressed. One leftover from the API cleanup below.

pkg/controllers/placementpolicy/controller.go
fgksgf: The new UpdateFunc covers paused/suspendAction changes (thanks), but a deletionTimestamp-only update is still dropped by this filter. So when a Cluster with finalizers is being deleted, a concurrently-deleting policy still discovers `clusterDeletingOrGone` only through error-backoff retries instead of an event. Minor latency issue only; fine to fold into the follow-up PR if you prefer.

fgksgf (approved):
/lgtm

Approving. The core mechanics have been verified against PD v8.5.2 source: `$`-prefixed exclusive label semantics, the `notIn ["__null__"]` constraint trick, keyspace range encoding (including the max-ID fencepost), store label / rule / batch API usage, and the diff-based rule sync all check out. The paused/suspend gate, field-indexer usage, `ids` required, and the API cleanup from the latest push address the earlier review rounds.

Two design decisions confirmed offline by the author, recording them here so the threads are closed:

1. Overlapping policies on the same keyspace are expected to stack: rules in the shared `tidb-operator` group all apply and their counts are additive, which matches standard PD placement-rule semantics. Please also note this in the RFC's Risks section (or the open P1 thread) so future readers don't re-raise it.
2. Raw-mode key ranges are intentionally kept alongside txn ranges (the earlier "Not necessary" reply in the raw-KV thread is superseded).

Remaining open items are follow-up material and not blocking: TiKVGroup deletion lifecycle (author plans a separate PR), e2e placement-rule checks comparing only rule IDs, periodic re-sync of PD-side drift, the unused `ListRegions` client method, and a couple of P3 test-robustness nits.

PR #6999, opened 2026-07-15
fix: use pd default location labels for placement rules

liubog2008:
## Summary
- read `location_labels` from the PD default placement rule `pd/default`
- set those location labels on all placement policy managed rules before syncing them to PD
- add `location_labels` support to the PD placement rule API type

## Tests
- `go test ./pkg/controllers/placementpolicy ./pkg/pdapi/v1`

### DIFF
--- a/pkg/controllers/placementpolicy/controller.go
+++ b/pkg/controllers/placementpolicy/controller.go
@@ -421,9 +421,9 @@ func defaultRuleLocationLabels(ctx context.Context, pdc pdapi.PDClient) ([]strin
 		return nil, fmt.Errorf("default placement rule bundle %q is nil", pdDefaultRuleGroupID)
 	}
 
-	for _, rule := range bundle.Rules {
-		if rule.ID == pdDefaultRuleID {
-			return rule.LocationLabels, nil
+	for i := range bundle.Rules {
+		if bundle.Rules[i].ID == pdDefaultRuleID {
+			return bundle.Rules[i].LocationLabels, nil
 		}
 	}
 
--- a/pkg/controllers/placementpolicy/controller.go
+++ b/pkg/controllers/placementpolicy/controller.go
@@ -49,6 +49,8 @@ import (
 
 const (
 	pdPlacementRuleGroupIndex = 99
+	pdDefaultRuleGroupID      = "pd"
+	pdDefaultRuleID           = "default"
 
 	reasonSynced     = "Synced"
 	reasonSyncFailed = "SyncFailed"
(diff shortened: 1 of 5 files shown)


pkg/controllers/placementpolicy/controller.go
```diff
@@ -396,12 +398,45 @@ func syncPolicyRules(
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get default placement rule bundle: %w", err)
+	}
+	if bundle == nil {
+		return nil, fmt.Errorf("default placement rule bundle %q is nil", pdDefaultRuleGroupID)
+	}
+
+	for i := range bundle.Rules {
+		if bundle.Rules[i].ID == pdDefaultRuleID {
+			return bundle.Rules[i].LocationLabels, nil
```
fgksgf, line 426: Normalize empty to nil here, e.g.:

```go
if labels := bundle.Rules[i].LocationLabels; len(labels) > 0 {
    return labels, nil
}
return nil, nil
```

If the default rule carries `location_labels: []`, this returns `[]string{}`, and `setPlacementRulesLocationLabels` stamps it on every rule. PD omits empty `location_labels` in its response, so the current rules unmarshal with nil - `reflect.DeepEqual(current, expected)` in `placementRuleDiffOps` then fails on every reconcile and the operator re-posts all rules to PD each time (permanent no-op write amplification).

pkg/controllers/placementpolicy/controller_test.go
```diff
@@ -251,6 +304,61 @@ func TestSyncPolicyRulesPostsRuleGroup(t *testing.T) {
...
+					LocationLabels: []string{},
+				},
+			},
+		}, nil),
+		pdc.EXPECT().SetPlacementRuleGroup(gomock.Any(), &pdapi.PlacementRuleGroup{
+			ID:       "tidb-operator",
+			Index:    pdPlacementRuleGroupIndex,
+			Override: true,
+		}).Return(nil),
+		pdc.EXPECT().SetPlacementRuleGroupRulesByIDPrefix(gomock.Any(), "tidb-operator", "p:", rules).Return(nil),
```
fgksgf: This assertion is vacuous: `syncPolicyRules` mutates `rules` in place, and the gomock expectation holds a reference to the same underlying array, so at call time DeepEqual compares the mutated slice against itself. The test passes regardless of what `setPlacementRulesLocationLabels` writes - including today, where no normalization logic actually exists (`defaultRuleLocationLabels` passes `[]string{}` straight through).

Build an independent expected slice (as `TestSyncPolicyRulesPostsRuleGroup` does with `expectedRules`) so the assertion is real, and add the normalization to the production code.

pkg/controllers/placementpolicy/controller_test.go
```diff
@@ -237,7 +238,59 @@ func TestSyncPolicyRulesPostsRuleGroup(t *testing.T) {
...
+	rules := []pdapi.PlacementRule{
+		{
+			GroupID: "tidb-operator",
+			ID:      "p:r1-1-raw",
+			Role:    v1alpha1.PlacementPolicyRoleVoter,
+			Count:   3,
+		},
+	}
+	pdc := pdapi.NewMockPDClient(gomock.NewController(t))
+	pdc.EXPECT().GetPlacementRuleBundle(gomock.Any(), "pd").Return(&pdapi.PlacementRuleGroupBundle{
```
fgksgf, line 293: Same aliasing issue as in TestSyncPolicyRulesNormalizesEmptyDefaultLocationLabels: the SetPlacementRuleGroupRulesByIDPrefix expectation at the end of this test references the same `rules` slice that `syncPolicyRules` mutates in place, so the assertion would still pass if non-nil labels were set. Use an independently constructed expected slice.

fgksgf (commented):
The fix direction looks right - rules synced without location labels lose PD's isolation-level scheduling. Two issues to address before merge:

1. `defaultRuleLocationLabels` passes `[]string{}` through as-is. Since PD returns `location_labels` with omitempty (unmarshals to nil), `reflect.DeepEqual` in `placementRuleDiffOps` will never match `[]string{}` against nil, so every reconcile re-posts the rules to PD. Normalize empty to nil (see inline comment).
2. Two of the new tests assert against the same slice that `syncPolicyRules` mutates in place, so the `LocationLabels` assertions are vacuous (see inline comments). Notably, `TestSyncPolicyRulesNormalizesEmptyDefaultLocationLabels` passes today even though no normalization logic exists.

Minor notes:
- The `LocationLabels` branch added to `comparePlacementRule` is currently unreachable in practice: the comparator is only used for sorting in `buildRules`, which runs before labels are set in `syncPolicyRules`. Idempotency is decided by `reflect.DeepEqual` in `placementRuleDiffOps`, not here. Harmless to keep for completeness.
- A test for the `GetPlacementRuleBundle` error path (sync should fail) would round out coverage.
- The `for _, rule := range` -> `for i := range` rewrites in client.go are unrelated mechanical churn; if not forced by lint, consider splitting them out.

fgksgf:
I found one additional correctness boundary on the current head, separate from the existing empty-slice and test-alias threads:

- PD's default rule carries both `location_labels` and `isolation_level`. The former drives best-effort physical spreading, while the latter is the hard minimum isolation constraint ([PD v8.5.2 source](https://github.com/tikv/pd/blob/v8.5.2/pkg/schedule/placement/rule.go#L68-L70)). Because the `tidb-operator` group is written with `override: true`, PlacementPolicy-managed ranges do not inherit those fields from `pd/default`. This patch restores only `location_labels`, so a cluster with a non-empty `isolation_level` can still lose its hard isolation guarantee for these ranges. Please either carry `isolation_level` alongside the labels (including the PD API type and diff), or explicitly document why the hard isolation setting is intentionally out of scope.

Two non-blocking coverage/documentation follow-ups are also distinct from the existing threads:

- Update `docs/rfcs/260707-placement-policy.md` so the reconciliation steps and generated-rule JSON include the inherited `location_labels`.
- Add an HTTP-level client test showing that a labels-only difference produces a batch add/update op whose JSON contains `location_labels`. The current tests verify bundle decoding and the controller-to-client mock boundary separately, but not the complete write path.

fgksgf (approved):

PR #7007, opened 2026-07-21
monitor: skip semver parsing when remote_write is not configured

time-and-fate:

`generateRemoteWrite` parsed `monitor.Spec.Prometheus.Version` as a semver string unconditionally, even though the parsed version is only used inside the `remote_write` loop to gate version-specific fields (`headers`, `name`, `min_shards`, `metadata_config`). When `remote_write` is not configured, the loop never runs and the version is never used — but the parse still runs and fails if the prometheus image tag is not semver-compatible.

The prometheus image tag is user-supplied and not guaranteed to be a semver string. Operator still expects a semver tag in the common case; this PR does not change that. But a non-semver tag — for example a hex string such as a git commit hash pinned as the image tag (i.e. `image:<hex>`) — is a valid image reference, and Operator should not fail reconciliation outright over it. In that case, with no `remote_write` configured, TidbMonitor reconciliation fails with `Invalid Semantic version`, and the `basic-monitor` StatefulSet is never created, even though `remote_write` plays no part in the failure.

The failure happens early in the reconcile path: `SyncMonitor` → `syncTidbMonitorStatefulset` → `syncTidbMonitorConfig` → `getPromConfigMap` → `generateRemoteWrite`. The semver parse fails before the StatefulSet creation step, so the StatefulSet is never created.

- Gate the semver parse behind `len(monitor.Spec.Prometheus.RemoteWrite) > 0`. When `remote_write` is empty, `generateRemoteWrite` returns `nil` without parsing the version.
- Change the return type of `generateRemoteWrite` from `yaml.MapItem` to `*yaml.MapItem` so the caller can assign the (possibly nil) pointer directly to `model.RemoteWriteCfg`. The renderer already skips the `remote_write` key when `RemoteWriteCfg` is `nil`, so this avoids emitting a redundant `remote_write: []` (Prometheus tolerates it, but omitting the key is cleaner).
- When `remote_write` IS configured, the version must still be parsed to gate the version-specific fields, so a non-semver tag remains a real error in that case. This is the intended boundary: parsing is only skipped when `remote_write` is unused.

### Code changes

- [x] Has Go code change

### Tests

- [x] Unit test
  - `TestGenerateRemoteWriteWithEmptyRemoteWrite`: empty `remote_write` + a non-semver (hex) tag does not error and returns `nil`.
  - `TestGenerateRemoteWriteWithNonSemVerVersionAndRemoteWrite`: non-empty `remote_write` + a non-semver (hex) tag still errors (intended boundary).
  - `TestGetPromConfigMapWithoutRemoteWrite`: end-to-end, `getPromConfigMap` does not fail and the rendered `prometheus.yml` has no `remote_write` key.
  - Existing tests `TestGenerateRemoteWrite` and `TestGenerateRemoteWriteWithHighVersion` continue to pass.
- [ ] E2E test
- [ ] Manual test
- [ ] No code

### Side effects

- [ ] Other side effects

No behavior change for users who configure `remote_write` or use a semver-compatible prometheus image tag. Only the case where `remote_write` is unused and the image tag is not semver-compatible is fixed.

### Related changes

This bugfix should be cherry-picked to `release-1.6` as well.

### Release Notes

```release-note
Fix TidbMonitor reconciliation failure ("Invalid Semantic version") when the prometheus image tag is not semver-compatible and `remote_write` is not configured.
```

### DIFF
--- a/pkg/monitor/monitor/util.go
+++ b/pkg/monitor/monitor/util.go
@@ -1503,24 +1503,14 @@ func buildExternalLabels(monitor *v1alpha1.TidbMonitor) model.LabelSet {
 }
 
 func generateRemoteWrite(monitor *v1alpha1.TidbMonitor, store *Store) (*yaml.MapItem, error) {
-	// remote_write is optional. The prometheus version is only used inside the loop
-	// below, to decide whether version-specific remote_write fields (headers, name,
-	// min_shards, metadata_config) should be rendered. So the version only needs to be
-	// parsed when at least one remote_write spec is configured.
+	// Operator expects a semver tag in the common case, but it's not guaranteed (for example a git commit hash pinned
+	// as the image tag), and we see such cases in some OP customers. So we choose not to parse the version
+	// unconditionally. Otherwise, if the version is not a semver tag, reconciliation fails with an "Invalid Semantic
+	// version" error and the basic-monitor StatefulSet is never created, even though remote_write is not configured
+	// and the version is not needed at all.
 	//
-	// Parsing the version unconditionally is wrong because the prometheus image tag
-	// is not guaranteed to be a semver string. Operator expects a semver tag in the
-	// common case, but the tag is ultimately user-supplied and may be a non-semver
-	// value such as a hex string (for example a git commit hash pinned as the image
-	// tag, i.e. image:<hex>). Parsing such a value fails. When that happens while
-	// remote_write is unused, reconciliation fails with an "Invalid Semantic version"
-	// error and the basic-monitor StatefulSet is never created, even though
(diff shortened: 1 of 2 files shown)


fgksgf (approved):
LGTM. The fix correctly skips Prometheus semver parsing only when `remote_write` is not configured, while preserving the existing version-gated behavior when it is configured. The tests cover both boundaries and the end-to-end config rendering path.

Non-blocking: this also changes the rendered output for every monitor without `remote_write` from `remote_write: []` to omitting the key. These forms are equivalent to Prometheus, but the PR description statement that semver-tag users see no behavior change could mention this config-output change.

PR #7008, opened 2026-07-22
feat: block tikvgroup deletion if ref placementpolicies are not deleted

liubog2008:
block tikvgroup deletion if placement policies are not deleted

### DIFF
--- a/pkg/apicall/placement_policy.go
+++ b/pkg/apicall/placement_policy.go
@@ -1,0 +1,65 @@
+// Copyright 2024 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package apicall
+
+import (
+	"cmp"
+	"context"
+	"slices"
+
+	"github.com/pingcap/tidb-operator/api/v2/core/v1alpha1"
+	coreutil "github.com/pingcap/tidb-operator/v2/pkg/apiutil/core/v1alpha1"
+	"github.com/pingcap/tidb-operator/v2/pkg/client"
+)
+
+func ListPlacementPolicies(ctx context.Context, c client.Client, ns, cluster string) ([]*v1alpha1.PlacementPolicy, error) {
+	var list v1alpha1.PlacementPolicyList
+	if err := c.List(ctx, &list,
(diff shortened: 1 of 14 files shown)


fgksgf:
Review at head `4abbe4e` — no blocking findings. The deletion-guard mechanism is sound and fail-safe on the paths I traced.

**Verified correct**

- Fail-safe on list errors: if `ListPlacementPoliciesReferencingTiKVGroup` fails, the task returns `Fail`, `blockTask` stops the deletion branch, and `TaskGroupFinalizerDel` never runs (`pkg/utils/task/v3/task.go:140-159`). Teardown cannot start on API errors.
- Unblock path is complete: `Wait` maps to no-requeue (`pkg/utils/task/v3/runner.go:73-76`), so the new PlacementPolicy watch is the resume trigger — Update enqueues both old and new ref sets, Delete enqueues refs, and status-only updates are filtered out (`pkg/controllers/tikvgroup/handler.go:60-88`). Operator restart also re-enqueues deleting groups.
- No deadlock with cluster teardown: the block is bypassed when the Cluster is deleting (`pkg/controllers/tikvgroup/tasks/placement_policy.go:33-35`), and the policy's own finalizer never waits on TiKVGroups (`pkg/controllers/placementpolicy/controller.go:275-285`).
- Prerequisites in place: field index `spec.cluster.name` for PlacementPolicy (`cmd/tidb-operator/main.go:334`), RBAC get/list/watch on placementpolicies (`api/core/v1alpha1/doc.go:48`). Namespace-local + same-cluster ref semantics match `resolveTargets` (`pkg/controllers/placementpolicy/controller.go:347-372`).
- Ran the touched unit packages locally at head (`task/v3`, `apicall`, `apiutil/core/v1alpha1`, `tikvgroup`, `tikvgroup/tasks`) — all pass; `go vet` clean on `tests/e2e/placementpolicy`.

**Non-blocking**

1. **Blocked deletion is invisible to users.** When the gate holds, the runner breaks before `TaskGroupConditionReady`/`TaskStatusPersister` (`pkg/controllers/tikvgroup/builder.go:42-47`), so the block reason (`"blocked by referencing placement policies: [p1]"`) only exists in operator logs. `kubectl delete tikvgroup` leaves a stuck object with no condition or event explaining why. Suggest setting a condition (e.g. `Ready=False, Reason=BlockedByPlacementPolicies`) or emitting an Event before breaking.
2. **Residual create-race.** The gate is level-triggered per reconcile: a policy created (or a ref added) after `TaskPlacementPolicyRefBlock` passes but before/during `TaskGroupFinalizerDel` will reference a group whose stores are going away — the rules become silent no-ops. The watch re-queues the group, but teardown may already be past the gate. If that window matters, a validating webhook rejecting `GroupRefs` pointing at deleting TiKVGroups would close it. Acceptable as-is; candidate follow-up.
3. **e2e read-modify-write** (`tests/e2e/placementpolicy/placementpolicy.go:251-259`): `Get` + `Client.Update` on the policy has a small conflict window if the placementpolicy controller writes status in between; the repo's `action.mergePatch` pattern avoids it. Minor flake risk only.

PR #7022, opened 2026-08-06
pdms: use evict API for TSO primary transfer

bufferflies:

TSO service may have multiple keyspace groups, each with its own primary. The old transfer API only handles a single group's primary. This PR uses the new evict API (introduced in tikv/pd#10968) that transfers all keyspace group primaries from the TSO node at once.

- Add `pdMSPrimaryEvictPrefix = "api/v1/primary/evict"` constant
- In `TransferPrimary`, for TSO service, try the evict API first (POST /tso/api/v1/primary/evict). If it succeeds, return. On failure (old PD without evict support), fall back to the old transfer API (POST /tso/api/v1/primary/transfer).
- Non-TSO services (router, scheduler) continue to use the transfer API directly.

### Code changes

- Has Go code change

### Tests

- Unit test (existing tests pass)

### Related changes

Related to tikv/pd#10968

### DIFF
--- a/pkg/pdapi/pdms_api.go
+++ b/pkg/pdapi/pdms_api.go
@@ -35,6 +35,7 @@ type PDMSClient interface {
 
 var (
 	pdMSHealthPrefix          = "api/v1/health"
+	pdMSPrimaryEvictPrefix    = "api/v1/primary/evict"
 	pdMSPrimaryTransferPrefix = "api/v1/primary/transfer"
 )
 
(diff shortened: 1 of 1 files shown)


liubog2008:
Add a feature gate to change current api call. Hence this api is not available for all pd versions.

liubog2008 (changes requested):
add a new feature gate for this incompatible change

PR #7033, opened 2026-08-17
tiproxy: handle failure for mark unhealthy or graceful-shutdown-begin-time

djshow832:
This PR handles 2 problems:
- As mentioned in https://github.com/pingcap/tidb-operator/pull/6964#issuecomment-5112367570, if SIGTERM is not sent to a legacy TiProxy, the operator will try to mark it unhealthy forever and never delete it. This PR skips marking when the API is not unsupported and deletes it anyway.
- As mentioned in https://github.com/pingcap/tidb-operator/pull/6964#issuecomment-5176385099, if marking unhealthy succeeds but marking graceful-wait-before-shutdown fails, then a scale-out happens, the TiProxy pod will be unhealthy forever and never be deleted. This PR marks the annotation before marking unhealthy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TiProxy graceful shutdown by recording its start time before health status changes.
  * Shutdown now honors the configured deletion delay when health-management APIs are unsupported.
  * Prevented unhealthy status changes when saving shutdown state fails.
  * Improved behavior across deletion, monitoring, and autoscaling workflows.

* **Tests**
  * Expanded coverage for delayed deletion, health API responses, metrics, and scale-in/scale-out scenarios.

### DIFF
--- a/api/core/v1alpha1/common_types.go
+++ b/api/core/v1alpha1/common_types.go
@@ -182,9 +182,9 @@ const (
 	// Last instance template is recorded to check whether the pod should be restarted because of changes of instance template
 	AnnoKeyLastInstanceTemplate = AnnoKeyPrefix + "last-instance-template"
 
-	// TiProxy graceful shutdown begin time is recorded on the pod when graceful shutdown
-	// begins (after MarkUnhealthy succeeds). HPA and monitoring may exclude pods
-	// carrying this annotation.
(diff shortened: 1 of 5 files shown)


YangKeao (approved):
LGTM

liubog2008 (approved):

PR #7057, opened 2026-09-03
feat: end evict leader only when pod is available

liubog2008:
## What problem does this PR solve?

During a TiKV rolling restart, the evict leader scheduler could be removed immediately after the replacement Pod was created. At that point the Pod might not be Ready, or might not have remained Ready long enough to be considered available, allowing leaders to be scheduled to an unstable TiKV instance.

## What is changed and how does it work?

- Keep the evict leader scheduler until the restarted TiKV Pod is available.
- Consider the Pod available only after its Ready condition has remained true for the existing `minReadySeconds` window of 15 seconds.
- Requeue reconciliation when the Pod is Ready but has not yet passed the minimum readiness window.
- Add `podutil.IsReady` and `podutil.IsAvailable`, and use them throughout the TiKV tasks instead of depending directly on the vendored StatefulSet helpers.
- Preserve the existing deletion behavior when the TiKV store itself is being removed.

## Tests

- Unit tests verify that the scheduler is retained while the restarted Pod is unready or Ready but not yet available, and removed after it becomes available.
- Unit tests cover Pod readiness and availability boundary cases in `podutil`.
- Local checks:
  - `go test ./pkg/controllers/tikv/... ./pkg/utils/podutil -count=1`
  - `go vet ./pkg/controllers/tikv/... ./pkg/utils/podutil`

## Summary by CodeRabbit

* **Bug Fixes**
  * Leader eviction now continues until a restarted pod is ready and has remained available for the configured minimum duration.
  * Pod readiness and availability checks now handle missing pod data consistently.
* **New Features**
  * Added standardized checks for pod readiness and sustained availability.
* **Tests**
  * Added coverage for readiness, availability duration, and leader-eviction retry behavior.

### DIFF
--- a/pkg/controllers/tikv/tasks/ctx.go
+++ b/pkg/controllers/tikv/tasks/ctx.go
@@ -25,8 +25,8 @@ import (
 	"github.com/pingcap/tidb-operator/v2/pkg/runtime/scope"
 	pdv1 "github.com/pingcap/tidb-operator/v2/pkg/timanager/apis/pd/v1"
 	pdm "github.com/pingcap/tidb-operator/v2/pkg/timanager/pd"
+	"github.com/pingcap/tidb-operator/v2/pkg/utils/podutil"
 	"github.com/pingcap/tidb-operator/v2/pkg/utils/task/v3"
-	"github.com/pingcap/tidb-operator/v2/third_party/kubernetes/pkg/controller/statefulset"
 )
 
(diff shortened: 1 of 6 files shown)


fgksgf (approved):
Reviewed at 00f3e6e. The scheduler now remains active until the replacement TiKV Pod is available, and the readiness helpers preserve the existing semantics. Focused tests and vet pass.

PR #7062, opened 2026-09-10
feat: support pausing component group reconciliation

WangLe1321:

Pausing a rolling restart through `Cluster.spec.paused` also stops instance reconciliation. A Pod whose deletion has already been submitted may remain absent, and TiKV leader-eviction cleanup may remain unfinished. Operators need to pause a component Group while allowing its existing instances to finish recovery.

### What changed and how does it work?

Add optional `spec.progressing` (`*bool`, default `true`) to the 14 component Group APIs (including DMGroup and DMWorkerGroup) and generated CRDs. Each Group TaskRunner checks it through a shared generic condition immediately after `Cluster.spec.paused`. All Groups expose `Progressing()` through the shared runtime Group interface; generated accessors treat an omitted value as true. Setting it to `false` stops Group reconciliation before instance listing, finalizer handling, revision or Service updates, scaling, rollout actions, and Group status updates. RouterGroup and ResourceManagerGroup deletion handling now also runs after both pause checks.

Independent instance controllers continue running, allowing Pod recreation, readiness recovery, and TiKV leader-eviction cleanup. `Cluster.spec.paused` takes precedence and also stops instance reconciliation. Already-issued operations are not canceled.

Setting `progressing` to `true`, or omitting it, resumes Group reconciliation toward the latest desired template and replica count. Scaling, Group deletion, and deferred cleanup wait until resume; existing surge instances are retained. Group status is frozen while paused, and there is no pause-completion acknowledgement. The flag is outside the instance template, so toggling it does not change the rollout revision.

Old typed clients must preserve this field when performing full-object updates: dropping an existing `false` value allows the API server default of `true` to resume reconciliation. This PR does not add server-side compatibility enforcement.

Validation:
- Shared-condition tests cover omitted/true/false values for PDGroup, DMGroup, and DMWorkerGroup. CRD defaulting tests cover all 14 Groups; revision-stability tests are retained.
- Race-enabled controller, updater, revision, and hash tests passed.
- Race-enabled API module and CRD validation tests passed.
- Targeted golangci-lint and `git diff --check` passed.
- CRDs were regenerated with `make crd`, and runtime accessors with `make runtimegen`.
- Race-enabled controller, runtime, runtime-generator, hash, revision, API, and CRD validation tests passed with DM Group support.

Live-cluster end-to-end validation was not run.

## Summary by CodeRabbit

- **New Features**
  - Added an optional `progressing` setting for component groups, defaulting to `true`.
  - Setting it to `false` pauses group updates, scaling, deletion, and status reconciliation.
  - Existing instances continue recovery and Pod reconciliation; already-issued operations are not canceled.
  - Cluster-level pausing takes precedence and also stops instance reconciliation.

- **Configuration**
  - TiKV cache TTL configuration now belongs to the template settings.
  - Added configurable minimum ready time for TiKV leaders.


api/core/v1alpha1/progressing_test.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2024 PingCAP, Inc.
```
liubog2008: remove useless test

pkg/controllers/pdgroup/tasks/updater.go
```diff
@@ -47,6 +48,10 @@ func TaskUpdater(state *ReconcileContext, c client.Client, af tracker.AllocateFa
 		logger := logr.FromContextOrDiscard(ctx)
 		pdg := state.PDGroup()
 
+		if !ptr.Deref(pdg.Spec.Progressing, true) {
```
liubog2008: Do not only stop updater. Stop at the beginning of the reconciler.

liubog2008: See cluster is paused as example.

api/core/v1alpha1/progressing_test.go
WangLe1321: Removed `api/core/v1alpha1/progressing_test.go` in e4e324506. Kept the CRD defaulting/explicit-value and revision-stability tests, and added behavior tests against all 12 real Group TaskRunners for pausing and resuming deletion. Local race-enabled controller/API/validation tests and targeted lint passed.

pkg/controllers/pdgroup/tasks/updater.go
WangLe1321: Addressed in e4e324506. All 12 kernel Group TaskRunners now check `progressing` immediately after `task.IfBreak(common.CondClusterIsPaused(state))` and exit before further Group reconciliation. Removed the updater-only guards. RouterGroup and ResourceManagerGroup deletion handling now also runs below both pause checks.

`progressing=false` pauses Group deletion, finalizer handling, revisions, Services, scaling, rollout actions, and Group status updates. Existing instance controllers continue independently; Cluster pause still takes precedence. Nil/true resumes reconciliation. The API/CRD descriptions and PR description now reflect these semantics.

Runner tests cover all 12 Groups with omitted/true/false values, Cluster pause precedence, normal and deleting Groups, no work beyond context reads while paused, and deletion completing after resume. Race-enabled controller, updater, revision, hash, API, and CRD validation tests passed; targeted lint reported 0 issues.

PR #7077, opened 2026-09-15
feat: support reducing group volume requests

liubog2008:
## Summary

Allow groups with VolumeAttributesClass enabled to request smaller volumes. For example, reducing TiKV storage from 2Gi to 1Gi keeps existing PVCs and Pods running, creates new 1Gi PVCs on scale-out, and prefers instances with excess capacity during scale-in.

- Add the informational `VolumeCapacityExceedsRequest` instance condition and a shared scale-in preference that respects existing topology constraints.
- Retain sufficient existing PVC requests through composable Apply transformers; allow smaller group requests in admission validation when VAC is enabled.
- Add the design document and a TiKV volume e2e case covering request reduction without Pod restarts and replacement through scale-out/scale-in.
- Extend the Pod rolling-update observer with a no-restart mode, migrate existing checks, and consolidate instance/PVC assertions in generic waiters.

## Validation

- Latest commit `69bff08c9`: all CI passed, including `pull-e2e`, build, image, lint, unit, and verify.
- Passed: unit tests for client (including transformer creation/update and unchanged status), coreutil, common controller tasks, updater, volumes, reloadable configuration, runtime, and e2e waiters.
- Passed: e2e suite compilation and upgrade e2e package compilation.
- Earlier implementation validation on the default Kubernetes cluster: TiKV volume e2e (1 passed); TiDB/TiProxy hot reload and TiDB rolling update cases (3 passed). These cluster runs preceded the final refactors and current consolidated commit.
- `git diff --check` passed.

## Release note

With VolumeAttributesClass enabled, group volume requests can be reduced for newly created PVCs. Existing PVCs retain their capacity, instances report `VolumeCapacityExceedsRequest`, and scale-in prefers instances with excess capacity.

## Summary by CodeRabbit

* **New Features**
  * Volume requests can be reduced when VolumeAttributesClass is enabled, without replacing instances or restarting Pods.
  * Existing PVC capacity is preserved while new PVCs use the smaller requested size.
  * Instance status reports when allocated volume capacity exceeds the requested amount.
  * Scale-in can prioritize instances with excess volume capacity.
  * Apply operations support transformation hooks for customized resource updates.

* **Bug Fixes**
  * Prevented unsafe PVC request reductions when existing capacity is already sufficient.


charts/tidb-operator/templates/webhook.yaml
```diff
@@ -66,9 +66,9 @@ spec:
...
-    messageExpression: "'Volume storage can only be increased, not decreased. One or more volumes have smaller storage than before'"
+  # Existing PVCs retain their capacity when a VAC-enabled group requests less storage.
+  - expression: "request.operation != 'UPDATE' || (has(object.spec.features) && 'VolumeAttributesClass' in object.spec.features) || !has(object.spec.template) || !has(object.spec.template.spec) || !has(object.spec.template.spec.volumes) || !has(oldObject.spec.template) || !has(oldObject.spec.template.spec) || !has(oldObject.spec.template.spec) || !has(oldObject.spec.template.spec.volumes) || object.spec.template.spec.volumes.all(newVol, oldObject.spec.template.spec.volumes.filter(oldVol, oldVol.name == newVol.name).all(oldVol, quantity(oldVol.storage).compareTo(quantity(newVol.storage)) <= 0))"
```
fgksgf, line 70: **Blocking:** This only checks that VAC is present on the update that lowers storage, so the safe path can be exited afterward. For example, lower a request from `2Gi` to `1Gi` with VAC, then remove VAC (or remove it from the Cluster, whose feature reconciler patches the group's `spec.features`). The second update is admitted because storage is unchanged. The instances still get the `1Gi` desired request, but `TaskPVC` switches to `LegacySyncPVCs`; for an existing `2Gi` PVC, `rawModifier.validate` rejects the shrink and `LegacySyncPVCs` then falls through to applying the smaller PVC request, which the API server rejects on every reconcile. The predicate also trusts the mutable new group feature mirror, so adding VAC and lowering storage in the same update can pass even when the Cluster gate is disabled. Please make feature transitions safe (or validate against authoritative gate state), and add regression coverage for removing VAC after a reduction.

- HuaxiClaw

fgksgf (commented):
Reviewed the full PR. I found one additional blocking feature-transition issue, noted inline. I also ran the focused unit suites and compiled all controller/e2e packages successfully. However, the real API-server integration test fails at `capacity_integration_test.go:93`: it gets an invalid PVC shrink instead of the expected conflict, confirming the unresolved resource-version guard issue already raised by CodeRabbit. `pull-e2e` is still pending.

- HuaxiClaw

pkg/controllers/common/task_volume_capacity.go
```diff
@@ -0,0 +1,62 @@
...
+// TaskInstanceConditionVolumeCapacityExceedsRequest observes PVC capacity and updates
+// the in-memory condition. The controller's status task persists the change.
+func TaskInstanceConditionVolumeCapacityExceedsRequest[
+	S scope.Instance[F, T],
+	F client.Object,
+	T runtime.Instance,
+](state InstanceCondVolumeCapacityExceedsRequestUpdater[F], c client.Client) task.Task {
+	return task.NameTaskFunc("CondVolumeCapacityExceedsRequest", func(ctx context.Context) task.Result {
+		obj := state.Object()
+		vols := coreutil.Volumes[S](obj)
```
fgksgf, line 44: **Correctness:** `coreutil.Volumes[S](obj)` only exposes `Spec.Volumes`, so this omits the special PVC-backed volumes used by DM. `DM.Spec.DataVolume` and `DMWorker.Spec.RelayVolume` are created separately in their `PVCNewer` implementations, but the generated `Volumes()` accessors return only `Spec.Volumes`. A normal DM with only its required data volume therefore reports `False/NoVolumes` even when that PVC exceeds the reduced request, and the scale-in preference can never select it; relay volumes have the same problem. Please evaluate each component's complete set of managed PVC-backed volumes and add DM data-volume / DMWorker relay-volume coverage.

- HuaxiClaw

tests/e2e/utils/waiter/pod.go
```diff
@@ -186,6 +225,9 @@ func generatePodInfoMapByWatch[
 		return nil, fmt.Errorf("watch stopped unexpected: %w", err)
 	}
 
+	if !initialized {
```
fgksgf, line 228: **Test reliability:** accepting an interrupted watch here does not guarantee that the observer consumed all events caused by the operation. Callers cancel as soon as an independent readiness/status waiter succeeds; if Pod deletion/add events are already queued but cancellation wins first, `podMap` still contains the original live Pod and `noRestart` can incorrectly pass after a real restart. The initial `synced` channel closes only the pre-action race, not this shutdown race. Please preserve baseline UIDs and perform a final authoritative List comparison, or establish a resource-version catch-up barrier before checking, with a test that delays replacement events until cancellation.

- HuaxiClaw

tests/e2e/utils/waiter/volume.go
```diff
@@ -0,0 +1,170 @@
...
+					continue
+				}
+				if pvc.Status.Phase != corev1.ClaimBound {
+					errList = append(errList, fmt.Errorf("PVC %s is %s, expected Bound", key, pvc.Status.Phase))
+				}
+				capacity := pvc.Status.Capacity.Storage()
+				if request := pvc.Spec.Resources.Requests.Storage(); request.Cmp(*capacity) != 0 {
+					errList = append(errList, fmt.Errorf("PVC %s requests %s, capacity is %s", key, request.String(), capacity.String()))
+				}
+				if capacity.Cmp(expected.Storage) != 0 {
```
fgksgf: **False positive:** `expectDifferent` is intended to validate `VolumeCapacityExceedsRequest=True`, but this accepts any unequal capacity, including an undersized PVC. For example, an instance requesting `1Gi` with request/capacity both `500Mi` sets `different=true` and passes. Require `capacity.Cmp(expected.Storage) > 0` for the excess branch (and equality for the non-excess branch), and add less-than/equal/greater-than unit cases.

- HuaxiClaw

fgksgf (commented):
Three additional findings from the cross-check are inline: one missing special-volume path in condition/scale-in behavior and two e2e assertions that can produce false positives.

- HuaxiClaw

pkg/controllers/common/task_volume_capacity.go
liubog2008: https://github.com/pingcap/tidb-operator/issues/7078

fgksgf:
Re-reviewed head `8ddb147d`.

Fixed in this update:
- The multi-PD leader assertion now propagates its error.
- Transformed applies now carry a resource-version precondition; the Kubernetes 1.32 API-server integration test passes locally.
- The single-volume undersized-PVC false positive is covered and fixed.

Still blocking from the previous review:
- The admission policy still trusts the mutable group-local `spec.features` and still permits leaving VAC after a reduced request has been admitted: https://github.com/pingcap/tidb-operator/pull/7077#discussion_r4015076507
- The no-restart Pod observer still treats cancellation as completion without an authoritative final UID/List check, so queued replacement events can be missed: https://github.com/pingcap/tidb-operator/pull/7077#discussion_r4015105004

Additional current-head concerns are already covered by the new inline reviews:
- With multiple volumes, one oversized PVC can still hide another undersized PVC in the excess assertion.
- The resource-version guard is currently applied to every SSA update, not only transformer-dependent writes, which changes the concurrency behavior of the shared apply primitive.

The DM special-volume gap is now tracked by #7078, so I am treating that item as explicitly deferred from this PR's `spec.template.spec.volumes[]` scope.

Local validation passed for the focused unit suites, all e2e package compilation, `git diff --check`, and `TestCapacityWithAPIServer` against Kubernetes 1.32 envtest. GitHub build/lint/unit/verify/image checks are green; `pull-e2e` is still pending.

- HuaxiClaw

pkg/client/client.go
```diff
@@ -123,6 +120,18 @@ func (p *applier) ApplyWithResult(ctx context.Context, obj client.Object, opts .
 
 		hasCreated = false
 	}
+	var currentObj client.Object
+	if hasCreated {
+		currentObj = obj
+	}
+	for _, t := range o.Transformers {
+		expectedObj = t.Transform(currentObj, expectedObj)
```
fgksgf, line 128: **Blocking regression:** removing the resource-version assignment entirely reopens the stale transformed-apply race. On this exact head, I restored the deleted `TestCapacityWithAPIServer` and ran it against Kubernetes 1.32 envtest; it fails at the concurrency assertion with `expected resource version conflict, got ... PVC ... field can not be less than status.capacity`. The concern about changing every SSA apply can be addressed by attaching the fetched resource version only when `o.Transformers` is non-empty (or otherwise tracking that the write depends on transformed current state), rather than dropping the precondition. Please restore the scoped guard and keep the API-server regression test.

- HuaxiClaw

fgksgf (commented):
Re-reviewed head `69bff08c`. The mixed-volume excess assertion is fixed and covered, the duplicate CEL guard is removed, and all CI checks including `pull-e2e` are green. However, this update removes the resource-version protection and its API-server regression test, which reintroduces the stale transformed-apply race; details and exact reproduction are inline. The previously reported mutable feature-gate admission bypass and no-restart observer cancellation race also remain unchanged.

Focused unit tests, e2e package compilation, Helm rendering, and `git diff --check` pass locally. The restored Kubernetes 1.32 API-server regression test fails on the current head.

- HuaxiClaw

pkg/client/client.go
fgksgf: Correction after tracing the actual writers: I do not see a normal second writer of PVC `spec.resources.requests` in this controller path. The integration test injects one with a direct `Update`; the Kubernetes external resizer normally updates PVC status, while the component controller is single-leader and serializes reconciliation for the instance key. A competing spec write would therefore require manual/independent automation or another unsupported writer. The test demonstrates a generic stale-write case, but I cannot establish it as a production-reachable race under the supported ownership model. I am withdrawing the blocking classification; removing the global resource-version guard is reasonable unless another real spec writer is identified.

- HuaxiClaw

fgksgf (approved):

PR #7082, opened 2026-09-18
feat: add feature logs to replay default feature gates at any time

liubog2008:
Feature defaults can change as features advance through stages across operator versions. This change records versioned feature definitions so historical default gates can be reconstructed by hash and used for upgrade compatibility checks.

- Define features as private iota constants with revision, stage, default, and explicit unreloadable annotations. A gengo-based generator produces API constants and stages, complete feature-log snapshots, a hash-to-revision index, and component restart metadata. Hashes use canonical JSON chained through the previous hash.
- Add constructors for historical defaults, Cluster gates with explicit spec overrides, and Group/Instance gates. `FeatureGate.Enabled` defaults to true and supports explicit disabling; update CRD validation accordingly.
- Add `status.featureGatesHash` and a reconciliation checkpoint that persists the adopted definition hash before changing finalizers or managed resources. Startup compatibility enforcement remains follow-up work.
- Add standalone feature-log lint that compares generated snapshots against a fetched baseline: published revisions must remain unchanged, and a PR may append only one revision. A baseline without feature logs is treated as empty history, allowing initial rev 0. Integrate lint into CI and `hack/lint.sh`, and share repository fetching with upgrade e2e through `hack/lib/repo.sh`.

Design: [feature log RFC](https://github.com/pingcap/tidb-operator/blob/fg-precheck/docs/rfcs/260917-feature-log.md).

Validation:

- Passed: `go test ./cmd/feature-gen ./cmd/feature-log-lint ./pkg/features`, including hash stability, history reconstruction, append-only checks, and local Git fetch integration tests.

## Summary by CodeRabbit

- **New Features**
  - Added feature-definition history and hashes for tracking adopted configurations.
  - Cluster status now records the active feature-definition hash.
  - Feature gates support optional `enabled` values, defaulting to enabled when omitted, with up to 256 entries.
  - Added historical definition lookup and restart handling for non-reloadable features.

- **Bug Fixes**
  - Prevented invalid feature-gate changes, including disabling protected features or altering unrelated gates.
  - Persisted feature-definition hashes before other reconciliation changes.
  - Improved upgrade cancellation, readiness checks, and failure diagnostics.
  - Added validation to preserve append-only feature history.

- **Documentation**
  - Added documentation covering feature history, hashing, gate behavior, and restart requirements.

- Passed: `go test ./pkg/controllers/cluster/...`, Cluster controller lint, and `go test ./...` in `tests/validation`.


fgksgf:
Thanks for the well-structured PR — the RFC is thorough and the implementation matches it closely. I went through the design and the key paths (generator + hash protocol, gate constructors, cluster checkpoint task, append-only lint). This is solid groundwork overall. A few findings, all verified against head `35321c1a`:

**Bugs (small, worth fixing before merge)**

1. `Makefile:147` — `lint/golangci: bin/golangci` references a nonexistent target. The `bin/%` pattern rule only covers `GO_TOOL_BIN`, which lists `golangci-lint`, so `make lint/golangci` fails with "No rule to make target 'bin/golangci'". Should depend on `bin/golangci-lint` (same as the aggregate `lint` target already does). Also flagged by CodeRabbit; still present at head.

2. The RFC (Append-only lint section) documents `make fetch/old-version-repo V_REPO_REF=...`, but no such target exists in the Makefile — anyone following the docs will hit "No rule to make target".

**Nits**

3. `tests/e2e/upgrade/upgrade.go:74` — on timeout, `exec.CommandContext` only kills the `bash` parent and leaves `kubectl` children running; the `cd %s && ...` also embeds an unquoted path. Suggest setting `cmd.Dir = absProjectRoot` and running the command in a process group that gets killed on context cancellation.

4. `FeatureGateStatus` in `api/meta/v1alpha1/feature_types.go` is now dead code since `status.featureGates` was replaced by `status.featureGatesHash` — could be dropped here or in a follow-up.

**Design notes (fine as-is, just calling out)**

- The feature-log lint intentionally ignores the generated `hashToRev` hashes (RFC says so explicitly). Within a PR, hash/source consistency is enforced by `make verify`, but a future change to the hash generation itself would pass this lint — meaning the hash protocol is effectively frozen once this lands, since historical hashes recorded in `status.featureGatesHash` must keep resolving via `DefinitionsAt`. Worth stating that freeze explicitly in the RFC if it's intended.

- Since `FeatureGate.Enabled` now defaults to `true`, any external consumer of `spec.featureGates` that treats presence as "enabled" needs to learn about `enabled: false` (not an issue in-repo; `ClusterFeatures`/`EnabledFeatures` handle it correctly).

**Highlights**

- `TaskFeatureGatesHash` re-checking `updated.Status.FeatureGatesHash` after the status write to catch an outdated CRD silently pruning the field is a nice touch.
- Checkpoint ordering (persist the hash before finalizer/propagation/service changes, fail-stop on write errors) is correct.
- Golden tests pinning the protocol bytes plus the CI append-only lint integration look good; patch coverage is solid.

cmd/feature-log-lint/fetch_test.go
```diff
@@ -0,0 +1,103 @@
...
+		output, err := cmd.Output()
+		return strings.TrimSpace(string(output)), err
+	}
+	assertFetch := func(ref, want string) {
+		t.Helper()
+		actual, err := fetch(ref)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if actual != checkout {
```
fgksgf, line 73: **[P2] Resolve symlinks before comparing checkout paths**

`repo::fetch` returns `pwd -P`, but `checkout` comes from `t.TempDir()` without resolving symlinks. On macOS, `/var/...` resolves to `/private/var/...`, so this assertion fails despite a successful checkout and breaks `go test ./cmd/feature-log-lint`.

Reproduced at `35321c1a7bdc`: the test fails with the default temporary directory and passes when `TMPDIR` is set to its physical path. Please normalize the expected checkout path with `filepath.EvalSymlinks` after the directory exists, before comparing it with stdout.

